PR #8, opened 2018-04-20
ServiceMonitor for Logging Elasticsearch

lukas-vlcek:
The rules are taken from <https://github.com/lukas-vlcek/prometheus-elasticsearch-rules> the rest of the code is implemented according to instructions in [contributing-new-component-integrations](https://github.com/openshift/cluster-monitoring-operator/blob/7d23d197f2a5937d66e450b55979d52eebeec3f4/README.md#contributing-new-component-integrations).

I am unsure the most about the `assets/prometheus-k8s/prometheus-k8s-service-monitor-logging-elasticsearch.yaml` file.

I should also say that in general I expect this PR to take some time to mature and settle down.

//cc @richm @jcantrill @portante


lukas-vlcek:
@ironcladlou What do I need to debug this locally? Is just `go` enough or do I also need Kubernetes/OpenShift? It is not clear to me what went wrong with the travis CI job, the log does not seem to point out anything obvious. May be I just need fresh pair of eyes.

assets/prometheus-k8s/prometheus-k8s-service-monitor-logging-elasticsearch.yaml
```diff
@@ -0,0 +1,22 @@
...
+  labels:
+    k8s-app: logging-elasticsearch
+spec:
+  selector:
+    matchLabels:
+      k8s-app: logging-elasticsearch
+      prometheus: TBD
+  namespaceSelector:
+    matchNames:
+    - this-namespace-is-automatically-replaced-at-runtime
```
ironcladlou: This should be an openshift-* namespace

pkg/manifests/manifests.go
```diff
@@ -77,6 +77,7 @@ var (
 	PrometheusK8sKubeControllersServiceMonitor    = "assets/prometheus-k8s/prometheus-k8s-service-monitor-kube-controllers.yaml"
 	PrometheusK8sKubeDNSServiceMonitor            = "assets/prometheus-k8s/prometheus-k8s-service-monitor-kube-dns.yaml"
 	PrometheusK8sPrometheusOperatorServiceMonitor = "assets/prometheus-k8s/prometheus-k8s-service-monitor-prometheus-operator.yaml"
+	PrometheusK8sLoggingElasticsearchServiceMonitor = "assets/prometheus-k8s/prometheus-k8s-service-monitor-logging-elasticsearch.yaml"
```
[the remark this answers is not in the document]
ironcladlou: I thought platform components were supposed to be immutable and prefixed with `openshift-`? /cc @smarterclayton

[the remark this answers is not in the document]
richm: Starting in OCP 3.10, the default logging namespace will be `openshift-logging`.

assets/prometheus-k8s/prometheus-k8s-service-monitor-logging-elasticsearch.yaml
richm: However - if we want to use this with OCP 3.9 and earlier, we will have to be able to use namespace `logging` for Elasticsearch and the other logging components, and `openshift-logging` in OCP 3.10 and later.

ironcladlou:
@lukas-vlcek

>@ironcladlou What do I need to debug this locally? Is just go enough or do I also need Kubernetes/OpenShift? It is not clear to me what went wrong with the travis CI job, the log does not seem to point out anything obvious. May be I just need fresh pair of eyes.

I put some of my own notes here: https://gist.github.com/ironcladlou/24339633f1d5f6c431a951083a363ba4

The development workflow covers a lot of different moving pieces, not sure yet how best to represent it or where it should live.

assets/prometheus-k8s/prometheus-k8s-service-monitor-logging-elasticsearch.yaml
ironcladlou: Cluster Monitoring Operator is designed for OCP 3.10+

derekwaynecarr: @ironcladlou is correct, it must be fixed to openshift-logging and we should not allow any choice.

pkg/manifests/manifests.go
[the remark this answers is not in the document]
derekwaynecarr: @portante we need to remove the option to make logging namespace configurable.

elad661:
@lukas-vlcek fyi, the Travis build failed because you need to run `make generate` and add the modified files to the commit.

lukas-vlcek:
@elad661 Do the `make` and `generate` have any dependencies like Golang or Kubernetes or anything that I might need to install in advance?

lukas-vlcek:
I guess at least go and Docker is required...
```
$ make -f Makefile generate
docker build -t tpo-generate -f Dockerfile.generate .
Cannot connect to the Docker daemon. Is the docker daemon running on this host?
make: *** [generate] Error 1
```

elad661:
Just docker should be sufficient.

lukas-vlcek:
@elad661 This is what I get now (I installed go rpm prior to that, not sure if that is needed anyway...):

```
$ make -f Makefile generate
docker build -t tpo-generate -f Dockerfile.generate .
Sending build context to Docker daemon  37.56MB
Step 1/2 : FROM golang:1.9.2
 ---> 138bd936fa29
Step 2/2 : RUN apt-get update &&     apt-get install -y python-yaml
 ---> Using cache
 ---> ebd4e43dcf63
Successfully built ebd4e43dcf63
Successfully tagged tpo-generate:latest
docker run --rm -v `pwd`:/go/src/github.com/openshift/cluster-monitoring-operator -w /go/src/github.com/openshift/cluster-monitoring-operator tpo-generate make merge-cluster-roles assets docs
python2 hack/merge_cluster_roles.py manifests/cluster-monitoring-operator-role.yaml.in `echo assets/*/*role.yaml` > manifests/cluster-monitoring-operator-role.yaml
go get -u github.com/jteeuwen/go-bindata/...
hack/generate-rules-configmap.sh k8s > assets/prometheus-k8s/prometheus-k8s-rules.yaml
#hack/generate-rules-configmap.sh kube-system etcd > assets/prometheus-etcd/prometheus-etcd-rules.yaml
hack/generate-alertmanager-secret.sh > assets/alertmanager/alertmanager-config.yaml
# Using "-modtime 1" to make generate target deterministic. It sets all file time stamps to unix timestamp 1
go-bindata -mode 420 -modtime 1 -pkg manifests -o pkg/manifests/bindata.go assets/...
embedmd -w `find Documentation -name "*.md"`
```
And I can see there are two modified files now:

```
$ git status
On branch add_logging_elasticsearch_support
Your branch is up-to-date with 'origin/add_logging_elasticsearch_support'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   assets/prometheus-k8s/prometheus-k8s-rules.yaml
	modified:   pkg/manifests/bindata.go

no changes added to commit (use "git add" and/or "git commit -a")
```

There two files need to be part of the commit?

brancz:
The problem is that go-bindata has changed often in the past, so significantly that it broke things, which is why we started including it in the review process.

Adding it to the documentation would definitely be good.

lukas-vlcek:
@elad661 Tests are passing now, great!

portante:
> fyi, the Travis build failed because you need to run make generate and add the modified files to the commit.

Why would we ever need to keep track of files that can be auto generated by running `make generate`?  Why wouldn't that be part of the normal build and test process?

elad661:
@portante I think @brancz explained the reason for this 3 hours ago.

portante:
Doesn't that mean the mechanism is a bit broken and fixing it by checking in auto generated files is just a band-aid that leads to problems like this?  Doesn't this just move the burden from an automated system that *could* do this correctly for everybody out to the developers who won't always do it correctly?

mxinden:
@portante I would like to keep the `docs` target, which is part of `generate` target. Checking in generated docs makes it possible to read them via the Github markdown view.

portante:
@mxinden, have we considered [commit hooks](
https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)?

rules/k8s/logging-elasticsearch.rules.yaml
```diff
@@ -0,0 +1,190 @@
...
+  #     elected yet, and this can take some time if the cluster load is high...)
+  #
+  # If the natural cause takes long or the cause is different than those listed above then it is considered
+  # the critical cause.
+  #
+  # Cluster health becomes YELLOW if not all index replica shards are allocated. If the goal is to have GREEN cluster
+  # health (ie number of replica shards is configured accordingly) but the status stays YELLOW for too long then this
+  # is considered serious.
+  #
+  - alert: Cluster_Health_Status_RED
```
brancz: Severity of alerts is encoded in their labels, not alert name. Alerts are camel case and should describe the alert precisely, more something like "ElasticSearchClusterNotHealthy".

rules/k8s/logging-elasticsearch.rules.yaml
```diff
@@ -0,0 +1,190 @@
...
+    expr: sum by (cluster) (es_cluster_status == 2)
+    for: 2m
+    labels:
+      severity: critical
+    annotations:
+      summary: 'Cluster health status is RED'
+      description: 'Cluster {{ $labels.cluster }} health status has been RED for at least 2 minutes'
+
+  - alert: Cluster_Health_Status_YELLOW
```
brancz: Alerts with checking the same condition with a different threshold are named the same and their severity indicates their urgency.

rules/k8s/logging-elasticsearch.rules.yaml
```diff
@@ -0,0 +1,190 @@
...
+  # is considered serious.
+  #
+  - alert: Cluster_Health_Status_RED
+    expr: sum by (cluster) (es_cluster_status == 2)
+    for: 2m
+    labels:
+      severity: critical
+    annotations:
+      summary: 'Cluster health status is RED'
+      description: 'Cluster {{ $labels.cluster }} health status has been RED for at least 2 minutes'
```
brancz: "RED" has no meaning, describe what this means to the consumer of the elasticsearch cluster, eg. "Cluster {{ $labels.cluster }} is unhealthy, no writes possible."

rules/k8s/logging-elasticsearch.rules.yaml
```diff
@@ -0,0 +1,190 @@
...
+              ) * 100, 0.001)
+            ) > 85
+    for: 5m
+    labels:
+      severity: alert
+    annotations:
+      summary: 'Low Watermark Reached - disk saturation is {{ $value }}%'
+      description: 'Low Watermark Reached at {{ $labels.node }} node in {{ $labels.cluster }} cluster'
+
+  - alert: Disk_High_Watermark_Reached
```
brancz: Would it be possible to predict when this happens with predict linear? It seems like this is a situation we would like to resolve before it actually occurs.

lukas-vlcek: `Disk_High_Watermark_Reached` is quite serious and we want to avoid it for sure. I think it would be possible to use predictive functions provided by Prometheus but it will take some more time to test it properly.

The real ES disk saturation over time is more like saw chain shape and we need to test what kind of predictive function works best in our use case (depends on number of index segments, indexing activity, ... etc). The prediction formula may also be influenced by our needs like how may minutes/hours ahead we need to know this limit is approaching (btw do you know SLA react times here?).

I want to keep it simple for the first round but improving in the future is welcome (and expected 😄).

rules/k8s/logging-elasticsearch.rules.yaml
lukas-vlcek: Got it. I will push new update soon.

rules/k8s/logging-elasticsearch.rules.yaml
brancz: just a convention, I personally couldn't care less but prefer consistency

rules/k8s/logging-elasticsearch.rules.yaml
brancz: No, it just gets turned into a label by the rule engine (but is in some regards special which is why it's not just a label). Only the combination of labels + alertname must be unique.

rules/k8s/logging-elasticsearch.rules.yaml
brancz: > Prometheus but it will take some more time to test it properly.

In that case at least do a warning severity alert before it reaches this level. In general, a small amount of very good alerts is a lot better than any number of poor ones, as it otherwise easily causes alert fatigue and people start ignoring them.

ant31:
shouldn't the rules/alerts be shipped with the component rather than the cluster-monitoring now? 
cc @brancz

ironcladlou:
@ant31 
>shouldn't the rules/alerts be shipped with the component rather than the cluster-monitoring now?
cc @brancz

Yes, although we're not quite ready to deploy support for externalized rules for 3.10. However we are currently able to support adding new scrape targets in the component (via `ServiceMonitor resources).

lukas-vlcek:
@ironcladlou @brancz I am suggesting we close this PR.

I moved follow up work to https://github.com/lukas-vlcek/elasticsearch-mixin for the time being. I also tried to apply the review comments there as well.

The repo also contains initial Grafana dashboard for Elasticsearch. I need to figure out what the future of this repo will be, whether I will create PR against https://github.com/kubernetes-monitoring/kubernetes-mixin or where to move this repo before it is adopted for OCP 3.11.

brancz:
A PR against the kubernetes-mixin doesn't make a lot of sense, as it's not mandatory to run elasticsearch on Kubernetes, and in fact a lot of people don't. For future versions of OpenShift those resources should be created in the logging namespace and the cluster-monitoring stack just picks up the resources from there.

Apart from that, yes we should close this :slightly_smiling_face: .

lukas-vlcek:
@brancz Sure, the reason why I mentioned creating a new PR is the fact that I would like to get new review of the code based on you previous review. If there is no PR then there is not PR review as well 😄

brancz:
That’s a perfectly legitimate thing to request from the team and we can plan it into our sprint :)

Besides that I would recommend publishing a thread on the Prometheus users mailing list to let everyone know of its existence and ask to collaborate.

PR #11, opened 2018-04-24
Add OpenShift app create availability ServiceMonitor

ironcladlou:
Add a ServiceMonitor which targets the app create availability metrics
application[1] which is part of the OpenShift availability monitoring suite[2].

[1] https://github.com/openshift/monitor-project-lifecycle
[2] https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_monitor_availability

### DIFF
--- a/assets/prometheus-k8s/prometheus-k8s-service-monitor-availability-app-create.yaml
+++ b/assets/prometheus-k8s/prometheus-k8s-service-monitor-availability-app-create.yaml
@@ -1,0 +1,22 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: monitor-app-create
+  labels:
+    k8s-app: monitor-app-create
+spec:
+  jobLabel: k8s-app
+  endpoints:
+  - port: http-metrics
+    interval: 30s
+    scheme: https
+    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    tlsConfig:
+      caFile: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
+      serverName: monitor-app-create.openshift-monitor-availability.svc
+  selector:
+    matchLabels:
+      k8s-app: monitor-app-create
+  namespaceSelector:
+    matchNames:
+    - openshift-monitor-availability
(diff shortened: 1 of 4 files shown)


ironcladlou:
Tested this with https://github.com/openshift/openshift-ansible/pull/8138

elad661:
I haven't tested this on an actual cluster, but the file looks fine and the change is simple enough.
/lgtm

assets/prometheus-k8s/prometheus-k8s-service-monitor-availability-app-create.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: monitor-app-create
+  labels:
+    k8s-app: monitor-app-create
+spec:
+  jobLabel: k8s-app
+  endpoints:
+  - port: http-metrics
```
brancz: `http-metrics` -> `https-metrics`

ironcladlou: @brancz 

> http-metrics -> https-metrics

This is actually [referentially correct](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_monitor_availability/files/monitor-app-create.yaml#L79) but I agree I goofed with the name.

>I'm struggling to understand the purpose of this without alerting rules or other interactions with the metrics. Do we just not have an idea yet what we would want to alert on or is it purely for reporting purposes?

Primary interaction with the metrics at this moment would be manual analysis. We're still iterating on the metrics themselves and the subsequent alerts.

I suggest it would be helpful clearly articulate the acceptance criteria for ServiceMonitor and rule additions to the operator.

ironcladlou: My goal was to get the bare minimum integration set up to facilitate experimentation and iteration, but I'm sure there's an argument to be made for WIP'ing the PR until such experimentation is done, which is fine. What's your take on it? In any case I would like to make it clearly stated for the benefit of others.

ironcladlou:
Closing this for now- we're going to commit to a new approach which allows ServiceMonitors and rules to be bundled with the applications outside this repo.

assets/prometheus-k8s/prometheus-k8s-service-monitor-availability-app-create.yaml
brancz: > Primary interaction with the metrics at this moment would be manual analysis.

This is a perfectly reasonable argument to have, just wanted to make sure we're not creating something that doesn't end up not having targets or something like that.

Issue #13, opened 2018-05-03
Unable to run `make generate`


jhadvig:
When running `make generate` fails with following issue:

>$ make generate
docker build -t tpo-generate -f Dockerfile.generate .
Sending build context to Docker daemon 37.55 MB
Step 1/2 : FROM golang:1.9.2
 ---> 138bd936fa29
Step 2/2 : RUN apt-get update &&     apt-get install -y python-yaml
 ---> Using cache
 ---> e0f3c4ea011e
Successfully built e0f3c4ea011e
docker run --rm -v `pwd`:/go/src/github.com/openshift/cluster-monitoring-operator -w /go/src/github.com/openshift/cluster-monitoring-operator tpo-generate make merge-cluster-roles assets docs
make: stat: Makefile: Permission denied
make: *** No rule to make target 'merge-cluster-roles'.  Stop.
make: *** [Makefile:48: generate] Error 2

Not really sure how to solve it.

elad661:
This is an SELinux issue, the files need to be labled as container_t (if I remember correctly) for this to work. I've been working around this locally by running the generate outside a container, since my environment is already set up with all the required dependencies (go, pyyaml):

`make merge-cluster-roles assets docs`

Another option is to set SELinux to permissive mode before running `make generate`. Generally though I think it's another sign we should improve the process to not have any hurdles like this one.

jhadvig:
disablins SELinux does the trick. Thanks @elad661 !!!

rhatdan:
Disabling SELinux should not be the solution.  We should figure out what is mislabeled and fix it.  RUnning something like containers with SELinux disabled is asking for trouble.

The label to assign is container_file_t,  container_t is a process label, and can not be assigned to content.

elad661:
@rhatdan Nothing is "mislabeled" in a way that we can fix properly. What happens here is simple: `make generate` mounts the cluster-monitoring-operator source tree as a volume for the container (see https://github.com/openshift/cluster-monitoring-operator/blob/master/Makefile#L48 ), and because all the files are (probably) labeled `user_home_t` SELinux doesn't allow the container to access any of them.

I think the proper way to fix this would be moving away from using a container for `make generate`, since it's reasonable to expect people who contribute to cluster-monitoring-operator will have Go installed on their development machine.

rhatdan:
Can you disable enforcement in the container IE Run with label:disable rather then disable the selinux on the host.

elad661:
[...]

PR #14, opened 2018-05-03
[WIP] Add ServiceMonitor for Openshift web-console

jhadvig:
Registering Openshift web-console ServiceMonitor.
Set the scraping interval to 5 minutes since we by default we are running the web-console smoke tests every 5 minutes, but I guess the frequency could be higher.

Also wasnt able to run `make generate` due to https://github.com/openshift/cluster-monitoring-operator/issues/13

@ironcladlou PTAL


assets/prometheus-k8s/prometheus-k8s-service-monitor-web-console.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: web-console
+  labels:
+    k8s-app: web-console
+spec:
+  selector:
+    matchLabels:
+      k8s-app: web-console
+  namespaceSelector:
+    matchNames:
+    - this-namespace-is-automatically-replaced-at-runtime
```
elad661: I'm pretty sure this should match the actual namespace in which the web console is installed

ironcladlou:
Since the web console is an optional component, I propose we mark this WIP for now: @mxinden is actively working on enabling us to bundle the ServiceMonitors and alerting rules with the app itself in another namespace. When that's done, we can implement this with a drastically simpler PR to openshift-ansible.

PR #16, opened 2018-05-15
Upgrade to prometheus-operator:v0.19.0

ironcladlou:
Upgrade to prometheus-operator:v0.19.0.

### DIFF
--- a/assets/prometheus-operator/prometheus-operator.yaml
+++ b/assets/prometheus-operator/prometheus-operator.yaml
@@ -23,7 +23,7 @@ spec:
         - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
         - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.0.3
         - --disable-auto-user-group
-        image: quay.io/coreos/prometheus-operator:v0.18.1
+        image: quay.io/coreos/prometheus-operator:v0.19.0
         name: prometheus-operator
         ports:
         - containerPort: 8080


elad661:
Also don't forget running `make generate` or `make merge-cluster-roles` if you modify [assets/prometheus-operator/prometheus-operator-cluster-role.yaml ](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/prometheus-operator/prometheus-operator-cluster-role.yaml)

ironcladlou:
Updated role and regenerated files.

ironcladlou:
Something is wrong here; prometheus-operator v0.19.0 ends up deploying prometheus-config-reloader:v0.0.3 but with arguments compatible only compatible with prometheus-config-reloader:v0.0.4...

elad661:
@ironcladlou oh, oops. I completely missed that.

https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/prometheus-operator/prometheus-operator.yaml#L24

This line needs to be updated to 0.0.4

ironcladlou:
It seems like more than that- [prometheus-operator](https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/statefulset.go#L468) master doesn't seem to be compatible with config reloader in the same branch.

ironcladlou:
I'll fix the CMO as part of another PR i have going

ironcladlou:
Confirmed that the reloader version bump in CMO fixes stuff. Thanks!

PR #18, opened 2018-05-16
Merge v0.0.2 into release-0.0 release branch

mxinden:
@ironcladlou I would wait for your "ok" here, in regards to the e2e tests or a quick manual check.


brancz:
I think you meant v0.0.2?

mxinden:
> I think you meant v0.0.2?

@brancz right, thanks, updated.

ironcladlou:
I've been bouncing back and forth between this and another higher priority issue, so haven't got a test run yet.

mxinden:
The Rule CRD feauture is merged into the Prometheus Operator master branch,
but is not yet released. We are still missing a migration path between old
configmaps and the new CRDs, as well as improved documentation.

ironcladlou:
I think we need to bring in https://github.com/openshift/cluster-monitoring-operator/pull/19, because without the prometheus-config-reloader version bump master is broken by default.

ironcladlou:
Just did a manual test of this, looks good

ironcladlou:
Found one more we'll want to go ahead and bring in.

https://github.com/openshift/cluster-monitoring-operator/pull/20

ironcladlou:
Not sure how you want to handle the version cut commit since I dragged in more changes, but this is ready to go.

PR #19, opened 2018-05-18
Remove resource limits

ironcladlou:
Monitoring component memory requirements are highly variable; remove default
limits to ensure a functional stack on a variety of cluster sizes.


ironcladlou:
@brancz, anything else required here beyond `make generate`?

brancz:
If we’re already at it we should do the same for all other components. Kube-state-metrics and the Prometheus operator also oomed in the scalability tests, so we should do the same.

ironcladlou:
Removed limits across the board, PTAL

assets/kube-state-metrics/kube-state-metrics-deployment.yaml
```diff
@@ -78,13 +78,7 @@ spec:
         image: quay.io/coreos/kube-state-metrics:v1.3.1
         imagePullPolicy: IfNotPresent
         name: kube-state-metrics
-        resources:
-          limits:
-            cpu: 102m
-            memory: 180Mi
-          requests:
-            cpu: 102m
-            memory: 180Mi
```
brancz: We should also remove the add on resizer here, I think it will also require changes in the kube state metrics task in pkg/tasks as there is some manual “merge” logic there because of the addon-resizer.

ironcladlou:
Removed auto-resizer stuff- I still need to do e2e tests for this

ironcladlou:
Round 1 of manual testing on this is looking good to me; still want to do another cluster from scratch for verification.

ironcladlou:
Manual tests look good, but I want to go ahead and fix the e2e tests as part of this.

ironcladlou:
Okay, I decided e2e has enough random work to do to fix that I'd like to separate it out. I'm confident in the manual testing.

jeremyeder:
Just deployed 3.10.0-52 and don't see this change yet. Just want to make sure that this is flagged correctly to end up in 3.10 GA? @ironcladlou @brancz

PR #20, opened 2018-05-22
Remove SCC from prometheus-operator

ironcladlou:
Remove the SCC from prometheus-operator for compatibility.


elad661:
It's weird that we didn't notice this when deploying on GCP. Maybe free-int has a more strict default security policy or something?

mxinden:
@ironcladlou @elad661 Do I understand correctly, that with this change the `securityContext` is set to the cluster default, which would enforce a non root user? How is that user created? We are not adding any users in the [Prometheus Operator Dockerfile](https://github.com/coreos/prometheus-operator/blob/master/Dockerfile).

PR #23, opened 2018-05-24
Add a default namespace selector for service monitors

ironcladlou:
Add a default namespace selector to consume servicemonitors from any namespace
containing an `openshift.io/cluster-monitoring` label (with a any value). This
allows optional cluster components in other namespaces to opt-in to scraping.


ironcladlou:
I made some progress on e2e but it's turning into enough work that I still want to keep it separate. Tested this manually.

assets/prometheus-k8s/prometheus-k8s.yaml
```diff
@@ -12,6 +12,9 @@ spec:
   serviceMonitorSelector:
     matchExpressions:
     - {key: k8s-app, operator: Exists}
```
brancz: I would say we should just change this `serviceMonitorSelector` to the "all" selector (`{}`). `k8s-app` is an old artifact of ours that we should get rid of.

brancz:
just one comment feel free to apply here or in a follow up, otherwise lgtm :+1:

assets/prometheus-k8s/prometheus-k8s.yaml
ironcladlou: Agree, also commented on this [elsewhere](https://github.com/openshift/openshift-ansible/pull/8518#discussion_r190882403). I vote followup for now.

PR #24, opened 2018-05-24
Add namespace label to pod cpu, memory, and disk pod recording rules

chancez:
This adds the `namespace` label to the pod level recording rules. Without this, it's actually incorrect due to pod names not being unique across namespaces, and namespace is important information for other tools.


brancz:
This needs a `make generate`.

Through the effort that we're working on regarding syncronizing these things consistently across the board, the names of these recording rules may change in the future, but we'll make sure to let you know when/if they do.

Otherwise this lgtm.

chancez:
@brancz Updated

PR #25, opened 2018-05-24
[WIP] Convert e2e tests to a Job

ironcladlou:
Fix the e2e tests and convert them to a k8s Job.

### DIFF
--- a/test/e2e/e2e-job-template.yaml
+++ b/test/e2e/e2e-job-template.yaml
@@ -1,0 +1,27 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: cluster-monitoring-operator-test
+parameters:
+- name: IMAGE
+  description: The application container image to use.
+  required: true
+objects:
+- apiVersion: batch/v1
+  kind: Job
+  metadata:
+    name: cluster-monitoring-operator-test
+    namespace: openshift-monitoring
+  spec:
+    parallelism: 1
+    completions: 1
+    backoffLimit: 1
+    activeDeadlineSeconds: 100
+    template:
+      spec:
+        serviceAccountName: cluster-monitoring-operator
+        containers:
+        - name: test
+          image: ${IMAGE}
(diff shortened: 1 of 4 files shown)


ironcladlou:
Not sure if this is totally the right way to go about things, but it's something to talk about.

test/e2e/e2e-job-template.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: template.openshift.io/v1
```
brancz: As our goal with this project is to eventually be compatible with vanilla I would prefer not to introduce any openshift specific means if not absolutely necessary.

test/e2e/main_test.go
```diff
@@ -67,53 +62,72 @@ func TestQueryPrometheus(t *testing.T) {
...
 	}
 
-	err = wait.Poll(5*time.Second, 10*time.Minute, func() (bool, error) {
+	err = wait.Poll(5*time.Second, 20*time.Second, func() (bool, error) {
 		defer t.Log("---------------------------\n")
 
 		for _, q := range queries {
-			n, err := f.QueryPrometheus("prometheus-k8s-0", q.query)
+			response, err := client.query(q.query)
+			res, err := gabs.ParseJSONBuffer(response.Body)
```
brancz: We actually do know the format of this response so we should actually parse it correctly and not use this dependency.

test/e2e/e2e-job-template.yaml
ironcladlou: Good point- also, @smarterclayton later yesterday evening pointed me to docs for the new ci-operator process which I think would provide the path forward here. I'll probably redo this PR along those lines.

test/e2e/main_test.go
ironcladlou: I wasn't entirely sure why we weren't unmarshaling into a little struct, but I was just doing a quick refactor. Will take care of this too.

PR #28, opened 2018-05-31
manifests: Use emptydir to shadow Dockerfile volume directive

brancz:
Unfortunately we missed that kube-state-metrics also has an (unused) volume directive which keeps it from being created in clusters that don't allow non shadowed containers. I already opened the upstream fix, but for the mean time, until that lands and is released, we can fix it on our side with this.

@mxinden @ironcladlou @elad661 

This still needs to be tested on an actual cluster with these constraints, but tests are passing.


ironcladlou:
Tested this approach in free-int (which is now correctly failing).

/lgtm

ironcladlou:
We'll need a release with this commit and a PR to openshift-ansible to upgrade to fix https://bugzilla.redhat.com/show_bug.cgi?id=1584386

PR #30, opened 2018-06-01
Cleanup rules

brancz:
This removes the old `rules/` directory and moves some rules over that were not present in the jsonnet based approach.

@ironcladlou @elad661 @mxinden


mxinden:
\lgtm

PR #32, opened 2018-06-05
Update to prometheus-operator v0.20.0

ironcladlou:
Update to prometheus-operator v0.20.0 and add various compatibility fixes.


ironcladlou:
Tested this manually in GCP with openshift-ansible

PR #34, opened 2018-06-05
Fix creation of default rules

ironcladlou:
Fix the creation of the default rules. Default rules are now expressed as a
PrometheusRule object, but the manifest was still being unmarshalled into a
ConfigMap causing the creation of an empty ConfigMap rather than a
PrometheusRule.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -190,6 +190,21 @@ func (c *Client) CreateOrUpdatePrometheus(p *monv1.Prometheus) error {
 	return errors.Wrap(err, "updating Prometheus object failed")
 }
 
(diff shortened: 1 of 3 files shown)


ironcladlou:
While testing all the other fixes and compatibility with prometheus-operator v0.20.0 I noticed no default alerting rules were present in my cluster...

Also, quay.io went down while I was testing this patch, so I'm marking it WIP until I know for sure the change works.

ironcladlou:
Tested in GCP, things are looking good now.

brancz:
Great catch, this looks perfect.

/lgtm

PR #37, opened 2018-06-14
Grafana

brancz:
This pull request integrates Grafana into the existing stack. In order for Grafana to authenticate with Prometheus the oauth proxies basic auth functionality is used. A strong random password is auto generated for the stack and can be rotated, by deleting the object that contains it (similar to the serving certs controller for example).

In the future we would like to have Grafana use the users oauth token, in order to use the users identity for tenancy and authZ purposes. For now the authentication story for Grafana is the same as for Prometheus: if the user has the appropriate RBAC permissions, then the user can view all dashboards and metrics of the cluster monitoring stack.

@ironcladlou @elad661 @mxinden


pkg/manifests/manifests.go
```diff
@@ -784,6 +814,187 @@ func (f *Factory) KubeControllersService() (*v1.Service, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	d := &GrafanaDatasources{}
+	err = json.Unmarshal([]byte(c.Data["prometheus.yaml"]), d)
+	if err != nil {
+		return nil, err
+	}
+	d.Datasources[0].BasicAuthPassword, err = GeneratePassword(43)
```
mxinden: Where does the number 43 originate from?

mxinden: There is no max-length specified in the HTTP basic auth spec. This is up to each individual implementation. Apache seems to do [255](https://stackoverflow.com/questions/4027825/is-there-a-maximum-length-for-a-http-basic-authentication-username), so this seems to be a value worth trying.

What do you think @brancz?

I don't think we are going to hit the [HTTP header size limit](https://stackoverflow.com/questions/686217/maximum-on-http-header-values) here.

brancz:
Updated the entire stack once more and testing it out on a cluster now. If all succeeds then I will merge this.

brancz:
Works like a charm. Merging :tada:

PR #38, opened 2018-06-15
assets: Update kube-prometheus dependency

mxinden:
Kube-prometheus recently introduced a regression deploying node-exporter
in the pod network instead of the host network. This updates the
vendored kube-prometheus version in the cluster-monitoring-operator to
include a patch for above described problem.

This is a follow up on https://github.com/coreos/prometheus-operator/pull/1484.


brancz:
This has been superseeded by #37, as that also involved updating the stack. Closing here, but thanks for taking care of the kube-prometheus changes!

PR #39, opened 2018-06-26
Disable the wifi collector in node_exporter

simonpasquier:
This is a similar change as https://github.com/openshift/openshift-ansible/pull/8914

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1593211


brancz:
Thanks for the contribution! :tada: 

There is actually one more step, which is that we synchronize everything we do using [jsonnet](https://jsonnet.org/) from the "upstream" [kube-prometheus](https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus) project (a sub project of the prometheus operator).

tl;dr we need to add `args+: ['--no-collector.wifi']`, here: https://github.com/openshift/cluster-monitoring-operator/blob/b14517fa264e14b8c669002f134c10e8da2a2104/jsonnet/node-exporter.jsonnet#L87

Then you run the following commands (and make sure you have [`jb` installed](https://github.com/jsonnet-bundler/jsonnet-bundler#install)):

```
cd jsonnet && jb install # this installs all jsonnet libraries into `jsonnet/vendor/`
hack/build-jsonnet.sh # from the root of the repo again
```

For some context: This allows us to syncronize everything with all upstream jsonnet code with these last two commands which in the past we have done manually which was a huge time effort every time we had to sync things (like kubernetes manifests, alerting rules, dashboard definitions, etc.).

brancz:
Perfect, thank you so much for taking care of this!

/lgtm

PR #40, opened 2018-06-27
Use openshift/origin-base as base container image

brancz:
This has been tested on a GCE cluster. Aside from changing the base image of the cluster-monitoring-operator this also uses the openshift images for Prometheus, Alertmanager and node-exporter.

@mxinden @ironcladlou


manifests/cluster-monitoring-operator.yaml
```diff
@@ -17,7 +17,7 @@ spec:
     spec:
       serviceAccountName: cluster-monitoring-operator
       containers:
-      - image: quay.io/coreos/cluster-monitoring-operator:v0.0.3
+      - image: quay.io/coreos/cluster-monitoring-operator-dev:b14517f
```
brancz: need to revert this

Dockerfile
```diff
@@ -1,5 +1,13 @@
...
+RUN mkdir $GOPATH
 
-ENTRYPOINT ["/bin/operator"]
+COPY . $GOPATH/src/github.com/openshift/cluster-monitoring-operator
+
+RUN yum install -y golang make && \
+   cd $GOPATH/src/github.com/openshift/cluster-monitoring-operator && \
+   make build && cp $GOPATH/src/github.com/openshift/cluster-monitoring-operator/operator /usr/bin/ && \
+   yum erase -y golang make && yum clean all
+
```
brancz: still clarifying which labels should go here

Dockerfile
```diff
@@ -1,5 +1,21 @@
...
+FROM openshift/origin-base
 
-ADD operator /bin/operator
+ENV GOPATH /go
+RUN mkdir $GOPATH
 
-ENTRYPOINT ["/bin/operator"]
+COPY . $GOPATH/src/github.com/openshift/cluster-monitoring-operator
+
+RUN yum install -y golang make && \
```
mxinden: Why use our production image to build the project?

PR #41, opened 2018-06-29
Update jsonnet and use openshift based images

brancz:
This PR updates all jsonnet dependencies and changes the provided configuration to use the openshift/origin-base based images. For now these images are under my quay user (quay.io/brancz), but once the release team sets up building all of these images for us we will use those from an "official" openshift source.

@mxinden @ironcladlou


jsonnet/main.jsonnet
```diff
@@ -9,6 +9,8 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
                },
                versions+:: {
                  openshiftOauthProxy: 'v1.1.0',
+                 prometheus: 'v2.2.1',
```
mxinden: @brancz do you prefer to not update these quite yet or are they overwritten somewhere else?

brancz: The openshift images have not been updated so we can't use newer versions yet.

PR #43, opened 2018-07-06
*: Add etcd monitoring

brancz:
@mxinden 

cc @ant31


hack/generate-etcd-secret.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+cat <<-EOF
+apiVersion: v1
+data:
+  etcd-client-ca.crt: "$(ssh -i ~/go/src/github.com/openshift/release/cluster/test-deploy/gcp-dev/ssh-privatekey cloud-user@35.237.160.60 sudo -E cat /etc/origin/master/master.etcd-ca.crt | base64 --wrap=0)"
```
mxinden: How about making the IP address an argument to the script?

brancz: Definitely! Good catch.

pkg/tasks/prometheus.go
```diff
@@ -230,6 +232,40 @@ func (t *PrometheusTask) Run() error {
 		return errors.Wrap(err, "reconciling Prometheus kube-controllers ServiceMonitor failed")
 	}
 
+	if t.config.EtcdConfig != nil && t.config.EtcdConfig.Enabled != nil && *t.config.EtcdConfig.Enabled {
```
mxinden: General question: Why do we have the `Enabled` property? Why don't we see etcd as enabled when the `EtcdConfig != nil`?

brancz: You're right let's stay more consistent with the rest of our components.

brancz:
One more re-review @mxinden ? :slightly_smiling_face:

PR #44, opened 2018-07-09
*: Update Prometheus and Alertmanager

brancz:
As rhel based images are now built for these versions we can now use the latest versions of Prometheus and Alertmanager.

@mxinden


mxinden:
@brancz `bindata.go` seems to need updating.

PR #46, opened 2018-07-10
Tag overrides

brancz:
This pull request allows individual tags of respective images to be overridden. Typically the cluster-monitoring-operator handles image tags, but in certain situation they need to be overridden (for example when non-upstream image tags are used, such as it is the case in OCP).

@mxinden 

cc @mrsiano


pkg/manifests/config.go
```diff
@@ -35,14 +35,18 @@ type Config struct {
 }
 
 type PrometheusOperatorConfig struct {
-	BaseImage                string `json:"baseImage"`
-	PrometheusConfigReloader string `json:"prometheusConfigReloaderBaseImage"`
-	ConfigReloaderImage      string `json:"configReloaderBaseImage"`
+	BaseImage                   string `json:"baseImage"`
+	Tag                         string `json:"-"`
```
mxinden: When is this struct ever marshalled to json? Why exclude the `Tag`?

brancz: It's actually for the other way around, so that when this struct is unmarshaled, users can't use the config struct to set the tag. Instead the tag is enforced by the flag. Basically this means that the flag is configured for the lifetime of the version whereas the config can be adapted at runtime.

mxinden: :+1: neat!

PR #47, opened 2018-07-16
Normalize Grafana setup

brancz:
This normalizes the Grafana setup, so Grafana does not depend on any
"default" paths, but instead the only interface for the container image
is that the entrypoint execs a Grafana server binary.

@mxinden @squat


hack/ocp-images.sh
```diff
@@ -0,0 +1,111 @@
...
+    ose-configmap-reloader
+    ose-cluster-monitoring-operator
+    ose-kube-state-metrics
+    ose-kube-rbac-proxy
+    ose-prometheus-config-reloader
+    ose-prometheus-operator
+)
+
+for i in ${IMAGES[@]}; do
+    docker pull $SRC_REGISTRY/$i:$TAG
```
mxinden: `s/$SRC_REGISTRY/$i:$TAG/"{$SRC_REGISTRY}/${i}:${TAG}"`

hack/ocp-images.sh
```diff
@@ -0,0 +1,111 @@
...
+        name: cluster-monitoring-operator
+        args:
+        - "-namespace=openshift-monitoring"
+        - "-configmap=cluster-monitoring-config"
+        - "-logtostderr=true"
+        - "-v=4"
+        - "-tags=prometheus-operator=$TAG"
+        - "-tags=prometheus-config-reloader=$TAG"
+        - "-tags=config-reloader=$TAG"
+        #- "-tags=prometheus=$TAG"
```
squat: Why do we need these commented out args?

hack/ocp-images.sh
```diff
@@ -0,0 +1,111 @@
...
+    ose-prometheus-config-reloader
+    ose-prometheus-operator
+)
+
+for i in ${IMAGES[@]}; do
+    docker pull $SRC_REGISTRY/$i:$TAG
+done
+
+for i in ${IMAGES[@]}; do
+    docker tag $SRC_REGISTRY/$i:$TAG $DST_REGISTRY/$i:$TAG
```
mxinden: See above.

hack/ocp-images.sh
```diff
@@ -0,0 +1,111 @@
...
+done
+
+for i in ${IMAGES[@]}; do
+    docker push $DST_REGISTRY/$i:$TAG
```
mxinden: See above.

hack/ocp-images.sh
```diff
@@ -0,0 +1,111 @@
...
+set -e
+set -x
+# only exit with zero if all commands of the pipeline exit successfully
+set -o pipefail
+
+# This script synchronizes OCP images into a cluster-local registry of an
+# OpenShift cluster, and writes the appropriate configuration into the
+# cluster-monitoring-operator manifests in `manifests/`.
+
+SRC_REGISTRY=registry.reg-aws.openshift.com:443/openshift3
```
squat: do we need to specify 443 here? and why not on `DST_REGISTRY`?

hack/ocp-images.sh
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env bash
+set -e
+set -x
+# only exit with zero if all commands of the pipeline exit successfully
+set -o pipefail
+
+# This script synchronizes OCP images into a cluster-local registry of an
+# OpenShift cluster, and writes the appropriate configuration into the
+# cluster-monitoring-operator manifests in `manifests/`.
```
mxinden: Maybe worth mentioning that it expects the `TAG` environment variable.

hack/ocp-images.sh
brancz: I intentionally did this, as this script syncs all the other images except prometheus, alertmanager, node-exporter and oauth-proxy. But I wanted to leave these here for documentation purposes.

hack/ocp-images.sh
brancz: Good point, let me check if that makes a difference.

brancz: It is not. Updating.

PR #48, opened 2018-07-18
Step templating.

mrsiano:
template the step \ resolutions and change the default value to 15s
to support large scale graphs in reliable timeframe.

@brancz @jeremyeder


brancz:
All the dashboards are synchronized from upstream sources using jsonnet, we would first have to adapt the upstream dashboards and then vendor them into this repository.

PR #50, opened 2018-07-20
config: changed operator config to include images from the coreos org

sichvoge:
Previous config included images from Frederics personal Quay account, this change will switch it to the CoreOS account.


brancz:
Note, this was only a configuration, the code doesn't do this.

/lgtm

PR #52, opened 2018-07-24
*: add basic instrumentation

squat:
This commit implements basic instrumentation for the Cluster Monitoring
Operator. The CMO will report how many errors it has gotten while
attempting to reconcile. The metrics endpoint is exposed on port 8080,
for consistency with the Prometheus Operator.

cc @mxinden @brancz

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -18,12 +18,15 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 	"os/signal"
 	"strings"
 	"syscall"
 
 	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"golang.org/x/sync/errgroup"
 
 	cmo "github.com/openshift/cluster-monitoring-operator/pkg/operator"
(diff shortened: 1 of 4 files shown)


pkg/operator/operator.go
```diff
@@ -120,11 +135,7 @@ func (c *Operator) keyFunc(obj interface{}) (string, bool) {
 	return k, true
 }
 
-func (o *Operator) handleConfigMapAdd(obj interface{}) {
```
squat: this method was actually unused :/

cmd/operator/main.go
```diff
@@ -88,12 +91,20 @@ func Main() int {
 		fmt.Fprint(os.Stderr, "`--configmap` flag is required, but not specified.")
 	}
 
+	r := prometheus.NewRegistry()
+	r.MustRegister(prometheus.NewGoCollector())
```
brancz: I think we might have done the same mistake in the Prometheus Operator, but we should also add the process collector here. https://github.com/prometheus/client_golang/blob/bcbbc08eb2ddff3af83bbf11e7ec13b4fd730b6e/prometheus/registry.go#L55

pkg/operator/operator.go
```diff
@@ -183,8 +194,9 @@ func (o *Operator) processNextWorkItem() bool {
 		return true
 	}
 
+	o.reconcileErrors.Inc()
```
brancz: we should additionally have a counter to track all reconcile attempts, so we can alert on the error rate

cmd/operator/main.go
squat: Sounds good :+1: I'll make sure to add this to Prometheus Operator as well.

pkg/operator/operator.go
squat: :tada:

brancz:
/lgtm

Do you want to do a follow up, and add these metrics to be collected and an alert for when the error rate is too high?

PR #53, opened 2018-07-25
*: create CMO service monitor and alerting rule

squat:
This PR created a service monitor for the Cluster Monitoring Operator as well as an alerting rule for the metrics generated by the CMO. The CMO will create the service monitor after it creates the prometheus operator deployment. The service monitor is defined in jsonnet.

This PR also contributes changes to the Makefile to improve static asset generation.

cc @brancz @mxinden


Makefile
```diff
@@ -1,20 +1,27 @@
...
 NAMESPACE=openshift-monitoring
 KUBECONFIG?=$(HOME)/.kube/config
-PKGS   = $(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
-GOOS   = linux
+PKGS=$(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
+GOOS=linux
 VERSION=$(shell cat VERSION | tr -d " \t\n\r")
+SRC=$(shell find . -type f -name '*.go') pkg/manifests/bindata.go
+GOBINDATA=$(GOPATH)/bin/go-bindata
+ASSETS=$(shell grep -oh 'assets/.*\.yaml' pkg/manifests/manifests.go)
```
squat: We use this heuristic to determine which asset files we should be depending on.

Makefile
```diff
@@ -37,15 +44,20 @@ embedmd:
...
+$(ASSETS): $(JSONNET)
+	./hack/build-jsonnet.sh
 
 generate:
 	docker build -t tpo-generate -f Dockerfile.generate .
-	docker run --rm  --security-opt label=disable -v `pwd`:/go/src/github.com/openshift/cluster-monitoring-operator -w /go/src/github.com/openshift/cluster-monitoring-operator tpo-generate make merge-cluster-roles assets docs
+	docker run --rm --security-opt label=disable -v `pwd`:/go/src/github.com/openshift/cluster-monitoring-operator -w /go/src/github.com/openshift/cluster-monitoring-operator tpo-generate make merge-cluster-roles bindata docs
 
-gobindata:
+gobindata: $(GOBINDATA)
```
squat: this way we can know when we need go-bindata installed and avoid re-generating static assets if go-bindata was already present

Makefile
```diff
@@ -37,15 +44,20 @@ embedmd:
 docs: embedmd
 	embedmd -w `find Documentation -name "*.md"`
 
-assets: gobindata
```
squat: assets was actually a phony target that duplicated the name of an real directory in the repo. this was confusing

assets/kube-state-metrics/cluster-role.yaml
```diff
@@ -34,6 +34,9 @@ rules:
   - apps
   resources:
   - statefulsets
+  - daemonsets
```
squat: updated due to `jb install`

assets/prometheus-k8s/secret-etcd-certs.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
squat: This file was actually depended on by `pkg/manifests/manifests.go` but we were removing it with the `hack/build-jsonnet.sh` script

hack/build-jsonnet.sh
```diff
@@ -51,5 +51,4 @@ rm -rf "assets/prometheus-k8s/service-monitor-core-d-n-s.yaml"
 rm -rf "assets/prometheus-k8s/service-monitor-kube-controller-manager.yaml"
 rm -rf "assets/prometheus-k8s/service-monitor-kube-scheduler.yaml"
 rm -rf "assets/prometheus-k8s/service-monitor-kube-scheduler.yaml"
-rm -rf "assets/prometheus-k8s/secret-etcd-certs.yaml"
```
squat: This file was actually a dependency of `pkg/manifests/manifests.go`

jsonnet/main.jsonnet
```diff
@@ -26,9 +26,11 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
            (import 'kube-state-metrics.jsonnet') +
            (import 'grafana.jsonnet') +
            (import 'alertmanager.jsonnet') +
-           (import 'prometheus.jsonnet') + {
+           (import 'prometheus.jsonnet') +
+           (import 'cluster-monitoring-operator.jsonnet') + {
   _config+:: {
     namespace: 'openshift-monitoring',
+    name: 'cluster-monitoring-operator',
```
squat: This is the name of the project. I am open to suggestions to renaming the field

scripts/deploy-on-openshift.sh
```diff
@@ -24,4 +24,5 @@ oc apply -f manifests/cluster-monitoring-operator-role.yaml
 oc apply -f manifests/cluster-monitoring-operator-role-binding.yaml
 oc apply -f manifests/cluster-monitoring-config.yaml
 oc apply -f manifests/secret-etcd-certs.yaml
-oc apply -f manifests/cluster-monitoring-operator.yaml
\ No newline at end of file
+oc apply -f manifests/cluster-monitoring-operator.yaml
+oc apply -f manifests/cluster-monitoring-operator-service.yaml
```
squat: we need to deploy a service so we can monitor the CMO

Makefile
```diff
@@ -1,20 +1,27 @@
 all: build
 
 APP_NAME=cluster-monitoring-operator
+BIN=operator
```
mxinden: What do we need `BIN` for, if it is only used once?

Makefile
```diff
@@ -37,15 +44,20 @@ embedmd:
 docs: embedmd
 	embedmd -w `find Documentation -name "*.md"`
 
-assets: gobindata
+bindata: pkg/manifests/bindata.go
```
mxinden: How about getting rid of the `bindata` target and just adding `pkg/manifests/bindata.go` as a dependency to `build`?

squat: previously we had a phony target that was used as a shortcut for the `generate` target. this preserves that functionality while using a better name

Makefile
squat: well, we currently use it twice, but we should actually use it in three places. I will update the third use in the `run` target

assets/cluster-monitoring-operator/service-monitor.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app: cluster-monitoring-operator
```
brancz: I don't think this `ServiceMonitor` is actually being picked up: https://github.com/openshift/cluster-monitoring-operator/blob/e111e7b3ce84c40f4e336a53d5af3248c2f4ae03/assets/prometheus-k8s/prometheus.yaml#L71-L74

Makefile
squat: also, for better locality, I think it's preferable to have a single reference to external files so that updating a file name means we don't have to chase down multiple references everywhere, but one single reference. This meas either using another variable or a phony target.

squat: talked offline with @mxinden. we will do a followup PR to further cleanup the Makefile

assets/cluster-monitoring-operator/service-monitor.yaml
brancz: hmm .. not sure we can do that, multiple expressions in that array are treated as a logical `AND`.

squat: updated the service monitor to use the `k8s-app` label. we will do a follow up to update all labels to use the new openshift keys.

scripts/deploy-on-openshift.sh
brancz: maybe we should put this in the task as well? I'd prefer not to have to add this to openshift-ansible or whatever is going to deploy cluster-monitoring in the future

jsonnet/main.jsonnet
brancz: we should nest this in a cluster-monitoring-operator config

Makefile
```diff
@@ -1,21 +1,30 @@
...
 KUBECONFIG?=$(HOME)/.kube/config
-PKGS   = $(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
-GOOS   = linux
+PKGS=$(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
+GOOS=linux
 VERSION=$(shell cat VERSION | tr -d " \t\n\r")
+SRC=$(shell find . -type f -name '*.go') pkg/manifests/bindata.go
+GOBINDATA_BIN=$(GOPATH)/bin/go-bindata
+GOJSONTOYAML_BIN=$(GOPATH)/bin/gojsontoyaml
+JSONNET_BIN=$(if $(shell which jsonnet),$(shell which jsonnet),$(GOPATH)/bin/jsonnet)
```
squat: we need to be able to have jsonnet on travis. here we default to the user's installed jsonnet binary; if nothing is installed, then install go-jsonnet

chancez: Your PR comment could probably be a comment in the Makefile.

Dockerfile.generate
```diff
@@ -1,4 +1,4 @@
 FROM golang:1.9.2
 
 RUN apt-get update && \
-    apt-get install -y python-yaml
+    apt-get install -y python-yaml jq gawk
```
squat: `make generate` never actually worked :/
By default, Debian uses an alternative called `mawk`, which was not compatible with the GNU awk syntax in `hack/build-jsonnet.sh`. CI never caught this until now because the script was never actually triggered because the Make dependencies were funky.

brancz: oops. nice catch!

jsonnet/jsonnetfile.lock.json
```diff
@@ -8,7 +8,7 @@
                     "subdir": "contrib/kube-prometheus/jsonnet/kube-prometheus"
                 }
             },
-            "version": "a24386e2d5e66de353cb0e939936041fa3977283"
+            "version": "5318247b242823e6f45266f678855f0eef215f05"
```
brancz: The lock file now contains all versions of all resolved dependencies. As a hack you can specify this version in the `jsonnetfile.json`, then remove `jsonnetfile.lock.json` and the `vendor/` directory inside `jsonnet/`, and then run `jb install`.

assets/prometheus-k8s/rules.yaml
```diff
@@ -304,6 +314,16 @@ spec:
       for: 15m
       labels:
         severity: critical
+    - alert: ClusterMonitoringOperatorDown
```
mxinden: Where is the jsonnet definition for this alert? I only see `ClusterMonitoringOperatorErrors` defined below.

brancz: From here: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/alerts/absent_alerts.libsonnet

jsonnet/cluster-monitoring-operator.jsonnet
```diff
@@ -0,0 +1,48 @@
+local k = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
+
+{
+  _config+:: {
+    clusterMonitoringOperatorSelector: 'job="cluster-monitoring-operator"',
+    jobs+: {
+      ClusterMonitoringOperator: $._config.clusterMonitoringOperatorSelector,
```
squat: @mxinden the absent job alert is created thanks to this field

squat:
approved by @mxinden. adding lgtm so we can merge

/lgtm

jsonnet/cluster-monitoring-operator.jsonnet
mxinden: Cool, thanks for clarifying @squat!

smarterclayton:
I *think* this broke api.ci prometheus - we deployed image 

```
docker.io/openshift/origin-cluster-monitoring-operator@sha256:a7fff1a8e02c964efa4a64b327bf8a9713398c3e5b9e08cb0c63ad389910b612
```

about 4 hours ago and we've been down since then.  The prometheus instances are not showing any targets (the rules are there, are being parsed, and check config passes, but nothing is listed in service discovery).

The previous image was `docker.io/openshift/origin-cluster-monitoring-operator@sha256:cf6ddffc970084c2c904d9b4a888c26bf42a27f64025603a8f06a32894d90407` which I think was previous master and didn't have this problem.

squat:
@smarterclayton @brancz @mxinden Im quite sure this is the same issue I found where the config is being generated but the scrape_configs section is `[]`, which causes the no targets to show. Clayton can you confirm this?

This is because this PR bumped the version of PO that is deployed. Here is the chart that shows when the issue is triggered:

```
|      x     | CMO master | CMO v0.1.0 | CMO v0.0.6 |
| ---------- | ---------- | ---------- | ---------- |
| PO v0.20.0 |      1     |      1     |      1     |
| PO v0.22.0 |      1     |      1     |      1     |
| PO v0.22.1 |      0     |      0     |      1     |
| PO v0.22.2 |      0     |      0     |      1     |
```

squat:
@smarterclayton the bug is almost certainly in the Prometheus operator, not CMO, since bumping that is what triggers the issue, and since the PO is what is responsible for generating the faulty config in the first place

PR #55, opened 2018-08-07
config: add http configuration

s-urbaniak:
Currently there is no possibility to add top-level HTTP configuration
to the operator.

This fixes it and applies the HTTP configuration to the oauth proxy
sidecar for alertmanager.

Fixes MON-255


pkg/manifests/manifests_test.go
```diff
@@ -276,6 +278,131 @@ func TestUnconfiguredManifests(t *testing.T) {
...
+					if a.Spec.Containers[c].Env[e].Name != name {
+						continue
+					}
+					if a.Spec.Containers[c].Env[e].Value == value {
+						return nil
+					}
+				}
+			}
+
+			return fmt.Errorf("containers expected to have env name %v=%v, but they don't", name, value)
```
squat: should `env name` here actually be `env var`?

s-urbaniak: right :+1: it should actually be `name env=var`, I suggest simply to reduce to `containers expected to have env %v=%v, but they don't`

s-urbaniak: ah, sorry, your suggestion actually makes perfect sense.

pkg/manifests/manifests_test.go
```diff
@@ -276,6 +278,131 @@ func TestUnconfiguredManifests(t *testing.T) {
...
+				}
+			}
+			return nil
+		}
+	}
+
+	hasContainers := func(expected bool) checkFunc {
+		return func(a *monv1.Alertmanager) error {
+			if got := len(a.Spec.Containers) > 0; got != expected {
+				return fmt.Errorf("expected spec to have containers %t, got %t", expected, got)
```
squat: small doubt: what would it mean for the error to say `expected spec to have containers false`? (or `true` for that matter)

s-urbaniak: yes, that would be the meaning ... hmm, do you have a better wording? :thinking:

squat: blah nothing short comes to mind. we could make it a bit more verbose and interpretable, like:
```go
var negative string
if !expected {
    negative = "not"
}
fmt.Errorf(strings.Join(" ", "expected spec", negative, "to have containers, got %d containers"), len(a.Spec.Containers))
```

But this is not necessary IMO

s-urbaniak: suggestion: as I am never asserting `hasContainers(false)` I can omit the parameter, always expecting `true`, which makes it more reasonable for now. Let's go with that.

```
	hasContainers := func() checkFunc {
		return func(a *monv1.Alertmanager) error {
			if len(a.Spec.Containers) == 0 {
				return errors.New("expected spec to have containers, but it doesn't have any")
			}
			return nil
		}
	}
```

PR #56, opened 2018-08-07
Bump Prometheus Operator and Kube-Prometheus

squat:
This PR bumps the kube-prometheus jsonnet dependencies as well as the prometheus operator go dependencies to regenerate the manifests and incorporate the fixes that went into prometheus operator 0.23.0.

Blocked by: https://github.com/coreos/prometheus-operator/pull/1762


manifests/cluster-monitoring-operator-role.yaml
```diff
@@ -10,9 +10,7 @@
 # 	assets/node-exporter/cluster-role.yaml
 # 	assets/prometheus-k8s/cluster-role.yaml
 # 	assets/prometheus-k8s/role-config.yaml
-# 	assets/prometheus-k8s/role-default.yaml
-# 	assets/prometheus-k8s/role-kube-system.yaml
-# 	assets/prometheus-k8s/role-namespace.yaml
+# 	assets/prometheus-k8s/role-specific-namespaces.yaml
```
squat: This file is a list of roles, so we need special logic in `hack/merge_cluster_roles.py` to treat it correctly.

brancz: We also need to handle creating the list of roles and rolebindings now. Previously we handled them individually.

pkg/tasks/prometheus.go
```diff
@@ -112,26 +112,6 @@ func (t *PrometheusTask) Run() error {
 		return errors.Wrap(err, "reconciling Prometheus ClusterRoleBinding failed")
 	}
 
-	rd, err := t.factory.PrometheusK8sRoleDefault()
```
brancz: We need to add creating/updating the role/rolebinding list here.

PR #61, opened 2018-08-13
Dockerfile: install jq from Github

squat:
This commit modifies the Dockerfile to download the jq binary directly
from the Github release page rather than relying on the `epel-release`
repository, which is not available when building OCP images.

cc @brancz 

Tested `make container` locally and everything builds correctly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,11 +6,11 @@ RUN mkdir $GOPATH
 
 COPY . $GOPATH/src/github.com/openshift/cluster-monitoring-operator
 
-RUN yum install -y epel-release && \
-    yum install -y golang make git jq && \
+RUN curl -L https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64 -o /usr/local/bin/jq && chmod +x /usr/local/bin/jq && \
+    yum install -y golang make git && \
     cd $GOPATH/src/github.com/openshift/cluster-monitoring-operator && \
     make build && cp $GOPATH/src/github.com/openshift/cluster-monitoring-operator/operator /usr/bin/ && \
-    yum erase -y golang make git jq && yum remove -y epel-release && yum clean all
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,11 +6,11 @@ RUN mkdir $GOPATH
 
 COPY . $GOPATH/src/github.com/openshift/cluster-monitoring-operator
 
-RUN yum install -y epel-release && \
-    yum install -y golang make git jq && \
+RUN curl -L https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64 -o /usr/local/bin/jq && chmod +x /usr/local/bin/jq && \
```
brancz: I'm trying to follow the `build` target in the Makefile and I don't see where we need `jq` without also needing `jsonnet`, so I think we should be good removing it.

squat: The way the dockerfile is currently set up, all building (jsonnet, manifests, bindata, binary) occurs during the docker build. Here is the dependency graph that necessitates jq:
binary > bindata > manifests > go-jsonnet
Because there is no go-jsonnet in the Docker container, make installs it, meaning the jsonnet binary is now newer than the rest of the chain, so the manifests get recompiled, which require jq.

If we want to avoid rebuilding all generated assets inside the Docker container build process, then there are a few options:

1. `make build && make container` don’t build the binary inside the container but simply inside CI, then COPY the binary into the container.

2. Remove binary installation targets from Makefile, then we can add a PHONY target that installs all needed and missing binaries before running `make build`, e.g. `make dependencies`

brancz: yeah, I think we should just ensure the bindata generation on CI, not at build time, then we won't have this problem

squat:
closing in favor of https://github.com/openshift/cluster-monitoring-operator/pull/62

PR #66, opened 2018-08-14
pkg/manifests: add nodeSelector to prometheus operator config

s-urbaniak:
Currently, the node selector in the prometheus operator config is
ignored.

This fixes it
Fixes MON-270

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -831,6 +831,10 @@ func (f *Factory) PrometheusOperatorDeployment() (*appsv1.Deployment, error) {
 		return nil, err
 	}
 
+	if len(f.config.PrometheusOperatorConfig.NodeSelector) > 0 {
+		d.Spec.Template.Spec.NodeSelector = f.config.PrometheusOperatorConfig.NodeSelector
+	}
+
 	if f.config.PrometheusOperatorConfig.BaseImage != "" {
 		image, err := imageFromString(d.Spec.Template.Spec.Containers[0].Image)
 		if err != nil {
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -41,12 +41,13 @@ type HTTPConfig struct {
 }
 
 type PrometheusOperatorConfig struct {
-	BaseImage                   string `json:"baseImage"`
-	Tag                         string `json:"-"`
-	PrometheusConfigReloader    string `json:"prometheusConfigReloaderBaseImage"`
-	PrometheusConfigReloaderTag string `json:"-"`
-	ConfigReloaderImage         string `json:"configReloaderBaseImage"`
-	ConfigReloaderTag           string `json:"-"`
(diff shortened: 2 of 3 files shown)


s-urbaniak:
setting WIP as I do want to test this manually.

/cc @brancz it seems jsonnet does create a default nodeSelector:
https://github.com/openshift/cluster-monitoring-operator/blob/f44aad55f779260c4b8f3ea811ca9a28c3253b04/assets/prometheus-operator/deployment.yaml#L34-L35

While looking at the file https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/prometheus-operator.jsonnet it seems pretty minimalistic, inheriting container properties from `super.containers`. Do you mind to give me a pointer, where the `super` root is declared, so I can reason about this very inheritance?

brancz:
`super` simply comes from the "previous" declaration, which is just the kube-prometheus files: 

https://github.com/coreos/prometheus-operator/blob/356dce265f3cbc6a66ba121711410d8c36931973/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus.libsonnet#L8

Which in turn imports the prometheus-operator jsonnet files, which sets the nodeSelector here:

https://github.com/coreos/prometheus-operator/blob/356dce265f3cbc6a66ba121711410d8c36931973/jsonnet/prometheus-operator/prometheus-operator.libsonnet#L137

squat:
@s-urbaniak the `containers` field is defined upstream in the Prometheus operator jsonnet code: 
https://github.com/coreos/prometheus-operator/blob/master/jsonnet/prometheus-operator/prometheus-operator.libsonnet#L117-L133.

s-urbaniak:
@brancz @squat ahhhh ... the light bulbs are turning on, thank you! :+1:

s-urbaniak:
ok, tested successfully :heavy_check_mark: 
@squat @brancz PTAL

pkg/manifests/manifests.go
```diff
@@ -831,6 +831,10 @@ func (f *Factory) PrometheusOperatorDeployment() (*appsv1.Deployment, error) {
 		return nil, err
 	}
 
+	if len(f.config.PrometheusOperatorConfig.NodeSelector) > 0 {
```
squat: Just curious, why do the `if` check at all?

s-urbaniak: (I am open to suggest we have an overwrite-always mode, but then we should change it consistently.

s-urbaniak:
just a test proof/reference point:
```
$ kubectl -n openshift-monitoring get pod prometheus-operator-5ff8d64f98-wt6z8 -o yaml | grep -A2 nodeSelector
  nodeSelector:
    node-role.kubernetes.io/infra: "true"
  priority: 0
```

pkg/manifests/manifests.go
squat: :+1:

PR #69, opened 2018-08-15
Makefile: container build operator with deps

squat:
This commit allows the Docker build flow to compile the operator binary without any Make dependencies. This means the compilation during container build time depends exactly on the static assets that are checked into the repository and will not regenerate any assets if they are actually out of date.

cc @brancz


brancz:
Can you add a new job on travis ci to make sure we won't run into a problem again with building the container from a fresh clone?

PR #70, opened 2018-08-15
prometheus-operator: Update to v0.23.1

mxinden:


assets/prometheus-k8s/rules.yaml
```diff
@@ -899,7 +899,7 @@ spec:
       for: 10m
       labels:
         severity: warning
-    - alert: PrometheusTargetScapesDuplicate
```
squat: good catch here!

PR #73, opened 2018-08-17
add recording rule for build metrics

bparees:


bparees:
@brancz per our discussion the other day, I think this is what you proposed.  not sure if you wanted me to put this rule in the common file, or if we should be creating a second rules file for "openshift" metric rules.

brancz:
That file is actually generated, so you need to put your recording rule into `jsonnet/rules.jsonne`, like this:

https://github.com/openshift/cluster-monitoring-operator/blob/87f2270982a8ff8753f37856ff394715495d850f/jsonnet/rules.jsonnet#L43-L46

bparees:
thanks @brancz, fixed.

bparees:
@brancz bump

PR #75, opened 2018-08-21
Update jsonnet dependencies

brancz:
Among others, this most notably pulls in the changes from https://github.com/coreos/prometheus-operator/pull/1812

@mxinden @squat @s-urbaniak


brancz:
@squat could you give this another review? (I just removed some Makefile changes that I thought were for the better but I'll keep them for a follow up as it broke the build)

PR #82, opened 2018-08-24
WIP *: Add metering service monitors

chancez:
This adds metering service monitors to the cluster monitoring operator.
Currently this is a WIP because I need to add some rolebindings to allow
CMO to monitor the openshift-metering namespace, but I wasn't 100% sure
if that even happens in this repo. I worry if these service monitors are
created when metering isn't installed it might cause a lot of errors, so
having it created by the openshift-metering install might also be
problematic.


mxinden:
Related PR: https://github.com/openshift/cluster-monitoring-operator/pull/14

@chancez could these also be created by the metering operator?

chancez:
@mxinden Potentially, if cluster monitoring is currently capable of discovering service monitors outside of it's namespace. This needs to work for 3.11 though.

chancez:
@mxinden I have no idea how the operator works, so I can't say. If that's what you want, I can do it. I worry about the operator potentially blocking on RBAC creation before it proceeds to other tasks. This was an issue I've ran into in the past. Can you verify that won't be an issue?

mxinden:
The Cluster Monitoring Operator runs all its tasks in sequence. Hence if the `openshift-metering` namespace does not exist the Operator blocks, waiting for the namespace before creating Alertmanager, Node Exporter or Kube State Metrics (See [`RunAll()`](https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/tasks/tasks.go#L36)).

While this is not perfect, eventually the stack should come up correctly.

@brancz what are your thoughts on this?

chancez:
That won't work then, because metering isn't installed by default, it's optional.

brancz:
I realize this is not ideal, but how about the metering ansible roles create the rolebinding? This repository would just deploy the ServiceMonitor into the openshift-monitoring namespace, then and we wouldn't have this problem. For 4.0 we can then move everything to the metering namespace.

PR #96, opened 2018-09-10
Fix cluster monitoring operator alert

brancz:
I noticed that on one of our clusters the cluster monitoring operator was failing on every reconcile, but the alert still didn't trigger so I looked into it and it was because the errors were not occurring often enough as a reconcile backs off after failures, which produced this graph:

screenshot from 2018-09-10 16-31-44

With these changes it does work correctly: 

screenshot from 2018-09-10 17-33-39


squat:
@brancz were you able to fix the CMO configuration on the cluster in question so the operator no longer failed to reconcile?

assets/prometheus-k8s/rules.yaml
```diff
@@ -511,8 +511,8 @@ spec:
       annotations:
         message: Cluster Monitoring Operator is experiencing {{ printf "%0.0f" $value
           }}% errors.
-      expr: sum(rate(cluster_monitoring_operator_reconcile_errors_total[5m])) * 100
-        / sum(rate(cluster_monitoring_operator_reconcile_attempts_total[5m])) > 10
+      expr: sum(rate(cluster_monitoring_operator_reconcile_errors_total[15m])) * 100
+        / sum(rate(cluster_monitoring_operator_reconcile_attempts_total[15m]))
```
squat: @brancz don't we need to compare this value to something? Or are we simply testing that it is non-zero, in which case we don't need to turn it into a percentage, no?

brancz: copy and paste mistake, you're right

brancz: Good catch! Fixed that.

squat: :+1:

PR #103, opened 2018-09-21
add Telemeter client

squat:
This pull request adds the Telemeter client to the Cluster Monitoring Operator stack.

One of the key details to note about the Telemeter client is that its deployment requires a secret containing the cluster's pull secret and ID. These values are only available for 4.x clusters created with github.com/openshift/installer. That is, trying to run the CMO on a non-4.x cluster will result in the deployment of a Telemeter client with an invalid secret, meaning it will not be able to authenticate against the Telemeter server. Specifically, the cluster ID and pull secret can be found in a ConfigMap in the `kube-system` namespace named `cluster-config-v1`, e.g. [0].

Once the URL for the production Telemeter server is known, this pull request must be followed by a PR to the github.com/openshift/telemeter repo to set the default Telemeter server URL and another PR to this repo to bump the Telemeter client jsonnet dependency and regenerate the manifests.

cc @brancz @s-urbaniak 

[0] https://github.com/openshift/installer/blob/master/installer/pkg/config-generator/fixtures/kube-system.yaml


brancz:
@squat could you elaborate on what is still work-in-progress here? Just looking at the code I would say tollbooth token handling, could you just confirm that? :slightly_smiling_face:

squat:
@brancz tollbooth token handling is a big one. I am waiting for input from the installer team for details on obtaining the token components. The second piece is the generation of the actual telemeter client manifests to deploy with the CMO. I am currently writing jsonnet in the telemeter repo to generate the manifests and will then import that dependency in this repo.

squat:
Once https://github.com/openshift/telemeter/pull/25 is in, this PR can vendor the Telemeter client jsonnet and render the manifests. The outstanding work is still identifying the source of the tollbooth authentication token and cluster ID.

assets/telemeter-client/deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+        k8s-app: telemeter-client
+    spec:
+      containers:
+      - command:
+        - /usr/bin/telemeter-client
+        - --id=$(ID)
+        - --from=https://prometheus-k8s.openshift-monitoring.svc:9091
+        - --from-ca-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
+        - --from-token-file=/var/run/secrets/kubernetes.io/serviceaccount/token
+        - --to=
```
brancz: this should have a value no? (either here or in the code below, but I don't see it in either)

squat: this is what I was referring to in the PR description when I spoke about follow ups. We need to determine the production URL for the telemeter server; once we have that we'll configure it in the telemeter jsonnet and regenerate the manifests here. Does that sound good?

brancz: Ack. My feeling is that this will need to be configurable in some way (maybe a flag on the cluster-monitoring-operator?).

squat: Yes this sounds completely reasonable. We need to be able to modify CMO deployments for testing, production, etc and be able to report to different telemeter servers. Either in a flag or as another field in the CMO configmap. I'll plumb the URL into a an environment variable

pkg/manifests/manifests.go
```diff
@@ -1426,6 +1431,68 @@ func (f *Factory) NewClusterRole(manifest io.Reader) (*rbacv1beta1.ClusterRole,
...
+
+// TelemeterClientSecret generates a new Secret for Telemeter client.
+func (f *Factory) TelemeterClientSecret() (*v1.Secret, error) {
+	s, err := f.NewSecret(MustAssetReader(TelemeterClientSecret))
+	if err != nil {
+		return nil, err
+	}
+
+	s.StringData["id"] = base64.StdEncoding.EncodeToString([]byte(f.config.TelemeterClientConfig.ClusterID))
+	s.StringData["token"] = base64.StdEncoding.EncodeToString([]byte(f.config.TelemeterClientConfig.PullSecret))
```
s-urbaniak: :+1: this resolves the secret mystery, thanks! :-)

assets/telemeter-client/deployment.yaml
s-urbaniak: ack :+1:

s-urbaniak:
LGTM from my side so far!

pkg/manifests/manifests.go
s-urbaniak: I wonder if we should rename this also to `pullSecret` on telemeter side. On second thought, this is fine'ish though, as this token is pretty much opaque to telemeter.

pkg/operator/operator.go
```diff
@@ -271,5 +274,27 @@ func (o *Operator) Config() *manifests.Config {
 		return manifests.NewDefaultConfig()
 	}
 
+	cmap, err = o.client.KubernetesInterface().CoreV1().ConfigMaps("kube-system").Get("cluster-config-v1", metav1.GetOptions{})
```
s-urbaniak: I feel we should make aware that storing the pull secret in a configmap might not be the best place?!

pkg/manifests/manifests.go
brancz: If you just set Data to the byte slice of what you want the marshaling will automatically encode it as base64. I generally prefer that as i always forget which base64 encoding is used :)

squat: TIL!!!

s-urbaniak: very true, good catch!

pkg/operator/operator.go
brancz: Agreed

assets/prometheus-k8s/rules.yaml
```diff
@@ -522,14 +522,6 @@ spec:
       record: build_error_rate
   - name: kubernetes-absent
     rules:
-    - alert: AlertmanagerDown
```
brancz: This doesn’t look right. I think we accidentally didn’t append the telemetry Job but instead overwrote the whole array.

squat: This is what I was thinking. I’ll fix this in the telemeter repo

pkg/operator/operator.go
squat: Absolutely, we should raise this in openshift/installer

squat:
needs another bump after https://github.com/openshift/telemeter/pull/29 merges

pkg/manifests/manifests.go
```diff
@@ -1426,6 +1431,69 @@ func (f *Factory) NewClusterRole(manifest io.Reader) (*rbacv1beta1.ClusterRole,
...
+// TelemeterClientSecret generates a new Secret for Telemeter client.
+func (f *Factory) TelemeterClientSecret() (*v1.Secret, error) {
+	s, err := f.NewSecret(MustAssetReader(TelemeterClientSecret))
+	if err != nil {
+		return nil, err
+	}
+
+	s.Data["id"] = []byte(f.config.TelemeterClientConfig.ClusterID)
+	s.Data["to"] = []byte(f.config.TelemeterClientConfig.TelemeterServerURL)
+	s.Data["token"] = []byte(f.config.TelemeterClientConfig.Token)
```
s-urbaniak: :ok_hand:

pkg/operator/operator.go
```diff
@@ -271,5 +274,47 @@ func (o *Operator) Config() *manifests.Config {
...
+		}
+	}
+	if c.TelemeterClientConfig.Token == "" {
+		ps := struct {
+			Auths struct {
+				COC struct {
+					Auth string `json:"auth"`
+				} `json:"cloud.openshift.com"`
+			} `json:"auths"`
+		}{}
```
s-urbaniak: /cc @smarterclayton this is the current parsing logic for the pull secret

s-urbaniak:
looks super fine to me :tada:

pkg/operator/operator.go
smarterclayton: ok, can we make sure tollbooth starts generating the secret in this form asap?

PR #108, opened 2018-09-28
jsonnet: remove duplicate alertmanager down alert

squat:
This PR bumps the kube-prometheus jsonnet dep to remove the duplicat AlertmanagerDown alert and regenerates assets.

cc @sichvoge @brancz 

targeted for 3.11.z


squat:
Hold until 3.11 release is out.

squat:
@mxinden @metalmatze as discussed OOB regarding https://github.com/openshift/cluster-monitoring-operator/pull/117, it is not possible to apply this patch to remove the one duplicate AlertmanagerDown alert without bumping all of kube-prometheus. If we really want this patch in 3.11.z, then it may mean we have to bump it entirely

mxinden:
I would vote for not including this patch in Openshift 3.11, what are your thoughts @metalmatze @squat?

metalmatze:
As it only removes a duplicate I don't either see a reason to patch this.

squat:
Good. It sounds like we all agree that this should not be applied to 3.11.z, since applying the patch produces too large of a footprint.

PR #115, opened 2018-10-02
jsonnet: Pin versions for OpenShift images

metalmatze:
For the images that OpenShift builds itself, we should make sure that the images actually exist before upgrading. Hence relying on the upstream dependencies is unsafe, as we might not yet have a Prometheus v2.4.2 available.

In the future this should be catched by the CI nonetheless. 

/cc @mxinden @squat @s-urbaniak @simonpasquier


simonpasquier:
:+1:

PR #117, opened 2018-10-04
Add antiaffinity to Prometheus & Alertmanager

metalmatze:
This will set the anti affinity for Prometheus & Alertmanager.
https://bugzilla.redhat.com/show_bug.cgi?id=1635802

/cc @mxinden @squat

### DIFF
--- a/assets/alertmanager/alertmanager.yaml
+++ b/assets/alertmanager/alertmanager.yaml
@@ -6,6 +6,20 @@ metadata:
   name: main
   namespace: openshift-monitoring
 spec:
+  affinity:
+    podAntiAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+      - podAffinityTerm:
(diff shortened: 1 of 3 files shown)


squat:
i'll manually make a cherry-pick pr @mxinden

mxinden:
Thanks @squat. Can you add the `do-not-merge` label for now?

squat:
@mxinden @metalmatze the cherry-pick is actually a little more sensitive. Adding this new libsonnet means bumping kube-prometheus, which in turn bumps several assets, including the prometheus server version to 2.4.2. To avoid this, I propose adding a new dep to the jsonnetfile to pull in kube-prometheus at the correct version but only for the anti-affinity libsonnet. This would mean having two kube-prometheus dependencies in the jsonnet, each at a different version (and with a different name). Thoughts?

PR #118, opened 2018-10-04
*: Add antiaffinity to Prometheus & Alertmanager

squat:
this is a manual patch of https://github.com/openshift/cluster-monitoring-operator/pull/117 for 3.11

cc @mxinden @metalmatze


jsonnet/main.jsonnet
```diff
@@ -1,4 +1,5 @@
 local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
+           (import 'anti-affinity/kube-prometheus-anti-affinity.libsonnet') +
```
mxinden: Can you add a comment here why we are doing it?

mxinden:
Could you add some details to the commit message why we are doing the hack via the second jsonnetfile entry?

squat:
updated the commit message and added a note in the main.jsonnet

brancz:
I'm not sure how significant this is but this changes our requirement of 1 infra node to at least 3 infra nodes (meaning things like minishift will never work entirely correct). I'm expecting this to cause bugs to be opened, but I do prefer having this. So what do you say we make the replicas of alertmanager/prometheus configurable when a bug does get opened about this? Then for those types of clusters we can recommend setting the replicas to 1 for Prometheus and Alertmanager.

PR #77, opened 2018-08-22
Add Openshift dashboard jsonnet.

rockash:
The Openshift dashboard displays metrics from the Openshift control plane involving the API server, Openshift volumes, SDN, builds.

Apart from the dashboards added from the kubernetes-mixin project, this dashboard is an addition to show metrics just from the control plane.

The metrics displayed from this dashboard have been added after analyzing the need from this document -> [Dashboard Gaps openshift-monitoring](https://docs.google.com/document/d/1wkXm7huHOIPkv9u99aFqeWk-Cfmd18fzYR7drIxoUvM/edit)

@mrsiano @ashishkamra @mxinden

PR #119, opened 2018-10-04
Introduce openshift dashboard

mrsiano:
this is a follow up PR to #77


mrsiano:
the guy who write the original PR has left RH already, and do want to merge it

PR #121, opened 2018-10-09
Add option to disable Telemeter

squat:
This PR adds an option to the Cluster Monitoring Operator
configuration file to disable the Telemeter client. Disabling the
Telemeter client will not just stop reconciling the deployment or set
the replica count to zero but rather will delete all related resources.

Note: The Cluster Monitoring Operator needs to be able to delete Service
objects in order to delete all Telemeter resources.

cc @s-urbaniak @mxinden @metalmatze


squat:
ping @s-urbaniak

PR #122, opened 2018-10-10
stop modifying etcd service

deads2k:
The etcd operator owns the service for etcd.  For now, that job is being done by the installer (see https://github.com/openshift/installer/pull/431), but this cluster-monitoring-operator is overwriting the values the fake etcd-operator is setting.  The selector in particular, but the entire service in general.

This causes the endpoints to be re-computed, which clears the values laid down the fake etcd-operator.  Which in turn causes the the openshift-apiserver (and eventually the kube-apiserver) to lose their connection to etcd.

See https://github.com/openshift/installer/issues/448 for the discovery of the problem.

@abhinavdahiya @mfojtik @sttts @derekwaynecarr


squat:
Discussed offline, I will add additional patches to finish the PR. Thanks @deads2k for tracking this down and the for initial changeset!

xref: https://github.com/openshift/installer/pull/450

sttts:
Sgtm.

squat:
this is ready for review cc @mxinden @s-urbaniak

pkg/manifests/config.go
```diff
@@ -211,6 +201,9 @@ func (c *Config) applyDefaults() {
 	if c.TelemeterClientConfig.BaseImage == "" {
 		c.TelemeterClientConfig.BaseImage = "quay.io/openshift/origin-telemeter"
 	}
+	if c.EtcdConfig == nil {
+		c.EtcdConfig = &EtcdConfig{}
+	}
```
s-urbaniak: :+1:

pkg/manifests/manifests.go
```diff
@@ -726,7 +682,7 @@ func (f *Factory) PrometheusK8s(host string) (*monv1.Prometheus, error) {
 		}
 	}
 
-	if f.config.EtcdConfig == nil {
+	if f.config.EtcdConfig.Enabled != nil && *f.config.EtcdConfig.Enabled == false {
```
s-urbaniak: that's a lot of duplicated `f.config.EtcdConfig.Enabled != nil && *f.config.EtcdConfig.Enabled == false` checks, can we implement this via a `nil` aware pointer receiver method on `EtcdConfig`?

squat: yeah, I was thinking that earlier. I just added it now :)

squat:
tests failed with `error: unable to check your credentials - pass --skip-check to bypass this error: Get https://registry.svc.ci.openshift.org/v2/: net/http: TLS handshake timeout`.

/retest

deads2k:
Thanks guys.

PR #123, opened 2018-10-10
Clarified the description for the Prometheus Operator

sichvoge:
The README falsely stated that someone could create their own Prometheus instances using the Prometheus Operator. That is not the case with OCP 3.11.


README.md
```diff
@@ -10,9 +10,7 @@ It contains the following components:
 * [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics)
 * [node_exporter](https://github.com/prometheus/node_exporter)
 
-The deployed Prometheus Operator is meant to be leveraged by users to easily deploy new Prometheus setup for their application monitoring.
-The Prometheus instance (`prometheus-k8s`) is responsible for monitoring and alerting on cluster and OpenShift components. It should not be extended to monitor user applications.
-Alertmanager is a cluster-global component for handling alerts generated by all Prometheus instances deployed in that cluster.
+The deployed Prometheus Operator is meant to manage the lifecycle for all Prometheus and Alertmanager instances part of the cluster monitoring stack setup. The Prometheus instances (starting with `prometheus-k8s`) are responsible for monitoring and alerting on cluster and OpenShift components. It should not be extended to monitor user applications. Alertmanager is a cluster-global component for handling alerts generated by all Prometheus instances deployed in that cluster.
```
squat: * should be "for all Prometheus and Alertmanager instances __that are part of__ part of".
* I think we can omit "setup" so it is just "part of the cluster monitoring stack setup."
* right now the `It` in "It should not" does not agree with the plural instances from the previous sentence; I think we can revert this sentence for now, there is only one Prometheus and don't have to talk in the plural yet.

sichvoge:
Do we still need this PR? It seems that the text has been updated already.

PR #131, opened 2018-10-18
Use golang builder

sosiouxme:
This pattern will work better for upstream and production builds.


squat:
/lgtm

thanks @sosiouxme

PR #132, opened 2018-10-18
Add route, service and proxies for metrics in Console

metalmatze:
This adds the [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) and the [prom-label-proxy](https://github.com/openshift/prom-label-proxy) as sidecars to the prometheus-k8s pod.
The Console will talk a Route that will forward the traffic to a Service that then fowards it through these proxies to Prometheus, making sure that the Console receives only relevant and secured data.

This is an additional ingress to Prometheus for this specific use case.

/cc @brancz @squat


metalmatze:
I've tested this on my minishift cluster with curl:
```
curl -s -k -H "Authorization: Bearer $TOKEN" 'https://localhost:9092/api/v1/query?query=up&namespace=default' | jq .
```

The `$TOKEN` was base64 decoded from a valid ServiceAccount.

jsonnet/prometheus.jsonnet
```diff
@@ -54,6 +54,27 @@ local namespacesRole = policyRule.new() +
       },
     },
 
+    routeConsole:{
```
brancz: the openshift console actually proxies the requests into the cluster, so we don't need this route

jsonnet/prometheus.jsonnet
```diff
@@ -82,6 +103,14 @@ local namespacesRole = policyRule.new() +
       service.mixin.spec.withType('ClusterIP') +
       service.mixin.spec.withPorts(servicePort.newNamed('web', 9091, 'web')),
 
+    serviceConsole+:
```
brancz: I think we should just add an additional `tenancy` port to the existing service

jsonnet/prometheus.jsonnet
```diff
@@ -313,6 +366,32 @@ local namespacesRole = policyRule.new() +
                 },
               ],
             },
+            {
+              name: 'kube-rbac-proxy',
+              image: 'quay.io/brancz/kube-rbac-proxy-dev:v0.3.1-a12535f',
```
brancz: this needs an update with the now available v0.4.0 release

jsonnet/main.jsonnet
```diff
@@ -20,6 +20,7 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
                  prometheus: 'openshift/prometheus',
                  alertmanager: 'openshift/prometheus-alertmanager',
                  nodeExporter: 'openshift/prometheus-node-exporter',
+                 promLabelProxy: 'quay.io/openshift/prom-label-proxy',
```
brancz: this needs to be updated

metalmatze:
I addressed your earlier comments and squashed all commits.

assets/kube-state-metrics/deployment.yaml
```diff
@@ -21,7 +21,7 @@ spec:
         - --upstream=http://127.0.0.1:8081/
         - --tls-cert-file=/etc/tls/private/tls.crt
         - --tls-private-key-file=/etc/tls/private/tls.key
-        image: quay.io/coreos/kube-rbac-proxy:v0.3.1
+        image: quay.io/brancz/kube-rbac-proxy:v0.4.0
```
brancz: pushed images to the coreos repo :slightly_smiling_face:

jsonnet/prometheus.jsonnet
```diff
@@ -313,6 +340,32 @@ local namespacesRole = policyRule.new() +
                 },
               ],
             },
+            {
+              name: 'kube-rbac-proxy',
+              image: $._config.imageRepos.kubeRbacProxy + ':' + $._config.versions.kubeRbacProxy,
+              args: [
+                '--secure-listen-address=0.0.0.0:9092',
+                '--upstream=http://127.0.0.1:9095',
+                '--resource-attributes-file=/etc/kube-rbac-proxy/resource-attributes.yaml',
```
brancz: this flag is renamed in v0.4.0 to `--config-file`

jsonnet/prometheus.jsonnet
```diff
@@ -156,6 +159,29 @@ local namespacesRole = policyRule.new() +
       secret.mixin.metadata.withNamespace($._config.namespace) +
       secret.mixin.metadata.withLabels({ 'k8s-app': 'prometheus-k8s' }),
 
+    kubeRbacProxySecret:
+      local config = {
+        'resource-attributes.yaml': std.base64(std.manifestYamlDoc({
```
brancz: as this now contains more than resource attribute configuration, let's rename this to `config.yaml`

metalmatze:
@brancz Fixed the wrong image repositories and updated the config to be called `config.yaml` now.

brancz:
@metalmatze looks like there was already another merge that caused a conflict :S

brancz:
if merged in time, maybe we can include https://github.com/coreos/prometheus-operator/pull/2055 in this PR?

metalmatze:
It was a conflict in `bindata.go`. :grin: 
I've also updated the kube-prometheus dependency to the latest commit on its master branch.

brancz:
Nice job! :+1: 

/lgtm

PR #133, opened 2018-10-20
improved operator to be able to run isolated 

FlorinPeter:
The goal of this PR is to allow this operator to run in a different namespace then openshift-monitoring to allow multiple installations on a single cluster.

The changes keep backward compatibility.

New features:
- Allow additionalScrapeConfigs in PrometheusK8sConfig
- Allow to disable default rules
- Allow to disable node-exporter
- Allow to disable kube-state-metrics


FlorinPeter:
@brancz is there a chance to get this PR reviewed and merged?

brancz:
There are lots of changes here that are not really desirable within the current scope of this project (changing the RBAC roles required to a much less privileged scope for example). The stack is always meant to be run in the `openshift-monitoring` namespace, the only reason it's configurable, is because it used to be a part of CoreOS Tectonic, and therefore was part of the `tectonic-system` namespace, and to avoid having to rewrite the stack again in case it's necessary we made it configurable.

That said the larger goal is to eventually have a more general stack that would be able to create "monitoring stacks", and then the cluster-monitoring-operator would just be a consumer of that. While this PR goes in the direction of that, it takes architectural choices that I'm not comfortable with at this point, as it mixes the above ideas, but at the same time couples itself tightly.

Allowing additoinal scrape configs is a non goal for the cluster-monitoring-operator, as that would allow users to configure plain Prometheus configuration preventing us from delivering upgrade guarantees (because users could use features that are broken in subsequent Prometheus releases).

While I understand what you want to achieve and I'm on your side regarding the larger goal, I don't think we can accept this PR as is.

FlorinPeter:
Hi @brancz,
thx for your reply and you are right about the larger goal ;)
Currently I am working on a operator that is able to create multiple monitoring namespaces based on this cluster-monitoring-operator and also allows via additional scrape config to federate only namespace relevant metrics.
I am a bit lost with my next steps as I see here a lot of good work in the right direction and I want to avoid forking this repo because most of the work is done and really useful. 
Maybe we can have a call about architectural choices and future plans?

brancz:
I took some time to think about this, and I think I have thought of a productive way forward. We can start a new controller that provisions a monitoring stack (prometheus, alertmanager, grafana) and use it internally. That way we can figure out the API to configure this stack without making it a public API immediately, but make progress on extracting this provisioning logic. What do you think?

FlorinPeter:
Hi @brancz,
sounds like a good idea to push this topic forward. I am willing to invest some time on this topic but need somehow a stating point as I think you may have already a structure in mind?

PR #135, opened 2018-10-26
bump Telemeter Client

squat:
commit 1:
This commit bumps the jsonnet dependencies in order to bring in updates
to the Telemeter client. Notably, the Telemeter client now has a default
URL configured for the Telemeter server. This is configured by default
to the production URL. This can always be overridden using the Cluster
Monitoring Operator ConfigMap.

commit 2:
This commit integrates the new Telemeter client ClusterRoleBinding to
allow it to use the `cluster-monitoring-view` ClusterRole.

cc @brancz @s-urbaniak


pkg/manifests/manifests.go
```diff
@@ -1410,6 +1411,17 @@ func (f *Factory) TelemeterClientClusterRoleBinding() (*rbacv1beta1.ClusterRoleB
 	return crb, nil
 }
 
+// TelemeterClientClusterRoleBindingView generates a new ClusterRoleBinding for Telemeter client
+// for the cluster monitoring view ClusterRole.
+func (f *Factory) TelemeterClientClusterRoleBindingView() (*rbacv1beta1.ClusterRoleBinding, error) {
+	crb, err := f.NewClusterRoleBinding(MustAssetReader(TelemeterClientClusterRoleBindingView))
+	if err != nil {
+		return nil, err
```
s-urbaniak: unrelated: we seem to be inconsisten in the usage of `errors.Wrap`. Here we don't, but below we do wrap errors.

squat: that's correct. We could do this consistently for everything. It's mechanical and lazy, but would be nice.

PR #136, opened 2018-10-26
Telemeter salt

squat:
This commit causes the operator to generate a random salt for the
Telemeter client to anonymize the metrics it sends.

cc @brancz @s-urbaniak


brancz:
Nice! Makes sense.

/lgtm

PR #138, opened 2018-10-31
*: Enable multi-namespace monitoring

brancz:
This pull request enables monitoring all namespaces, more specifically any namespace labelled with `openshift.io/cluster-monitoring: "true"`.

This has yet to be tested on a 4.0 cluster, but putting this out there for a review already.


cmd/operator/main.go
```diff
@@ -79,6 +79,7 @@ func (t *tags) Type() string {
 func Main() int {
 	flagset := flag.CommandLine
 	namespace := flagset.String("namespace", "openshift-monitoring", "Namespace to deploy and manage cluster monitoring stack in.")
+	namespaceSelector := flagset.String("namespace-selector", "openshift.io/cluster-monitoring=true", "Selector for namespaces to monitor.")
```
s-urbaniak: very very small nit: not sure if this needs to be configurable

pkg/client/client.go
```diff
@@ -176,6 +177,22 @@ func (c *Client) CreateRouteIfNotExists(r *routev1.Route) error {
 	return nil
 }
 
+func (c *Client) NamespaceToMonitor() ([]string, error) {
```
s-urbaniak: nit: s/NamespaceToMonitor/Namespace**s**ToMonitor (plural)

assets/prometheus-k8s/prometheus.yaml
```diff
@@ -98,12 +98,6 @@ spec:
...
   serviceAccountName: prometheus-k8s
-  serviceMonitorNamespaceSelector:
-    matchExpressions:
-    - key: openshift.io/cluster-monitoring
-      operator: Exists
-  serviceMonitorSelector:
-    matchExpressions:
-    - key: k8s-app
-      operator: Exists
+  serviceMonitorNamespaceSelector: {}
```
mxinden: In case I would like to have parts of my namespace monitored by the `prometheus-k8s` Prometheus, and other parts monitored by my own Prometheus, how do I prevent the `prometheus-k8s` to pick up the ServiceMonitors of the second part?

cmd/operator/main.go
brancz: As a side note, within openshift users will never be able to configure this, this is only configurable so we have the ability to change it should we decide on a different labeling/selection scheme.

assets/prometheus-k8s/prometheus.yaml
brancz: As we are limiting the namespaces to monitor the “all” selection here is restricted to “all” core openshift component namespaces.

cmd/operator/main.go
tobru: If I happen to run the monitoring operator on any other K8s cluster which is probably not OpenShift, it could be that I want to change this setting.

brancz: As of today, this operator will not work on anything but openshift. This may change in the future, but may take a long time to be a reality, or maybe never.

pkg/client/client.go
brancz: fixed, please take a look :slightly_smiling_face:

PR #142, opened 2018-11-01
fix makefile issue when GOPATH contains colon

pgier:
Only use the first path from GOPATH, otherwise make fails to correctly parse the Makefile
because of the colon separating muliple paths


brancz:
Thanks! :+1: 

/lgtm

PR #143, opened 2018-11-07
Update kube-prometheus and Prometheus to v2.5.0

metalmatze:
This updates the cluster monitoring operator to use the latest Prometheus version.

/cc @krasi-georgiev @mrsiano @squat @brancz


assets/prometheus-k8s/rules.yaml
```diff
@@ -980,6 +980,16 @@ spec:
       for: 10m
       labels:
         severity: warning
+    - alert: AlertmanagerMembersInconsistent
```
mrsiano: this addition is needed for 2.5? or it is something else?

smarterclayton:
Looks like this regressed new installer and existing ansible in 4.0 - that image doesn't exist:

```
  Warning  Failed     2m (x2 over 3m)  kubelet, ip-10-0-168-118.ec2.internal  Failed to pull image "openshift/prometheus:v2.5.0": rpc error: code = Unknown desc = Error reading manifest v2.5.0 in docker.io/openshift/prometheus: manifest unknown: manifest unknown
```

smarterclayton:
As a side note we'll need to get an `e2e-aws` job set up for this operator and make sure the tests that verify that prometheus comes up are running so you don't regress others.  I'll help with that this week (if you don't have folks familiar with how to do that).

metalmatze:
That would be awesome. I'm not sure if @mxinden wants to do something like that already?

mxinden:
>  I'll help with that this week (if you don't have folks familiar with how to do that).

@smarterclayton that would be great, thanks. I am not aware of anyone in our team.

In addition we might look into something simple with an minishift setup.

PR #144, opened 2018-11-07
targets: Add cluster version operator service as a target

smarterclayton:
Add the cluster version operator as a service monitor target. This will
allow telemetry to report metrics back to cincinnati and the update
monitoring system.

Depends on https://github.com/openshift/cluster-version-operator/pull/49


metalmatze:
The ServiceMonitor needs to be available via jsonnet as the generated assets might be overridden.

You can pretty much copy & paste the etcd ServiceMonitor and put this one beneath it.

https://github.com/openshift/cluster-monitoring-operator/blob/df1677167dac20a9ea5bdd48401f2e61be6712d8/jsonnet/prometheus.jsonnet#L243-L253

Afterwards run `make generate` in the root of this repository.

s-urbaniak:
@smarterclayton I fixed this PR with the jsonnet semantics we have in place here and regenerated everything. PTAL

@brancz @squat @metalmatze PTAL

brancz:
I don’t see code to actually apply these.

s-urbaniak:
@brancz https://github.com/openshift/cluster-monitoring-operator/pull/144/commits/978e1926a4a6e322fd8b8d4e1f18aa422071b99c

s-urbaniak:
@brancz oh wait, I see what you mean. I thought the whole RoleList/RoleBindingList is being applied, I might be wrong :thinking:

brancz:
For the lists yes, but not for `serviceMonitorClusterVersionOperator` (although we should probably make those a list as well)

brancz:
@s-urbaniak can you add a unit test to make sure there are no parse errors (now or in the future)?

s-urbaniak:
@brancz added test and regenerated all the things.

PR #151, opened 2018-11-14
pkg/manifests: etcd defaults to disabled

squat:
cc @brancz

### DIFF
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -723,7 +723,7 @@ func TestPrometheusEtcdRulesFiltered(t *testing.T) {
 }
 
 func TestPrometheusEtcdRules(t *testing.T) {
-	c, err := NewConfigFromString(`etcd: {}`)
+	c, err := NewConfigFromString(`etcd: {enabled: true}`)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/manifests/manifests_test.go
```diff
@@ -723,7 +723,7 @@ func TestPrometheusEtcdRulesFiltered(t *testing.T) {
 }
 
 func TestPrometheusEtcdRules(t *testing.T) {
-	c, err := NewConfigFromString(`etcd: {}`)
+	c, err := NewConfigFromString(`etcd: {enabled: true}`)
```
brancz: Are these changes actually necessary? The empty object is also non-nil, I think we actually dropped the `enabled` field.

squat: The enables field still exists: https://github.com/squat/cluster-monitoring-operator/blob/1bb2db07a7aac0773d14457d4a74d880914db990/pkg/manifests/config.go#L104. We can re-adjust the semantics to always enable etcd Monitoring if the object is non-nil, but I wanted to make the pr small

brancz: Sorry, I was under the impression we did remove it. Let's keep it as is and fix the bug at hand.

squat:
@brancz I just rebased this after it went stale due to some recent changes. PTAL :)

brancz:
thanks for taking care of this! :+1: 

/lgtm

PR #153, opened 2018-11-14
Update recording rules to filter metrics where container_name="" causing double counting

chancez:
I've heard of a double counting report for metering which isn't using
these rules directly, but runs these equivalent queries directly, and
recently I noticed some values looked larger than I know they are
and looking in Grafana I noticed the values were closer to what I
expected. They key difference is Grafana is also not using the recording
rules, but is running the aggregations directly, and I noticed it had
`container_name!=""` in it's label filters.

When testing these queries manually with container_name!="" I found this
to produce more accurate values that are closer to what I was expecting
and measuring manually.


chancez:
@brancz thank you for explain it. I was really sure this wasn't an issue until recently but the query has always been like this.

Reamer:
Can we cherry pick this commit to branch 3.11?

PR #157, opened 2018-11-19
pkg/tasks/telemeter: enable only cluster ID and token is given

s-urbaniak:
This prevents from starting telemeter if no config, cluster ID or
token is given.

cc @squat @brancz

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -123,6 +123,20 @@ type TelemeterClientConfig struct {
 	Token              string `json:"token"`
 }
 
+func (cfg *TelemeterClientConfig) IsEnabled() bool {
+	if cfg == nil {
+		return false
+	}
+
+	if (cfg.Enabled != nil && *cfg.Enabled == false) ||
+		cfg.ClusterID == "" ||
+		cfg.Token == "" {
+		return false
+	}
+
+	return true
+}
+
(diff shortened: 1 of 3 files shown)


s-urbaniak:
xref: https://github.com/openshift/origin/pull/21496

s-urbaniak:
@squat good point :+1: let me change this here.

pkg/manifests/config.go
```diff
@@ -123,6 +123,14 @@ type TelemeterClientConfig struct {
 	Token              string `json:"token"`
 }
 
+func (cfg *TelemeterClientConfig) IsEnabled() bool {
+	return cfg != nil &&
+		cfg.Enabled != nil &&
+		*cfg.Enabled &&
+		cfg.ClusterID != "" &&
+		cfg.Token != ""
+}
```
squat: something like this
```go
if (cfg != nil && cfg.Enbled != nil && *cfg.Enabled == false) || cfg.ClusterID == "" || cfg.Token == "" {
    return false
}
return true
```

s-urbaniak:
@squat addressed, PTAL

squat:
as discussed our options are:
1. opt-out with default to opted in
1. opt-out with default to opted out
1. opt-in with default to opted in
1. opt-in with default to opted out

we want 1 or 3, preferably 1

PR #159, opened 2018-11-20
assets: bump telemeter, node-exporter

squat:
This commit bumps Telemeter client and adjusts the Jsonnet configuration
so that it matches the new layout of the project. This finally bumps
the Telemeter client ServiceMonitor to use the new serving certs CA
bundle.

This commit also pulls in the new changes for the node-exporter which
ensure that the Pod only listens on port 9100.

cc @brancz @s-urbaniak 

@smarterclayton this should fix any final issues with the Telemeter client crashlooping due to not being able to find the serving certs CA bundle on disk.


s-urbaniak:
post-mortem comment: good catch :+1: nevertheless I am puzzled why telemeter client then works on my cluster :thinking:

PR #163, opened 2018-11-21
assets,jsonnet: add host to grafana probe

squat:
cc @brancz


jsonnet/grafana.jsonnet
```diff
@@ -161,7 +161,9 @@ local authorizationRole = policyRule.new() +
           template+: {
             spec+: {
               containers: [
-                super.containers[0] + container.withPorts(containerPort.newNamed('http', 3001)),
+                super.containers[0] +
```
brancz: I actually didn't quite realize what you did here on the last PR, nice :+1: have to remember this one, I usually did a `std.map(super.containers, func(c) {...})`. Although I guess with a map we wouldn't rely on ordering, but this seems fine, let's only change things here should we actually ever run into a problem.

PR #169, opened 2018-11-28
configsharing: Expose full URLs instead of hosts

brancz:
The console team requested this to be the full URLs instead of just hosts.

@s-urbaniak @squat @mxinden @metalmatze @kyoto 

cc @spadgett

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -16,6 +16,7 @@ package client
 
 import (
 	"fmt"
+	"net/url"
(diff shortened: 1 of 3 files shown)


kyoto:
On the Console frontend we normally leave acronyms as upper case in variable names. So `alertmanagerURL`, `grafanaURL` and `prometheusURL` would be consistent with that.

brancz:
Will do a follow up in that case.

PR #171, opened 2018-11-29
jsonnet/grafana: disable http probe against grafana, ...

s-urbaniak:
enable a tcp probe against grafana-proxy instead.

Currently, the k8s health check fails:
``` t=2018-11-29T10:36:31+0000 lvl=info msg="HTTP Server Listen"
logger=http.server address=127.0.0.1:3001 protocol=http subUrl= socket=
```

The reason is, grafana binds against 127.0.0.1:3001 in its network
namespace which k8s cannot probe from the outside.

Instead we probe the grafana proxy instead.

cc @brancz @squat


brancz:
/lgtm

thanks for finding and the follow up fix!

PR #172, opened 2018-11-30
pkg/tasks/telemeter: wait for serving cert secret

s-urbaniak:
Currently, when deploying telemeter client the following error event
is being logged:

```
Warning  FailedMount  4m55s  kubelet, test1-worker-0-t8rrv  MountVolume.SetUp failed for volume "telemeter-client-tls" : secrets "telemeter-client-tls" not found
```

This fixes it by waiting for the secret to be available before
proceeding with the telemeter deployment.

cc @brancz @squat

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -367,6 +367,27 @@ func (c *Client) DeleteSecret(s *v1.Secret) error {
 	return err
 }
 
+func (c *Client) WaitForServingCertSecret(svc *v1.Service) error {
+	return wait.Poll(time.Second*10, time.Minute*5, func() (bool, error) {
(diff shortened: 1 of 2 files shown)


squat:
At the moment several deployments depend on serving cert secrets being available: grafana, alertmanager, prometheus k8s, telemeter client etc. All of these will create the service, and then create the pod that consumes te expected secret assuming that k8s will reattempt creation u til the secret is available. We should either standardize so all of these tasks do the same waking you are proposing OR continue without this botch and let kubernetes do the waiting for us. In both cases the pods will not be created until the secret is available, this is te beauty of k8s.

s-urbaniak:
@squat your suggestion sgtm, let's continue then without waiting for it. Closing this one out.

PR #174, opened 2018-12-03
[WIP] Add Prometheus Adapter to the monitoring stack

metalmatze:
Adding the Prometheus Adapter to the monitoring stack to replace the metrics server as Prometheus is deployed anyway.

Currently a few things are blocking/missing:

* [ ] Publish a new release of Prometheus Adapter with custom CA certificate support 
* [ ]  Verify that the adapter fully works. Some pieces are still missing 

Note: The kubelet's ports are being scrape with `insecureSkipVerify: true` because they are currently using self-signed certs.

/cc @s-urbaniak @brancz @squat


assets/prometheus-adapter/api-service.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apiregistration.k8s.io/v1beta1
+kind: APIService
+metadata:
+  name: v1beta1.metrics.k8s.io
+spec:
+  group: metrics.k8s.io
+  groupPriorityMinimum: 100
+  insecureSkipTLSVerify: true
```
brancz: this can get the serving cert injected as well, just add this annotation:

```
metadata:
  annotations:
    service.alpha.openshift.io/inject-cabundle: "true"
```

assets/prometheus-adapter/deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+      volumes:
+      - emptyDir: {}
+        name: tmpfs
+      - emptyDir: {}
+        name: volume-serving-cert
+      - configMap:
+          name: adapter-config
+        name: config
+      - configMap:
+          name: prometheus-serving-certs-ca-bundle
```
brancz: this was just renamed to just `serving-certs-ca-bundle` (without the `prometheus-` prefix) in #175

assets/prometheus-adapter/service.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
```
brancz: add this annotation to get a serving TLS cert:

```
metadata:
  annotations:
    service.alpha.openshift.io/serving-cert-secret-name: prometheus-adapter-tls
```

then mount the `prometheus-adapter-tls` secret into the deployment and use the cert within that. The keys are called `tls.crt` and `tls.key`.

assets/prometheus-adapter/api-service.yaml
```diff
@@ -1,6 +1,8 @@
 apiVersion: apiregistration.k8s.io/v1beta1
 kind: APIService
 metadata:
+  annotations:
+    'service.alpha.openshift.io/inject-cabundle:': "true"
```
brancz: there is a `:` in that string too much, also I don't think you need the quotes

metalmatze: Yup. I saw that `:` earlier as well... We do need the string for the key.

metalmatze:
I've addressed the issues above. 

The adapter now uses it's own ca-bundle for serving TLS (thanks @s-urbaniak for your help).
Additionally, I've rebased to the latest `master` branch.

Currently I'm stuck at a 403 response by the prometheus-proxy sidecar. It prints something along the lines of `no RBAC policy matched`. Am I missing sending the correct ServiceAccount token? Any ideas?

s-urbaniak:
@metalmatze 
> Currently I'm stuck at a 403 response by the prometheus-proxy sidecar. It prints something along the lines of no RBAC policy matched. Am I missing sending the correct ServiceAccount token? Any ideas?

This is because the adapter doesn't send any service account token along to prometheus, causing the rbac proxy to reject the request.

Additionally, we need to swap the service and deployment creation, otherwise the adapter deployment will deadlock on the tls secret creation which happens after the service creation.

Using the following patch: https://github.com/DirectXMan12/k8s-prometheus-adapter/pull/143 I have working resource metrics:

```
$ kubectl top pod
NAME                                   CPU(cores)   MEMORY(bytes)   
alertmanager-main-0                    6m           37Mi            
alertmanager-main-1                    5m           36Mi            
alertmanager-main-2                    4m           34Mi            
grafana-58456d859d-7ncbg               5m           63Mi            
kube-state-metrics-dcf7dc56d-grjhw     4m           109Mi           
node-exporter-58mq9                    2m           39Mi            
node-exporter-679pm                    2m           42Mi            
node-exporter-f7m2k                    2m           34Mi            
node-exporter-fvppq                    2m           42Mi            
node-exporter-gb7s8                    2m           35Mi            
node-exporter-npvst                    2m           38Mi            
prometheus-adapter-6bcf6746b6-9gtxh    4m           38Mi            
prometheus-k8s-0                       128m         1671Mi          
prometheus-k8s-1                       101m         1637Mi          
prometheus-operator-649f779b78-chz4s   14m          96Mi            
telemeter-client-747b776f55-4bjtn      1m           48Mi        
```

metalmatze:
This now works. `kubectl top node` and `kubectl top pods` print the correct metrics.
Now waiting on a new Prometheus Adapter release.

metalmatze:
All comments have been addressed. The adapter's Prometheus clients configured via a kubeconfig styled config for now.
LGTM when green. :slightly_smiling_face:

brancz:
So awesome! Congrats for delivering this very important feature! :tada: 

/lgtm

jsonnet/main.jsonnet
```diff
@@ -22,6 +22,7 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
                  nodeExporter: 'openshift/prometheus-node-exporter',
                  promLabelProxy: 'quay.io/coreos/prom-label-proxy',
                  kubeRbacProxy: 'quay.io/coreos/kube-rbac-proxy',
+                 prometheusAdapter: 'quay.io/surbania/k8s-prometheus-adapter-amd64',
```
s-urbaniak: @metalmatze let's get rid of this in a follow-up ;-)

jsonnet/main.jsonnet
```diff
@@ -32,6 +33,10 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
                  nodeExporter: 'v0.16.0',
                  promLabelProxy: 'v0.1.0',
                  kubeRbacProxy: 'v0.4.0',
+                 prometheusAdapter: '326bf3c',
```
s-urbaniak: see above

jsonnet/prometheus-adapter.jsonnet
```diff
@@ -0,0 +1,133 @@
...
+            spec+: {
+              containers:
+                std.map(
+                  function(c)
+                    if c.name == 'prometheus-adapter' then
+                      c
+                      {
+                        args+: [
+                          // '--prometheus-ca-file=%s/%s' % [servingCertsCABundleMountPath, servingCertsCABundleFileName],
+                          // '--prometheus-token-file=/var/run/secrets/kubernetes.io/serviceaccount/token',
```
s-urbaniak: @metalmatze I guess these are left-overs? If so, let's get rid of commented code.

metalmatze: I left that in there for now until we now how the adapter moves on. 😉

brancz: Maybe add the discussion issue to the comment

PR #177, opened 2018-12-05
Change kube-controllers targets to separated scheduler and controller-manager

brancz:
Depends on https://github.com/openshift/installer/pull/791.

@s-urbaniak @squat @metalmatze 

cc @smarterclayton


brancz:
agreed

/hold

PR #182, opened 2018-12-10
pkg/client: fix namespaceSelector being ignored when creating client

abhinavdahiya:
The namespace selector is being ignored when creating a client. It is getting defaulted to ""
which is equivalent to select all. During e2e test this is causing churn on prom operator as the pod is getting
redeployed continuosly with more new namespaces like this:
```console
piVersion: extensions/v1beta1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "13"
  creationTimestamp: 2018-12-10T08:47:45Z
  generation: 20
  labels:
    k8s-app: prometheus-operator
  name: prometheus-operator
  resourceVersion: "53430"
  selfLink: /apis/extensions/v1beta1/namespaces/openshift-monitoring/deployments/prometheus-operator
  uid: 4356b33f-fc58-11e8-852c-121e8f0b6b56
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      k8s-app: prometheus-operator
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        k8s-app: prometheus-operator
    spec:
      containers:
      - args:
        - --kubelet-service=kube-system/kubelet
        - --logtostderr=true
        - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
        - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.26.0
        - --namespaces=default,e2e-tests-namespaces-hbpvn,e2e-tests-nsdeletetest-6xf7n,kube-public,kube-system,openshift,openshift-apiserver,openshift-cluster-api,openshift-cluster-dns,openshift-cluster-dns-operator,openshift-cluster-kube-apiserver-operator,openshift-cluster-kube-controller-manager-operator,openshift-cluster-kube-scheduler-operator,openshift-cluster-machine-approver,openshift-cluster-network-operator,openshift-cluster-node-tuning-operator,openshift-cluster-openshift-apiserver-operator,openshift-cluster-openshift-controller-manager-operator,openshift-cluster-samples-operator,openshift-cluster-version,openshift-config,openshift-config-managed,openshift-console,openshift-controller-manager,openshift-core-operators,openshift-csi-operator,openshift-image-registry,openshift-infra,openshift-ingress,openshift-ingress-operator,openshift-kube-apiserver,openshift-kube-controller-manager,openshift-kube-scheduler,openshift-machine-config-operator,openshift-monitoring,openshift-node,openshift-operator-lifecycle-manager,openshift-operators,openshift-sdn,openshift-service-cert-signer
        image: quay.io/coreos/prometheus-operator:v0.26.0
        imagePullPolicy: IfNotPresent
        name: prometheus-operator
        ports:
        - containerPort: 8080
          name: http
          protocol: TCP
        resources: {}
        securityContext: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      nodeSelector:
        beta.kubernetes.io/os: linux
      priorityClassName: system-cluster-critical
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      serviceAccount: prometheus-operator
      serviceAccountName: prometheus-operator
      terminationGracePeriodSeconds: 30
status:
  availableReplicas: 1
  conditions:
  - lastTransitionTime: 2018-12-10T08:48:00Z
    lastUpdateTime: 2018-12-10T08:48:00Z
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available
  - lastTransitionTime: 2018-12-10T08:47:45Z
    lastUpdateTime: 2018-12-10T09:11:04Z
    message: ReplicaSet "prometheus-operator-7b7997c968" has successfully progressed.
    reason: NewReplicaSetAvailable
    status: "True"
    type: Progressing
  observedGeneration: 20
  readyReplicas: 1
  replicas: 1
  updatedReplicas: 1
```

The generation is `20` because of new namespaces are being added and removed during e2e.

/cc @wking

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -100,14 +100,15 @@ func New(cfg *rest.Config, namespace string, namespaceSelector string, appVersio
 	}
 
 	return &Client{
-		namespace:      namespace,
-		appVersionName: appVersionName,
-		kclient:        kclient,
-		ossclient:      ossclient,
-		osrclient:      osrclient,
-		mclient:        mclient,
-		eclient:        eclient,
-		aggclient:      aggclient,
+		namespace:         namespace,
+		namespaceSelector: namespaceSelector,
+		appVersionName:    appVersionName,
+		kclient:           kclient,
+		ossclient:         ossclient,
+		osrclient:         osrclient,
+		mclient:           mclient,
+		eclient:           eclient,
+		aggclient:         aggclient,
 	}, nil
 }
 


wking:
Looks like this was omitted in #138.  I don't see any discussion of the missing `namespaceSelector` there, so maybe the ommission was accidental?  CC @brancz 

/lgtm

wking:
[The previous e2e-aws run][1] included errors like:

```
fail [k8s.io/kubernetes/test/e2e/framework/framework.go:319]: Dec 10 09:58:20.288: Couldn't delete ns: "e2e-tests-clientset-5gk2m": Error while fetching pod metrics for selector  in namespace "e2e-tests-clientset-5gk2m": no pods to fetch metrics for (&errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:""}, Status:"Failure", Message:"Error while fetching pod metrics for selector  in namespace \"e2e-tests-clientset-5gk2m\": no pods to fetch metrics for", Reason:"", Details:(*v1.StatusDetails)(nil), Code:500}})
```

variations of which show up many times:

```console
$ curl -s https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/182/pull-ci-openshift-cluster-monitoring-operator-master-e2e-aws/26/build-log.txt | grep 'no pods to fetch metrics for' | wc -l
504
```

It looks like the code generating that error is [here][2].  And even though that test is running on @abhinavdahiya's fix here, it still has `metrics for selector  in namespace` (with the empty-string selector).  So somewhere up the chain there's probably a similar selector break.  Or maybe we're just missing a default?  But we probably want to figure out why that string is still empty before banging away on too many future e2e-aws runs ;).

[1]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/182/pull-ci-openshift-cluster-monitoring-operator-master-e2e-aws/26/build-log.txt
[2]: https://github.com/openshift/k8s-prometheus-adapter/blob/eee898e17a3bedb4883baf9f5a2e670fcae50a39/pkg/resourceprovider/provider.go#L121-L123

wking:
Hmm.  Trying to walk up that error message, `Error while fetching pod metrics` is upstream [here][1] (since kubernetes-incubator/metrics-server#122), but not in the OpenShift fork (which has [`No metrics for pod`][2] instead).  Which metrics server are we using?  Or am I looking in the wrong place entirely?

[1]: https://github.com/kubernetes-incubator/metrics-server/blob/ca29d12df6c984c2ed3816772ef92fbb0be16fe5/pkg/storage/podmetrics/reststorage.go#L90
[2]: https://github.com/openshift/kubernetes-metrics-server/blob/cf90cf955090399228166efd897b06e6263bc4f5/metrics/storage/podmetrics/reststorage.go#L93

squat:
@wking hmm we're not deploying the metrics server in openshift, only the prometheus adapter. however, per @metalmatze the error could be coming from the upstream metrics api. we'll dig in more after lunch

squat:
OpenShift does not deploy any metrics-server (be it fork or upstream). I suspect that the only reason we see that metrics-server error in the logs is because the Prometheus Adapter vendors the metrics-server package in question. In any case, it only justifies the symptom but not the cause.

smarterclayton:
this is absolutely must fix, but I'm going to work around it in https://github.com/openshift/origin/pull/21641

There's probably 3-4 bugs underpinning this.

1. e2e test validation should be ignoring list of things that don't have delete resources (would prevent this from being called)
2. apiservice for prom adapter shouldn't be added until deployment for adapter is 100% ready (not sure if that's happening, but could also cause this failure)
3. prom adapter needs to be sure to return empty list even if no pods (it would be better for prom adapter to return empty list over 500 in many cases, although we don't want to hide legit errors)
4. need to double check prom adapter isn't reporting having a DELETE endpoint or flakes like this would break namespace deletion
5. obviously having the prom adapter go down and be "unavailable" whenever namespaces change is b ad.  we should verify that prom adapter has 100% uptime, and rolling updates work, etc.

squat:
Leaving some documentation here for sanity in response to @smarterclayton's list of potential sources of error:
1. not specific to monitoring infrastructure
2. the cluster monitoring operator waits for the prometheus adapter deployment's replicas to be ready before creating the apiservice object [0]
3. the prometheus adapter does not directly implement the k8s apiserver interface; instead, it implements the kubernetes-incubator CustomMetricsProvider interface [1], so that means that if this is indeed an error source, we will need to dig into the custom metrics server code and fix this for the whole project [2]
4. same as 3
5. we should consider tweaking the deployment's rollingUpdate strategy maxunavailable and other parameters.

[0] https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/client/client.go#L439-L446
[1] https://github.com/kubernetes-incubator/custom-metrics-apiserver/blob/master/pkg/provider/interfaces.go#L84:6
[2] https://github.com/kubernetes-incubator/custom-metrics-apiserver/blob/master/pkg/apiserver

smarterclayton:
5. as well as graceful deletion (you don't want to race with your load balancers) - pods are removed from endpoints as soon as they are marked for deletion, but that could take time to propagate, so if the pod is faster than propagation API calls start failing

metalmatze:
Tackling 5:
I've created a PR against Prometheus Operator upstream and will let it trickle down once merged. This will make sure that at least on Prometheus Adapter is always running. 
https://github.com/coreos/prometheus-operator/pull/2209

PR #185, opened 2018-12-13
test/e2e: Replace e2e tests with basic stub

mxinden:
In order to reintroduce the end-to-end tests and their continuous
execution, this patch disables the tests for now and replaces them with
a simple test connecting to the Kubernetes cluster.

Relates to https://github.com/openshift/release/pull/2383.

Let me know if this approach sounds reasonable to you. Happy for alternative suggestions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,12 +89,10 @@ $(JSONNET_BIN):
 test-unit:
 	go test $(PKGS)
 
-test: e2e-test
-
 vendor:
 	govendor add +external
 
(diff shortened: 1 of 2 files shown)


s-urbaniak:
let's iterate gently, this is a good starter :+1: Nevertheless, maybe get rid of commented code, we can blame it back from history.

mxinden:
@s-urbaniak @metalmatze @squat would one of you mind giving this another look?

PR #189, opened 2018-12-19
route: Use insecureEdgeTerminationPolicy to redirect from http to https

smarterclayton:
As a minor convenience to users on the platform, redirect http links for
alertmanager, prometheus, and grafana to https automatically.


s-urbaniak:
/lgtm

Thanks! this just needed regeneration of bindata which I added.

smarterclayton:
Is there a make target I should have used?  I called the targets that are
run inside the pod and didn't get biodata update

PR #191, opened 2019-01-07
use latest oauth-proxy images

pgier:


brancz:
those files are auto generated, you will need to update those images in the jsonnet files and re-generate

pgier:
@brancz Thanks, I've updated the jsonnet file and regenerated.

PR #193, opened 2019-01-08
Set Images by image reference

brancz:
This PR enables the cluster monitoring operator to use images as defined by the added image-references.

@s-urbaniak @squat @metalmatze @mxinden 

cc @pgier


assets/grafana/dashboard-definitions.yaml
```diff
@@ -6122,4 +6122,347 @@ items:
   metadata:
     name: grafana-dashboard-k8s-resources-pod
     namespace: openshift-monitoring
+- apiVersion: v1
+  data:
+    persistentvolumesusage.json: |-
```
metalmatze: Do we want to ship this dashboard? I still haven't tested it, as I still do not have these metrics.
I suggest we ignore this for now as well, just like we do with others:
https://github.com/openshift/cluster-monitoring-operator/blob/bf7f82146647419af396f0bdf08b6c17ed6f10eb/jsonnet/main.jsonnet#L78

manifests/04-deployment.yaml
```diff
@@ -24,16 +24,17 @@ spec:
...
-        #- "-tags=prometheus-config-reloader=master"
-        #- "-tags=config-reloader=master"
-        #- "-tags=prometheus=master"
-        #- "-tags=alertmanager=master"
-        #- "-tags=grafana=master"
-        #- "-tags=oauth-proxy=master"
-        #- "-tags=node-exporter=master"
-        #- "-tags=kube-state-metrics=master"
-        #- "-tags=kube-rbac-proxy=master"
+        - "-images=prometheus-operator=docker.io/openshift/origin-prometheus-operator:latest"
```
brancz: @pgier can you check that we indeed have all of these images and that we do also have a latest tag for each of them that we are pushing to?

manifests/04-deployment.yaml
```diff
@@ -24,16 +24,17 @@ spec:
...
+        - "-images=prometheus-config-reloader=docker.io/openshift/origin-prometheus-config-reloader:latest"
+        - "-images=config-reloader=docker.io/openshift/origin-configmap-reload:latest"
+        - "-images=prometheus=docker.io/openshift/origin-prometheus:latest"
+        - "-images=alertmanager=docker.io/openshift/origin-prometheus-alertmanager:latest"
+        - "-images=grafana=docker.io/grafana/grafana:latest"
+        - "-images=oauth-proxy=docker.io/openshift/oauth-proxy:latest"
+        - "-images=node-exporter=docker.io/openshift/origin-prometheus-alertmanager:latest"
```
pgier: Probably copy/paste error.  Should be openshift/origin-prometheus-node-exporter:latest.

manifests/image-references
```diff
@@ -6,3 +6,47 @@ spec:
...
+      kind: DockerImage
+      name: docker.io/openshift/grafana:latest
+  - name: oauth-proxy
+    from:
+      kind: DockerImage
+      name: docker.io/openshift/oauth-proxy:latest
+  - name: node-exporter
+    from:
+      kind: DockerImage
+      name: docker.io/openshift/origin-prometheus-alertmanager:latest
```
pgier: Should be node-exporter.

manifests/04-deployment.yaml
```diff
@@ -24,16 +24,17 @@ spec:
...
-        #- "-tags=prometheus=master"
-        #- "-tags=alertmanager=master"
-        #- "-tags=grafana=master"
-        #- "-tags=oauth-proxy=master"
-        #- "-tags=node-exporter=master"
-        #- "-tags=kube-state-metrics=master"
-        #- "-tags=kube-rbac-proxy=master"
+        - "-images=prometheus-operator=docker.io/openshift/origin-prometheus-operator:latest"
+        - "-images=prometheus-config-reloader=docker.io/openshift/origin-prometheus-config-reloader:latest"
+        - "-images=config-reloader=docker.io/openshift/origin-configmap-reload:latest"
```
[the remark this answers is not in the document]
brancz: updated

manifests/04-deployment.yaml
brancz: absolutely! updated :slightly_smiling_face:

manifests/image-references
brancz: updated

brancz:
I believe now all image references are appropriately configured. Now the only thing missing is that we need to ensure each of the images actually contain the correct versions we want to ship. To get this PR passing most important is the v0.27.0 release of the Prometheus Operator.

cc @pgier could you try to get a PR out to update the prometheus operator fork to v0.27.0? Then we should be able to take care of the rest tomorrow.

brancz:
This is unlikely to pass right now. I'm still doing some testing.

pkg/client/client.go
```diff
@@ -760,6 +762,10 @@ func (c *Client) CreateOrUpdateAPIService(apiService *apiregistrationv1beta1.API
 		return errors.Wrap(err, "retrieving APIService object failed")
 	}
 
+	apiService.ResourceVersion = oldAPIService.ResourceVersion
+	if len(oldAPIService.Spec.CABundle) > 0 {
```
squat: eventually we could try to build a patch using the api machinery pkg rather than manually

pkg/manifests/manifests_test.go
```diff
@@ -634,6 +658,26 @@ ingress:
 	}
 }
 
+func TestK8sPrometheusAdapterConfiguration(t *testing.T) {
```
squat: Great job with these tests :)

manifests/image-references
```diff
@@ -5,4 +5,56 @@ spec:
...
+      name: quay.io/openshift/origin-prometheus-node-exporter:latest
+  - name: kube-state-metrics
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-kube-state-metrics:latest
+  - name: kube-rbac-proxy
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-kube-rbac-proxy:latest
+  - name: telemeter
```
s-urbaniak: here we call it `telemeter`, above we call it `telemeter-client`, those don't have to be in sync?

brancz: this has to be consistent with the prow job so it must be telemeter

PR #199, opened 2019-01-14
manifests: add config API to clusterrole

squat:
cc @s-urbaniak


brancz:
for reference, this fixes permissions required by telemeter (not a feature)

PR #203, opened 2019-01-15
configmap-reloader is usually published as 'reloader', not 'reload'

smarterclayton:
OCP was created with `configmap-reloader` and that is more readable.
Origin is being changed to publish as configmap-reloader, and this
changes CMO to also be updated.

The v3.11 image location is not updated.

See https://github.com/openshift/release/pull/2603

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -40,7 +40,7 @@ type Images struct {
 	PromLabelProxy           string
 	PrometheusOperator       string
 	PrometheusConfigReloader string
-	ConfigmapReload          string
+	ConfigmapReloader        string
 	Prometheus               string
 	Alertmanager             string
 	Grafana                  string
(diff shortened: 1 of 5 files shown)


pkg/manifests/config.go
```diff
@@ -185,7 +185,7 @@ func (c *Config) applyDefaults() {
 func (c *Config) SetImages(images map[string]string) {
 	c.Images.PrometheusOperator = images["prometheus-operator"]
 	c.Images.PrometheusConfigReloader = images["prometheus-config-reloader"]
-	c.Images.ConfigmapReload = images["configmap-reload"]
+	c.Images.ConfigmapReload = images["configmap-reloader"]
```
s-urbaniak: super small nit: let's rename the var to `c.Images.ConfigmapReloader` too, in order to have it consistent

smarterclayton: ok

PR #200, opened 2019-01-14
ClusterOperator status

brancz:
This pull request adds syncing status into the cluster-monitoring-operator's `ClusterOperator` object.

@s-urbaniak @mxinden @squat @metalmatze 

Note this is a feature and first needs to be announced to the mailing list and approved.

/hold

PR #207, opened 2019-01-16
manifests: remove default cluster monitoring config

s-urbaniak:
Currently, any changes done to the configuration of cluster monitoring operator
will be reconciled by the cluster version operator.

As the current default configuration (manifested in code in the operator)
is supposed to work anyways, there is no need to ship a default configuration at all.

That way the user can create this config post-install, disabling telemeter,
or changing any other operational parameters of the cluster monitoring operator.

Fixes Bugzilla 1666626
Fixes Jira MON-515

cc @smarterclayton @brancz @squat @juzhao


juzhao:
@s-urbaniak 
cluster monitoring is enabled by default, I think this change will not install cluster monitoring by default now

brancz:
superseeded by #200

PR #208, opened 2019-01-16
set cpu request and remove limits for pods and operator

sjenning:
@smarterclayton @derekwaynecarr


sjenning:
wow!
/hold
something got messed up here :P

brancz:
We had `20m` until recently and even with that frequently encountered CPU throttling. Are we sure this is a good idea, and based on actual usage data?

sjenning:
i'm only adding a request for the operator itself and `10m` is the expected usage for an operator (operators should be mostly idle).  removing the cpu limits for the other components should actually do away with the throttling and make things better.

PR #209, opened 2019-01-17
manifests: remove 03-etcd-secret.yaml

s-urbaniak:
Following the same scheme as 03-config.yaml it is safe to remove
03-etcd-secret.yaml.

If etcd is disabled, we remove that secret from the spec anyways as per
https://github.com/openshift/cluster-monitoring-operator/blob/9ec3b14d4993596f095f8f954caada4aab1a4a2d/pkg/manifests/manifests.go#L685-L694.

Hence it doesn't make sense to deploy an empty default config in the
first place.

Fixes MON-506

cc @brancz @metalmatze @squat


squat:
Even though the secrets are empty, it might be nice to deploy it anyways to provide an easy way for those manually deploying CMO to OpenShift to configure the stack.

squat:
Also, as we are past code freeze, we should refrain from touching any code unrelated to a bug and there is currently no bug surrounding etcd configuration. This should stay here for now.

s-urbaniak:
hmm, let's @brancz chime in to break ties, as removing the file was his suggestion ;-)

brancz:
This is user configuration, and the problem would be that the CVO overwrites any configuration a user attempts to do with this empty configuration, essentially making etcd monitoring defunct.

brancz:
superseeded by #200

PR #211, opened 2019-01-18
pkg/client: return error from WaitForPrometheusOperatorCRDsReady

simonpasquier:


brancz:
Good catch. Maybe we should introduce vet and staticcheck :)

/lgtm

PR #216, opened 2019-01-22
manifests: Delete user provided config from payload

brancz:
Extracting this from #215 to see if e2e tests pass individually.


pkg/operator/operator.go
```diff
@@ -130,10 +130,31 @@ func (o *Operator) Run(stopc <-chan struct{}) error {
...
 
 	go o.worker()
-
 	go o.cmapInf.Run(stopc)
 
-	<-stopc
+	ticker := time.NewTicker(5 * time.Minute)
+	defer ticker.Stop()
+
+	time.Sleep(10 * time.Second)
```
squat: Normally we can use waitforcachesync here, however let's refactor it later if tests are passing right now

PR #217, opened 2019-01-22
pkg/operator: cleanup waitforcachesync

squat:
This commit cleans up the Run and sync methods of the operator so that
they wait for the cache to be properly synced before trying to
reconcile and so that they actually use the `key` argument that is
passed.

cc @brancz @metalmatze


pkg/operator/operator.go
```diff
@@ -129,28 +129,35 @@ func (o *Operator) Run(stopc <-chan struct{}) error {
 		return nil
 	}
 
+	glog.V(4).Info("Waiting for initial cache sync.")
+	ok := cache.WaitForCacheSync(stopc, o.cmapInf.HasSynced)
```
brancz: can this ever succeed before we start it?

squat: you're right, we need to run the informer first.

pkg/operator/operator.go
```diff
@@ -129,28 +129,36 @@ func (o *Operator) Run(stopc <-chan struct{}) error {
...
 	go o.cmapInf.Run(stopc)
 
+	glog.V(4).Info("Waiting for initial cache sync.")
+	ok := cache.WaitForCacheSync(stopc, o.cmapInf.HasSynced)
+	if !ok {
+		return errors.New("failed to sync informers")
+	}
+	glog.V(4).Info("Initial cache sync done.")
+
+	go o.worker()
```
brancz: I think I would also star the worker already as well, just so we're not in a situation where waiting for cache sync wants to enqueue and is blocked on a worker accepting

squat: we don't want to start the worker yet, or we will start processing work items and may try to get things out of the cache, which has not finished syncing yet. What will happen here is that the queue will grow until the cache is synced and then the workers will start processing it. The other option is to delay adding the handlers after the sync is finished.

squat: the queue does not block adding work items on workers accepting AFAIK.

mxinden:
Mh, that is an interesting order of events. But e2e-aws-operator failure seems unrelated to our (pretty much non-existing) test suite.

squat:
agreed, tests DID pass, and was being merged by prox, but I accidentally re-kicked the tests somehow :$

PR #218, opened 2019-01-22
Drop cardinality explosion metric from kube-controller-manager

brancz:
Until the 1.12 rebase lands we're subject to this, so to make Prometheus not OOM, we're dropping these for now.

@smarterclayton @squat @mxinden @metalmatze


smarterclayton:
@soltysh fyi this is fixed 1.12 so nothing for you, just be aware

PR #219, opened 2019-01-22
remove limits for pods

sjenning:
@smarterclayton @brancz 

follow on to my previous PR to make components run as Burstable, without limits

### DIFF
--- a/assets/grafana/deployment.yaml
+++ b/assets/grafana/deployment.yaml
@@ -24,9 +24,6 @@ spec:
         - containerPort: 3001
           name: http
         resources:
-          limits:
-            cpu: 200m
-            memory: 200Mi
           requests:
             cpu: 100m
             memory: 100Mi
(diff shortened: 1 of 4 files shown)


sjenning:
ugh, i did it again.  these are generated files that i can't find the source for.

sjenning:
doing this in https://github.com/openshift/prometheus-operator/pull/20

squat:
@sjenning this one is a little more tricky :p 
These manifests are likewise all generated from jsonnet, however they come mostly from kube-prometheus, a distribution of monitoring components that we put together and are a dependency of this project [0]. The correct way to do this would be to write a mixin that removes limits for any object used in cluster monitoring operator. (We don’t want to remove the limits upstream in the kube-prometheus distro and other projects we are importing as they have valid use cases for limits). Please let me know if you need any help or want me to take over this and I can make a PR tomorrow. 

[0] 
https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/main.jsonnet#L1

squat:
Note: kube-prometheus is being imported from the upstream repo, not the openshift repo 
https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/jsonnetfile.json#L7

squat:
finished this in #273

PR #220, opened 2019-01-23
Add cluster mem|cpu|storage total and utilization rules

yaacov:
Add aggregated cluster storage mem and cpu rules.

In the telemeter [1] we want to federate only one metric per cluster [2] to reduce bandwidth and storage.
This new recording rules will create one cluster metrics for N nodes in the cluster.

[1] https://github.com/openshift/telemeter
[2] https://github.com/openshift/telemeter/pull/114


yaacov:
@jhernand, @kbsingh, @oourfali,  @squat please review.

assets/prometheus-k8s/rules.yaml
```diff
@@ -334,6 +334,18 @@ spec:
     - expr: sum(rate(container_cpu_usage_seconds_total{container_name!="POD",container_name!="",pod_name!=""}[5m]))
         / sum(machine_cpu_cores)
       record: cluster:container_cpu_usage:ratio
+    - expr: sum({__name__="machine_cpu_cores"})
+      record: cluster_cpu_cores
```
brancz: The best practice is to use `:` to show this is a recording rule and put the result and used aggregation into perspective. So this would be `cluster:machine_cpu_cores:sum`. You can look at examples of the other recording rules defined here.

assets/prometheus-k8s/rules.yaml
```diff
@@ -334,6 +334,18 @@ spec:
...
+    - expr: sum({__name__="node:node_cpu_utilisation:avg1m"})
+      record: cluster_cpu_utilisation:avg1m
+    - expr: sum({__name__="machine_memory_bytes"})
+      record: cluster_memory_bytes
+    - expr: sum({__name__="node:node_memory_utilisation:ratio"})
+      record: cluster_memory_utilisation:ratio
+    - expr: sum({__name__="node_filesystem_size_bytes"})
```
brancz: This heavily relies on the underlying metrics to only be collected from physical filesystems, which I don't think is true. Could you share an example of the time-series returned by this metric?

assets/prometheus-k8s/rules.yaml
```diff
@@ -334,6 +334,18 @@ spec:
     - expr: sum(rate(container_cpu_usage_seconds_total{container_name!="POD",container_name!="",pod_name!=""}[5m]))
         / sum(machine_cpu_cores)
       record: cluster:container_cpu_usage:ratio
+    - expr: sum({__name__="machine_cpu_cores"})
+      record: cluster_cpu_cores
+    - expr: sum({__name__="node:node_cpu_utilisation:avg1m"})
```
brancz: `__name__` is an internal label, you can just skip it for all of these and just do

```
sum(node:node_cpu_utilisation:avg1m)
```

assets/prometheus-k8s/rules.yaml
yaacov: @brancz, this metrics was chosen because this is the metrics uesd in Grafana dahsboard [1].

[1] https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/use.libsonnet#L79

> Could you share an example of the time-series returned by this metric?

This example show size vs. free per node:
`{__name__=~"node_filesystem_free_bytes|node_filesystem_size_bytes"}`
This example show cluster:
`sum({__name__=~"node_filesystem_free_bytes"})`

yaacov: @brancz we are not set on what metrics to use here, if you think of a better metric to show for cluster `storage` totals and utilization I will be happy to change this metrics to something that will be more helpful to end users.

yaacov:
@brancz hi, thanks for the review, fixed the "easy" things, about storage metrics I'm open to suggestions :-) 

pleas re-review.

assets/prometheus-k8s/rules.yaml
brancz: That's the metric name used, but the whole query we end up using is:

```
sum(max(node_filesystem_size_bytes{fstype=~"ext[234]|btrfs|xfs|zfs"} - node_filesystem_avail_bytes{fstype=~"ext[234]|btrfs|xfs|zfs"}) by (device,pod,namespace)) by (pod,namespace)
/ scalar(sum(max(node_filesystem_size_bytes{fstype=~"ext[234]|btrfs|xfs|zfs"}) by (device,pod,namespace)))
* on (namespace, pod) group_left (node) node_namespace_pod:kube_pod_info:
```

brancz:
For what it's worth these rules would be accepted in the kubernetes-mixin directly as well, I think it would make sense to put them there and use them in the Grafana dashboards directly.

yaacov:
@brancz hi, changed the storage metrics to only include sensible file system types, please re-review.

assets/prometheus-k8s/rules.yaml
```diff
@@ -334,6 +334,14 @@ spec:
     - expr: sum(rate(container_cpu_usage_seconds_total{container_name!="POD",container_name!="",pod_name!=""}[5m]))
         / sum(machine_cpu_cores)
       record: cluster:container_cpu_usage:ratio
+    - expr: sum(node:node_cpu_utilisation:avg1m)
```
brancz: This would give us the sum of average CPU utilization in %. The sum of multiple % is not accurate out of 100. We need to use the raw data and not aggregated by node.

assets/prometheus-k8s/rules.yaml
```diff
@@ -334,6 +334,14 @@ spec:
...
+      record: cluster:node_cpu_utilisation:avg1m
+    - expr: sum(node:node_memory_utilisation:ratio)
```
brancz: same as above

assets/prometheus-k8s/rules.yaml
```diff
@@ -334,6 +334,14 @@ spec:
...
+      record: cluster:node_memory_utilisation:ratio
+    - expr: sum(node_filesystem_size_bytes{fstype=~"ext[234]|btrfs|xfs|zfs"})
```
brancz: This metric shows each filesystem even if the same device is mounted on multiple mountpoints. I think we're looking for:

```
sum(max by(instance, device) (node_filesystem_size_bytes{fstype=~"ext[234]|btrfs|xfs|zfs"}))
```

assets/prometheus-k8s/rules.yaml
```diff
@@ -334,6 +334,14 @@ spec:
...
+      record: cluster:node_filesystem_size_bytes:userfstypes
+    - expr: sum(node_filesystem_free_bytes{fstype=~"ext[234]|btrfs|xfs|zfs"})
```
brancz: this needs similar aggregations as above

assets/prometheus-k8s/rules.yaml
```diff
@@ -334,6 +334,18 @@ spec:
     - expr: sum(rate(container_cpu_usage_seconds_total{container_name!="POD",container_name!="",pod_name!=""}[5m]))
         / sum(machine_cpu_cores)
       record: cluster:container_cpu_usage:ratio
+    - expr: sum(node:node_num_cpu:sum)
+      record: cluster:node_num_cpu:sum
+    - expr: sum(node:node_cpu_utilisation:avg1m * node:node_num_cpu:sum)
+      record: cluster:node_cpu_usage:avg1m
+    - expr: sum(node:node_memory_bytes_total:sum)
+      record: cluster:node_memory_bytes_total:sum
+    - expr: sum(node:node_memory_utilisation:ratio * node:node_memory_bytes_total:sum)
```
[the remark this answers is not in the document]
yaacov: > What you want to do is sum up all the memory being used (as in the raw values, not ratios) and divide that by the total memory available.

I just want to get he `memory being used ` no need to devide.

my reasoning is `node:node_memory_utilisation:ratio * node:node_memory_bytes_total:sum`
returns a vector of `memory being used ` per node,
summing it up will give `memory being used ` per cluster.

What am I missing here ?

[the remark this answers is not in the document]
yaacov: @brancz ^^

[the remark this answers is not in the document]
brancz: bytes used total would be:

```
sum(node_memory_MemFree_bytes{job="node-exporter"} + node_memory_Cached_bytes{job="node-exporter"} + node_memory_Buffers_bytes{job="node-exporter"})
```

total amount of memory:

```
sum(node_memory_MemTotal_bytes{job="node-exporter"})
```

This gets more complicated with CPU though, as CPU across servers can be heterogeneous.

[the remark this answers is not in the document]
brancz: Sorry that actually wasn't the total used, it was the total _unused_. This is the total used:

```
sum(node_memory_MemTotal_bytes{job="node-exporter"}) - (sum(node_memory_MemFree_bytes{job="node-exporter"} + node_memory_Cached_bytes{job="node-exporter"} + node_memory_Buffers_bytes{job="node-exporter"}))
```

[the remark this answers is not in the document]
yaacov: @brancz , sorry to neg, I am not familiar with the metrics collected by the openshift monitoring:
 
a - do we want to show people `cached` and `buffered` mem ?
I think a more sensible metrics will be just the `working set` ?

b - why only the "node exporter job" , shouldn't we give all memory used by the cluster ?

kbsingh:
@jhernand note that @brancz has offered to pickup this PR and get it through, then exposed into the endpoints via client

jhernand:
Thanks @brancz, please let us know when you know what will be the names of the metrics that we will be able to query in the telementer Prometheus.

yaacov:
Adding the new rules for cluster mem|cpu|storage will be done by the telemeter team ?

@jhernand  @brancz , should I close this one ?

brancz:
Yes closing here @squat is currently working on this.

oourfali:
@brancz thanks!
@squat Is there a card tracking that?

PR #221, opened 2019-01-23
Fix duplicate CPUThrottlingHigh alert

metalmatze:
This fixes the bug which causes a CPUThrottlingHigh alert to be fired twice for one container and pod.

Fixes MON-521

/cc @brancz @squat @mxinden


metalmatze:
PTAL, I simply generated the missing bindata.go

PR #226, opened 2019-01-30
Exclude network interfaces instead of selecting them

metalmatze:
Instead of selecting specific interfaces and try to match as much as possible, we should rather exclude the we don't want to include.

https://bugzilla.redhat.com/show_bug.cgi?id=1670330

/cc @mxinden @brancz @squeed


mxinden:
Not sure `master` is open again. Thereby we will probably need to wait merging this. What do you think?

metalmatze:
I'm fine with waiting. I just thought it was basically a bug fix, is it not?

metalmatze:
You're totally right. I'll update the upstream mixin and then it should trickle down to kube-promtheus like always. Thanks for the hint

PR #228, opened 2019-01-31
Add crio metrics as an endpoint of nodes

brancz:
As this needs a bit of a hack, I went ahead and did the relabelling rules to make this work.

@squat @s-urbaniak @mxinden @metalmatze 

cc @runcom


brancz:
Still testing this so putting a hold on this for now.

/hold

brancz:
I am currently seeing a problem where consistently there are nodes that cannot be scraped. @runcom is investigating.

screenshot from 2019-01-31 17-00-28

runcom:
so, I've been testing this on a brand new cluster from today and I'm perfectly able to grab metrics from each node

brancz:
Got it, in that case I will give it another try.

brancz:
I think I have an idea what's happening. This seems to only happen to workers, I think the security group rules are probably off.

runcom:
latest installer now has https://github.com/openshift/installer/pull/1201 which should open ports for worker<->worker communication for metrics

brancz:
Awesome. I'm going to give this a try soon.

s-urbaniak:
@brancz can we add the `up` query to the existing e2e tests too?

https://github.com/openshift/cluster-monitoring-operator/blob/e536f4fea32bf02944f583d57eb29951e8c70481/test/e2e/main_test.go#L93-L128

PR #230, opened 2019-02-01
Override tlsConfig.serverName in kube-controllers ServiceMonitor

jim-minter:


brancz:
We try to not set things at runtime when we can. Could you instead set this value directly in the jsonnet files and re-generate? Then this would lgtm.

jim-minter:
@brancz I respect your approach; I tried and was not able to do that because the jsonnet file in question is vendored.  I was hoping that given we're doing exactly the same thing that the function above already does, that I might get away with it?

squat:
@jim-minter the difference is that the function above needs runtime information whereas the field in your patch is statically known. Please let me know if you need a hand with the jsonnet; I’d be happy to help you make the change.

jim-minter:
@squat help would be appreciated - afaics the jsonnet file that would need to be modified does not reside in this repo.

brancz:
As this is 3.11, it actually does reside in this repository: https://github.com/openshift/cluster-monitoring-operator/blob/release-3.11/jsonnet/prometheus.jsonnet#L131

jim-minter:
Thanks @brancz @squat.  I am not sure how I missed that.  Perhaps I thought all of /jsonnet was vendored.  Anyway, PTAL again.

brancz:
We're almost there. This just needs a `make pkg/manifests/bindata.go`. Then CI should be passing.

jim-minter:
Done - Travis is flaking on something unrelated I think

brancz:
yeah that looked like a flake, restarted it

jim-minter:
@brancz this keeps flaking and we need it in - please can you take a look?

jim-minter:
I think the travis config is wrong for the release-3.11 branch?

brancz:
That's not a flake. You need to rebase this pull request, the build infrastructure has been changed heavily between 4.0 and 3.11. `HEAD` of `release-3.11` has the needed changes, just rebase against that.

jim-minter:
Got it - thanks.  Repushed, @brancz over to you

PR #231, opened 2019-02-04
added recording rules for KubeState metrics

SriKrishnaPaparaju:
Added recording rules for 'pod:kube_pod_container_status_restarts_total' and 'pod:kube_pod_container_status_terminated_reason'.


brancz:
I’m assuming you want the total sum to do rates, but that doesn’t quite work as the rate function takes counter resets into account, which would be incorrect in this case. What we want here is a recording rule that sums all rates of container restarts.

SriKrishnaPaparaju:
@brancz Have changed the recording rule to pick up rates and sum them up for container restarts related metrics.

brancz:
This PR needs a

```
hack/build-jsonnet.sh
make pkg/manifests/bindata.go
```

brancz:
As I asked on the telemeter PR related to this. Is this still relevant?

SriKrishnaPaparaju:
@brancz Its not relevant at this time as the POC for predicting upgrades strictly rely on code changes across the releases.

PR #232, opened 2019-02-04
manifests: fix cluster operator naming

s-urbaniak:
Fixes https://coreos.slack.com/archives/C3VS0LV41/p1549234605180400

cc @brancz @squat @clayton


brancz:
Whops we missed coordinating this. Fortunately it's a small change. I also opened #232 which I believe addresses the other spots this needs fixing at.

PR #235, opened 2019-02-06
Update jsonnet deps

brancz:
This primarily pulls in a few metric relabelling rules to drop garbage and high cardinality metrics.

@mxinden @s-urbaniak @squat @metalmatze


squat:
It looks like this needs a bindata regeneration.

squat:
Otherwise, looks good!

brancz:
Whops, good catch. Updated.

PR #238, opened 2019-02-07
Bump KubeletTooManyPods alert threshold to 250

brancz:
To make this based on the actual limit configured we need an upstream change that is only landing in Kubernetes 1.14, so going with this for now, which is the default limit in OpenShift.


jsonnet/main.jsonnet
```diff
@@ -73,6 +73,8 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
     kubeSchedulerSelector: 'job="scheduler"',
 
     namespaceSelector: 'namespace=~"(openshift-.*|kube-.*|default|logging)"',
+
+    kubeletPodLimit: 250,
```
squat: Ah yes, we recently exposed this config variable in mixins, no?

brancz: yes!

PR #239, opened 2019-02-08
*: Enable etcd monitoring by default

brancz:
This PR removes the leftover configuration options, that are unnecessary now as etcd is always setup as static pods and we consistently monitor it.

As discussed on slack, we can either merge this as soon as e2e goes green on this, giving the etcd team an easier path to test this, or wait for them to provision the `openshift-monitoring/kube-etcd-client-certs` Secret with the appropriate client certificates, and then verify this works. I favor the first, as we can treat everything else as follow up/bugfix, but have no strong opinion (this is also a bugfix as we have an open bug, that these configurations don't work).

@mxinden @s-urbaniak @squat @metalmatze 

cc @hexfusion

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -34,7 +34,7 @@ type Config struct {
 	AlertmanagerMainConfig   *AlertmanagerMainConfig   `json:"alertmanagerMain"`
 	KubeStateMetricsConfig   *KubeStateMetricsConfig   `json:"kubeStateMetrics"`
 	GrafanaConfig            *GrafanaConfig            `json:"grafana"`
-	EtcdConfig               *EtcdConfig               `json:"etcd"`
+	EtcdConfig               *EtcdConfig               `json:"-"`
 	HTTPConfig               *HTTPConfig               `json:"http"`
 	TelemeterClientConfig    *TelemeterClientConfig    `json:"telemeterClient"`
 	K8sPrometheusAdapter     *K8sPrometheusAdapter     `json:"k8sPrometheusAdapter"`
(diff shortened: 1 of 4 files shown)


brancz:
this actually fails legitimately, it needs a few more tweaks

brancz:
I changed the strategy a little. As we rely on the etcd management controller to provision a certificate for us, "enabling" etcd monitoring happens by the etcd certificates being provisioned appropriately instead of explicitly user configured.

pkg/operator/operator.go
```diff
@@ -382,5 +384,23 @@ func (o *Operator) Config(key string) *manifests.Config {
 		glog.Warningf("Error loading proxy from API. Proceeding without it: %v", err)
 	}
 
+	s, err := o.client.GetSecret(o.namespace, prometheusEtcdCertSecretName)
+	if err != nil {
+		glog.Warningf("Error loading etcd certificates for Prometheus. Proceeding with etcd disabled. Error: %v", err)
```
mxinden: I guess we could do an early return here, but maybe this function grows in the future, so this is also fine with me.

pkg/manifests/config.go
```diff
@@ -96,8 +96,7 @@ type K8sPrometheusAdapter struct {
 }
 
 type EtcdConfig struct {
-	Enabled    *bool  `json:"enabled"`
-	ServerName string `json:"serverName"`
+	Enabled *bool `json:"-"`
```
mxinden: Is this still necessary, given that `EtcdConfig` in `Config` is disabled anyways?

brancz: I intentionally left this to modify the least amount of code, but also seems like a reasonable strategy to me. Essentially what's happening here is that when the right certificates are present in the `Secret`, then this field gets set to enabled.

pkg/operator/operator.go
brancz: Agreed, I think the early return would cause future problems.

mxinden:
Seems like your _retest_ was not picket up @brancz. Trying again.

/retest

PR #240, opened 2019-02-08
pkg/client: update waits to report lastError than the timed-out error

abhinavdahiya:
During an installation failure issue [1], the monitoring was reporting following error:
```
Error: running task Updating Grafana failed: waiting for Grafana Route to become ready failed: timed out waiting for the condition
```

`timed out waiting for the condition` does not indicate what was the last state that caused the wait to timeout. Reporting lastErr in cases of
timeout should provide better errors.

[1]: https://github.com/openshift/installer/issues/1219

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -313,7 +313,8 @@ func (c *Client) DeletePrometheus(p *monv1.Prometheus) error {
 		return errors.Wrap(err, "deleting Prometheus object failed")
 	}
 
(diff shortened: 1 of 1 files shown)


s-urbaniak:
@mxinden ping, as there was a force-push here

brancz:
As this improves debugging a lot I think this is ok to merge and I can even see this being a bug as currently we provide insufficient information for triaging.

/lgtm
/hold cancel

PR #241, opened 2019-02-11
Makefile: Include jsonnet and gobindata in 'generate' target and regenerate

mxinden:
Include both jsonnet templating and gobindata generation in the Makefile
'generate' target, which is included in the Travis builds.


mxinden:
Once CI fails on this, I will push the commit updating gobindata.

Makefile
```diff
@@ -62,9 +62,15 @@ $(ASSETS): $(JSONNET_SRC) $(JSONNET_VENDOR) hack/build-jsonnet.sh
 $(JSONNET_VENDOR): jsonnet/jsonnetfile.json
 	cd jsonnet && jb install
 
-generate: clean
+generate: clean $(GOBINDATA_BIN)
```
squat: This change should not be here. Note that the target is run in a docker container, but this dependency is installed on the calling host. The make invocation in the target already includes "dependencies" ensuring that all binaries are installed in the docker container.

mxinden: Right. I meant to add this dependency to `pkg/manifests/bindata.go`. That should work, right? Why do we need our homegrown dependency management in that case?

mxinden:
@squat would you mind taking another look?

mxinden:
Validated on an Openshift 3.11 cluster via quay.io/mxinden/cluster-monitoring-operator:3.11-rbac-proxy.0

mxinden:
Thanks @squat

PR #244, opened 2019-02-13
 operator: rotate api authentication certs

s-urbaniak:
Currently, when the extension api authentication certificates are
being rotated, prometheus adapter is not being updated.

This fixes it by watching the extension api authentication configmap
in the kube-system namespace in the cluster monitoring operator
directly. When this configmap is being updated (this happens, when the
CAs rotate), a new secret in the openshift-monitoring namespace is
generated with a synthesized name based on the CA hashes. Finally the
prometheus adapter deployment is being updated and rolled out.

Fixes MON-219

cc @squat @mxinden PTAL
~@mxinden bear with me, I have to write up some e2e tests still~


s-urbaniak:
The `up{job=\"node-exporter\"} == 1" but got 0` seem to be unrelated to this PR :thinking:

pkg/manifests/errmapreader.go
```diff
@@ -0,0 +1,85 @@
...
+
+package manifests
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+// errMapReader wraps a map[string]string and exposes read method.
```
mxinden:
```suggestion
// errMapReader wraps a map[string]string and exposes a read method.
```

pkg/tasks/prometheusadapter.go
```diff
@@ -130,7 +130,27 @@ func (t *PrometheusAdapterTask) Run() error {
...
 	{
-		dep, err := t.factory.PrometheusAdapterDeployment()
+		apiAuthConfigmap, err := t.client.GetAPIServerAuthenticationConfigmap()
+		if err != nil {
+			return errors.Wrap(err, "failed to load kube-system/extension-apiserver-authentication configmap")
+		}
+
+		apiAuthSecret, err := t.factory.PrometheusAdapterAPIAuthSecret(apiAuthConfigmap.Data)
+		if err != nil {
+			return errors.Wrap(err, "initializing PrometheusAdapter Deployment failed")
```
mxinden:
```suggestion
			return errors.Wrap(err, "failed to create prometheus adapter api auth secret")
```

Something along the lines?

pkg/operator/operator.go
```diff
@@ -41,6 +41,9 @@ var (
 
 const (
 	resyncPeriod = 5 * time.Minute
+
+	// see https://github.com/kubernetes/apiserver/blob/b571c70e6e823fd78910c3f5b9be895a756f4cbb/pkg/server/options/authentication.go#L239
```
mxinden:
```suggestion
	// > authenticationConfigMapName is the name of ConfigMap in the kube-system namespace holding the root certificate
	// > bundle to use to verify client certificates on incoming requests before trusting usernames in headers specified
	// > by --requestheader-username-headers. This is created in the cluster by the kube-apiserver.
        // > "WARNING: generally do not depend on authorization being already done for incoming requests.")
	// see https://github.com/kubernetes/apiserver/blob/b571c70e6e823fd78910c3f5b9be895a756f4cbb/pkg/server/options/authentication.go#L239
```
Would you mind adding the comment directly?

pkg/manifests/errmapreader_test.go
```diff
@@ -0,0 +1,191 @@
+// Copyright 2018 The Cluster Monitoring Operator Authors
```
mxinden:
```suggestion
// Copyright 2019 The Cluster Monitoring Operator Authors
```

pkg/manifests/errmapreader.go
```diff
@@ -0,0 +1,85 @@
+// Copyright 2018 The Cluster Monitoring Operator Authors
```
mxinden:
```suggestion
// Copyright 2019 The Cluster Monitoring Operator Authors
```

pkg/operator/operator.go
s-urbaniak: hm, I didn't want to refer to the comment, but the constant per se. I suggest to leave this for reference.

pkg/tasks/prometheusadapter.go
s-urbaniak: sounds good :+1:

pkg/manifests/errmapreader.go
s-urbaniak: good catch :+1:

pkg/manifests/errmapreader_test.go
s-urbaniak: time is moving so fast ..... ;-)

s-urbaniak:
@mxinden thanks for the review, I'll address the comments along with the e2e test with the logic we discussed yesterday.

s-urbaniak:
@mxinden PTAL, also added an e2e test

test/e2e/prometheusadapter_test.go
```diff
@@ -0,0 +1,85 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+func TestPrometheusAdapterCARotation(t *testing.T) {
+	// Wait for Prometheus adapter
+	err := wait.Poll(time.Second, 5*time.Minute, func() (bool, error) {
+		_, err := f.KubeClient.Apps().Deployments(f.Ns).Get("prometheus-adapter", metav1.GetOptions{})
+		if err != nil {
+			return false, nil
```
mxinden: Let's surface this error to the user for better debugging. Something along the lines of https://github.com/openshift/cluster-monitoring-operator/pull/240/files.

test/e2e/prometheusadapter_test.go
```diff
@@ -0,0 +1,85 @@
...
+	newSecret, err := factory.PrometheusAdapterAPIAuthSecret(cm.Data)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Wait for the new secret to be created
+	err = wait.Poll(time.Second, 5*time.Minute, func() (bool, error) {
+		_, err := f.KubeClient.CoreV1().Secrets(f.Ns).Get(newSecret.Name, metav1.GetOptions{})
+		if err != nil {
+			return false, nil
```
mxinden: Same as above, let's surface this error to the user, see: https://github.com/openshift/cluster-monitoring-operator/pull/240/files

test/e2e/prometheusadapter_test.go
```diff
@@ -0,0 +1,85 @@
...
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Wait for new Prometheus adapter to roll out
+	err = wait.Poll(time.Second, 5*time.Minute, func() (bool, error) {
+		d, err := f.KubeClient.Apps().Deployments(f.Ns).Get("prometheus-adapter", metav1.GetOptions{})
+		if err != nil {
+			return false, nil
```
mxinden: See above.

test/e2e/prometheusadapter_test.go
s-urbaniak: very good idea :+1:

s-urbaniak:
@mxinden addressed comments, PTAL

metalmatze:
Looking good to me with the latest changes! :+1:

mxinden:
@s-urbaniak This needs to target the `release-4.0` branch as it is supposed to land in Openshift 4.0, right?

s-urbaniak:
@mxinden changed the base to `release-4.0`

pkg/manifests/errmapreader.go
```diff
@@ -0,0 +1,86 @@
...
+	result, ok := r.src[key]
+	if !ok {
+		r.err = fmt.Errorf("key %s is missing", key)
+		return ""
+	}
+
+	return result
+}
+
+// Slice calls errMapReader.Value(key) and parses it as a string slice in JSON format.
```
squat: JSON doesn't have a concept of slices, only arrays. For clarity, maybe consider:
```
Slice calls errMapReader.Value(key) and parses the resulting string as a JSON list of strings.
The list of strings is joined with commas and returned.
```

pkg/manifests/errmapreader.go
```diff
@@ -0,0 +1,86 @@
...
+// It returns the slice as a string joined with commas to be consumed for command line arguments.
+// If an error already occured in a previous call, empty string is returned.
+// If an error occurs during parsing, errMapReader.err is updated
+// and empty string is returned.
+func (r *errMapReader) Slice(key string) string {
```
squat: It's quite confusing that this is called `Slice` when it returns a string. Perhaps it would be better to directly return a slice and let the caller do the `strings.Join`. Otherwise, consider renaming this method.

s-urbaniak: ack, I was hesitant to leave the Join in here or not, I can do this at consumer side :+1:

pkg/manifests/errmapreader.go
s-urbaniak: ack, good catch :+1:

pkg/manifests/errmapreader.go
```diff
@@ -0,0 +1,86 @@
...
+// errMapReader wraps a map[string]string and exposes read methods
+// for values and slices.
+// It tracks the first occured error in errMapReader.err
+// when reading values from the underlying map.
+type errMapReader struct {
+	src map[string]string
+	err error
+}
+
+func NewErrMapReader(src map[string]string) *errMapReader {
```
squat: This should not be an exported func, plus, it returns an unamable type to callers from other packages.

s-urbaniak: agreed :+1:

pkg/client/client.go
```diff
@@ -222,6 +224,36 @@ func (c *Client) GetProxy(name string) (*configv1.Proxy, error) {
 	return c.oscclient.ConfigV1().Proxies().Get(name, metav1.GetOptions{})
 }
 
+func (c *Client) GetAPIServerAuthenticationConfigmap() (*v1.ConfigMap, error) {
+	return c.kclient.CoreV1().ConfigMaps("kube-system").Get("extension-apiserver-authentication", metav1.GetOptions{})
+}
+
+func (c *Client) DeletePrometheusAdapterAPIAuthenticationSecret(newName string) error {
```
squat: Consider naming this `DeleteOld...` to be clear that is is not deleting the secret with the argument as the name, but an old secret that explicitly is NOT the argument. Otherwise, it sort of looks like it is designed to delete the specified resource.

pkg/tasks/prometheusadapter.go
```diff
@@ -130,7 +130,27 @@ func (t *PrometheusAdapterTask) Run() error {
 		}
 	}
 	{
-		dep, err := t.factory.PrometheusAdapterDeployment()
+		apiAuthConfigmap, err := t.client.GetAPIServerAuthenticationConfigmap()
+		if err != nil {
+			return errors.Wrap(err, "failed to load kube-system/extension-apiserver-authentication configmap")
```
squat: It's a little messy specifying the string `kube-system/extension-apiserver-authentication` which is actually a const in another package. This task knows nothing about that namespace/name combination and should only say something like `failed to load the apiserver authentication configuration`. The specifics of what resource name are not a concern of this package, which is clear because the details are not even known (the package doesn't have the const). Makes sense?

pkg/client/client.go
s-urbaniak: sounds good, I'll fix the naming

pkg/tasks/prometheusadapter.go
s-urbaniak: yep, sounds good :+1:

pkg/client/client.go
```diff
@@ -222,6 +224,36 @@ func (c *Client) GetProxy(name string) (*configv1.Proxy, error) {
 	return c.oscclient.ConfigV1().Proxies().Get(name, metav1.GetOptions{})
 }
 
+func (c *Client) GetAPIServerAuthenticationConfigmap() (*v1.ConfigMap, error) {
```
squat: These two methods are somewhat out of place in this package. Note that no other methods in this package hardcore an explicit namespace or resource name. The fact that these two do indicates that these are utility methods rather than some general purpose client method. Instead, this method should probably be `GetConfigmap(cm *v1.ConfigMap)` or something similar.

pkg/client/client.go
squat: The more I think about it, the more it seems to me that perhaps this func should be part of the prometheus-adapter task rather than the client package. It seems like very specific application logic encoding a core operator task than a general purpose client func. Thoughts?

s-urbaniak: I am fine with factoring those methods into the task itself, agreed.

pkg/client/client.go
s-urbaniak: agreed, factoring this into a more abstract method.

s-urbaniak:
@squat addressed comments, PTAL

s-urbaniak:
@mxinden i just switched temporarily back to master to kick off e2e tests.

s-urbaniak:
oki, all green, reset to `release-4.0`

pkg/client/client.go
```diff
@@ -35,7 +35,7 @@ import (
 	openshiftsecurityclientset "github.com/openshift/client-go/security/clientset/versioned"
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1beta2"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
squat: why did we include this change? the name of this package is already `v1` by default: https://github.com/kubernetes/api/blob/master/core/v1/doc.go#L21

s-urbaniak: hm, good question, this was reformatted automagically by `goimports` :thinking:

pkg/manifests/manifests.go
```diff
@@ -20,16 +20,19 @@ import (
...
+	"strconv"
 	"strings"
 
 	monv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	securityv1 "github.com/openshift/api/security/v1"
+	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1beta2"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
squat: here too

pkg/client/client.go
s-urbaniak: ack, even if I remove that qualifier goimports insists to reinsert it

s-urbaniak:
@squat i edited the files outside of goimports control, PTAL

PR #246, opened 2019-02-13
jsonnet/main.jsonnet: Bump Prometheus to v2.7.1

mxinden:
This version should only give the Prometheus Operator the hint, to configure the given Prometheus instance with command line flags compatible with Prometheus v2.7.1.


simonpasquier:
LGTM but I'm not sure what the [version](https://github.com/openshift/cluster-monitoring-operator/pull/246/files#diff-0e89493bf083b0ef3344b10269a96578R31) in `jsonnet/main.jsonnet` is for.

mxinden:
> I'm not sure what the [version](https://github.com/openshift/cluster-monitoring-operator/pull/246/files#diff-0e89493bf083b0ef3344b10269a96578R31) in `jsonnet/main.jsonnet` is for.

Hope I understand your question correctly: `main.jsonnet` holds the values that are going to be templated, resulting in the manifests in `assets`. The change in `promtheus.yaml` results directly from the change in `main.jsonnet`.

As this is not super urgent, I would wait for @brancz to give this a review as well.

simonpasquier:
@mxinden thanks for the explanation, I missed that `assets/prometheus-k8s/prometheus.yaml` was generated which sounds obvious now that you said it.

brancz:
fwiw this version here is only the version specified in origin, OCP overrides the version through the payload. Even for origin I'm not entirely sure starting with 4.0. Either way it doesn't hurt to update this as it's what we're already using in OCP.

brancz:
looks like this needs a rebase

PR #249, opened 2019-02-14
operator: rotate api authentication certs

s-urbaniak:
Currently, when the extension api authentication certificates are being
rotated, prometheus adapter is not being updated.

This fixes it by watching the extension api authentication configmap in the
kube-system namespace in the cluster monitoring operator directly. When
this configmap is being updated (this happens, when the CAs rotate), a new
secret in the openshift-monitoring namespace is generated with a
synthesized name based on the CA hashes. Finally the prometheus adapter
deployment is being updated and rolled out.

Fixes MON-219

cc @mxinden @squat @metalmatze this is a resubmission of #244 as the 4.0 branch is not respected yet anywhere in the OpenShift build system.


mxinden:
`/test e2e-aws`

pkg/tasks/prometheusadapter.go
```diff
@@ -154,3 +178,29 @@ func (t *PrometheusAdapterTask) Run() error {
...
+	var name string
+	vs := deployment.Spec.Template.Spec.Volumes
+	for i := range vs {
+		if vs[i].Name == "api-auth" && vs[i].Secret != nil {
+			name = vs[i].Secret.SecretName
+			break
+		}
+	}
+
+	if name == "" || name == newName {
```
brancz: I think this might be a bit too fragile, typically pruning logic like this is done by using label selectors. As in list all Secrets that have the label `name: api-auth`, and `not` the `hash: <latest-hash>`. That way even if for whatever reason we miss a rotation we still properly clean up in the future.

s-urbaniak: agreed, I'll add this in a subsequent PR along with the serving certs.

brancz: Sounds perfect.

PR #250, opened 2019-02-15
jsonnet,pkg: watch kubelet CA

squat:
This commit addds a new configmap-reloader container to the Prometheus
K8s pods that watch the configmap holding the CAs used to sign the
Kubelet CSRs. These CSRs are signed by the CSR controller. The CA is
rolled frequently and so must be watched by the
cluster-monitoring-operator and mirrored into the openshift-monitoring
namespace whenever it changes.

These changes fix scraping of all kubelets on worker nodes, however, scraping
master kubelets will be broken until
https://github.com/openshift/cluster-kube-apiserver-operator/pull/247 lands and
makes it into the installer. Once that is in, we can change the CA configmap to
`kubelet-serving-ca`.

cc @s-urbaniak @deads2k @brancz


assets/prometheus-k8s/prometheus.yaml
```diff
@@ -89,6 +90,18 @@ spec:
     - --label=namespace
     image: quay.io/coreos/prom-label-proxy:v0.1.0
     name: prom-label-proxy
+  - args:
```
brancz: I don't think this is necessary. Prometheus reloads the [scape managers every 5 seconds](https://github.com/prometheus/prometheus/blob/b41d6d54f2262f089ff1b7151f67a00fcdfb8e0f/scrape/manager.go#L96), which reloads all scrape pools, and reloading a scrape pool [re-creates the http client](https://github.com/prometheus/prometheus/blob/b41d6d54f2262f089ff1b7151f67a00fcdfb8e0f/scrape/scrape.go#L267), which also re-[reads the TLS configuration](https://github.com/prometheus/prometheus/blob/b41d6d54f2262f089ff1b7151f67a00fcdfb8e0f/vendor/github.com/prometheus/common/config/http_config.go#L255:6). 5 seconds delay is acceptable I think as kubelets reloading their certificates and the secret being re-mounted is racy anyways.

squat: TIL!

s-urbaniak: fwiw, let's persist this fact in a comment inside jsonnet, otherwise it is not obvious why a configmap reloader is not needed.

squat: done and done

pkg/manifests/manifests.go
```diff
@@ -612,6 +613,18 @@ func (f *Factory) PrometheusK8sServingCertsCABundle() (*v1.ConfigMap, error) {
 	return c, nil
 }
 
+func (f *Factory) PrometheusK8sCSRControllerCABundle(data map[string]string) (*v1.ConfigMap, error) {
```
s-urbaniak: mini-nit: i think we are unit-testing all those factory methods. let's add a test for that one too.

sjenning:
kubelet metrics targets are working once again :+1: thanks!

PR #251, opened 2019-02-15
pkg/manifests: move telemeter config to env

squat:
This commit moves certain Telemeter configuration values to environment
variables rather than fields in a secret. This way, the Telemeter pod
will automatically be rolled when the configuration in the CMO configmap
is changed. Without this, the CMO would modify the Telemter secret, but
the pod was never updated because the secret fields were read using the
downward API only once during pod initialization.

part two of: https://github.com/openshift/telemeter/pull/127


squat:
Validated working :)

squat:
This PR has been 100% green twice but prow didn’t merge and the e2e test suite is being quite flaky :/

PR #252, opened 2019-02-16
fix(operatorgroup): add missing static configuration

njhale:
It looks like there are some missing fields on the OperatorGroup manifest. I've added them here:

- olm.providedAPIs: An annotation that contains a comma delimited list of GVKs provided by members of the OperatorGroup
- staticProvidedAPIs: True if OLM should not mutate the `olm.providedAPIs` annotation

Together, these fields prevent other intersecting OperatorGroups from installing any operators that provide the same GVKs, even if there are no CSVs that provide operators in the OperatorGroup's namespace.

The fix that provides this feature was merged early this week, staticProvidedAPIs is an additional field that was added to address CMO's specific usecase.

### DIFF
--- a/manifests/06-operatorgroup.yaml
+++ b/manifests/06-operatorgroup.yaml
@@ -3,6 +3,9 @@ kind: OperatorGroup
 metadata:
   name: openshift-cluster-monitoring
   namespace: openshift-monitoring
+  annotations:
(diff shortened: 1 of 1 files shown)


brancz:
Thanks for doing the fix!

PR #253, opened 2019-02-18
jsonnet: Fix incorrectly referenced oauth-proxy image

smarterclayton:
oauth-proxy was not being replaced by the payload. This sets it to
be equivalent to what is referenced in image-references.


smarterclayton:
Login to prometheus via the console is broken as @enj informed me because of breaking compatibility changes in oauth-proxy.

jsonnet/main.jsonnet
```diff
@@ -75,6 +75,8 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
     namespaceSelector: 'namespace=~"(openshift-.*|kube-.*|default|logging)"',
 
     kubeletPodLimit: 250,
+
+    tlsCipherSuites: ["TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256","TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256","TLS_RSA_WITH_AES_128_CBC_SHA256","TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256","TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256"],
```
brancz: this shouldn't be necessary, you should just need to run `jb install` in the jsonnet directory (the jsonnet-bundler is the jsonnet package manager)

smarterclayton:
Removed tlsConfig and rebased.

PR #259, opened 2019-02-21
manifests: Fix grafana and oauth-proxy image-references

brancz:
These image-references don't seem to align with the images specified in the cluster-monitoring-operator deployment, so this PR fixes that. Hopefully this fixes the last issues regarding getting the right images replaced by the CVO.

https://bugzilla.redhat.com/show_bug.cgi?id=1678645

@squat @s-urbaniak @metalmatze @mxinden @smarterclayton

### DIFF
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -29,11 +29,11 @@ spec:
   - name: grafana
     from:
       kind: DockerImage
-      name: quay.io/openshift/grafana:latest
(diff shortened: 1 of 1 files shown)


smarterclayton:
I should have mentioned the jobs extract out the payload into the artifacts so you can verify pre merge that it does the replace you expect

PR #260, opened 2019-02-22
pkg/operator: set creationTimestamp for PVCs, add e2e tests

s-urbaniak:
Currently, when a volume claim template is specified for either
prometheus or alertmanager, a CRD validation error is generated.

See
https://github.com/coreos/prometheus-operator/issues/2399#issuecomment-466320464
for an envisioned resolution in prometheus operator.

Until then an upstream fix is ready, this is a bandaid.

Fixes MON-579

cc @brancz @mxinden @metalmatze

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -264,6 +264,9 @@ func (c *Client) CreateOrUpdatePrometheus(p *monv1.Prometheus) error {
 	}
 
 	p.ResourceVersion = oldProm.ResourceVersion
+	if p.Spec.Storage != nil {
+		p.Spec.Storage.VolumeClaimTemplate.CreationTimestamp = metav1.Unix(0, 0)
+	}
 	_, err = pclient.Update(p)
 	return errors.Wrap(err, "updating Prometheus object failed")
 }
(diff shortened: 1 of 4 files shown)


s-urbaniak:
cc @ant31

pkg/client/client.go
```diff
@@ -256,6 +256,9 @@ func (c *Client) CreateOrUpdatePrometheus(p *monv1.Prometheus) error {
 	}
 
 	p.ResourceVersion = oldProm.ResourceVersion
+	if p.Spec.Storage != nil {
+		p.Spec.Storage.VolumeClaimTemplate.CreationTimestamp = metav1.Now()
```
brancz: This means we're actually going to be updating every time as in fully modifying and bumping the `resourceVersion`, can we just set this to the 0 timestamp?

s-urbaniak: I tried this, but this effectively becomes the zero value, being marshaled as `null`

brancz: I'm just looking for any value that's static, so 1970-01-01T00:01Z works for me as well.

ant31: isn't that a bit too hacky a fixed date? are we sure there's no side-effect on any of k/k controllers?

brancz: judging by the output, the statefulset controller drops it

s-urbaniak: @brancz doh, of course, let me update the PR.
@ant31 yes, this _is_ a hacky solution but gives us some air to fix prometheus operator.

s-urbaniak:
@brancz updated, PTAL

PR #262, opened 2019-02-22
Fix rule namespace selector

brancz:
I chose to e2e test here that we are in fact picking up our alerting rules, by testing that the `Watchdog` alert is firing.

@metalmatze @mxinden @s-urbaniak @squat 

cc @josefkarasek


test/e2e/main_test.go
```diff
@@ -124,6 +124,9 @@ func TestQueryPrometheus(t *testing.T) {
 		}, {
 			query:   `up{job="crio"} == 1`,
 			expectN: 1,
+		}, {
+			query:   `ALERTS{alertname="Watchdog"} == 1`,
```
mxinden: This only tests, whether rules from the same namespace are loaded, right?

josefkarasek:
Looks great, thanks!

test/e2e/main_test.go
brancz: What does "same" namespace mean, when you request "all" namespaces? Sure the object is in the openshift-monitoring namespace, but the Prometheus object doesn't know that at selection time

brancz: Added another e2e test that creates a new, non `openshift-monitoring` namespace, that should be picked up from.

test/e2e/multi_namespace_test.go
```diff
@@ -0,0 +1,73 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func TestMultinamespacePrometheusRule(t *testing.T) {
+	t.Parallel()
+
+	nsName := "openshift-test-prometheus-rules"
+
+	err := f.OperatorClient.CreateOrUpdateNamespace(&v1.Namespace{
```
mxinden: Given that the namespace name is not randomized, would you mind also removing it at the end of the tests, making sure two consecutive tests run on personal clusters don't influence each other.

brancz: https://github.com/openshift/cluster-monitoring-operator/pull/263

PR #265, opened 2019-02-25
test/e2e: Refactor framework setup & wait for query logic

mxinden:
- Instead of counting the returned time series, inspect the first and only
  returned value from a query. This enables checking specific timeseries values.
  The previous behavior can still be achieved by using the `count` PromQL
  function.

- Introduce `waitForQueryReturn{GreaterEqualOne,One}` helper functions for easier
  access.

- Intialize Prometheus client at test framework creation time, as more function
  will need to use it in the long run.

- Wait for Prometheus k8s before running any tests as multiple of them depend on
  its existence.

I will follow up with a patch checking for firing alerts in a bit. This lays the groundwork for it.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -35,63 +35,76 @@ import (
 var namespaceName = "openshift-monitoring"
 
 type Framework struct {
-	OperatorClient       *client.Client
-	CRDClient            crdc.CustomResourceDefinitionInterface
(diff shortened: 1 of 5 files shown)


brancz:
Sorry it looks like I already merged something and caused a conflict. Otherwise this lgtm.

/lgtm

mxinden:
@brancz rebased. Would you mind taking another look?

PR #268, opened 2019-02-26
pkg/operator: fix progressing state of cluster operator

s-urbaniak:
Currently, we always set the progressing state of the cluster operator
status.

The desired behavior is to set the cluster operator status progressing
status only if the operator is not available or if the available status
has not been set at all (unknown state).

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1669258

cc @mxinden @brancz


brancz:
Thanks for taking care of this!

/lgtm

PR #269, opened 2019-02-26
jsonnet: Update dependencies

brancz:
This pulls in updates from kube-prometheus, primarily the node-exporter tolerating any taint on linux nodes and Grafana dashboard breakdown for rss/cache/swap memory, which had been reported to be confusing if the entire memory is shown as one.

@mxinden @s-urbaniak


assets/prometheus-operator/deployment.yaml
```diff
@@ -20,9 +20,9 @@ spec:
         - --kubelet-service=kube-system/kubelet
         - --logtostderr=true
         - --config-reloader-image=quay.io/coreos/configmap-reload:v0.0.1
-        - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.28.0
+        - --prometheus-config-reloader=quay.io/coreos/prometheus-config-reloader:v0.29.0
         - --namespaces=openshift-monitoring
-        image: quay.io/coreos/prometheus-operator:v0.28.0
+        image: quay.io/coreos/prometheus-operator:v0.29.0
```
mxinden: This pull request also updates the Prometheus Operator. Would you mind adding that to the commit message for future git bisecting? I guess changing the description of the pull request should include it in the merge commit?

brancz: it doesn't actually, this is overridden at runtime

PR #270, opened 2019-02-26
assets: do not set cpu limits

derekwaynecarr:
Do not set cpu limits as it causes more harm than good.

Needed so we can pass:
https://github.com/openshift/origin/pull/22095


squat:
closing in favor of #273

PR #274, opened 2019-02-27
pkg/tasks: create service monitor at the end of each tasks.

s-urbaniak:
This gives the possibility to postpone waiting for the prometheus
operator is actually available before proceeding with the actual
deployment artifacts of the task at hand.

This should give another performance gain during installation time.

Another potential improvement for https://bugzilla.redhat.com/show_bug.cgi?id=1678475

cc @mxinden @brancz @squat

### DIFF
--- a/pkg/tasks/alertmanager.go
+++ b/pkg/tasks/alertmanager.go
@@ -48,16 +48,6 @@ func (t *AlertmanagerTask) Run() error {
 		return errors.Wrap(err, "waiting for Alertmanager Route to become ready failed")
 	}
 
-	smam, err := t.factory.AlertmanagerServiceMonitor()
-	if err != nil {
-		return errors.Wrap(err, "initializing Alertmanager ServiceMonitor failed")
-	}
-
-	err = t.client.CreateOrUpdateServiceMonitor(smam)
-	if err != nil {
-		return errors.Wrap(err, "reconciling Alertmanager ServiceMonitor failed")
-	}
-
 	s, err := t.factory.AlertmanagerConfig()
 	if err != nil {
 		return errors.Wrap(err, "initializing Alertmanager configuration Secret failed")
@@ -129,5 +119,15 @@ func (t *AlertmanagerTask) Run() error {
 	}
 
 	err = t.client.WaitForAlertmanager(a)
-	return errors.Wrap(err, "waiting for Alertmanager object changes failed")
+	if err != nil {
+		return errors.Wrap(err, "waiting for Alertmanager object changes failed")
+	}
+
+	smam, err := t.factory.AlertmanagerServiceMonitor()
+	if err != nil {
+		return errors.Wrap(err, "initializing Alertmanager ServiceMonitor failed")
+	}
+
+	err = t.client.CreateOrUpdateServiceMonitor(smam)
+	return errors.Wrap(err, "reconciling Alertmanager ServiceMonitor failed")
 }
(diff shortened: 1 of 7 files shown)


pkg/tasks/alertmanager.go
```diff
@@ -129,5 +119,19 @@ func (t *AlertmanagerTask) Run() error {
 	}
 
 	err = t.client.WaitForAlertmanager(a)
-	return errors.Wrap(err, "waiting for Alertmanager object changes failed")
+	if err != nil {
+		errors.Wrap(err, "waiting for Alertmanager object changes failed")
```
brancz: missing return

pkg/tasks/clustermonitoringoperator.go
```diff
@@ -53,11 +63,5 @@ func (t *ClusterMonitoringOperatorTask) Run() error {
...
 	}
 
-	cr, err := t.factory.ClusterMonitoringClusterRole()
-	if err != nil {
-		return errors.Wrap(err, "initializing cluster-monitoring ClusterRole failed")
-	}
-
-	err = t.client.CreateOrUpdateClusterRole(cr)
-	return errors.Wrap(err, "reconciling cluster-monitoring ClusterRole failed")
+	return nil
```
squat: for consistency, this should not return nil, but we should return an `errors.Wrap` above on https://github.com/openshift/cluster-monitoring-operator/blob/43a0a62c2733737fc39634ac3c507d9f46d26742/pkg/tasks/clustermonitoringoperator.go#L62, which itself checks for nil.

pkg/tasks/kubestatemetrics.go
```diff
@@ -89,5 +79,19 @@ func (t *KubeStateMetricsTask) Run() error {
...
+	if err != nil {
+		return errors.Wrap(err, "initializing kube-state-metrics ServiceMonitor failed")
+	}
+
+	err = t.client.CreateOrUpdateServiceMonitor(sm)
+	if err != nil {
+		return errors.Wrap(err, "reconciling kube-state-metrics ServiceMonitor failed")
+	}
+
+	return nil
```
squat: same here

pkg/tasks/telemeter.go
```diff
@@ -129,7 +119,21 @@ func (t *TelemeterClientTask) create() error {
...
+	if err != nil {
+		return errors.Wrap(err, "initializing Telemeter client ServiceMonitor failed")
+	}
+
+	err = t.client.CreateOrUpdateServiceMonitor(sm)
+	if err != nil {
+		return errors.Wrap(err, "reconciling Telemeter client ServiceMonitor failed")
+	}
+
+	return nil
```
squat: same with each file

squat:
We should hold off on putting this in 4.0 as it's not a bug fix

pkg/tasks/alertmanager.go
s-urbaniak: whops, good catch

pkg/tasks/clustermonitoringoperator.go
s-urbaniak: yes, i am not a fan of the nil swallowing of `errors.Wrap`, but I'll revert for consistency.

s-urbaniak:
@squat @brancz PTAL
@squat afaik @brancz discussed with the perf team to get this into 4.0

pkg/tasks/clustermonitoringoperator.go
squat: ack, I get it but we should be consistent so reading the codebase is sane :/

s-urbaniak: yep yep agreed :+1:

squat:
> @squat @brancz PTAL
> @squat afaik @brancz discussed with the perf team to get this into 4.0

ack :+1:

s-urbaniak:
@brancz build is green, please take a look :-)

PR #275, opened 2019-02-28
jsonnet: Decrease KubeClientCertificateExpiration expiration threshold

mxinden:
Given that Openshift rotates certificates each issued for ~4h, the
default rule threshold is not correct. With this patch a warning is
fired 1.5h before expiration and a critical alert is fired 1h before
expiration.

I think here the jsonnet mixins really shine. Very cool!


assets/prometheus-k8s/rules.yaml
```diff
@@ -800,17 +800,17 @@ spec:
     - alert: KubeClientCertificateExpiration
       annotations:
         message: A client certificate used to authenticate to the apiserver is expiring
-          in less than 7 days.
+          in less than 0 days.
```
squat: we should change the alert message format to use hours rather than days, as this is not quite helpful right now.

mxinden: Sure, I will prepare a patch upstream in the kubernetes-mixins.

mxinden:
Given that the lowest bucket of the `certificate_expiration_seconds` histogram is [6h](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go#L42) we need to make a patch upstream first. I will do that next.

assets/prometheus-k8s/rules.yaml
mxinden: Xref: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/168

s-urbaniak:
@mxinden since upstream is merged, can you create a downstream cherry-pick too?

mxinden:
> @mxinden since upstream is merged, can you create a downstream cherry-pick too?

@s-urbaniak Downstream is already merged here: https://github.com/openshift/origin/pull/22205, I am just waiting for https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/168 currently.

brancz:
This just needs a regenerate of bindata.go, then good to go.

assets/prometheus-k8s/rules.yaml
```diff
@@ -815,36 +815,58 @@ spec:
...
+          /
+        sum(rate(apiserver_request_count{job="apiserver"}[5m])) by (resource,subresource,verb) * 100 > 5
       for: 10m
       labels:
         severity: warning
     - alert: KubeClientCertificateExpiration
       annotations:
         message: A client certificate used to authenticate to the apiserver is expiring
-          in less than 7 days.
+          in less than 1 hours.
```
mxinden: Once https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/179 merged this will be `1.5 hours`. Even though this is a bit confusing I would like to continue here and get https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/179 into cluster-monitoring-operator in a follow up patch. What do you think?

mxinden:
@squat @s-urbaniak @metalmatze would you mind taking another look? Sorry for the noise on the manifests. Given that I had to update kube-prometheus, it includes unrelated changes as well.

assets/prometheus-k8s/rules.yaml
metalmatze: I just merged kubernetes-monitoring/kubernetes-mixin#179. We can continue with this PR as it is or get 179 in here as well. You decide.

brancz:
I think I mentioned before, but the warning alert is probably not all that useful with such low rotation intervals. We could think about excluding it in the kubernetes-mixin if 0 or filter it out in here. I'm tending a little towards the latter I think, as it's quite specific. I'm happy with either being a follow up though, so this lgtm :+1: .

/lgtm

assets/prometheus-k8s/rules.yaml
mxinden: As it involves a patch on Prometheus Operator as well, let's move on here and do the update as a follow up. Thanks @metalmatze.

PR #277, opened 2019-02-28
hack: Adapt generate etcd secret script to 4.0

brancz:
This adapt the `hack/` script we have to OpenShift 4.0, to immitate what the etcd management should do, so we can test the etcd integration out of band of the team doing this work.

(note there is more to getting etcd working, like the etcd team creating the right `Service` entries, but this is what enables etcd monitoring on our side so sufficient for testing https://bugzilla.redhat.com/show_bug.cgi?id=1662114 for example)

@s-urbaniak @squat @mxinden

### DIFF
--- a/hack/generate-etcd-secret.sh
+++ b/hack/generate-etcd-secret.sh
@@ -4,17 +4,18 @@ set -x
 # only exit with zero if all commands of the pipeline exit successfully
 set -o pipefail
 
-IP=${1}
+NAMESPACE=openshift-kube-apiserver
+APISERVERPODNAME="$(kubectl -n ${NAMESPACE} get pod -lapp=openshift-kube-apiserver -ojsonpath='{.items[0].metadata.name}')"
 
 cat <<-EOF
 apiVersion: v1
-data:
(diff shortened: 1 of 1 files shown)


hack/generate-etcd-secret.sh
```diff
@@ -4,17 +4,18 @@ set -x
 # only exit with zero if all commands of the pipeline exit successfully
 set -o pipefail
 
-IP=${1}
+NAMESPACE=openshift-kube-apiserver
+APISERVERPODNAME="$(kubectl -n ${NAMESPACE} get pod -lapp=openshift-kube-apiserver -ojson | jq -r '.items[0].metadata.name')"
```
s-urbaniak: small nit to reduce the dependency on `jq`:
```
APISERVERPODNAME="$(kubectl -n ${NAMESPACE} get pod -lapp=openshift-kube-apiserver -o jsonpath='{.items[0].metadata.name}')
```

brancz: done

PR #278, opened 2019-02-28
clusteroperator: Report the appropriate injected status version

smarterclayton:
In order to indicate we have reached our desired level, we accept an opaque
token from the payload and pass it through to status only after we have
fully rolled out our new components. This allows the CVO to know that we
have hit the desired state.

See https://github.com/openshift/cluster-image-registry-operator/pull/219 for more context and discussion

PR #280, opened 2019-03-01
*: Cleanup release version plumbing

brancz:
#278 introduced the actual mechanism that reports status as configured by the CVO. As we had previously created some plumbing for setting the version through a compiler flag, we should clean up our mechanism and align it with what's actually being done.

I'm putting a hold label on this, just so we can explicitly verify in the logs, that the right version is being reported. As this is dynamically set via the CVO I have a hard time coming up with a test for this, but if you have an idea, I'm more than happy to add a test.

/hold

@s-urbaniak @mxinden @squat


s-urbaniak:
According to the logs this seems to work:
```
I0301 18:28:43.825083       1 main.go:106] Release version set to 0.0.1-2019-03-01-175845
```

brancz:
Indeed.

/hold cancel

Please review :slightly_smiling_face: .

PR #281, opened 2019-03-01
add cluster instance type rules

tschuy:
For OSD and UHC reports, we need to know CPU information per node role. 

This pull request adds two new metrics. The first is CPUs per instance type/role:

cluster:cluster_capacity_cpu_cores:sum
```
{label_beta_kubernetes_io_instance_type="n1-standard-8",label_node_role_kubernetes_io_compute="true"} 48
{label_beta_kubernetes_io_instance_type="n1-standard-4",label_node_role_kubernetes_io_master="true"} 4
```

The second is node count per instance type/role:

cluster:node_instance_type_count:sum
```
{label_beta_kubernetes_io_instance_type="n1-standard-4",label_node_role_kubernetes_io_master="true"} 1
{label_beta_kubernetes_io_instance_type="n1-standard-8",label_node_role_kubernetes_io_compute="true"} 6
```

This should be sufficient for reports needing CPU counts per cluster, split by compute/master (UHC), and for reports needing node counts per cluster, split by type and compute/master (OSD).

~~This adds a recording rule for capacity_cpu_cores that also includes label_node_role_kubernetes_io_compute/master and the label_beta_kubernetes_io_instance_type.~~

This information is needed in Telemeter. My understanding of the process for this is I need to get recording rules added here first, and then add them as `__name__`s to telemeter. 

~~I've added them in a branch but have not opened a PR for that: https://github.com/tschuy/telemeter/tree/instance_type~~

~~Please let me know if I have any part of this process wrong, or if I need to make changes to my queries.~~


tschuy:
/hold

this doesn't expose the metrics in a way we can use them with machine_cpu_core information. hmm.

tschuy:
I know that you all have been reluctant to add more per-node labels, since it can balloon the amount of data being collected. 

I'm open to proposals for how to get per-role node count and per-role cpu count without adding this. This should also practically replace machine_cpu_cores' purpose, so it would be possible to collect this *as* that, and remove it from telemeter, perhaps.

brancz:
Could you share the result of a query like this? I think the `node` dimension is something we don't need here, we can just aggregate by type of node, and have constant to low cardinality.

brancz:
I tried out some queries and it looks to me that dropping the node dimension would be acceptable, and we would get the sum by each role type instead of on a per node granularity. Ultimately you need to tell though, as you need to work with this data. I don't think per node cardinality is something that is acceptable at this point.

bildschirmfoto vom 2019-03-07 10-40-46

tschuy:
If we go the fully aggregated route, we'll also need the raw count. I've added a second recording rule for that information, and updated the first with your suggestion of summing the CPUs in the query.

I'd like to get a +1 from someone on the UHC side as a verification.

tschuy:
I've also added a pull request to the Telemeter repository. Just the optimist in me hoping that you all might approve and merge this overnight ;) https://github.com/openshift/telemeter/pull/137

tschuy:
Thanks for the update. I'll sit tight until you have that worked out, then.

tschuy:
@squat I'm also not attached to the names I've chosen for these rules. If you have thoughts on better names please add those while you're at it

squat:
cc @smarterclayton

smarterclayton:
Adding some context here.

The minimal requirement is that we identify the CPU time spent on the infrastructure that roughly corresponds to "CPU cores that can't run user workloads".

The definition of "cores for infrastructure" historically was cores running the master and cores running the infrastructure nodes.  In both cases we didn't allow user workloads on the machines, so those were "off limits".  

However, infrastructure nodes create perverse incentives for the project in 4.0 because they encourage users to run poorly-matched workloads (router, elastic search, prometheus) on the same small set of nodes and not take advantage of things like machine compute pools and cluster autoscaling.  Also, we want to target a slightly smaller footprint by default than the previous 8 node recommendation - the 3 / 3 configuration.  Furthermore, we will have even smaller 3 master metal clusters with 16+ cores each that will run user workloads.  So for 4.x we don't want to create infra nodes by default, but we want them to be possible to setup.   We need to handle the 3 master, 3+3 master+worker, and 3+N+M master+worker+misc-infra use cases with some consistency.

For all those reasons, the metric we want to gather is "cores generally reserved for infrastructure".  Admins that label nodes as `node-role.kubernetes.io/infra` is a 3.x signal that could reasonably carry over.  However, in the 3 master cluster with user workloads on masters, selecting only masters won't make sense.  Nor will in a 3+3 cluster accounting for all the cores on both boxes.  This metric can be inaccurate (maybe within 10-20%) as long as we round up (it's better to overestimate infrastructure usage rather than underestimate).

The metric "CPU used by things in namespaces starting with `openshift-*` is actually a pretty good proxy, with some inaccuracies.  Another proxy would be "capacity - (Idle CPU + CPU given to processes *not* in the openshift namespace)".  Tracking master and infra node role CPU is potentially useful, but won't be accurate on 3 master clusters.

squat:
fixed in #291

PR #285, opened 2019-03-04
added SLI metrics for openshift-sre

rafael-azevedo:
This change will add metrics that will be used in reports for SLI/SLOs for openshift-sre


assets/prometheus-k8s/rules.yaml
```diff
@@ -362,6 +362,14 @@ spec:
     rules:
     - expr: sum(openshift_build_total{job="kubernetes-apiservers",phase="Error"})/(sum(openshift_build_total{job="kubernetes-apiservers",phase=~"Failed|Complete|Error"}))
       record: build_error_rate
+  - name: openshift-sre.rules
+    rules:
+    - expr: sum(apiserver_request_count{job="apiserver"}) by (code)
```
brancz: The rule is first rate then sum, as otherwise counter resets cannot be accounted for correctly. We need to sum the rate of requests here, as in:

```
sum(rate(apiserver_request_count{job="apiserver"}[5m])) by (code)
```

And for what it's worth, we already have an [alerting rule for this](https://github.com/openshift/cluster-monitoring-operator/blob/1b2c8054cbe37c686b50301aee2ba05bad8d2c8b/assets/prometheus-k8s/rules.yaml#L780-L799), and we're already reporting all firing alerting rules through telemetry.

assets/prometheus-k8s/rules.yaml
```diff
@@ -362,6 +362,14 @@ spec:
...
+      record: code:apiserver_request_count:sum
+    - expr: kube_pod_status_ready{condition="true",namespace="kube-system",pod=~"etcd.*"}
```
brancz: Could we instead of reporting the raw granularity report the aggregates of each state? As in

```
sum(kube_pod_status_ready{condition="true",namespace="kube-system",pod=~"etcd.*"}) by(condition)
```

That way the cardinality of this is constantly 3 (as [condition is `true|false|unknown`](https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md#pod-metrics)).

assets/prometheus-k8s/rules.yaml
```diff
@@ -362,6 +362,14 @@ spec:
...
+      record: kube_pod_status_ready:etcd
+    - expr: kube_pod_status_ready{condition="true",namespace="openshift-image-registry",pod=~"image-registry.*"}
```
brancz: Same as for etcd pods.

rafael-azevedo:
I removed the last two metrics after speaking with our team, It seems we will be able to get things into telemeter without them. The first metric will still be required I changed that to sum(rate(apiserver_request_count{job="apiserver"}[5m])) by (code)

assets/prometheus-k8s/rules.yaml
rafael-azevedo: updated the metric to sum(rate(apiserver_request_count{job="apiserver"}[5m])) by (code)

assets/prometheus-k8s/rules.yaml
rafael-azevedo: we decided we can grab current metrics to represent this data in telemeter so this will be removed

assets/prometheus-k8s/rules.yaml
rafael-azevedo: we decided we can grab current metrics to represent this data in telemeter so this will be removed same as the etcd pods metric

jsonnet/rules.jsonnet
```diff
@@ -66,6 +66,15 @@
           },
         ],
       },
+      {
+        name: 'openshift-sre.rules',
+        rules: [
+          {
+            expr: sum(rate(apiserver_request_count{job="apiserver"}[10m])) by (code)
```
brancz: both of these need to be quoted I believe (at least CI seems to be complaining)

rafael-azevedo: yup I see I missed the formatting completely !

brancz:
looks like this flaked, but this lgtm

/lgtm
/retest

PR #286, opened 2019-03-06
operator: fetch the pull-secret directly from secret than cluster-config-v1

abhinavdahiya:
The installer already creates a secret in the cluster [1] and there is no requirement to fetch the pull secret from cluster-config-v1 config map.

This also unblocks https://github.com/openshift/installer/pull/1379 to fix https://bugzilla.redhat.com/show_bug.cgi?id=1686119

[1]: https://github.com/openshift/installer/blob/13a752ea0fcae927cba6795782f87ffa332d5b75/data/data/manifests/bootkube/pull.yaml.template

/cc @deads2k

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -17,12 +17,10 @@ package manifests
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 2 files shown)


wking:
/lgtm

(for what it's worth ;).

PR #291, opened 2019-03-19
jsonnet: add cluster resource utilization rules

squat:
This PR adds a number of new recording rules to measure cluster resource capacity and utilization:

1. cluster:capacity_cpu_cores:sum
2. cluster:capacity_memory_bytes:sum
3. cluster:node_instance_type_count:sum
1. cluster:cpu_usage_cores:sum
2. cluster:memory_usage_bytes:sum
3. openshift:cpu_usage_cores:sum
4. openshift:memory_usage_bytes:sum

supersedes #281

Note that due to https://github.com/prometheus/prometheus/issues/3726, a new recording rule was added called `cluster:master_nodes` to be able to group by node type and have a
`label_node_role_kubernetes_io="master"` label where appropriate.

cc @brancz @tschuy @brenton


s-urbaniak:
/approve
I have no concerns, but a 3rd/4th eye would make sense cc @metalmatze @brancz

metalmatze:
lgtm. However, I'm still not 100% confident with the conventions for naming recording rules on my side. So yet another pair of eyes is good.

assets/prometheus-k8s/rules.yaml
```diff
@@ -349,6 +349,30 @@ spec:
...
+        or on(node) (kube_node_labels * on(node) group_left kube_node_status_capacity_memory_bytes))
+        BY (label_beta_kubernetes_io_instance_type, label_node_role_kubernetes_io)
+      record: cluster:capacity_memory_bytes:sum
+    - expr: sum(1 - rate(node_cpu_seconds_total{mode="idle"}[1m]) * on(namespace,
+        pod) group_left(node) node_namespace_pod:kube_pod_info:{})
+      record: cluster:cpu_usage_cores:sum
+    - expr: sum(node:node_memory_bytes_total:sum - node:node_memory_bytes_available:sum)
+      record: cluster:memory_usage_bytes:sum
+    - expr: sum(rate(container_cpu_usage_seconds_total{namespace=~"openshift-.+"}[1m]))
+      record: cluster:infra_cpu_usage_cores:sum
```
brancz: maybe "openshift:cpu_usage_cores:sum" would be more appropriate here

assets/prometheus-k8s/rules.yaml
```diff
@@ -349,6 +349,30 @@ spec:
...
+    - expr: sum(container_memory_working_set_bytes{namespace=~"openshift-.+"})
+      record: cluster:infra_memory_usage_bytes:sum
```
brancz: same as above regarding openshift prefix as opposed to cluster

assets/prometheus-k8s/rules.yaml
```diff
@@ -349,6 +349,30 @@ spec:
     - expr: sum(rate(container_cpu_usage_seconds_total{container_name!="POD",container_name!="",pod_name!=""}[5m]))
         / sum(machine_cpu_cores)
       record: cluster:container_cpu_usage:ratio
+    - expr: kube_node_labels and on(node) kube_node_spec_taint{key="node-role.kubernetes.io/master"}
```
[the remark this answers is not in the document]
squat: :bowing_man:

assets/prometheus-k8s/rules.yaml
squat: ack sounds good

assets/prometheus-k8s/rules.yaml
squat: ack

cben:
To check my understanding: is the usage measured `openshift:cpu_usage_cores:sum`, `openshift:memory_usage_bytes:sum` supposed to _included_ in `cluster:cpu_usage_cores:sum`, `cluster:memory_usage_bytes:sum` respectively?

I'm seeing `openshift:cpu_usage_cores:sum / cluster:cpu_usage_cores:sum` and `openshift:memory_usage_bytes:sum / cluster:memory_usage_bytes:sum` fractions vary widely, mostly between 0.4~1.8 (NOT capped at 1 as I expected), and some even exceed total cluster:capacity!

squat:
@cben your intuition is correct: `openshift:...` should always be part of `cluster:...` and strictly << than it. Can you plz share more info to help us debug, e.g. samples from a running openshift cluster where this issue is present?

smarterclayton:
I think this is going to have to change slightly for node labels but I'm not sure exactly why the current model is chosen.  Nodes can have multiple roles, and we need to be keying off the label, not the taint.  If the gap is that empty label strings are bad, then we should set a value in the label string on clusters.  But for older clusters that won't be possible.

squat:
Keying off taints was used as a quick shortcut to workaround the unusable node role labels with empty values in OpenShift. We used the node role in the value so that we could aggregate over node roles without needing to explicitly enumerate all possible roles in the promql expression. This was a naive assumption to simplify the query. Since this was merged, support to detect the existence of kubernetes node labels was added to Prometheus k8s discovery. We can use this now

PR #295, opened 2019-03-22
pkg/client/condition: respect current and target version

s-urbaniak:
Currently, we ignore the versions of the operator
and only set to progressing if the operator has been failing before.

This fixes it by also also respecting the current version of the operator.
If the target version is the same, the existing rules apply.
If the current version is different, we always set to progressing.

Fixes a part of https://bugzilla.redhat.com/show_bug.cgi?id=1690162
where the operator was launched with a new version but did not report
a progressing state.

cc @brancz @squat @metalmatze

### DIFF
--- a/pkg/client/clusteroperator.go
+++ b/pkg/client/clusteroperator.go
@@ -35,7 +35,7 @@ func (r *StatusReporter) SetDone() error {
 
 	time := metav1.Now()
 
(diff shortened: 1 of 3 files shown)


metalmatze:
/approve

Much :heart: for writing that many tests! :+1:

PR #302, opened 2019-04-02
pkg/manifests/operator: synthesize etcd secret

s-urbaniak:
Currently, we do assume that a kube-etcd-client-certs secret exists
in the same namespace as the cluster monitoring stack.

The reality is etcd client certificates are created in the openshift-config
namespace, separately in a configmap for the etcd client CA and a secret including
the TLS client certificate and key.

This synthesizes the kube-etcd-client-certs secret in the monitoring namespace
or updates it when changes in the underlying etcd client CA happen.
We rely on Prometheus' ability to detect and react to those changes in-process.

cc @brancz @hexfusion


pkg/operator/operator.go
```diff
@@ -112,6 +112,14 @@ func New(config *rest.Config, version, namespace, namespaceSelector, configMapNa
 		UpdateFunc: func(_, newObj interface{}) { o.handleEvent(newObj) },
 	})
 
+	o.openshiftConfigCmapInf = cache.NewSharedIndexInformer(
```
squat: separate from your changes, we have so many informers now with handlers that require getting objects that we should consider using a cache.Store to reduce api load and make our reconciliation faster.

pkg/manifests/manifests.go
```diff
@@ -634,6 +634,42 @@ func (f *Factory) PrometheusK8sEtcdServiceMonitor() (*monv1.ServiceMonitor, erro
 	return s, nil
 }
 
+func (f *Factory) PrometheusK8sEtcdSecret(tlsClient *v1.Secret, ca *v1.ConfigMap) (*v1.Secret, error) {
+	data := make(map[string]string)
+
+	for k, v := range tlsClient.Data {
+		data[k] = string(v)
```
squat: instead of doing these conversions from []byte to string and then back to []byte, consider getting and setting using the `stringData` field of the Secret. This way, you don't even need to construct an intermediate `data` map.

s-urbaniak: agreed, fixed

s-urbaniak: oh wait, StringData can only be used for writing:
```
	// stringData allows specifying non-binary secret data in string form.
	// It is provided as a write-only convenience method.
```

s-urbaniak: ok, at least changed the write-path

squat: ah too bad :/

s-urbaniak:
/hold
I'd like to fully test this on a cluster

s-urbaniak:
verified the following points on a running cluster:
- successful creation of etcd client secrets:
```
$ kubectl -n openshift-monitoring get secret kube-etcd-client-certs
NAME                     TYPE     DATA   AGE
kube-etcd-client-certs   Opaque   3      10m
```
- successful creation of service monitor:
```
$ kubectl -n openshift-monitoring get servicemonitor etcd
NAME   AGE
etcd   10m
```
- successful creation of prometheus targets:

*Note*: the connection is refused, as the proxy is not up yet, as long as https://github.com/openshift/machine-config-operator/pull/517 is merged yet

image

- the etcd Grafana dashboard is rendered, but shows no data yet as the metrics proxy is not running yet:

image

cc @brancz @hexfusion : once the proxy lands, we should be good to go, unless I missed something, PTAL

s-urbaniak:
fwiw I also checked if `etcdInsufficientMembers` is firing, it is

s-urbaniak:
Travis fails with the following unrelated issue :-(
```

E: Failed to fetch http://security.ubuntu.com/ubuntu/dists/trusty-security/main/source/Sources.gz  Hash Sum mismatch
E: Failed to fetch http://security.ubuntu.com/ubuntu/dists/trusty-security/universe/source/Sources.gz  Hash Sum mismatch
E: Failed to fetch http://security.ubuntu.com/ubuntu/dists/trusty-security/main/binary-amd64/Packages.gz  Hash Sum mismatch
E: Failed to fetch http://security.ubuntu.com/ubuntu/dists/trusty-security/universe/binary-amd64/Packages.gz  Hash Sum mismatch
E: Some index files failed to download. They have been ignored, or old ones used instead.
The command "sudo apt-get -qq update" failed and exited with 100 during .
```

paulfantom:
> Travis fails with the following unrelated issue :-(

It looks like updating apt repo list is needed only to install yamllint. Since we are also using docker in CI, it might be better to use a script like this one: https://github.com/openshift/machine-api-operator/blob/master/hack/yaml-lint.sh to run yamllint instead of installing it on a system.

Should I create a PR with this?

s-urbaniak:
@paulfantom sounds great, yes, go for it! :+1:

s-urbaniak:
@hexfusion I also added an e2e test which will fail until the proxy in https://github.com/openshift/machine-config-operator/pull/517 is enabled, do you estimate with merging that soon?

s-urbaniak:
@brancz this is good for a final pass. as noted above, I also enabled the e2e unit test for etcd here, which will fail until the dependency is ready. I fyou wish I can submit the e2e in a subsequent PR, but I'd prefer to leave this open until all requisites are merged, otherwise we have defunct grafana etcd dashboards/firing etcd alerts/failing etcd targets.

brancz:
That sounds good to me. Let's wait until the dependencies are "fixed". (note for CI to pass we'll need to clean the namespace to get the "fresh" dependency versions, I don't remember how to do this, you'll need to ask Clayton)

/lgtm
/hold

s-urbaniak:
Independent of CI I can confirm this is working very nicely on my local cluster:

image

image

s-urbaniak:
@brancz do you mind to re-ptal? this is looking fine ... at least on my machine :tm:

s-urbaniak:
cc @hexfusion thanks a lot for the great collaboration here!! :bowing_man:

brancz:
Loooooooove this :+1: 

/lgtm

hexfusion:
Great work fellas, happy day!!!

PR #304, opened 2019-04-03
Patch recording rules to only pending and running cpu & memory requests

metalmatze:
This patches the recording rules for requests CPU and memory to only take pending and running Pods into account.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1691893

/cc @brancz @squat @s-urbaniak @mxinden @paulfantom


squat:
xref: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/180

PR #313, opened 2019-04-08
Add SessionAffinity to Prometheus and Alertmanager services

metalmatze:
To avoid confusion with graphs being load balances to different Prometheuses and for Alertmanager getting a 404 page when creating a silence because the state has not yet been gossiped, we want to set a session affinity based on client IP to both services.

https://bugzilla.redhat.com/show_bug.cgi?id=1697295

/cc @mxinden @brancz @paulfantom @s-urbaniak @squat


brancz:
this needs a re-gen of the bindata file

PR #315, opened 2019-04-09
Bug 1697644: operator: use openshift-config/pull-secret to fetch the pull secret for cloud.openshift.com

abhinavdahiya:
The operator should be using the openshift-config/pull-secret as the kube-system/coreos-pull-secret has been marked deprecated.

/cc @derekwaynecarr @smarterclayton

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -365,7 +365,7 @@ func (o *Operator) Config(key string) *manifests.Config {
 		}
 
 		err = c.LoadToken(func() (*v1.Secret, error) {
-			return o.client.KubernetesInterface().CoreV1().Secrets("kube-system").Get("coreos-pull-secret", metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


pkg/operator/operator.go
```diff
@@ -104,6 +105,14 @@ func New(config *rest.Config, version, namespace, namespaceSelector, configMapNa
 		UpdateFunc: func(_, newObj interface{}) { o.handleEvent(newObj) },
 	})
 
+	o.openshiftConfigCmapInf = cache.NewSharedIndexInformer(
+		o.client.ConfigMapListWatchForNamespace("openshift-config"),
```
squat: Why are we watching for configmap changes in this namespace?

abhinavdahiya: Oh, i brain fart. didn't see that the informer was for configmap.

is there any other informer that needs setup so that the operator is responding to changes to this secret or no sync was setup for previous secret location too.

abhinavdahiya:
Friendly ping @squat

PR #321, opened 2019-04-11
Update cluster-monitoring-operator image

carlosedp:
With previous version (v0.0.6) when the CMO got deployed into a `oc cluster up` cluster, Grafana was not deployed. With version v0.1.1, the monitoring stack was correctly deployed.

This was tested on a "Openshift on Footloose" that is a Docker in Docker deployment for it. Ref. https://github.com/carlosedp/openshift-on-footloose


brancz:
As I mentioned this is only the image that was built when 3.11 was initially released, it does not include the fixes available in that branch. You should build an image from `HEAD` of the branch yourself.

/lgtm

PR #331, opened 2019-04-19
Port changes from #330 to master branch - Add replicas config

carlosedp:
Add configmap option to define the number of replicas for Prometheus and Alertmanager StatefulSets.

### DIFF
--- a/Documentation/user-guides/configuring-cluster-monitoring.md
+++ b/Documentation/user-guides/configuring-cluster-monitoring.md
@@ -8,7 +8,7 @@ Configuring Cluster Monitoring is optional. If the config does not exist, or is 
 
 In certain environments it may be required that container images are downloaded from a custom registry rather than from the canonical container image repositories on [quay.io][quay].
 
-This is an example configuration with all image parameters set to a custom registry:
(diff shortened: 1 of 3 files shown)


brancz:
Can you explain why this should be configurable? In my opinion this is a part of the stack that is intended to be 100% managed, thus not user configurable.

carlosedp:
The idea is to be able to deploy it with a small resource requirement for non-HA scenarios like a LAB, POC, DEMO or testing environment.

brancz:
I'd say that's out of scope for this project and not something we'd like to support. I'd say if this is something you want to do, then disable the cluster-monitoring-operator explicitly and change the replicas yourself. That way we ensure people know they're explicitly moving in unsupported territory.

carlosedp:
Ok, no problem since it's not aligned with the product strategy. I created this to be used on my https://github.com/carlosedp/openshift-on-footloose project mainly focused on the cases I pointed previously.

The idea is having an environment as close as possible to production but with a small set of components.

carlosedp:
Any news on this? Should I close?

brancz:
Yeah, I don't think we are gonna move forward with this.

PR #332, opened 2019-04-23
pkg/tasks/prometheusadapter: deploy aggregated cluster role

s-urbaniak:
Currently, only cluster wide admins have the permissions to view metrics 
resources. This fixes it by adding a read-only cluster role which includes 
aggregation labels to synthesize permission rules for standard user-facing 
roles according to [1].

Note that only the "pods" resource is granted as reading "nodes" metrics 
requires a cluster wide permission.

[1] https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1701343

### DIFF
--- a/manifests/02-role.yaml
+++ b/manifests/02-role.yaml
@@ -9,6 +9,7 @@
 # 	assets/kube-state-metrics/cluster-role.yaml
 # 	assets/kube-state-metrics/role.yaml
 # 	assets/node-exporter/cluster-role.yaml
+# 	assets/prometheus-adapter/cluster-role-aggregated-metrics-reader.yaml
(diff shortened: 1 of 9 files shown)


manifests/02-role.yaml
```diff
@@ -223,6 +224,14 @@ rules:
...
   verbs:
   - use
+- apiGroups:
+  - metrics.k8s.io
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
```
brancz: wow I'm thankful for this script

PR #333, opened 2019-04-24
*: Change ClusterOperator failing status to degraded

brancz:
This is the API change that was recently merged.

@s-urbaniak @squat @metalmatze @mxinden @paulfantom


s-urbaniak:
nit: do you mind to split the actual code changes and the vendoring in separate commits?

PR #336, opened 2019-04-24
pkg/manifests: Update etcd namespace

brancz:
I haven't actually found the pull request that changed this, but looking at new clusters, this seems to be where etcd lives.

@s-urbaniak @squat @metalmatze @mxinden @paulfantom


s-urbaniak:
I think we have to look further :-( According to CI
```
	prometheus_client.go:181: failed to extract first value from query response for query "max(up{job=\"etcd\"})": expected body to contain single timeseries but got 0
```

deads2k:
I was about to make this change (just found it).  It happened here https://github.com/openshift/machine-config-operator/pull/648 .

Be aware that certs are moving too https://github.com/openshift/installer/pull/1666

wking:
/hold

[Sounds like][1] this needs adjustments before it can get through CI?  I'm dropping a hold so the restest-bot doesn't pound away at this overnight (although it's been quiet so far).  Pull the hold whenever (maybe I'm misunderstanding the earlier comment).

[1]: https://github.com/openshift/cluster-monitoring-operator/pull/336#issuecomment-486362383

s-urbaniak:
/hold cancel

just a quick confirmation point that the etcd target is now up with this PR:
image

The CI failure must have had stale payloads I guess.

PR #340, opened 2019-04-30
manifests: Remove etcd gRPC calls failed alerts

brancz:
These alerts are firing constantly due to some issues around how to etcd
increases its metrics. See https://github.com/etcd-io/etcd/issues/10289

@metalmatze @squat @s-urbaniak @paulfantom @mxinden


pierreozoux:
I'm using OpenShift 3.11, do you know when this will land there?
I just want to make sure it will land with the next version of OpenShift.

zhigwang:
I have the same question, when will this fix be available to use in production (3.11)?

PR #343, opened 2019-05-01
plumb a resource version for SCC updates

deads2k:
Prereq for https://bugzilla.redhat.com/show_bug.cgi?id=1698625

https://github.com/openshift/origin/pull/22737 indicates that when we try to use a CRD handler for SCC, the behavior around resourceVersion and unconditional update isn't quite the same.  We can see 

```
    message: 'Failed to rollout the stack. Error: running task Updating node-exporter
      failed: reconciling node-exporter SecurityContextConstraints failed: updating
      SecurityContextConstraints object failed: securitycontextconstraints.security.openshift.io
      "node-exporter" is invalid: metadata.resourceVersion: Invalid value: 0x0: must
      be specified for an update'
```
from https://storage.googleapis.com/origin-ci-test/pr-logs/pull/22737/pull-ci-openshift-origin-master-e2e-aws/8266/artifacts/e2e-aws/must-gather/cluster-scoped-resources/config.openshift.io/clusteroperators/monitoring.yaml

(nice message btw).

We may eventually deal with that, but for now we need to unblock our move since it relates to storage and cannot be completed post-GA.  This sets a resourceVersion into a copy of the SCC that we're updating.  That update becomes conditional, but the chances of a conflict are fairly low and the operator should reconcile anyway.

@sallyom

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -178,7 +178,7 @@ func (c *Client) WaitForPrometheusOperatorCRDsReady() error {
 
 func (c *Client) CreateOrUpdateSecurityContextConstraints(s *secv1.SecurityContextConstraints) error {
 	sccclient := c.ossclient.SecurityV1().SecurityContextConstraints()
-	_, err := sccclient.Get(s.GetName(), metav1.GetOptions{})
+	existing, err := sccclient.Get(s.GetName(), metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		_, err := sccclient.Create(s)
 		return errors.Wrap(err, "creating SecurityContextConstraints object failed")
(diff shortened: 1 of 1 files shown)


pkg/client/client.go
```diff
@@ -187,7 +187,11 @@ func (c *Client) CreateOrUpdateSecurityContextConstraints(s *secv1.SecurityConte
 		return errors.Wrap(err, "retrieving SecurityContextConstraints object failed")
 	}
 
-	_, err = sccclient.Update(s)
+	// the CRD version of SCC appears to require this.  We can try to chase why later.
+	required := s.DeepCopy()
+	required.ResourceVersion = existing.ResourceVersion
```
deads2k: you can see this done in this file for other resources like https://github.com/openshift/cluster-monitoring-operator/pull/343/files#diff-fb87476fa66849c8514ba700e715b4b6R270

deads2k:
I've manually approved this one to get us unstuck.

@brancz @s-urbaniak feel free to get up with me tomorrow if you have any questions or concerns.

smarterclayton:
This looks ok to me.

0x0 implies that resource version is being set as an int64, not a string.

smarterclayton:
If so, that's something that some client libraries can cause.

enj:
They are using client-go, do we need to do anything to make sure they are using json and not proto?

deads2k:
On further review it cannot be proto because we got a good message.

deads2k:
AWS failed on known flake fixed by https://github.com/openshift/library-go/issues/381 the rest passed. It won't block merge.

PR #344, opened 2019-05-02
client: force JSON client for security client

mfojtik:
As SCC is moving to CRD resource we need to force the `application/json` content type for the client that create the SCC manifests in order for it to create the SCC.

/cc @deads2k 
/cc @sallyom


mfojtik:
/cc @openshift/sig-monitoring 

This blocks the auth team making progress on 4.1 blocker bug.

PR #361, opened 2019-05-28
grafana: Set default admin user to impossible openshift user

brancz:
Currently Grafana can be hijacked by creating a user in OpenShift called
"admin", which is the default admin user in Grafana having superuser
privileges in Grafana, allowing that user to modify dashboards, when
they should be entirely read only.

@s-urbaniak @squat @paulfantom @metalmatze 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1668441#c18


jsonnet/grafana.jsonnet
```diff
@@ -58,6 +58,13 @@ local authorizationRole = policyRule.new() +
             http_addr: '127.0.0.1',
             http_port: '3001',
           },
+          security: {
+            // OpenShift users are limited to 63 characters, with this we are
+            // setting the Grafana user to something that can never be created
+            // in OpenShift. This prevents users from getting proxied with an
+            // identity that has superuser permissions in Grafana.
+            admin_user: 'WHAT_YOU_ARE_DOING_IS_VOIDING_SUPPORT_0000000000000000000000000000000000000000000000000000000000000000',
```
s-urbaniak: additionally: as we can't modify grafana, we could theoretically modify the proxy in front of it. Would it make sense additionally to respond with 403 if this user is trying to access it?

brancz: We're using the openshift oauth proxy. Are you aware of functionality to disable certain identities? I'd like that as well.

s-urbaniak: I am not aware of that. I suggest to merge this in and submit a feature request towards oauth proxy.

s-urbaniak:
I am actually fine with this, let's have another pair of eyes on this though, @metalmatze @paulfantom @squat

PR #384, opened 2019-06-20
Bump jsonnet & prometheus-operator migration

brancz:
Bumping these manifests includes a change where the prometheus operator changes it's `.spec.selector`, which is why an additional migration is necessary.

Among others, this bumps the kubernetes-mixin to 1.14 compatibility closing MON-665.

@paulfantom @s-urbaniak @squat @metalmatze


brancz:
CI failure looks like a timeout, the cluster-monitoring-operator was still in the middle of rolling out (1/2 prometheus rolled out and 2/3 alertmanagers). It doesn't look like this is a problem in the cmo.

/retest

PR #393, opened 2019-06-28
Remove low quality dashboards

brancz:
This removes the accidentally pulled in dashboards.

Random caveat, in jsonnet to use `std.setMember` the array passed in as the "set" must be sorted (added a comment as it actually turns out this is the second time I encountered this here, but I didn't quite understand what was happening last time, but now I know, although I wish I didn't).

@squat @paulfantom


jsonnet/main.jsonnet
```diff
@@ -115,7 +115,8 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
   grafanaDashboards:: {
     [k]: d[k]
     for k in std.objectFields(d)
-    if !std.setMember(k, ['nodes.json', 'persistentvolumesusage.json', 'pods.json', 'statefulset.json'])
+                         // This array must be sorted for `std.setMember` to work.
```
squat: 🤦‍♀️

PR #395, opened 2019-07-01
manifests: Allow configuring remote write endpoint

brancz:
This adds the ability to configure remote write endpoints for the cluster-monitoring Prometheus server to replicate its data to a remote write backend.

@squat @sichvoge

### DIFF
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -607,6 +607,8 @@ func TestPrometheusK8sConfiguration(t *testing.T) {
       memory: 750Mi
   externalLabels:
     datacenter: eu-west
+  remoteWrite:
+  - url: "https://test.remotewrite.com/api/write"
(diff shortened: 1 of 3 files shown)


pkg/manifests/manifests_test.go
```diff
@@ -686,6 +689,16 @@ ingress:
 	if storageRequestPtr.String() != "15Gi" {
 		t.Fatal("Prometheus volumeClaimTemplate not configured correctly, expected 15Gi storage request, but found", storageRequestPtr.String())
 	}
+
+	if p.Spec.RemoteWrite[0].URL != "https://test.remotewrite.com/api/write" {
+		t.Fatal("Prometheus external labels are not configured correctly")
```
squat: Should the error say something about remote-write instead?

brancz: yes

PR #396, opened 2019-07-04
Add gosec for security code analysis

JAORMX:
This adds a "test-sec" target to the Makefile, which will trigger gosec. gosec will do static analysis on the go code to check for common security issues.

This PR also tags some low priority warnings so they don't appear as errors when triggering this test.


brancz:
/lgtm

Thanks!

PR #398, opened 2019-07-11
 Added "code:registry_api_request_count:rate:sum" recording rule 

abyrne55:
OpenShift SRE needs to be able to track the rates of API requests to the image registry service for Service Level Agreement compliance. The existing "code:apiserver_request_count:rate:sum" rule doesn't differentiate between the service each API call is going to, so it isn't specific enough for this purpose.

To remedy this, I added another recording rule that selects only the API call counters that correspond to the registry service, and then performs a rate-and-sum similar to "code:apiserver_request_count:rate:sum"


abyrne55:
Seems like CI got rate-limited. Requesting a /retest

assets/prometheus-k8s/rules.yaml
```diff
@@ -425,6 +425,9 @@ spec:
     rules:
     - expr: sum(rate(apiserver_request_count{job="apiserver"}[10m])) BY (code)
       record: code:apiserver_request_count:rate:sum
+    - expr: sum(rate(apiserver_request_count{job="apiserver",resource=~"image.*",verb!="WATCH"}[10m]))
```
brancz: If it is intentional to only record the image resource here then that should be reflected in the name of the recording rule.

brancz: :facepalm: you already did that. Out of curiosity, why specifically this API?

abyrne55: Managed OSDv4 instances have an SLO regarding "the proportion of successful responses to API requests to the registry as measured from Prometheus"

smarterclayton: This query doesn't select only requests made by the registry.  Lots of other components in openshift make many calls to the API.  

What are you trying to measure here?  What you're actually measuring is "how many API requests are made to things called image" which includes many things the registry has no impact on.

abyrne55: AFAIK, the goal here is to measure "the proportion of successful responses to API requests to the registry as measured from Prometheus" 
@jewzaam might be able to explain the objective with this SLI better than I can though

PR #400, opened 2019-07-16
Bug 1734540: etcd should still alert when a member disappears from the endpoints

smarterclayton:
Upon observation there are a number of clusters reporting a single degraded etcd member, but without an alert highlighting that behavior. Our current alerting reports when quorum is lost, but a single member being down is still important because you can no longer tolerate another master failure.  Add alert `EtcdMembersDown` that counts the number of suspected down members.

Further, since it is possible if the etcd metrics are collected under a Kubernetes service and the etcd instance is removed from the endpoint (due to implementation details of how etcd is run), extend the alert to also capture failure rate of requests to that cluster above a minimum threshold. This might potentially report a false positive, but a non-zero peer failure rate likely represents a serious failure.  Use the max of either the metrics from an instance, or the service up == 0 measurement to get a best guess of the number of down members.

QUESTION: I followed the convention for the rest of the alerts, but noticed that etcd alerts are not consistent naming wise (camelCase not CamelCase).  Is there a bug / item tracking fixing this?


smarterclayton:
Still working on the alert, do not review

smarterclayton:
@hexfusion as per discussion yesterday, this will allow SRE to use the alert vs having to add their own metric (tracking `up{service="etcd"}` by itself was not sufficient as we discovered in prod and in testing)

metalmatze:
Ultimately this should end up upstream in https://github.com/etcd-io/etcd/tree/master/Documentation/etcd-mixin. Once we merged it there we only need to update the etcd dependency of the jsonnet configuration.

smarterclayton:
Is the upstream going to be changed to be consistent with other alert name conventions?

smarterclayton:
Opened https://github.com/etcd-io/etcd/pull/10906

hexfusion:
Upstream is merged etcd-io/etcd#10906

smarterclayton:
Updated from latest etcd (I only bumped that dependency, because it picked up a more dramatic set).  This includes the new alert `etcdMembersDown`.

smarterclayton:
To recreate, `oc debug node/MASTER_1`, `mv /etc/kubernetes/manifests/etcd-member.yaml /tmp/`, go to dashboard:

image

PR #405, opened 2019-07-18
Makefile: fix openshift-state-metrics role rules

squat:
This commit fixes the rules added by the openshift-state-metics rule. It
seems like this file was built by hand instead of using the build
script. This is a result of the renaming of the generated role file: now
that the file has a different name, the make target was wrong.

This fixes the name of the target in the Makefile and regenerates the
file.

cc @paulfantom @wanghaoran1988


crawford:
:+1: I ran into this recently as well.

PR #406, opened 2019-07-18
metrics: Add ingress related recording rules for OpenShift

smarterclayton:
Summarize the key metrics for the platform for USE on the ingress
controllers - specifically the error rates for HTTP based traffic
and the incoming and outgoing bandwidth, and session concurrency.

@ironcladlou


ironcladlou:
Looks great to me, this is a big head start for us

jsonnet/rules.jsonnet
```diff
@@ -96,6 +96,27 @@
...
+            expr: 'sum by (code) (rate(haproxy_server_http_responses_total[5m]) > 0)',
+            record: 'code:cluster:ingress_http_request_count:rate:sum',
+          },
+          {
+            expr: 'sum by (frontend) (rate(haproxy_frontend_bytes_in_total[5m]))',
+            record: 'frontend:cluster:ingress_frontend_bytes_in:rate:sum',
+          },
+          {
+            expr: 'sum by (frontend) (rate(haproxy_frontend_bytes_out_total[5m]))',
+            record: 'frontend:cluster:ingress_frontend_bytes_out:rate:sum',
```
brancz: each of these should be `...rate5m...`, otherwise this looks good to me

smarterclayton: fixed

PR #415, opened 2019-07-29
Bug 1727024: Add OperatorUpgradeable status to StatusReporter

s-urbaniak:
Currently we don't report that status. This fixes it.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1727024

/cc @LiliC @paulfantom

### DIFF
--- a/pkg/client/condition.go
+++ b/pkg/client/condition.go
@@ -41,6 +41,11 @@ func newConditions(cos v1.ClusterOperatorStatus, targetVersion string, time meta
 			Status:             v1.ConditionUnknown,
 			LastTransitionTime: time,
 		},
+		v1.OperatorUpgradeable: {
+			Type:               v1.OperatorUpgradeable,
+			Status:             v1.ConditionUnknown,
+			LastTransitionTime: time,
+		},
 	}
 
 	for _, c := range cos.Conditions {
--- a/pkg/client/status_reporter.go
+++ b/pkg/client/status_reporter.go
@@ -48,6 +48,7 @@ func (r *StatusReporter) SetDone() error {
 	conditions.setCondition(v1.OperatorAvailable, v1.ConditionTrue, "Successfully rolled out the stack.", time)
 	conditions.setCondition(v1.OperatorProgressing, v1.ConditionFalse, "", time)
 	conditions.setCondition(v1.OperatorDegraded, v1.ConditionFalse, "", time)
+	conditions.setCondition(v1.OperatorUpgradeable, v1.ConditionTrue, "", time)
 	co.Status.Conditions = conditions.entries()
 
 	// If we have reached "level" for the operator, report that we are at the version
@@ -89,6 +90,10 @@ func (r *StatusReporter) SetInProgress() error {
 
 	conditions := newConditions(co.Status, r.version, time)
 	conditions.setCondition(v1.OperatorProgressing, v1.ConditionTrue, "Rolling out the stack.", time)
+	conditions.setCondition(v1.OperatorUpgradeable, v1.ConditionFalse,
+		"Rollout of the monitoring stack is progressing. Please wait until it finishes.",
+		time,
+	)
 	co.Status.Conditions = conditions.entries()
 	co.Status.RelatedObjects = newRelatedObjects(r.namespace)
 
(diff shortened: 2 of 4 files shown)


s-urbaniak:
/hold

let me add some more tests for the status reporter.

pkg/client/status_reporter.go
```diff
@@ -89,6 +90,7 @@ func (r *StatusReporter) SetInProgress() error {
 
 	conditions := newConditions(co.Status, r.version, time)
 	conditions.setCondition(v1.OperatorProgressing, v1.ConditionTrue, "Rolling out the stack.", time)
+	conditions.setCondition(v1.OperatorUpgradeable, v1.ConditionFalse, "", time)
```
LiliC: In the API docs for `OperatorUpgradeable`, it says we should set a reason for when this is false:

> and the message field should contain a human readable description of what the administrator should do to allow the operator to successfully update.

pkg/client/status_reporter.go
```diff
@@ -112,6 +114,7 @@ func (r *StatusReporter) SetFailed(statusErr error) error {
 	conditions.setCondition(v1.OperatorAvailable, v1.ConditionFalse, "", time)
 	conditions.setCondition(v1.OperatorProgressing, v1.ConditionFalse, "", time)
 	conditions.setCondition(v1.OperatorDegraded, v1.ConditionTrue, fmt.Sprintf("Failed to rollout the stack. Error: %v", statusErr), time)
+	conditions.setCondition(v1.OperatorUpgradeable, v1.ConditionFalse, "", time)
```
LiliC: Same as above, we should set the msg here to why this is not upgradable currently.

pkg/client/status_reporter.go
s-urbaniak: great catch :+1: fixed

pkg/client/status_reporter.go
s-urbaniak: fixed

s-urbaniak:
@LiliC addressed comments, PTAL

pkg/client/condition.go
```diff
@@ -41,6 +41,11 @@ func newConditions(cos v1.ClusterOperatorStatus, targetVersion string, time meta
 			Status:             v1.ConditionUnknown,
 			LastTransitionTime: time,
 		},
+		v1.OperatorUpgradeable: {
+			Type:               v1.OperatorUpgradeable,
+			Status:             v1.ConditionUnknown,
```
LiliC: So we are fine with setting this to default as `ConditionUnknown` as this means more or less that it's possible to upgrade?

pkg/client/status_reporter.go
```diff
@@ -89,6 +90,10 @@ func (r *StatusReporter) SetInProgress() error {
 
 	conditions := newConditions(co.Status, r.version, time)
 	conditions.setCondition(v1.OperatorProgressing, v1.ConditionTrue, "Rolling out the stack.", time)
+	conditions.setCondition(v1.OperatorUpgradeable, v1.ConditionFalse,
+		"Rollout of the monitoring stack is progressing. Please wait until it finishes.",
```
LiliC: Nit:
```suggestion
		"Rollout of the monitoring stack is in progress. Please wait until it finishes.",
```

pkg/client/condition.go
s-urbaniak: This is just the constructor for the transient set of states in this state machine. The actually persisted states are as follows (as is asserted in the tests):

1. "In Progress" (aka initally persisted state):
"Available": "Unknown",
"Degraded": "Unknown",
"Progressing": "True",
**"Upgradeable", "False"** <- this is debatable, we could also leave this as "Unknown", but we probably don't want to upgrade during rollout.

2. "Done" (aka happy case):
"Available", "True",
"Degraded", "False",
"Progressing", "False",
**"Upgradeable", "True"**

3. "Failed" (aka unhappy case):
"Available", "False",
"Degraded", "True",
"Progressing", "False",
**"Upgradeable", "False"**

pkg/client/status_reporter.go
s-urbaniak: oh good catch :+1:

PR #420, opened 2019-07-31
Bug 1730413: etcd should still alert when a member disappears from the endpoints

smarterclayton:
There is no alert that reports when an etcd quorum member is degraded.
Add an alert that reports if any etcd member is down or potentially
unreachable. Since it is possible if the etcd metrics are collected
under a Kubernetes service and the etcd instance is removed from the
endpoint (due to implementation details of how etcd is run), extend
the alert to also capture failure rate of requests to that cluster.


LiliC:
The shellcheck test is failing because it's missing the `hack/shellcheck ` file, this is fixed in master, but you need to rebase.

smarterclayton:
@LiliC PTAL (green)

LiliC:
Was not in the OWNER file in 4.1 branch, so can't approve this.

cc @s-urbaniak @squat

s-urbaniak:
ping @smarterclayton @eparis this needs an explicit cherry-pick-approval I think :thinking:

smarterclayton:
Yeah this weeks patch manager will do it.

PR #423, opened 2019-08-01
pkg/client,pkg/manifests: bump to k8s.io/api/apps/v1

s-urbaniak:
The Daemonset, and Deployment API promoted to apps/v1
but we are still referencing v1beta2.

This fixes it

/cc @LiliC


s-urbaniak:
oh, i need to update go modules.

LiliC:
Tried to find out why this was done in the first place, but it seems like this commit comes from the very first commit which included the entire project when it was open-sourced.

I don't see a problem with merging this, but since I think it doesn't fix anything urgent maybe would be good to wait what @brancz says when he is back from PTO, if there is a historical reason for this. WDYT? Also since its not really a fix, should we aim this for 4.3 instead?

s-urbaniak:
@LiliC sounds good, let's rebase this to 4.3 :+1:

PR #425, opened 2019-08-01
jsonnet/rules: Correctly account for infrastructure usage

smarterclayton:
This changes the recording rules for CPU and memory use related to
infrastructure to fix two issues:

1. double counting of containers and pod memory/CPU accounting
2. not accounting for node level services and kernel CPU

The new algorithm is to calculate infrastructure use by subtracting
user workload (!~"openshift-.+") from cluster usage. This accounts
for everything that is not end user workloads and more accurately
reflects what belongs to cluster vs user.

```
cluster:cpu_usage_cores:sum = total CPU usage of entire cluster
workload:cpu_usage_cores:sum = total CPU usage of all pods in non openshift-* namespaces
openshift:cpu_usage_cores:sum = all CPU not accounted for by workloads
```

Similar for mem_usage_bytes


smarterclayton:
Ready for review, @fabianofranz @luciddreamz

smarterclayton:
This will be a candidate for back port to 4.1 so we can get accurate infrastructure CPU.

fabianofranz:
/lgtm
@brancz @squat for approval

smarterclayton:
Yes, I’ll be taking that as follow up

PR #432, opened 2019-08-12
assets: Update telemeter deployment

LiliC:
@smarterclayton This should solve the problem of you not seeing the `workload:cpu_usage_cores:sum`, seems like telemeter was updated to whitelist but the change was not propagated to CMO.

Do we want all of the below changes?

cc @s-urbaniak @paulfantom @brancz


smarterclayton:
Thanks

smarterclayton:
Can we bump this in 4.1 as well?  Bug # is going to be 1732945

LiliC:
> Can we bump this in 4.1 as well? Bug # is going to be 1732945

@smarterclayton so for that to happen I opened https://github.com/openshift/telemeter/pull/221 and https://github.com/openshift/telemeter/pull/222

Once those two get merged (one just needs the label added) can open a PR to cherry-pick this to CMO release-4.1 branch.

(Also note had to change the bugzilla 1732945 to ASSIGNED status)

PR #441, opened 2019-08-15
Bug 1712912: jsonnet/rules: fix double accounting of metrics

s-urbaniak:
Backports: https://github.com/openshift/cluster-monitoring-operator/pull/153
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1712912


s-urbaniak:
/cc @brancz @LiliC @paulfantom 
Note: This does not introduce new metrics semantics (working set bytes), as for this the underlying recording rule has to be changed. This just fixes the double accounting in 3.11.

PR #448, opened 2019-08-22
Add Alertmanager trusted CA bundle support

LiliC:
Should be ready for review:

TODO:
- [x] Test local cluster again, we had some flakes with this once.
- [ ] Document what we want to do in the future to mount to a specified location.
- [x] Remove when other PR is merged https://github.com/openshift/cluster-monitoring-operator/pull/447

### DIFF
--- a/assets/alertmanager/trusted-ca-bundle.yaml
+++ b/assets/alertmanager/trusted-ca-bundle.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+data:
+  ca-bundle.crt: ""
+kind: ConfigMap
+metadata:
+  labels:
+    config.openshift.io/inject-trusted-cabundle: "true"
+  name: alertmanager-trusted-ca-bundle
+  namespace: openshift-monitoring
--- a/jsonnet/alertmanager.jsonnet
+++ b/jsonnet/alertmanager.jsonnet
@@ -24,6 +24,11 @@ local authorizationRole = policyRule.new() +
 {
   alertmanager+:: {
 
+    trustedCaBundle:
+      configmap.new('alertmanager-trusted-ca-bundle', { 'ca-bundle.crt': '' }) +
+      configmap.mixin.metadata.withNamespace($._config.namespace) +
+      configmap.mixin.metadata.withLabels({ 'config.openshift.io/inject-trusted-cabundle': 'true' }),
+
     // OpenShift route to access the Alertmanager UI.
 
     route: {
--- a/pkg/manifests/bindata.go
+++ b/pkg/manifests/bindata.go
@@ -9,6 +9,7 @@
 // assets/alertmanager/service-account.yaml
 // assets/alertmanager/service-monitor.yaml
 // assets/alertmanager/service.yaml
+// assets/alertmanager/trusted-ca-bundle.yaml
 // assets/cluster-monitoring-operator/cluster-role.yaml
 // assets/cluster-monitoring-operator/service-monitor.yaml
 // assets/cluster-monitoring-operator/service.yaml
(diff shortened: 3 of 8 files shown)


LiliC:
Note: For this to work we need to make some modifications in prometheus-operator to allow specifying custom mounts for ConfigMaps.

LiliC:
e2e test error:

>     alertmanager_test.go:161: no volume alertmanager-trusted-ca-bundle-cquddrmb6dfoh mounted

So it does find the ConfigMap and the hashed ConfigMap, but the volume is not mounted. 🤷‍♀

LiliC:
cc @paulfantom 
PTAL thanks :)

gfvirga:
It looks like that the path 

```
:~/cluster-monitoring-operator$ oc rsh -c alertmanager  alertmanager-main-0
sh-4.2$ curl https://oauth-openshift.apps.ose.company.com
curl: (60) Peer's Certificate issuer is not recognized.
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.

sh-4.2$ ls /etc/pki/alertmanager-ca-bundle/ca-bundle.crt 
/etc/pki/alertmanager-ca-bundle/ca-bundle.crt
```

```
update-ca-trust 
p11-kit: couldn't create file: /etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt: Permission denied
p11-kit: couldn't create file: /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem: Permission denied
p11-kit: couldn't create file: /etc/pki/ca-trust/extracted/pem/email-ca-bundle.pem: Permission denied
p11-kit: couldn't create file: /etc/pki/ca-trust/extracted/pem/objsign-ca-bundle.pem: Permission denied
p11-kit: couldn't create file: /etc/pki/ca-trust/extracted/java/cacerts: Permission denied
```

Should the line https://github.com/openshift/cluster-monitoring-operator/blob/2211aba6346c8536b958f2b5223f60312323fa2b/pkg/manifests/manifests.go#L367 be changed to /etc/pki/ca-trust/extracted/ ? Also I am guessing we need to add the configmap to the proxy sidecar.

s-urbaniak:
^ cc @LiliC

LiliC:
@gfvirga Hello, what is the reason for running the curl and then the `update-ca-trust` command? :)

We specify a different location than usual, but this works because the update of the CA is not done via that command but via updating the ConfigMap that has the CA bundle. Note that this location needs to be also specified in the Alertmanger configuration as well. Hopefully that answers any questions, otherwise feel free to ask some more! :)

gfvirga:
LiliC, 
The curl command is showing the the additional CAtrust bundle is not available for alertmanager. When trying to authenticate via oauth we are getting 
```
oc logs -c alertmanager-proxy alertmanager-main-2 | grep x509
2019/10/28 12:38:13 oauthproxy.go:645: error redeeming code (client:10.128.0.1:39918): Post https://oauth-openshift.apps.osesbx.mtb.com/oauth/token: x509: certificate signed by unknown authority
```

I was able to fix this by manually changing the path for the additional CA trust Bundle to "/etc/pki/ca-trust/extracted" (Not sure if it is the right approach, I copied it from the openshift-console deployment)

The reason for the update-ca-trust command was to show that it is not picking it up the additional ca bundle from the Config Map 

What is the Alertmanager configuration that needs to be set?

gfvirga:
LiliC,
The curl command is showing the the additional CAtrust bundle is not available for alertmanager. When trying to authenticate via oauth we are getting

oc logs -c alertmanager-proxy alertmanager-main-2 | grep x509
2019/10/28 12:38:13 oauthproxy.go:645: error redeeming code (client:10.128.0.1:39918): Post https://oauth-openshift.apps.ose.company.com/oauth/token: x509: certificate signed by unknown authority
I was able to fix this by manually changing the path for the additional CA trust Bundle to "/etc/pki/ca-trust/extracted" (Not sure if it is the right approach, I copied it from the openshift-console deployment)

The reason for the update-ca-trust command was to show that it is not picking it up the additional ca bundle from the Config Map

What is the Alertmanager configuration that needs to be set?

LiliC:
What configmap are your refereeing to? What version of openshift cluster are you using, might better to open a bugzilla or go through customer support for this, so we have more information.

The entire content of the trusted CA bundle ConfigMap that is created is mounted in the alertmanager container. If the rotation is done via the ca bundle ConfigMap, but for you, it looks like you did it manually and that is why it did not work.

gfvirga:
Sorry for not being well detailed.
I installed the latest Openshift 4.2 version. And I used the variable "additionalTrustBundle:" to add our internal intermediate and root chains.
I found this PR that it is mounting the TrustBundle under "/etc/pki/alertmanager-ca-bundle/" which it seems that is not being cached by the container, perhaps because of the path, the curl command fails when I try to use the CAtrustBundle. 

Not only that, the sidecar proxy container does not receive the bundle at all, we need to add the CAtrustBundle there too.
```
$ oc rsh -c alertmanager alertmanager-main-0
sh-4.2$ ls -l /etc/pki/alertmanager-ca-bundle/
total 0
lrwxrwxrwx. 1 root root 20 Oct 27 10:46 ca-bundle.crt -> ..data/ca-bundle.crt
sh-4.2$ exit
```
```
$ oc rsh -c alertmanager-proxy alertmanager-main-0
sh-4.2$ ls -l /etc/pki/alertmanager-ca-bundle/
ls: cannot access /etc/pki/alertmanager-ca-bundle/: No such file or directory
```

# Quote
> The entire content of the trusted CA bundle ConfigMap that is created is mounted in the alertmanager container. If the rotation is done via the ca bundle ConfigMap, but for you, it looks like you did it manually and that is why it did not work.

This change was applied to my current installation without me manually modifying anything does not work properly. I get the x509 errors and it breaks authentication.

I will open the bugzilla, I've always had more luck talking here, than using the support. (We pay for support)

If you would like to schedule something to talk about it over the phone and screen share, please let me know.

gfvirga:
https://bugzilla.redhat.com/show_bug.cgi?id=1766181

LiliC:
@gfvirga Which cluster version, can you do `oc version` please so i can recreate this.

I tested on
`OpenShift Version: 4.2.2` and get the following:

```
oc rsh -n openshift-monitoring alertmanager-main-0
Defaulting container name to alertmanager.
Use 'oc describe pod/alertmanager-main-0 -n openshift-monitoring' to see all of the containers in this pod.
sh-4.2$ ls /etc/pki/alertmanager-ca-bundle/
ca-bundle.crt
sh-4.2$
```

Thanks! We are looking int the proxy stuff as part of the bugzilla you opened.

gfvirga:
LiliC you are getting the same result as me, I am running 4.2.0.

Can you check if the ca-bundle is being loaded in the proxy with oc rsh -n openshift-monitoring -c alertmanager-proxy alertmanager-main-0 and then run "trust list" look for your certificate.

gfvirga:
I believe the linux container will not use /etc/pki/alertmanager-ca-bundle/ without modification, right now it is not working for me unless it is located under /etc/pki/ca-trust/extracted/

LiliC:
So the file was there correctly in the alertmanager pod, if I understand you correctly?

> Can you check if the ca-bundle is being loaded in the proxy with oc rsh -n openshift-monitoring -c alertmanager-proxy alertmanager-main-0 and then run "trust list" look for your certificate.

Yes, @paulfantom is working on the proxy problem. 

Thanks again for all the feedback and reporting the problem and sorry for running into these issues! We are trying to reproduce and working on it! :)

PR #449, opened 2019-08-23
Bug 1743257: pkg/client/status_reporter.go: Set OperatorUpgradeable to true

LiliC:
We should not set OperatorUpgradeable to false during rollout.
As per the docs:
An operator reports Upgradeable as false when it wishes to prevent
an upgrade for an admin-correctable condition. The component should
include a message that describes what must be fixed.

See docs for [this](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md
). From what I understand we should *never* really set upgradable to false. As this prevents an upgrade and even if we fail with any reason in CMO operator, we will always be able to upgrade as well we just reconcile to the desired state anyways. But I could be missing something?

CVO seems to be backporting the new ClusterOperatorFlapping to 4.1 but from what I could tell we did not backport the Upgradeable status to 4.1 in CMO?

cc @s-urbaniak @paulfantom 

https://bugzilla.redhat.com/show_bug.cgi?id=1743257

### DIFF
--- a/pkg/client/status_reporter.go
+++ b/pkg/client/status_reporter.go
@@ -106,7 +106,7 @@ func (r *StatusReporter) SetInProgress() error {
 	reasonInProgress := "RollOutInProgress"
 	conditions := newConditions(co.Status, r.version, time)
 	conditions.setCondition(v1.OperatorProgressing, v1.ConditionTrue, "Rolling out the stack.", reasonInProgress, time)
-	conditions.setCondition(v1.OperatorUpgradeable, v1.ConditionFalse,
+	conditions.setCondition(v1.OperatorUpgradeable, v1.ConditionTrue,
 		"Rollout of the monitoring stack is in progress. Please wait until it finishes.",
 		reasonInProgress,
 		time,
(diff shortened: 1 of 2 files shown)


pkg/client/status_reporter.go
```diff
@@ -136,7 +136,7 @@ func (r *StatusReporter) SetFailed(statusErr error, reason string) error {
 	conditions.setCondition(v1.OperatorAvailable, v1.ConditionFalse, "", "", time)
 	conditions.setCondition(v1.OperatorProgressing, v1.ConditionFalse, "", "", time)
 	conditions.setCondition(v1.OperatorDegraded, v1.ConditionTrue, fmt.Sprintf("Failed to rollout the stack. Error: %v", statusErr), reason, time)
-	conditions.setCondition(v1.OperatorUpgradeable, v1.ConditionFalse,
+	conditions.setCondition(v1.OperatorUpgradeable, v1.ConditionTrue,
```
LiliC: This is for when we call `SetFailed `, see description of this PR why I changed this as well. But could use double checking of my logic here :)

s-urbaniak:
Thanks @LiliC :tada: From my perspective this is now correct behavior.

Just a nit: https://bugzilla.redhat.com/show_bug.cgi?id=1740312 doesn't seem to be the correct BZ.

s-urbaniak:
holding to be able to fix BZ reference

LiliC:
Thanks! Too many bugzillas 🤦‍♀

LiliC:
/hold cancel

Fixed bugzilla link.

PR #450, opened 2019-08-26
Bug 1712293: assets/prometheus-adapter: use pod_name:container_memory_usage_bytes:sum recording rule

s-urbaniak:


s-urbaniak:
@LiliC afaik, we'll have to create a 4.1.z BZ, depend that on https://bugzilla.redhat.com/show_bug.cgi?id=1712293, and then we can backport.

s-urbaniak:
as discussed OOB with @brancz we change console to use working set bytes instead.

PR #451, opened 2019-08-26
Bug 1724248: Schedule prometheus-operator on master nodes

LiliC:
We want to make sure that the prometheus-operator is scheduled on the master nodes so that it can be created early enough so that other operators can use its CRDs.

cc @s-urbaniak @paulfantom @brancz 

https://bugzilla.redhat.com/show_bug.cgi?id=1724248
https://jira.coreos.com/browse/MON-732


squat:
Have we already changed the order in which CVO creates the CMO? Last I heard the CVO creates the CMO last (or close to last) in the install process, so this change may be moot

LiliC:
@squat Actually the master nodes are created about 9 minutes ahead of the worker nodes and in the events I seen the prometheus-operator pending as the nodes cannot be found due to (lack of) tolerations. And so the CMO is created and we wait for about 5 minutes or so additionally for the prometheus-operator to be created. (At least on the cluster I looked at). So I don't think this change is completely wrong.

As for the order, TBH we have a very strange order of creation, as we sometimes race with OLM so I would not bump that up sooner. The CVO is creating its prom CR resources in [90](https://github.com/openshift/cluster-version-operator/tree/master/install) and CMO is created in [50](https://github.com/openshift/cluster-monitoring-operator/tree/master/manifests) To be honest no idea if this is enough, but its a start? 🤷‍♀

LiliC:
/test e2e-aws-operator
/test e2e-aws

Unrelated failures....

droslean:
/hold 

until tests will pass

LiliC:
/hold cancel

🍏 it's all green

PR #458, opened 2019-08-29
Bug 1746793: manifests: Bump operators.coreos.com API version to v1

LiliC:
We have been seeing errors when trying to create OperatorGroup resource when using the old API version, I think that API bug might already be fixed in 4.2, but nonetheless we can bump to using the `v1` API. See [Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1740838) for background.

I asked the OLM devs and they said there should be no differences between the two versions so we are safe to bump to using the newer API version.

cc @s-urbaniak @paulfantom


LiliC:
I am not sure but maybe we should have a ticket for this in Jira or something so that QE can be done? WDYT?

s-urbaniak:
@LiliC let's simply create a BZ for this then.

PR #459, opened 2019-08-29
Bug 1746686: jsonnet/telemeter: fix client reload port 

s-urbaniak:
/cc @squat @LiliC @paulfantom


squat:
xref: https://github.com/openshift/telemeter/pull/229

s-urbaniak:
seems e2e-aws-operator failure is unrelated:
```
--- FAIL: TestAlertmanagerVolumeClaim (320.13s)
    alertmanager_test.go:86: waiting for StatefulsetRollout of alertmanager-main: statefulset alertmanager-main is not ready.
```

LiliC:
@s-urbaniak I have seen this error before in a PR recently, I think we might have a race in our tests, where the Alertmanager is not ready. It passed after retest, but not sure if we should keep track of these flakey failures somewhere?

s-urbaniak:
@LiliC ah interesting, ok, i'll submit a jira so we don't forget.

PR #461, opened 2019-08-29
jsonnet: Add terminationMessagePolicy to all pods

smarterclayton:
By setting fallback to logs on error we get a record of the last chunk
of stderr output for each pod when they crash, which is then captured
both by CI logging (the event monitor prints this, which makes debugging
upgrade / crash bugs much faster) as well as the insights operator which
will automatically grab this output. All cluster operators set this out
of the box, there's no downside for most services setting this as well.


smarterclayton:
Updated

PR #462, opened 2019-08-29
Capture sum of OS (RHCOS or RHEL) with version and role for the nodes.

chambridge:
* Intended to be utilized for subscription reconciliation when cluster has mixed set of nodes (bring your own worker)


smarterclayton:
What’s this going to be used for?  I thought we needed cores by role / os, not just count?

jsonnet/rules.jsonnet
```diff
@@ -102,6 +102,10 @@
               severity: 'critical',
             },
           },
+          {
+            expr: 'sum((cluster:master_nodes * on(node) group_left kube_node_labels) or on(node) (kube_node_info * on(node) group_left kube_node_labels)) BY (node, kernel_version, os_image, label_node_openshift_io_os_id, label_beta_kubernetes_io_instance_type, label_node_role_kubernetes_io)',
```
brancz: We don’t need the first part of the query, as master nodes are included in the kube_pod_info metric. The second half aggregates by the “node” label values which causes the cardinality of this recording rule to be the cardinality of number of nodes, which I don’t think you intended to do as these are meant to be aggregates. I think you want to remove the node label from the by clause.

smarterclayton:
/hold

We are going to have a discussion on monday, we might want to just add OS version to the existing node count.

jsonnet/rules.jsonnet
chambridge: Can do that but I thought that was how to introduce `label_node_role_kubernetes_io` into the output.

I only have CodeReadyContainers deployment but the output looks like this:
```
{kernel_version="4.18.0-80.7.2.el8_0.x86_64",node="crc-56mmj-master-0",os_image="Red Hat Enterprise Linux CoreOS 410.8.20190807.0 (Ootpa)"}    1
```

smarterclayton:
I think the metric we need to report is a variant of `cluster:node_instance_type_count:sum` but instead of instance type, split by `node role label` and `os version` and returning cores.  So should be called `node_role:cpu_capacity_cores:sum` (I think? correct me if I'm wrong Frederic) and have a label for `node_role_...master` and a label for `os_version`.

smarterclayton:
Actually, we also need to capture architecture in the same series.

brancz:
`node_role_os_version_machine:cpu_capacity_cores:sum` would be appropriate (machine being the machine's architecture).

chambridge:
@smarterclayton should I update this PR to be `node_role_os_version_machine:cpu_capacity_cores:sum` and capture the arch per the comment above or open a new PR?

chambridge:
output looks like the following:
```
{kernel_version="4.18.0-80.7.2.el8_0.x86_64",label_beta_kubernetes_io_arch="amd64",label_node_openshift_io_os_id="rhcos",node="crc-56mmj-master-0",os_image="Red Hat Enterprise Linux CoreOS 410.8.20190807.0 (Ootpa)"}	4
```

LiliC:
I think there was a problem with the rebase, seems like there are commits from master included in this PR.

chambridge:
yeah. Just going to kill this and  reopen. Don't know what happened 🤷‍♂ .

PR #465, opened 2019-09-03
pkg/tasks/tasks.go: Return error only when its not nil

LiliC:
This way we only return errors when they are present, so we can bubble
up the error and actually return the task error to the call function.

cc @s-urbaniak @paulfantom


LiliC:
We need to backport this to 4.1, as the problem is present there as well. Wonder if we can do it all part of the https://github.com/openshift/cluster-monitoring-operator/pull/440 wdyt?

s-urbaniak:
sounds good :+1:

PR #467, opened 2019-09-04
Bump prometheus-operator release to 0.32

LiliC:
We need these changes to use the new `Volumes` object in Alertmanager Custom Resource.

cc @s-urbaniak @paulfantom


assets/prometheus-k8s/rules.yaml
```diff
@@ -1049,17 +1049,6 @@ spec:
       for: 4h
       labels:
         severity: warning
-    - alert: PrometheusTSDBWALCorruptions
```
LiliC: Hmm, not sure why this was removed? Let me try to not include this...

PR #472, opened 2019-09-09
Use Alertmanager Volume instead of the ConfigMap

LiliC:
This also needed a bump to the latest release of prometheus-operator `v0.32` that also included that change. Because the client-go version was bumped in that release, some adjustments had to be made to the pinned deps in the `go.mod` file.

Some small refactoring work, to reuse the same generation for both telemeter and alertmanager.

Note: Still need to test this, opening early to see if tests pass, hence WIP.


LiliC:
/hold

Sadly this does not work, need to make some changes in prometheus-operator for this to work. Currently I am seeing the Alertmanager Statefulset we create have duplicated containers with name `alertmanager`, because of this the Statefulset only rolls out two pods out of three, no idea why it works for two pods but not the third one, but don't have time to look into this just yet. Will focus on making the change in prometheus-operator to be able to merge the containers so we end up with only one container named alertmanager, but merge the VolumeMounts into it.

LiliC:
Once https://github.com/coreos/prometheus-operator/pull/2755 gets merged and we do a release of prometheus-operator, bump here, etc. the tests should pass. :) 

I tested the whole thing and it works. But still on hold until the above PR gets merged.

LiliC:
Seems like a flaky test, could not delete ns.


/test e2e-aws

LiliC:
/remove lgtm

lets see if this works :D 

so the tests don't get retriggered

LiliC:
@s-urbaniak can you have another look, should be fine now :) Thanks!

pkg/manifests/manifests.go
```diff
@@ -2146,3 +2123,27 @@ func (f *Factory) HashTrustedCA(caBundleCM *v1.ConfigMap, prefix string) *v1.Con
 		},
 	}
 }
+
+func trustedCABundleVolumeMount(name string) v1.VolumeMount {
+	return v1.VolumeMount{
+		Name:      name,
+		ReadOnly:  true,
+		MountPath: "/etc/pki/ca-trust/extracted/ca-bundle/",
```
s-urbaniak: looks good, but we should verify in `#warroom-proxy` if this is a correct place.

PR #474, opened 2019-09-10
add /dev/vd* volume device name prefix in scraping rules

thpham:
This would be very useful to have to extend this device filter.

Thank you

CC. /assign @brancz


thpham:
Hello, I just notice this PR is a duplicate of the [PR-247](https://github.com/openshift/cluster-monitoring-operator/pull/247). The use case is the same and target specific device mapping on openstack. Any chance to make it merged for the release-3.11 branch ?

brancz:
These are auto generated from https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/d445c4d98fdf88fd3c59bb34ca4b0f82536f878c/rules/rules.libsonnet#L279, which as you can see are hard coded in that version. We would need to branch off on upstream at the version that we used here and make it configurable and then set that configuration (or add this to the hard coded values).

thpham:
Does it make sense to keep it auto-generated on the build process for this maintenance branch ?

brancz:
Yes I think so. There are still other sources of truth.

thpham:
@brancz , ok ! but could you explain me for example why this [PR-441](https://github.com/openshift/cluster-monitoring-operator/pull/441) has been accepted and merged ? which basically does the same !  maybe I'm missing something ... but why all this unnecessary complexity within the `bindata.go ` file. I might think it kind of wrong open source mindset, ... but I might be wrong. What do you think ?

LiliC:
@thpham hello, Frederic is on PTO. Hopefully, I can answer the questions. The PR you are referencing had a valid bugzilla, to be able to backport bugs you need a valid bugzilla, its the current process we have at openshift. This cannot be merged otherwise.

>  but why all this unnecessary complexity within the bindata.go file. 

Thats how the files get to the operator, it's the setup we have. It's not unnecessary complexity from our side, but I can see how it can be for someone who first starts to look at the project. :) 

Are you using openshift 3.11, or just cluster-monitoring-operator? :) We would be happy to help out and get the fix you need!

thpham:
@LiliC yes sorry you're right, I should have read the code a bit more :-)

Yes, I'm using openshift 3.11, the use case is that I deploy the cluster over openstack and the volume device name mapping is not included in the filter.
I updated the PR with my fork of the `kubernetes-monitoring/mixin`. But I think you might want to fork it yourself internally...

Let me know what to do next ! Thx

LiliC:
@thpham sorry for the late reply, busy times :) 

So I would rather go the approach of adding this to the hardcoded values, instead of forking of the upstream repo. Less to maintain that way. Do you mind making the adjustments for that? Thanks!

thpham:
@LiliC no worries.

But I could not hardcode those changes in the assets because they will be rewritten by the upstream repo (`jsonnet/jsonnetfile.lock.json`) when your CI build the version (`.travis.yml`). and the final image wont have the fix. That's what @brancz wanted to explain [here](https://github.com/openshift/cluster-monitoring-operator/pull/474#issuecomment-529852697)

s-urbaniak:
@thpham if you feel this is still needed, please submit a bugzilla ticket against https://bugzilla.redhat.com/. We unfortunately cannot merge things into 3.11 at free will. If you have a valid customer case, please also ping our PM @sichvoge on that bugzilla so we can assert priority and severity, thank you!

For the time being closing out.

PR #481, opened 2019-09-19
Bump openshift-state-meterics libsonet file

wanghaoran1988:


manifests/0000_50_cluster_monitoring_operator_02-role.yaml
```diff
@@ -240,6 +240,13 @@ rules:
   - securitycontextconstraints
   verbs:
   - use
+- apiGroups:
```
LiliC: Does the cluster-monitoring-operator really need this? If this is a change only in openshift-state-metrics, why do we need to update the CMO role as well.

brancz: the cluster-monitoring-operator needs the role in order to grant it onwards, this is correct I believe

LiliC: aha interesting, I thought that the operator just needs a role to be able to create roles, and nothing else?

brancz: nope, this protects against privilege escalation, otherwise as soon as you had the right to grant roles you would be able to make yourself cluster admin

LiliC: 😈

PR #488, opened 2019-09-23
Adds new tc, checks terminationMessagePolicy.

wmuha:


test/e2e/prometheus_test.go
```diff
@@ -167,3 +167,35 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
 }
+
+func TestPrometheusTerminationMessagePolicy(t *testing.T) {
+	pods, err := f.KubeClient.CoreV1().Pods(f.Ns).List(
+		metav1.ListOptions{FieldSelector: "status.phase=Running"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	var (
+		testNs = "openshift-monitoring"
```
LiliC: This should be `f.Ns` we don't need to specify this.

test/e2e/prometheus_test.go
```diff
@@ -167,3 +167,35 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+		clOper = "cluster-monitoring-operator"
+		v      string
```
LiliC: This doesn't need to be initialized here.

test/e2e/prometheus_test.go
```diff
@@ -167,3 +167,35 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+
+	for _, p := range pods.Items {
+		if strings.Contains(p.Namespace, testNs) &&
+			strings.Contains(p.Name, clOper) {
+			for _, msg := range p.Spec.Containers {
+				v = string(msg.TerminationMessagePolicy)
+				if strings.Contains(v, "FallbackToLogsOnError") {
+					msgOk = true
+				} else {
+					t.Fatal("Can not find terminationMessagePolicy message.")
```
LiliC: Something like that, because we are checking for this specific msg.

```suggestion
					t.Fatal("Could not find 'terminationMessagePolicy' message: FallbackToLogsOnError.")
```

test/e2e/prometheus_test.go
```diff
@@ -167,3 +167,35 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
 		t.Fatal("Can not find pods: prometheus-k8s or alertmanager-main")
 	}
 }
+
+func TestPrometheusTerminationMessagePolicy(t *testing.T) {
+	pods, err := f.KubeClient.CoreV1().Pods(f.Ns).List(
+		metav1.ListOptions{FieldSelector: "status.phase=Running"})
```
LiliC: Can't we just select Pod Name here to be "cluster-monitoring-operator", that way we don't need the if condition for namespace and pod name on line 185.

test/e2e/prometheus_test.go
wmuha: Thanks. Fixed

test/e2e/prometheus_test.go
wmuha: Fixed

test/e2e/prometheus_test.go
wmuha: Fixed

test/e2e/prometheus_test.go
wmuha: Thanks for the idea. Let me investigate that path please.

wmuha: I have no idea have how to do that. Try different cases without result. The problem is that cluster operator name  joined with randomly generated suffix and I had could not get it without knowing this random part.

wmuha: But If you know something, you are welcome to ping me )

wmuha: /assign @LiliC

LiliC: Since it should always be just one pod, you can just use a `.Get` as well.

lilic:
@wmuha are you interested in finishing up this PR, if so is it ready for review? :)

PR #491, opened 2019-09-26
Bug 1755242: jsonnet/rules: fix memory_usage_bytes recording rule

s-urbaniak:
Currently cluster:memory_usage_bytes:sum references
node:node_memory_bytes_total:sum and node:node_memory_bytes_available:sum.

These are not available any more.

This fixes it.


LiliC:
Ah we should merge this first in master I think, then bring it into 4.2, right?

s-urbaniak:
@LiliC it is already in master. i am still unsure how to proceed here, trying to clarify.

PR #489, opened 2019-09-24
Bump master to 4.3 branch

s-urbaniak:
This bumps `master` to be in sync with the current changes in `master-4.3`.
Once merged, we can delete `master-4.3` for good.

/cc @LiliC @paulfantom @brancz

PR #492, opened 2019-09-26
Pin dependencies to master

LiliC:
Now that release is over, we want to pin the dependencies to master again. This PR also updates all the dependencies and brings in the changes.

Also added a note on the process before an openshift release happens. As per @s-urbaniak suggestion.

cc @paulfantom @brancz


LiliC:
cc @s-urbaniak @brancz updated to include the prom-adapter fix, PTAL, thanks.

s-urbaniak:
from the latest e2e:
```
    prometheusadapter_test.go:155: number of running pods (137) doesn't match number of pods reporting metrics (63)
```
not zero, but still a mismatch :thinking:

s-urbaniak:
hmm, i can't see anything obvious, the pod count seems to be oscillating between 140 and 180:
image

s-urbaniak:
let's retrigger to make sure it's not a flake.

LiliC:
Could this be due to the 1.16 depreciation of APIs?

paulfantom:
I don't think any of used here APIs are deprecated in 1.16. Source: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

LiliC:
https://github.com/openshift/cluster-monitoring-operator/pull/489/commits/32bb312d1b2bb57e2b2bf3ff753a67b81b9cf3c4

PR #502, opened 2019-10-11
Use access to Alertmanager CR to gate access to AM

cblecker:
This changes the permissions needed to access Alertmanager's UI, and the ability to view its config or create silences. Instead of having to rely in the ability to see all namespaces, this changes the permissions to the ability to view the Alertmanager CR. This way, access can be specifically granted to users based on the ability to view that CR.

Also fix a makefile syntax issue on systems (such as Darwin) that use a different file path for the make binary.

cc: @RiRa12621 @brancz @sichvoge 
ref: https://jira.coreos.com/browse/SREP-1983

PR #506, opened 2019-10-15
assets/cluster-monitoring-view: add alertmanager policy

s-urbaniak:
This adds the alertmanager policy rule to the existing cluster-monitoring-view
cluster role such that users can access alertmanager with the new subject access review
policy added in https://github.com/openshift/cluster-monitoring-operator/pull/502.

/cc @LiliC @paulfantom @brancz


s-urbaniak:
(cc @cblecker)

assets/cluster-monitoring-operator/cluster-role.yaml
```diff
@@ -9,3 +9,9 @@ rules:
   - namespaces
   verbs:
   - get
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - alertmanagers
+  verbs:
+  - get
```
brancz: The SubjectAccessReview is only against the `openshift-monitoring` namespace: https://github.com/openshift/cluster-monitoring-operator/pull/502/files#diff-c4e970fc1be688ce39664190175bfd86

s-urbaniak: I think we cannot express that in a cluster role. That cluster role can be bound to a (namespaced) role limiting it, but afaik we cannot express such behavior here.

Alternative: Introduce a dedicated `alertmanager-view` Role (not ClusterRole) which has to be explicitly bound to the grantee.

/cc @cblecker: can you give more insight/direction on what path you wanted to go with this?

brancz: I think `alertmanager-access` was previously suggested. I think that's the right way to go. It might be slightly breaking, but I think it's for the better.

cblecker: Yes, I had created a `alertmanager-access` role for this purpose in my last PR: https://github.com/openshift/cluster-monitoring-operator/pull/502/files#diff-8a39df569f081fc563dde843fae1075e

The thing is, because this is looking at a CR, I don't think we expect or will act on instances of the CR that are created outside of the openshift-monitoring namespace, will we? I think in effect, it won't make a difference. I'm also okay changing the SAR to be a cluster-wide one (removing the namespace specifically) and make it more clear.. again, because we're looking at a custom resource (and not `secret get`) I think there is flexibility in how we express this.

s-urbaniak:
closing as the role already exists.

PR #507, opened 2019-10-15
Bug 1761887: tasks/prometheusoperator: Change deployment selector migration strategy

brancz:
It's not actually 100% clear whether this fixes the issues seen in https://bugzilla.redhat.com/show_bug.cgi?id=1761887, but it adds more logging so that we could see whether the migration was attempted at all when we expect it as I feel when we run into the problem described in the bug we are for some reason not attempting the migration, so changing the strategy with this.

@LiliC @s-urbaniak @paulfantom

### DIFF
--- a/pkg/tasks/prometheusoperator.go
+++ b/pkg/tasks/prometheusoperator.go
@@ -15,11 +15,14 @@
 package tasks
 
 import (
+	"reflect"
+
 	"github.com/openshift/cluster-monitoring-operator/pkg/client"
 	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
 )
 
 type PrometheusOperatorTask struct {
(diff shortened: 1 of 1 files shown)


s-urbaniak:
sounds good, we need to do a fix in master first i think :-/

s-urbaniak:
(to comply to the bugzilla process)

brancz:
@s-urbaniak https://github.com/openshift/cluster-monitoring-operator/pull/504 fixed this on master, as the migration is just not necessary for 4.3 anymore (it was a 4.1 -> 4.2 specific migration).

s-urbaniak:
ah, i see, all is good.

jwforres:
Can you all run a bunch of 4.1 to 4.2 upgrades with this PR. @runcom just did this with another BZ, he can tell you the steps for setting up your own release payload and spinning up the upgrade tests using cluster-bot in slack.

Since we have no way to verify this on master. Can you please do that a bunch of times to get the signal to see if the failure still occurs with this PR, and if it is then that should get the logging that you were looking for to help you debug.

brancz:
We've tried upgrading twice, both times successful. Do you think that's enough or would you like us to try more @jwforres ?

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-4.1/819
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-4.1/820

eparis:
@brancz how often did it fail before this fix? 100% of the time?

eparis:
I ask because simple probabilities would tell us if 2 tests were 'enough'. If it flaked 99% of the time then the likelihood you didn't fix it but we got 2 passes would be 0.01%. But if it flaked 1% of the time then the likelihood we didn't fix it but we got 2 passes would be 98%.'

brancz:
We actually haven’t seen an upgrade failure, the logs always suggested that the error disappeared after time but it put the cluster monitoring operator into degraded status unnecessarily.

PR #511, opened 2019-10-18
pkg/tasks/configsharing: add thanos route

s-urbaniak:
This is important for the OpenShift console is it reads that configmap,
sourcing route URLs for monitoring components.

/cc @LiliC @paulfantom @kyoto


kyoto:
👍 LGTM

PR #516, opened 2019-10-21
assets: move Kube*Down alerts to corresponding operators

sjenning:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/622
https://github.com/openshift/cluster-kube-controller-manager-operator/pull/298
https://github.com/openshift/cluster-kube-scheduler-operator/pull/180

### DIFF
--- a/assets/prometheus-k8s/rules.yaml
+++ b/assets/prometheus-k8s/rules.yaml
@@ -508,30 +508,6 @@ spec:
       for: 15m
       labels:
         severity: critical
-    - alert: KubeAPIDown
-      annotations:
-        message: KubeAPI has disappeared from Prometheus target discovery.
-      expr: |
-        absent(up{job="apiserver"} == 1)
-      for: 15m
-      labels:
-        severity: critical
-    - alert: KubeControllerManagerDown
-      annotations:
-        message: KubeControllerManager has disappeared from Prometheus target discovery.
-      expr: |
-        absent(up{job="kube-controller-manager"} == 1)
-      for: 15m
-      labels:
-        severity: critical
-    - alert: KubeSchedulerDown
-      annotations:
-        message: KubeScheduler has disappeared from Prometheus target discovery.
-      expr: |
-        absent(up{job="scheduler"} == 1)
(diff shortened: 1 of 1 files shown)


sjenning:
fyi @csrwng @derekwaynecarr

sjenning:
aaand of course this is a generated file :man_facepalming: 
down the rabbit hole I go to figure out where this should actually be changed.

sjenning:
likely either here
https://github.com/helm/charts/blob/master/stable/prometheus-operator/templates/prometheus/rules/kubernetes-absent.yaml
or here
https://github.com/coreos/kube-prometheus/blob/master/manifests/prometheus-rules.yaml

LiliC:
Yes this is indeed generated manifets, these alerts trickle down from https://github.com/kubernetes-monitoring/kubernetes-mixin so its not as easy to just remove.

cc @brancz

sjenning:
looks like this is the source
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/0d91b77bd64f9f83541fdbd394b9d8d6a1b87682/config.libsonnet#L15-L21

s-urbaniak:
@sjenning yes, these alerts are maintained in upstream. We have to change those alerts there, then vendor those changes in here.

brancz:
That's correct, we should refactor the alerts in the kubernetes-mixin so that they can be imported on a per component basis, and then individually remove them here and import them in the component operator repos.

sjenning:
@brancz they have already started merging into the component repos (see PRs in the description).  If you could do this refactor, or tell me how to do it, that would be great.

derekwaynecarr:
To provide context, this is desired to support deploying on IBM Cloud where we run control planes on management cluster, but still deploy the rest of cluster monitoring in the user cluster.

brancz:
I went ahead and refactored the kubernetes mixin to allow importing recording and alerting rules by component (https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/281) so what we can do here is just filter out the rule groups related to the apiserver.

PR #522, opened 2019-10-24
pkg/tasks: setup GRPC TLS for thanos querier

s-urbaniak:


pkg/manifests/tls.go
```diff
@@ -0,0 +1,142 @@
...
+func getSelfSignedCA(s *v1.Secret) (*crypto.CA, error) {
+	crt, crtPresent := s.Data["ca.crt"]
+	key, keyPresent := s.Data["ca.key"]
+
+	if crtPresent && keyPresent {
+		ca, err := crypto.GetCAFromBytes(crt, key)
+		if err != nil {
+			klog.Warningf("creating a new CA due to error reading CA: %v", err)
+		}
+		if !needsNewCert(ca.Config.Certs[0].NotBefore, ca.Config.Certs[0].NotAfter, time.Now) {
```
s-urbaniak: this needs a `if ca != nil`

s-urbaniak: (or simply an `else if`)

pkg/manifests/manifests.go
```diff
@@ -2586,6 +2650,42 @@ func (f *Factory) HashTrustedCA(caBundleCM *v1.ConfigMap, prefix string) *v1.Con
 	}
 }
 
+// HashSecret synthesizes a configmap just by copying "ca-bundle.crt" from the given configmap
```
s-urbaniak: fix this godoc

pkg/manifests/manifests.go
```diff
@@ -2586,6 +2650,42 @@ func (f *Factory) HashTrustedCA(caBundleCM *v1.ConfigMap, prefix string) *v1.Con
...
+// and naming it by hashing the contents of "ca-bundle.crt".
+// It adds "monitoring.openshift.io/name" and "monitoring.openshift.io/hash" labels.
+// Any other labels from the given configmap are discarded.
+//
+// It returns nil, if the given configmap does not contain the "ca-bundle.crt" the data key
+// or data is empty string.
+func (f *Factory) HashSecret(secret *v1.Secret, data ...string) (*v1.Secret, error) {
```
s-urbaniak: add some unit tests

pkg/tasks/prometheus_user_workload.go
```diff
@@ -139,8 +139,45 @@ func (t *PrometheusUserWorkloadTask) create() error {
 		return errors.Wrap(err, "reconciling UserWorkload Prometheus Service failed")
 	}
 
+	grpcTLS, err := t.factory.GRPCSecret(nil)
+	if err != nil {
+		return errors.Wrap(err, "initializing Prometheus GRPC secret failed")
```
s-urbaniak: fix error message: `s/Prometheus/UserWorkload Prometheus`

pkg/tasks/thanos_querier.go
```diff
@@ -61,12 +61,12 @@ func (t *ThanosQuerierTask) Run() error {
 		return errors.Wrap(err, "waiting for Thanos Querier Route to become ready failed")
 	}
 
-	s, err := t.factory.ThanosQuerierOauthCookieSecret()
+	grpcTLS, err := t.factory.ThanosQuerierOauthCookieSecret()
```
s-urbaniak: whops, this renamed a bit too much, rename it back to `s`

pkg/manifests/manifests.go
s-urbaniak: done

pkg/manifests/manifests.go
s-urbaniak: done

pkg/manifests/tls.go
s-urbaniak: done

pkg/tasks/prometheus_user_workload.go
s-urbaniak: done

pkg/tasks/thanos_querier.go
s-urbaniak: done

s-urbaniak:
@LiliC @paulfantom @brancz this is ready for a next pass

s-urbaniak:
Let's get some more pair of eyes.

pkg/manifests/tls.go
```diff
@@ -0,0 +1,141 @@
...
+		s.Data = make(map[string][]byte)
+		s.Annotations = make(map[string]string)
+	}
+
+	ca, err := getSelfSignedCA(s)
+	if err != nil {
+		return nil, errors.Wrap(err, "error rotating CA")
+	}
+
+	// no need to rotate
```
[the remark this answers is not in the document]
s-urbaniak: good catch, let's add some godoc to `getSelfSignedCA`, it returns a `nil` CA if no rotation is necessary.

pkg/client/client.go
```diff
@@ -592,6 +611,40 @@ func (c *Client) WaitForStatefulsetRollout(sts *appsv1.StatefulSet) error {
 	return nil
 }
 
+func (c *Client) WaitForSecret(s *v1.Secret) (*v1.Secret, error) {
+	var result *v1.Secret
+	var lastErr error
+	if err := wait.Poll(1*time.Second, 1*time.Minute, func() (bool, error) {
```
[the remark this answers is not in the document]
s-urbaniak: hah, that's a left-over from initial testing :+1: I need to increase it to `5*time.Minute` as we do everywhere else.

pkg/manifests/tls.go
```diff
@@ -0,0 +1,141 @@
...
+func getSelfSignedCA(s *v1.Secret) (*crypto.CA, error) {
+	crt, crtPresent := s.Data["ca.crt"]
+	key, keyPresent := s.Data["ca.key"]
+
+	if crtPresent && keyPresent {
+		ca, err := crypto.GetCAFromBytes(crt, key)
+		if err != nil {
+			klog.Warningf("creating a new CA due to error reading CA: %v", err)
+		} else if !needsNewCert(ca.Config.Certs[0].NotBefore, ca.Config.Certs[0].NotAfter, time.Now) {
+			return nil, nil
```
brancz: I feel like this check rather belongs in the calling function, I had to jump there and back to find the 

```
// no need to rotate
```

comment in the other function. At the very least we should comment this function with what the `nil` return means.

pkg/manifests/tls.go
```diff
@@ -0,0 +1,141 @@
...
+			return nil, errors.Wrap(err, "error getting PEM bytes for thanos querier client certificate")
+		}
+		s.Data["thanos-querier-client.crt"] = crt
+		s.Data["thanos-querier-client.key"] = key
+	}
+
+	{
+		cfg, err := ca.MakeServerCert(
+			sets.NewString("prometheus-grpc"),
+			crypto.DefaultCertificateLifetimeInDays,
```
brancz: This one doesn't need the `*24*time.Hour` like the other method?

pkg/manifests/tls.go
s-urbaniak: agreed, good catch. I will refactor for better readability.

pkg/manifests/tls.go
s-urbaniak: No, in that case the library expects an `int`: https://github.com/openshift/cluster-monitoring-operator/blob/c6195af44d80d4358f15f8e015c6cc608326f224/vendor/github.com/openshift/library-go/pkg/crypto/crypto.go#L665:15

brancz: Ack

s-urbaniak:
@LiliC @brancz @paulfantom ptal

pkg/manifests/manifests.go
```diff
@@ -2610,6 +2674,45 @@ func (f *Factory) HashTrustedCA(caBundleCM *v1.ConfigMap, prefix string) *v1.Con
 	}
 }
 
+// HashSecret synthesizes a secret by setting the given data
```
bwplotka: Hmm this sentence sounds odd to me ):

jsonnet/thanos-querier.jsonnet
```diff
@@ -177,29 +183,33 @@ local authorizationRole =
...
                       exec: {
-                        command: ["sh", "-c", "curl http://localhost:9090/-/healthy"],
+                        command: ['sh', '-c', 'curl http://localhost:9090/-/healthy'],
                       },
                     },
                     readinessProbe: {
                       httpGet:: {},
                       exec: {
-                        command: ["sh", "-c", "curl http://localhost:9090/-/healthy"],
+                        command: ['sh', '-c', 'curl http://localhost:9090/-/healthy'],
```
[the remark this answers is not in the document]
s-urbaniak: hmm .. does thanos querier have a dedicated `/-/ready` endpoint? :thinking: i think i took this over from observatorium, cc @metalmatze

pkg/client/client.go
```diff
@@ -592,6 +611,40 @@ func (c *Client) WaitForStatefulsetRollout(sts *appsv1.StatefulSet) error {
 	return nil
 }
 
+func (c *Client) WaitForSecret(s *v1.Secret) (*v1.Secret, error) {
+	var result *v1.Secret
+	var lastErr error
```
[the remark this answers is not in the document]
s-urbaniak: agreed, i will add it (we use that method/idiom quite often though in CMO in other methods too).

pkg/manifests/manifests.go
s-urbaniak: indeed :-) I will change this as a follow up.

PR #524, opened 2019-10-25
Bug 1750283: Bump telemeter-client

adelton:
Bring in changes from https://github.com/openshift/telemeter/pull/252.


assets/prometheus-operator/0podmonitor-custom-resource-definition.yaml
```diff
@@ -0,0 +1,235 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
LiliC: I think this might be our fault, because in telemeter we are using a different version of prom-operator. cc @brancz

LiliC:
Hmm where is the generate test job here 🤔 cc @paulfantom 

/hold

adelton:
It seems to be run in the Travis CI ... and it failed due to the prometheus-operator change ...

adelton:
If what https://travis-ci.com/openshift/cluster-monitoring-operator/jobs/249443543 produced is the correct output as opposed to the result of my local `make pkg/manifests/bindata.go`, I can certainly pick the result of that Travis CI run.

assets/prometheus-operator/0podmonitor-custom-resource-definition.yaml
brancz: This file needs to be removed as seen [here](https://github.com/openshift/cluster-monitoring-operator/blob/6ce17a90badc8b0d51862ff7c1916689636431db/hack/build-jsonnet.sh#L47) with the other CRDs.

adelton: Removed in updated commit -> 6e206008bbe5027bd0af6893ed9ec8702e7cb250.

adelton:
I've removed the prometheus-operator change -> 6e206008bbe5027bd0af6893ed9ec8702e7cb250.

adelton:
Thanks, @brancz. So it look like now it needs just `cherry-pick-approved` to make it into mergeable state.

brancz:
Correct. The release manager is automatically pinged and will look at it soon.

adelton:
Awesome. Thanks.

PR #521, opened 2019-10-24
Bug 1760393: Add metrics for collecting socket and hyperthreading enablement.

chambridge:
* cluster:hyperthread_enabled_nodes (kube_node_labels + label_node_hyperthread_enabled=true)
* cluster:capacity_cpu_cores_hyperthread_enabled:sum (similar to cluster:capacity_cpu_cores:sum but add categorizing by label_node_hyperthread_enabled)
* cluster:capacity_cpu_sockets_hyperthread_enabled:sum (similar to cluster:capacity_cpu_cores_hyperthread_enabled:sum but sockets)
* node_role_os_version_machine:cpu_capacity_cores:sum (updated to include label_node_hyperthread_enabled)
* node_role_os_version_machine:cpu_capacity_sockets:sum (similar to node_role_os_version_machine:cpu_capacity_cores:sum but sockets)

These metrics are intended to provide the necessary data for correct subscription management.

[cluster-hyperthread_enabled_nodes.txt](https://github.com/openshift/cluster-monitoring-operator/files/3768894/cluster-hyperthread_enabled_nodes.txt)
[cluster:capacity_cpu_cores_hyperthread_enabled:sum.txt](https://github.com/openshift/cluster-monitoring-operator/files/3768895/cluster.capacity_cpu_cores_hyperthread_enabled.sum.txt)
[cluster:capacity_cpu_sockets_hyperthread_enabled:sum.txt](https://github.com/openshift/cluster-monitoring-operator/files/3768896/cluster.capacity_cpu_sockets_hyperthread_enabled.sum.txt)
[node_role_os_version_machine-cpu_capacity_cores-sum.txt](https://github.com/openshift/cluster-monitoring-operator/files/3768897/node_role_os_version_machine-cpu_capacity_cores-sum.txt)
[node_role_os_version_machine-cpu_capacity_sockets-sum.txt](https://github.com/openshift/cluster-monitoring-operator/files/3768898/node_role_os_version_machine-cpu_capacity_sockets-sum.txt)

PR #532, opened 2019-10-30
Bug 1771389: Refactor cpu recording rules

pgier:
Trying a different approach to #521 to make the recording rules simpler.

The previous recording rules for sockets seems to count number of cores.  I changed this to count the number of sockets, but I'm not sure if this is the correct result.


pgier:
CC @chambridge

chambridge:
@pgier  FYI - the reason `label_node_role_kubernetes_io_infra: "true"` was used instead of `label_node_role_kubernetes_io: infra` was because a node could potentially have multiple role defined per Clayton's statement.

pgier:
@chambridge In the nodes_roles metric, I set the value of `label_node_role_kubernetes_io` to be `master,infra` if the node has both roles.  This way we should be able to filter by either master, infra, or both using regex matching.  Would this work for our use case?

chambridge:
@pgier Sounds like it might potentially work. Any chance you could provide output related to these metrics, so I can grasp what the changes would look like? 

Would it cause any issues if the output changes for a metric? *node_role_os_version_machine:cpu_capacity_cores:sum* is already in 4.2 and 4.1.z

I also need to thing how we would handle different output formats in the clusters-service if we change the produced metric (should be doable though).

pgier:
I'll try to get this running in a cluster and provide some sample output.  It should be basically the same as what we have now, just with the change to the `node_role` label.  As an alternative, I think it should be possible to produce the same output we have now with some minor changes, using the separate master and infra labels.

pgier:
Based on some more discussion it seems that we don't need specific hyperthreading info by node, we just need to know number of cores vs. cpus.  (2 cpu per core for hyperthreading systems, otherwise 1 cpu per core).  Note: the metric `cluster:capacity_cpu_cores:sum` has been changed from calculating the number of cpus to the number of cores.  If we need to keep the behaviour as-is for backwards compat reasons, I can restore the old one and re-name the new one.

[cpu_metrics.txt](https://github.com/openshift/cluster-monitoring-operator/files/3800923/cpu_metrics.txt)

LiliC:
FYI: The failing test is a known failure, nothing to do with this PR.

LiliC:
@chambridge do you mind having a look again, if this change works for your use case.

chambridge:
So my biggest concern would be around `cluster:capacity_cpu_cores:sum` and whether altering that value breaks any of the metrics consumers (i.e. clusters-service) 

Outside of that I'd want @smarterclayton to 👍 that he's happy with the metric updates. We'd also have to manage the different metric names between releases until everything is backported which would be some effort (but doable).

assets/prometheus-k8s/rules.yaml
```diff
@@ -292,31 +292,28 @@ spec:
...
         BY (label_beta_kubernetes_io_instance_type, label_node_role_kubernetes_io)
       record: cluster:capacity_memory_bytes:sum
+    - expr: |
+        sum(cluster:nodes_roles * on (node) group_left() kube_node_status_capacity_cpu_cores)
+        by (label_node_openshift_io_os_id, label_kubernetes_io_arch, label_beta_kubernetes_io_instance_type, label_node_role_kubernetes_io_master, label_node_role_kubernetes_io_infra)
+      record: cluster:capacity_cpus:sum
+    - expr: |
+        sum(cluster:nodes_roles * on (node) group_left() ( label_replace( (sum (max (node_cpu_info) by (instance, package, core) ) by (instance) ), "node", "$1", "instance", "(.*)" ) ) )
+        by (label_node_openshift_io_os_id, label_kubernetes_io_arch, label_beta_kubernetes_io_instance_type, label_node_role_kubernetes_io_master, label_node_role_kubernetes_io_infra)
+      record: cluster:capacity_cpu_cores:sum
```
smarterclayton: Have we manually verified this returns same results before and after?

assets/prometheus-k8s/rules.yaml
```diff
@@ -335,60 +332,6 @@ spec:
...
-        * on(node) group_left(label_node_hyperthread_enabled) cluster:hyperthread_enabled_nodes{label_node_hyperthread_enabled!="true"})
-        or (cluster:infra_nodes * on(node) group_left(label_node_hyperthread_enabled)
-        cluster:hyperthread_enabled_nodes{label_node_hyperthread_enabled="true"})
-        * on(node) group_left kube_node_status_capacity_cpu_cores) or on(node) (cluster:hyperthread_enabled_nodes
-        * on(node) group_left kube_node_status_capacity_cpu_cores) or on(node) ((kube_node_labels
-        unless on(node) cluster:hyperthread_enabled_nodes) * on(node) group_left kube_node_status_capacity_cpu_cores))
-        BY (label_node_openshift_io_os_id, label_kubernetes_io_arch, label_node_hyperthread_enabled,
-        label_node_role_kubernetes_io_master_infra, label_node_role_kubernetes_io_master,
-        label_node_role_kubernetes_io_infra)
-      record: node_role_os_version_machine:cpu_capacity_cores:sum
```
smarterclayton: We have to preserve the old name for consumers - we are already using this in layers on top of OCM.

assets/prometheus-k8s/rules.yaml
smarterclayton: I'm not worried about the additional labels here (because we already have to sum/max across the data in telemetry).

assets/prometheus-k8s/rules.yaml
```diff
@@ -292,31 +292,28 @@ spec:
...
+        cluster:infra_nodes or on (node) kube_node_labels
+      record: cluster:nodes_roles
     - expr: sum((cluster:master_nodes * on(node) group_left kube_node_status_capacity_memory_bytes)
         or on(node) (kube_node_labels * on(node) group_left kube_node_status_capacity_memory_bytes))
         BY (label_beta_kubernetes_io_instance_type, label_node_role_kubernetes_io)
       record: cluster:capacity_memory_bytes:sum
+    - expr: |
+        sum(cluster:nodes_roles * on (node) group_left() kube_node_status_capacity_cpu_cores)
+        by (label_node_openshift_io_os_id, label_kubernetes_io_arch, label_beta_kubernetes_io_instance_type, label_node_role_kubernetes_io_master, label_node_role_kubernetes_io_infra)
+      record: cluster:capacity_cpus:sum
```
smarterclayton: What is the distinction between this and cores?  The value reported by Kube node status is subtle and may not represent the real value.  I might call this as `cluster:node_capacity_cpu_cores`.  We should always use `cores` when referring to the units of CPU resources in kube.

assets/prometheus-k8s/rules.yaml
pgier: The existing rule returns number of real/virtual cpus
After the change this would return the number of cores
So I changed it purposely to better match the name, since it sounds like it's counting number of cores.  But it sounds like I should change it back for backwards compatility and just add a new rule for number of cores.

assets/prometheus-k8s/rules.yaml
pgier: This one records the number of CPUs as kube sees them.  The number of cores will either be the same as this, or half of this if hyperthreading is enabled.

assets/prometheus-k8s/rules.yaml
pgier: The new metrics provide basically the same info minus the hyperthreading label, since it can be determined from the number of cores vs. cpus.  So I could rename the new metrics to match these names if that would prevent issues in the higher layers.

smarterclayton: Hrm, are we sure that actually captures the use case?  We need the hyper threading label to be super obvious so we can account appropriately.  Is the argument here that we *can* query it separately, or that we *must* query it separately?

pgier: I think we don't need to query hyperthreading at all.  I discussed it a bit with @luciddreamz and IIUC we just needed to know the number of sockets, cores, and cpus.  The idea to get hyperthreading info was just an indirect way to get the number of cores.

pgier: There also might be an edge case where you could have two sockets on the same node, one with hyperthreading and one without.  In that case the hyperthreading label would be inaccurate.

pgier:
Closing since #544 resolves this without breaking backwards compatibliity.

PR #534, opened 2019-10-31
jsonnet: bump telemeter-client to latest revision

crawford:
This brings in the new cluster_control_plane_architecture metric.

Continuation from https://github.com/openshift/telemeter/pull/256.


crawford:
It turns out that we already have this data, so this isn't necessary.

/close.

PR #537, opened 2019-11-06
jsonnet: bump telemeter-client to latest version for multus telemeter …

aneeshkp:
This brings in the multus net-attach-definition metric.
Continuation from openshift/telemeter#262


pkg/manifests/bindata.go
```diff
@@ -588,7 +588,7 @@ func assetsGrafanaDashboardDefinitionsYaml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "assets/grafana/dashboard-definitions.yaml", size: 758829, mode: os.FileMode(420), modTime: time.Unix(1, 0)}
+	info := bindataFileInfo{name: "assets/grafana/dashboard-definitions.yaml", size: 758752, mode: os.FileMode(420), modTime: time.Unix(1, 0)}
```
LiliC: You commited too much here, this should not be part of this commit.

aneeshkp: Thanks @LiliC , let me fix this.  The doc says commit all the file generated, So I missed to understand.  let me go through this again.
https://docs.google.com/document/d/1a6n5iBGM2QaIQRg9Lw4-Npj6QY9--Hpx3XYut-BrUSY/edit#
 In the root directory of the repo run `make generate-in-docker` (or `MAKE=make generate-in-docker` if you are running this on mac) and commit the generated files.

LiliC: Make sure to run `make clean` before running `make generate-in-docker `, I added it to the docs. Thanks.

aneeshkp: @LiliC Thanks, I pushed in changes with correct files now. Thanks again.

PR #543, opened 2019-11-12
Bug 1739375; drop remote write dashboard

simonpasquier:


LiliC:
After https://github.com/openshift/cluster-monitoring-operator/pull/542 gets merged you need to rebase to fix the generate failures.

PR #545, opened 2019-11-13
test/e2e: add user workload monitoring

s-urbaniak:
Take 2, let's see if CI is going to tackle this now.

### DIFF
--- a/test/e2e/user_workload_monitoring_test.go
+++ b/test/e2e/user_workload_monitoring_test.go
@@ -77,6 +77,20 @@ func createUserWorkloadAssets(cm *v1.ConfigMap) func(*testing.T) {
 			t.Fatal(err)
 		}
 
+		err = framework.Poll(time.Second, 5*time.Minute, func() error {
+			_, err := f.KubeClient.AppsV1().StatefulSets(f.UserWorkloadMonitoringNs).Get("prometheus-user-workload", metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+
+			return nil
+		})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// this will only poll if the statefulset is there in the first place
+		// otherwise it will fail immediately
 		err = f.OperatorClient.WaitForPrometheus(&monitoringv1.Prometheus{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      "user-workload",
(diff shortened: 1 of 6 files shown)


test/e2e/user_workload_monitoring_test.go
```diff
@@ -0,0 +1,351 @@
...
+			Replicas: proto.Int32(1),
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func assertUserWorkloadMetrics(t *testing.T) {
+	// assert that the previously deployed user application metrics are available in thanos
```
bwplotka:
```suggestion
	// Assert that the previously deployed user application metrics are available in Thanos.
```

test/e2e/user_workload_monitoring_test.go
```diff
@@ -0,0 +1,351 @@
...
+			if i == 0 {
+				return nil
+			}
+
+			return fmt.Errorf("expected version metric from user application to be equal 0 but got %v", i)
+		},
+	)
+
+	{
+		// assert that the same metric is not scraped by the cluster monitoring stack
```
bwplotka: ditto

bwplotka: and below if we want to be matching Go style guide

test/e2e/user_workload_monitoring_test.go
```diff
@@ -0,0 +1,351 @@
...
+	for _, scenario := range []struct {
+		name string
+		f    func(*testing.T)
+	}{
+		{"enable user workload monitoring, assert prometheus rollout", createUserWorkloadAssets(cm)},
+		{"create and assert an user application is deployed", deployUserApplication},
+		{"create prometheus and alertmanager in user namespace", createPrometheusAlertmanagerInUserNamespace},
+		{"assert user workload metrics", assertUserWorkloadMetrics},
+		{"assert prometheus and alertmanager is not deployed in user namespace", assertPrometheusAlertmanagerInUserNamespace},
+		{"assert assets are deleted when user workload monitoring is disabled", assertDeletedUserWorkloadAssets(cm)},
```
bwplotka: It's kind of weird that `assert` function deletes inside and not only assert, but I it's fine, minor nit.

s-urbaniak:
The e2e test failure is legit, i have a small bug in `createUserWorkloadAssets`. We can only invoke `WaitForPrometheus` and `WaitForStatefulsetRollout` if we know that the statefulset objects have been created in the first place, otherwise the poll will fail immediately.

PR #546, opened 2019-11-13
grafana: Secure cookie true

brancz:
@s-urbaniak @paulfantom @LiliC @simonpasquier @pgier


brancz:
I actually forgot to update the bindata.go file. Another commit incoming...

PR #556, opened 2019-11-16
jsonnet: Pod recording rules are incorrectly summing resources

smarterclayton:
In kubernetes, the sum of resources used by a pod is not the sum
of containers, but is instead the value reported for the pod
cgroup `{container=""}`. The recording rules here were incorrectly
omitting CPU usage accounted to the pod but not accounted to a
container which would under report containers that run in the pod
context like builds (but which aren't kube containers) as well as
other metrics also accounted to the pod but not containers.

@derekwaynecarr @sjennings


smarterclayton:
I verified at least CPU and memory, but I want to double check that cadvisor is still correctly reporting sum metrics for the others.

PR #560, opened 2019-11-21
test/e2e/prometheusadapter: disable rotation test

s-urbaniak:

Let's disable that test for now as something has changed in origin that quickly reconciles away the changes we do on the `requestheader-client-ca-file` content of `kube-system/extension-apiserver-authentication`.

Generally this is good, but renders this test intrinsically racy and thus flaky.

### DIFF
--- a/test/e2e/prometheusadapter_test.go
+++ b/test/e2e/prometheusadapter_test.go
@@ -221,7 +221,7 @@ func TestAggregatedMetricPermissions(t *testing.T) {
 	}
 }
 
-func TestPrometheusAdapterCARotation(t *testing.T) {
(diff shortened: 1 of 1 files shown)


s-urbaniak:
fyi we started having failures for this test in https://github.com/openshift/cluster-monitoring-operator/pull/559 and https://github.com/openshift/cluster-monitoring-operator/pull/556.

s-urbaniak:
ok, I think I have found another good way of testing the rotation, it is described in https://bugzilla.redhat.com/show_bug.cgi?id=1777473#c6.

I will submit a follow-up PR re-enabling that test.

PR #561, opened 2019-11-21
Bug 1774621: Bump telemeter-client

awgreene:

PR #563, opened 2019-11-22
Bug 1774621: telemeter: bump dependencies

s-urbaniak:
/cc @awgreene @paulfantom


awgreene:
@s-urbaniak Thanks for closing #561.

Will this be backported to the 4.3 branch?

s-urbaniak:
@awgreene my understanding is that this should still land in 4.3 once green, cc @bparees for confirmation

awgreene:
>@awgreene my understanding is that this should still land in 4.3 once green, cc @bparees for confirmation

@s-urbaniak I believe that changed as of 9am EST this morning.

bparees:
> @awgreene my understanding is that this should still land in 4.3 once green, cc @bparees for confirmation

yeah since the branch happened this morning, this will only land in 4.4(master) currently.  You will need a second BZ targeting 4.3 and have to cherry-pick this to release-4.3.

The 4.4 BZ will need to be in "modified" (not verified) to get the cherry-pick approved.

PR #576, opened 2019-12-03
test/e2e/prometheusadapter: reenable CA rotation test

s-urbaniak:
This reenables the requestheader CA rotation e2e test
by using a different method.

It deletes the CSR signer secrets causing the extension-apiserver-authentication
configmap to be reissued.

### DIFF
--- a/test/e2e/framework/route_client.go
+++ b/test/e2e/framework/route_client.go
@@ -64,7 +64,12 @@ func NewRouteClient(
 	var token string
 
 	for _, secret := range secrets.Items {
-		if strings.Contains(secret.Name, "cluster-monitoring-operator-e2e-token-") {
+		_, dockerToken := secret.Annotations["openshift.io/create-dockercfg-secrets"]
+		e2eToken := strings.Contains(secret.Name, "cluster-monitoring-operator-e2e-token-")
+
+		// we have to skip the token secret that contains the openshift.io/create-dockercfg-secrets annotation
+		// as this is the token to talk to the internal registry.
+		if !dockerToken && e2eToken {
 			token = string(secret.Data["token"])
 		}
 	}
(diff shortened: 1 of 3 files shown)


s-urbaniak:
I have to look into this one :thinking: 
```
2019/12/03 14:58:05 wait for prometheus-k8s: invalid character '<' looking for beginning of value: timed out waiting for the condition
FAIL	github.com/openshift/cluster-monitoring-operator/test/e2e	61.985s
```

s-urbaniak:
I found at least the root cause for the above message: it is the OAuth proxy telling us we're not allowed to execute the request (403). As we don't check the response status code, we try to parse the OAuth login html which causes the above error to be generated.

This is unrelated to this PR as it happens in the framework setup but I'll try to fix it here as we're here.

s-urbaniak:
@LiliC now this is good to go, PTAL. I hopefully fixed the outstanding issues with our e2e test client in new commit.

test/e2e/framework/route_client.go
```diff
@@ -64,7 +64,10 @@ func NewRouteClient(
 	var token string
 
 	for _, secret := range secrets.Items {
-		if strings.Contains(secret.Name, "cluster-monitoring-operator-e2e-token-") {
+		_, dockerToken := secret.Annotations["openshift.io/create-dockercfg-secrets"]
+		e2eToken := strings.Contains(secret.Name, "cluster-monitoring-operator-e2e-token-")
+
+		if !dockerToken && e2eToken {
```
LiliC: Can we just add a comment that if its dockerToken that means this token is an invalid one, to avoid future confusion.

s-urbaniak: sounds good

s-urbaniak:
@LiliC pta

LiliC:
Seems like a flake
/retest

PR #579, opened 2019-12-05
jsonnet: Gather ~10 metrics that tell us workloads that are being used

smarterclayton:
These are the first metrics that show rough details of what is going on for
a cluster so that we can assess rough usage. It captures both scale (number
of particular workload types) and gives us a small hint about usage type
(are people using DCs or deployments, statefulsets or jobs). The containers
usage can be used to assess approx container per pod numbers when
contrasted to pod numbers, which is also useful.

For cardinality 10, we get a much better insight into whether this cluster
is in use.  We can approximate infrastructure resource count against the
user's workload resource count by looking at empty clusters, but sheer
scale helps us determine better values.


smarterclayton:
@derekwaynecarr @bparees this is the first attempt to capture a small number of high value metrics around usage.  We obviously can't get everything, but this answers some fundamental questions.  Suggestions on other super critical metrics (as seen by the whole product) might be useful.

LiliC:
Actually we don't need to hold as that is targeting 4.3 and this is targeting master, the PR for whitelisting for that has already been merged in telemeter master.

/hold cancel

assets/prometheus-k8s/rules.yaml
```diff
@@ -345,6 +345,14 @@ spec:
       record: cluster:node_instance_type_count:sum
     - expr: sum(etcd_object_counts) BY (instance)
       record: instance:etcd_object_counts:sum
+    - expr: max(etcd_object_counts{resource=~"builds\.build\.openshift\.io|deploymentconfigs\.apps\.openshift\.io|images\.image\.openshift\.io|statefulsets\.apps|deployments\.apps|cronjobs\.batch|jobs\.batch|pods|persistentvolumeclaims|persistentvolumes|services|namespaces|routes\.route\.openshift\.io|ingresses\.networking\.k8s\.io|horizontalpodautoscalers\.autoscaling|users\.user\.openshift\.io|machinesets\.machine\.openshift\.io|resourcequotas"})
+        by (resource)
+      record: cluster:usage:resources:sum
+    - expr: count(count (kube_pod_restart_policy{type!="Always",namespace!~"openshift-.+"})
+        by (namespace,pod))
+      record: cluster:usage:pods:terminal:workload:sum
+    - expr: sum(max (kubelet_containers_per_pod_count_sum) by (instance))
```
LiliC: Nit:
```suggestion
    - expr: sum(max(kubelet_containers_per_pod_count_sum) by (instance))
```

assets/prometheus-k8s/rules.yaml
```diff
@@ -345,6 +345,14 @@ spec:
       record: cluster:node_instance_type_count:sum
     - expr: sum(etcd_object_counts) BY (instance)
       record: instance:etcd_object_counts:sum
+    - expr: max(etcd_object_counts{resource=~"builds\.build\.openshift\.io|deploymentconfigs\.apps\.openshift\.io|images\.image\.openshift\.io|statefulsets\.apps|deployments\.apps|cronjobs\.batch|jobs\.batch|pods|persistentvolumeclaims|persistentvolumes|services|namespaces|routes\.route\.openshift\.io|ingresses\.networking\.k8s\.io|horizontalpodautoscalers\.autoscaling|users\.user\.openshift\.io|machinesets\.machine\.openshift\.io|resourcequotas"})
```
LiliC: I get the following error when trying to perform this query:

> Error executing query: parse error at line 1, char 34: unknown escape sequence U+002E '.'

So this might be the reason the tests are failing consistently.

bparees:
@adambkaplan fyi, there will be metrics about number of build objects on clusters in telemeter w/ this change

bparees:
/hold
per @LiliC's comment about failing tests

smarterclayton:
Fixed, applying label

PR #587, opened 2019-12-10
SREP-2140: alert when sre stays in cluster-admin more than 2 hours

boranx:
Fixes: SREP-2140

Currently, SRE can temporarily elevate permissions to cluster-admin by adding themselves to the 'osd-sre-cluster-admins' group.
The alert will be fired if SRE stays in `cluster admin` role for more than 2 hours.


RiRa12621:
This is a generated file.
Additionally we don't want our SRE specific alerts to go into the upstream project.

PR #605, opened 2020-01-09
assets,pkg: Update dependencies 

LiliC:
This primarily brings in changes from https://github.com/coreos/kube-prometheus/pull/365 but also bumps other dependencies.


assets/grafana/dashboard-definitions.yaml
```diff
@@ -148,44 +148,44 @@ items:
               {
                   "columns": [
                       {
-                          "text": "",
```
LiliC: Not sure about the grafana dashboards changes? 🤔

assets/kube-state-metrics/deployment.yaml
```diff
@@ -63,7 +63,7 @@ spec:
         - --port=8081
         - --telemetry-host=127.0.0.1
         - --telemetry-port=8082
-        image: quay.io/coreos/kube-state-metrics:v1.8.0
+        image: quay.io/coreos/kube-state-metrics:v1.9.0
```
LiliC: This is fine, we don't want 1.9.0 yet, as it still has a potential bug, but it will be overriden by CVO anyways, so "this is fine". :)

LiliC:

assets/thanos-querier/deployment.yaml
```diff
@@ -15,6 +15,20 @@ spec:
...
+              labelSelector:
+                matchExpressions:
+                - key: app.kubernetes.io/name
+                  operator: In
+                  values:
+                  - thanos-querier
+              namespaces:
+              - monitoring
+              topologyKey: kubernetes.io/hostname
+            weight: 100
```
s-urbaniak: oh I guess this is a good thing to bring in :+1:

LiliC:
/hold

will look at the failures, although from what i could tell they said marketplace failing

assets/thanos-querier/deployment.yaml
```diff
@@ -15,6 +15,20 @@ spec:
...
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: app.kubernetes.io/name
+                  operator: In
+                  values:
+                  - thanos-querier
+              namespaces:
+              - monitoring
```
LiliC: I think this namespace needs to be changed @s-urbaniak I will remove this change from this PR, as I don't have the background behind the change. s

LiliC: Nvm it comes from kube-prometheus :D

s-urbaniak: hmm .. the more i think about this the more i come to the conclusion that we will have to dependency-inject `openshift-monitoring` and `openshift-user-workload-monitoring` during runtime.

LiliC:
@s-urbaniak the file in question came in through prometheus-operator not kube-promethus, so I left it out, we can do that another time in a separate PR. Probably after the release of prometheus-operator. PTAL, thanks!

assets/prometheus-k8s/service-monitor-kubelet.yaml
```diff
@@ -7,54 +7,55 @@ metadata:
...
-    interval: 30s
-    port: https-metrics
-    relabelings:
-    - sourceLabels:
-      - __metrics_path__
-      targetLabel: metrics_path
-    scheme: https
-    tlsConfig:
-      caFile: /etc/prometheus/configmaps/kubelet-serving-ca-bundle/ca-bundle.crt
-      insecureSkipVerify: false
```
s-urbaniak: is this now left out by purpose?

manifests/0000_50_cluster_monitoring_operator_02-role.yaml
```diff
@@ -163,6 +170,15 @@ rules:
...
   - storageclasses
+  - volumeattachments
+  verbs:
+  - list
+  - watch
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - mutatingwebhookconfigurations
+  - validatingwebhookconfigurations
```
s-urbaniak: out of curiosity: do we also need those permissions inherited?

assets/prometheus-k8s/service-monitor-kubelet.yaml
LiliC: It's git being silly, it just moved to the end of the file, because we needed to apply the relabelling to the other endpoint.

manifests/0000_50_cluster_monitoring_operator_02-role.yaml
LiliC: These are new metrics based on those resources added to kube-state-metrics. We are in the process of switching to this version of ksm yes.

PR #617, opened 2020-01-20
A/B flag for remote write

brancz:
This implements the remote write feature, but turns it off by default, but at the same time allows using the new, much simpler configuration mechanism for selecting metrics while using the more conservative v1 protocol.

@LiliC @s-urbaniak


brancz:
/hold

Just to check in CI that indeed the v1 protocol is used.

brancz:
Validated that this works as expected.

/hold cancel

cmd/operator/main.go
```diff
@@ -86,10 +91,35 @@ func Main() int {
 	kubeconfigPath := flagset.String("kubeconfig", "", "The path to the kubeconfig to connect to the apiserver with.")
 	apiserver := flagset.String("apiserver", "", "The address of the apiserver to talk to.")
 	releaseVersion := flagset.String("release-version", "", "Currently targeted release version to be reconciled against.")
+	telemetryConfigFile := flagset.String("telemetry-config", "/etc/cluster-monitoring-operator/telemetry/metrics.yaml", "Path to telemetry-config.")
+	remoteWrite := flagset.Bool("enabled-remote-write", false, "Wether to use legacy telemetry write protocol or Prometheus remote write.")
```
metalmatze: So this is the real change now, is it?

LiliC:
@brancz don't forget to update the docs :)

cmd/operator/main.go
brancz: Yep

PR #623, opened 2020-01-23
pkg/manifests: don't overwrite telemetry remote write

s-urbaniak:
Currently, if a custom remote write configuration is provided
and telemetry remote write is enabled,
the latter is overwritten by the custom remote write configuration.

This fixes it by appending the custom remote write configuration
to the telemetry remote write configuration.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1165,8 +1165,9 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
 	}
 
 	if len(f.config.PrometheusK8sConfig.RemoteWrite) > 0 {
-		p.Spec.RemoteWrite = f.config.PrometheusK8sConfig.RemoteWrite
+		p.Spec.RemoteWrite = append(p.Spec.RemoteWrite, f.config.PrometheusK8sConfig.RemoteWrite...)
 	}
+
 	for _, rw := range p.Spec.RemoteWrite {
 		if f.config.HTTPConfig.HTTPProxy != "" {
 			rw.ProxyURL = f.config.HTTPConfig.HTTPProxy
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -16,8 +16,11 @@ package manifests
 
 import (
 	"reflect"
+	"sort"
 	"strings"
 	"testing"
+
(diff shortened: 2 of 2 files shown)


squat:
Then how do customers running OpenShift on prem or in some offline environment override telemeter? Do they have to override DNS?

s-urbaniak:
@squat currently, only the URL is replaceable as per https://github.com/openshift/cluster-monitoring-operator/blob/11945732b0e7195d189a1d3802f43c1727ac41fc/pkg/manifests/manifests.go#L1129, @brancz any further thoughts?

s-urbaniak:
The other way to accomplish this is to 1. disable telemetry (by unsetting cluster ID and token, or disabling explicitly) and 2. add a custom remote write configuration.

But the current semantic is confusing, as the user expects telemetry to "just" work, even if the user provided a custom remote write target (which is not necessarily telemetry).

squat:
@s-urbaniak yes but with this change, the url is no longer replaceable. That is the question I am raising. How do users running their telemetry on-prem point remote write at their on-prem stack? Do they have to make their internal DNS resolve the canonical telemetry name and point internally?

s-urbaniak:
/hold

> @s-urbaniak yes but with this change, the url is no longer replaceable.
I don't understand why with this change the url is no longer replaceable :thinking: 

In any way, let's hold this until we clarify.

brancz:
The URL used for remote write is taken from the cluster monitoring configmap, if provided it has precedence over the default one. I think this PR is correct as it is.

s-urbaniak:
just to be on a safe side i added a specific test case for this edge case: https://github.com/openshift/cluster-monitoring-operator/pull/623/commits/8099b1a9afaab28fc0d1ba7d33e4b83cbc7f214a

@squat ptal does that cover your concern?

squat:
The question is:

I am running OpenShift and dependencies on-prem. How do I point telemeter remote write ONLY at my local stack? Appending is not a valid because then prometheus will be failing to write to the canonical URL.

Can I still override the canonical remote-write url using the TelemeterServerURL config option? Or do I have to resolve the canonical DNS name using my internal DNS?

squat:
Yes that test covers what I was asking about

brancz:
> Can I still override the canonical remote-write url using the TelemeterServerURL config option?

Yes you can. @s-urbaniak just added a test for this.

s-urbaniak:
thanks @squat for raising that question, it's a good edge case to be covered :+1: ptal (the lgtm got lost due to the added test)

PR #628, opened 2020-01-24
pkg/manifests: Correctly default telemeter URL

brancz:
We were incorrectly defaulting the telemeter URL since we merged #617. 

@LiliC @s-urbaniak


brancz:
Works as expected.

PR #629, opened 2020-01-24
jsonnet/kube-state-metrics.jsonnet: Fix blacklist arguments

LiliC:


LiliC:
/hold

Will verify with cluster-bot as I only verified on vanilla ksm.

LiliC:
Verified with cluster-bot

/hold cancel

PR #638, opened 2020-01-30
Bug 1796473: Revert "jsonnet: regenerate jsonnetfile.* with latest jb"

s-urbaniak:
This reverts commit dba1836fa2bf5bb2308aa4db14ac31da7b2e246b.

### DIFF
--- a/jsonnet/jsonnetfile.json
+++ b/jsonnet/jsonnetfile.json
@@ -1,6 +1,7 @@
 {
   "dependencies": [
     {
+      "name": "kube-prometheus",
       "source": {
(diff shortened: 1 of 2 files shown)


s-urbaniak:
I guess what happened that we merged in a new jsonnet lock file with a non-yet-broken version of jb in master. Since jb broke on master we pinned to 0.2.0 in the jsonnet-ci image.
0.2.0 though had a different lock format and breaks.

That failure manifested in PRs that were opened before jb broke but used the newer lock format letting jb 0.2.0 fail.

PR #639, opened 2020-01-30
add oom metric

rphillips:
add an OOM rule and telemetry metric


rphillips:
`make generate-in-docker` produced the generated files... with the following image. I am not sure why CI doesn't like the generated files.

```
REPOSITORY                  TAG                 IMAGE ID            CREATED             SIZE
quay.io/coreos/jsonnet-ci   latest              7992e3f713a8        13 hours ago        1.07GB
```

rphillips:
rebased with the fix from earlier today...

assets/prometheus-k8s/rules.yaml
```diff
@@ -92,6 +92,11 @@ spec:
     - expr: |
         sum(rate(container_cpu_usage_seconds_total{job="kubelet", metrics_path="/metrics/cadvisor", image!="", container!="POD"}[5m])) by (namespace)
       record: namespace:container_cpu_usage_seconds_total:sum_rate
+    - expr: |
+        sum by (cluster, namespace, pod, container) (
+          rate(kube_pod_container_status_last_terminated_reason{reason="OOMKilled",job="kube-state-metrics"})
```
paulfantom: Why are you suddenly interested in the rate and not absolute value?

assets/prometheus-k8s/rules.yaml
```diff
@@ -92,6 +92,11 @@ spec:
...
+        ) * on (cluster, namespace, pod) group_left(node) max by(cluster, namespace, pod, node) (kube_pod_info)
+      record: node_namespace_pod_container_status:oom_total
```
paulfantom: Best practices dictate to have a name of the aggregator at the end of the recording rule. In this case it might be better to have something like `kube_pod_container_status_last_terminated_reason:oom_total:sum`

assets/prometheus-k8s/rules.yaml
rphillips: refactored... I think I did it right...

assets/prometheus-k8s/rules.yaml
rphillips: thanks... fixed.

assets/prometheus-k8s/rules.yaml
```diff
@@ -92,6 +92,10 @@ spec:
     - expr: |
         sum(rate(container_cpu_usage_seconds_total{job="kubelet", metrics_path="/metrics/cadvisor", image!="", container!="POD"}[5m])) by (namespace)
       record: namespace:container_cpu_usage_seconds_total:sum_rate
+    - expr: |
+        kube_pod_container_status_last_terminated_reason{reason="OOMKilled",job="kube-state-metrics"}
+        * on (cluster, namespace, pod) group_left(node) max by(cluster, namespace, pod, node) (kube_pod_info)
+      record: kube_pod_container_status_last_terminated_reason:oom_total:sum
```
paulfantom: Now there is no `sum` in the recording expression :thinking:

s-urbaniak:
@rphillips being in feature freeze phase this needs a valid bugzilla, is it addressing a bug?

assets/prometheus-k8s/rules.yaml
LiliC: > which gives you a split on namespaces and 39 new metrics into telemetry, or

That is most likely too many metrics, you get 10 series budget per component roughly, unless you get special approval from @smarterclayton and a valid reason why this is necessary.

rphillips: Thank you for the explanations. I changed it to the single metric so we can keep an eye on OOMs across the cluster.

rphillips:
aws limits...

/retest

LiliC:
You need to rebase, and have to just add some comments describing the metric you want to send.

assets/prometheus-k8s/rules.yaml
```diff
@@ -274,6 +274,8 @@ spec:
     - expr: sum(rate(container_cpu_usage_seconds_total{container="",pod!=""}[5m]))
         BY (pod, namespace)
       record: pod:container_cpu_usage:sum
+    - expr: sum(kube_pod_container_status_last_terminated_reason{reason="OOMKilled",job="kube-state-metrics"})
```
simonpasquier: To avoid double accounting in corner cases:

`sum(max by(namespace, pod, container) (kube_pod_container_status_last_terminated_reason{...}))`

assets/prometheus-k8s/rules.yaml
```diff
@@ -274,6 +274,8 @@ spec:
...
+      record: kube_pod_container_status_last_terminated_reason:oom_total:sum
```
simonpasquier: not sure about the metric name

smarterclayton:
/hold

I'd like an alert first, before we send the metric back.  What is a reasonable amount of oomkills for an admin to expect per minute/hour?

rphillips:
@smarterclayton I think we should get metrics for anything greater than zero... OOMs are indicative of an overloaded cluster, or pods that are over their resource limits. Either scenario needs user intervention.

smarterclayton:
/hold

I generally won't allow new exploratory metrics to go to prometheus until
you can demonstrate you have a working alert htat catches what we would
consider "too many OOMKills".  If that's one oomkill an hour, that's fine.
I just need to see a proposed alert before i ack any fishing expedition.

assets/prometheus-k8s/rules.yaml
metalmatze: Yep. This should be something like `namespace_pod_container:kube_pod_container_status_oom_total:max`
`sum` is mostly implicit and the `by` is the aggregation that's put to the front.

rphillips:
I added an alert in this separate PR https://github.com/openshift/cluster-monitoring-operator/pull/642 .

rphillips:
Going to close this in favor of bringing up an OOM metric at the sig node meeting. Perhaps it should live in both the container runtime and kubelet. Thank you for the comments and review.

PR #642, opened 2020-02-04
add ClusterOOMWarning alert

rphillips:
Adds an alert to warn if a cluster is experiencing OOMs.

ref: https://github.com/openshift/cluster-monitoring-operator/pull/639


assets/prometheus-k8s/rules.yaml
```diff
@@ -380,6 +380,13 @@ spec:
       record: node_role_os_version_machine:cpu_capacity_sockets:sum
     - expr: clamp_max(sum(alertmanager_notifications_total),1)
       record: cluster:alertmanager_routing_enabled:max
+    - alert: ClusterOOMWarning
```
LiliC: Hmm not sure this is related to the cluster 100% of the time, could be just wrong memory specifications, maybe something like "KubePodOOMWarning" naming would be better.

assets/prometheus-k8s/rules.yaml
```diff
@@ -380,6 +380,13 @@ spec:
...
+      annotations:
+        message: Cluster is experiencing out of memory (OOM) kills
```
LiliC: Can we have something like:
```suggestion
        message: Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) is experiencing out of memory termination. 
```

assets/prometheus-k8s/rules.yaml
```diff
@@ -380,6 +380,13 @@ spec:
...
+      expr: sum(max(kube_pod_container_status_last_terminated_reason{reason="OOMKilled",job="kube-state-metrics"}))
```
LiliC: Do we care about all namespaces, including user namespaces? I would limit this to openshift namespaces like we do for other pod alerts. https://github.com/openshift/cluster-monitoring-operator/blob/2d04b3a7bb69c8333f132d39f91488d60b02a7b4/assets/prometheus-k8s/rules.yaml#L588

assets/prometheus-k8s/rules.yaml
rphillips: Thank you for the great comments. Updated.

assets/prometheus-k8s/rules.yaml
rphillips: Yes. This makes a ton of sense!

assets/prometheus-k8s/rules.yaml
rphillips: critical namespaces makes sense. A customer can configure additional alerts for their namespaces. +1

assets/prometheus-k8s/rules.yaml
```diff
@@ -380,6 +380,14 @@ spec:
       record: node_role_os_version_machine:cpu_capacity_sockets:sum
     - expr: clamp_max(sum(alertmanager_notifications_total),1)
       record: cluster:alertmanager_routing_enabled:max
+    - alert: KubePodOOMWarning
+      annotations:
+        message: Pod {{ $labels.namespace  }}/{{ $labels.pod  }} ({{ $labels.container  }})
+          is experiencing out of memory termination.
+      expr: sum(max(kube_pod_container_status_last_terminated_reason{reason="OOMKilled",namespace=~"(openshift-.*|kube-.*|default|logging)",job="kube-state-metrics"}))
```
LiliC: kube_pod_container_status_last_terminated_reason only contains as it says last terminated reason, as its a gauge. This would alert everytime we have a OOMKilled which seems very drastic and would lead to alert fatigue. I would prefer to collect the OOM reasons in a counter, but since we do not have this metric right now I would suggest something along the lines of this, note this query is just a rough idea and has not been tested:

```
kube_pod_container_status_last_terminated_reason{reason=“OOMKilled”,} == 1 
    and on(container) rate(kube_pod_container_status_restarts_total[5m]) * 300 > 1
```

I think while this does not mean all of those containers were terminated because OOM, it still gives us a better indication that something is actually wrong in the cluster.

Second thing, we tend to like to contribute all alerts in upstream, so once we decide on this, we should have this be in the upstream repo and we can bring it in. https://github.com/kubernetes-monitoring/kubernetes-mixin

cc @openshift/openshift-team-monitoring for thoughts and improvements :)

brancz: This has also been discussed numerous times on sig-instrumentation. The consensus was that ideally the kubelet or container runtime could expose a counter of oomkilled containers, that way we would actually have a reliable counter instead of this heuristic. That said, the heuristic suggested by @LiliC might be "good enough :tm: ".

rphillips:
Going to close this in favor of bringing up an OOM metric at the sig node meeting. Perhaps it should live in both the container runtime and kubelet. Thank you for the comments and review.

brancz:
That sounds like the best outcome I could have imagined :)

LiliC:
Thank you, this is great!

smarterclayton:
I want something in 4.4, even best effort.  Can we get that closed out before end of week?

lilic:
@smarterclayton best effort basis would be the above alert suggestion I made but have not tested. I will try my best but currently swamped with stuff.

smarterclayton:
@rphillips can you assign someone to work with lili?

lilic:
Came up with this, will open a PR we can go on from there. But would be good to see some activity on the kube PR as well :) 

```
 sum( increase(kube_pod_container_status_restarts_total[15m]) and on (namespace, container, pod)
kube_pod_container_status_last_terminated_reason{reason="OOMKilled"} == 1 ) > 5
```

simonpasquier:
@lilic the expression looks good to me.

PR #645, opened 2020-02-06
Bug 1799539: Node exporter init collectors must be privileged

smarterclayton:
dmidecode (used by the VM type detection) requires access to
/dev/mem which requires privileged. Only the init container gets
that permission. Blocks being able to read physical machine BIOS
info.


smarterclayton:
1799540 is 4.3 and 1799541 is 4.2

smarterclayton:
Oh, it might be SCC. Hold.

smarterclayton:
Now with the appropriate SCC.  I think this is still scoped enough (host pid and network are already pretty powerful, and we will eventually need to gather things at init time from /dev).

smarterclayton:
Opened separate back ports, testing this change with actual code.

PR #646, opened 2020-02-06
Bug 1799540: Node exporter init collectors must be privileged

smarterclayton:
dmidecode (used by the VM type detection) requires access to
/dev/mem which requires privileged. Only the init container gets
that permission. Blocks being able to read physical machine BIOS
info.


LiliC:
Generate should be fixed by https://github.com/openshift/cluster-monitoring-operator/pull/649

assets/node-exporter/daemonset.yaml
```diff
@@ -87,6 +87,7 @@ spec:
         name: init-textfile
         resources: {}
         securityContext:
+          privileged: true
```
LiliC: Think you are missing `allowPrivilegedContainer: true` as well.

bparees:
skipping this one for approval this week due to bug severity and the number of other changes already accepted.

PR #648, opened 2020-02-07
Bug 1798941: added record rule for workload type deploymentconfig

abhi-kn:
Metric queries for `DeploymentConfig` resources return with ZERO results due to missing `record` rule for `workload_type` `deploymentconfig`. Hence `record` rule has been added for `deploymentconfig` to resolve this issue.

### DIFF
--- a/assets/prometheus-k8s/rules.yaml
+++ b/assets/prometheus-k8s/rules.yaml
@@ -171,6 +171,16 @@ spec:
       labels:
         workload_type: statefulset
       record: mixin_pod_workload
+    - expr: |
+        sum(
+          label_replace(
+            kube_pod_owner{job="kube-state-metrics", owner_kind="DeploymentConfig"},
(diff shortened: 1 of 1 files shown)


assets/prometheus-k8s/rules.yaml
```diff
@@ -171,6 +171,16 @@ spec:
       labels:
         workload_type: statefulset
       record: mixin_pod_workload
+    - expr: |
```
LiliC: We use jsonnet, so these changes should first go into https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/rules.jsonnet and after commiting you run:
```
make clean
make generate-in-docker
```
and this and other files might be generated, let me know if you have any questions. :)

abhi-kn:
This change alone doesn't resolve the issue as there is no way to currently get the replication controller owner and this need to be added to kube-state-metrics to relate that information. Hence closing this PR for now.

PR #650, opened 2020-02-10
Bug 1794616: replace up with count:up[0,1], remove node_uname_info

s-urbaniak:
This reduces size for the `up` metric by replacing it with `count:up0` and `count:up1` recording rules and removes `node_uname_info` as it is unused.

Once we reach final consensus here especially with @smarterclayton this will be backported to 4.3.

/cc @smarterclayton @openshift/openshift-team-monitoring


s-urbaniak:
Whoops, i actually wanted `/hold` it, until we reach final consensus, my bad. We can always revert if we find problems with the recording rules here.

brancz:
happy with this

LiliC:
Do we need to tell whoever added that info metric that we removed it?

brancz:
yeah probably a good idea :)

LiliC:
Opened https://issues.redhat.com/browse/MON-970 to make this easier in the future.

s-urbaniak:
The removal of the info metric was synced with @smarterclayton though as he was shepherding its removal in the first place.

PR #652, opened 2020-02-13
Bug 1787520: fix query showing metrics collected by telemetry

simonpasquier:
This isn't 100% accurate as it includes alerts in pending state (versus
firing alerts only) but it is better than what we had.

Generating the exact query using jq would be really cumbersome. The
alternative would be to write a Go program to parse the label selectors
used for federation.


lilic:
/hold

lets see if we can figure out the ALERTS first, if not we can remove this hold

PR #653, opened 2020-02-14
Bug 1803827: pkg: Generate recording rule to measure total number of series we are

lilic:
sending via telemetry.

This also includes the rule itself. Already spoke with Clayton and he approved we can send this info via telemetry.

/hold 
want to test the full thing but cluster-bot is not working currently.


Documentation/data-collection.md
```diff
@@ -175,6 +175,9 @@ data:
     # openshift-apiserver with request type which can be either 'mutating' or 'readonly'.
     # We want to have an idea of how loaded our api server(s) are globally.
     - '{__name__="cluster:apiserver_current_inflight_requests:sum:max_over_time:2m"}'
+    # (monitoring-team) cluster:telemetry_selected_series:count tracks the total number of series
```
lilic: @brancz suggested the following format:
>  This metric is owned by the monitoring team. Contact #forum-monitoring on slack for details.

pkg/promqlgen/promqlgen.go
```diff
@@ -64,3 +61,53 @@ func LabelSelectorsToRelabelConfig(matches []string) (*monv1.RelabelConfig, erro
...
+	for k := range newLabelSet {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	res := "{"
+	i := 0
+	for _, k := range keys {
+		res += k + `=~"`
+		res += strings.Join(newLabelSet[k], "|") + `"`
```
simonpasquier: We need to add a trailing `|` to the label selectors otherwise it will filter too many timeseries. Which means `{__name__=~"ALERTS|node_uname_info|csv_abnormal|",alertstate=~"firing|"}` instead of `{__name__=~"ALERTS|node_uname_info|csv_abnormal",alertstate=~"firing"}`.

lilic: For some reason, probably something obvious:
`{__name__=~"ALERTS|node_uname_info|csv_abnormal|",alertstate=~"firing|"}` 
does not work, but promql says:
> Error executing query: invalid parameter ‘query’: 1:1498: parse error: vector selector must contain at least one non-empty matcher

`{__name__=~"ALERTS|node_uname_info|csv_abnormal",alertstate=~"firing|"}` instead did this and it seems to produce the correct number.

lilic:
@openshift/openshift-team-monitoring PTAL, thanks!

pkg/promqlgen/promqlgen.go
simonpasquier: Right, removing the extra `|` in the `__name__` selector as you suggest is the right thing to do.

pkg/promqlgen/promqlgen.go
```diff
@@ -64,3 +61,55 @@ func LabelSelectorsToRelabelConfig(matches []string) (*monv1.RelabelConfig, erro
...
+	}
+	sort.Strings(keys)
+
+	res := "{"
+	i := 0
+	for _, k := range keys {
+		res += k + `=~"`
+		res += strings.Join(newLabelSet[k], "|")
+		i++
+		if i == len(newLabelSet) {
```
simonpasquier: wouldn't it be more correct to say that we add the trailing `|` unless `k == "__name__"`?

lilic: hmm good point, but what if we have multiple label keys?

lilic: hmm okay, yes I see what you mean, it just needs one so it will always work, applying suggestion.

pkg/promqlgen/promqlgen.go
```diff
@@ -23,13 +24,9 @@ import (
...
-	labelSets := make([][]*labels.Matcher, len(matches))
-	var err error
-	for i, m := range matches {
-		labelSets[i], err = promql.ParseMetricSelector(m)
-		if err != nil {
-			return nil, err
-		}
+	labelSets, err := parseMetricSelectorFromArray(matches)
+	if err != nil {
+		return nil, err
```
s-urbaniak: can we wrap the error here so we have better context for debugging?

lilic:
@s-urbaniak @simonpasquier PTAL, thanks!

PR #654, opened 2020-02-17
Bug 1803192: *: duplicate shared configmap in openshift-config-managed namespace

simonpasquier:


pkg/manifests/manifests.go
```diff
@@ -1043,11 +1048,19 @@ func (f *Factory) ThanosQuerierRoute() (*routev1.Route, error) {
 	return r, nil
 }
 
+func (f *Factory) SharingConfigDeprecated(promHost, amHost, grafanaHost, thanosHost *url.URL) *v1.ConfigMap {
```
s-urbaniak: let's add a unit test in `manifests_test.go`

s-urbaniak:
one mini-nit, else lgtm, thank you!

pkg/manifests/manifests.go
simonpasquier: done

s-urbaniak:
side note: i believe we need to set up a role binding against the `openshift-config-managed` namespace in https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster_monitoring_operator_02-role.yaml.

it might be that e2e fails because we have no permission to create a configmap in that namespace.

simonpasquier:
> it might be that e2e fails because we have no permission to create a configmap in that namespace.

I wasn't sure. If e2e tests fail because of this, I'll add the required permission for sure.

PR #655, opened 2020-02-17
Bug 1794616: pkg/manifests: set modified telemeter-client command line arguments

s-urbaniak:
Currently we are manipulating but not changing them. This fixes it.
The current symptom is that the new metrics configuration scheme
is effectively not enabled.


s-urbaniak:
@lilic yes indeed, I can implement an e2e test that verifies the command line arguments
@simonpasquier hmm ... good idea! maybe via some metric that telemeter-client exposes that it successfully sent metrics. does something like this also exist for prometheus remote-write?

lilic:
We should yes plan with remote-write in mind, and not invest too much time with the current stack as it will hopefully go away soon.

PR #658, opened 2020-02-18
Bug 1796040: pkg/manifests: change location of trustedCA bundle; add trustedCA to …

lilic:
…alertmanager oauth-proxy

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -331,6 +331,11 @@ func (f *Factory) AlertmanagerMain(host string, trustedCABundleCM *v1.ConfigMap)
 
 	if trustedCABundleCM != nil {
 		volumeName := "alertmanager-trusted-ca-bundle"
+		for in, container := range a.Spec.Containers {
+			if container.Name == "alertmanager-proxy" {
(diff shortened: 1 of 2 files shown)


s-urbaniak:
so far lgtm :+1:

lilic:
/retest

flake

lilic:
@s-urbaniak PTAL, thanks!

pkg/manifests/manifests.go
```diff
@@ -331,6 +331,11 @@ func (f *Factory) AlertmanagerMain(host string, trustedCABundleCM *v1.ConfigMap)
 
 	if trustedCABundleCM != nil {
 		volumeName := "alertmanager-trusted-ca-bundle"
+		for in, container := range a.Spec.Containers {
```
s-urbaniak: nit: isn't the container index deterministic. we rely on this invariant in a couple of other places.

lilic: I would not like to rely on an order of the containers, even if we are doing it other places. Don't see a problem with this current implementation. I would even suggestion to not make it deterministic, as its bad practice IMHO, as we clearly all don't know that. :D

PR #659, opened 2020-02-18
Bug 1787520: replace ad-hoc script generating telemeter doc

simonpasquier:
This PR leverages #653 so we're now in a situation where the generated documentation always matches with what we send to telemetry.


lilic:
Thanks! Had plan to do the same 😂

PR #668, opened 2020-02-20
jsonnet/rules.jsonnet: Add MultipleContainersOOMKilled alert

lilic:
Still need to test this again, but should be fine. :)


jsonnet/rules.jsonnet
```diff
@@ -196,6 +196,17 @@
               severity: 'warning',
             },
           },
+          {
+            expr: ' sum(increase(kube_pod_container_status_restarts_total[15m]) and on (namespace, container, pod) kube_pod_container_status_last_terminated_reason{reason="OOMKilled"} == 1 ) > 5',
```
simonpasquier: you might want this to avoid many-to-many errors:

```
sum(max by(namespace, container, pod) (increase(kube_pod_container_status_restarts_total[15m])) and max by (namespace, container, pod) (kube_pod_container_status_last_terminated_reason{reason="OOMKilled"}) == 1) > 5
```

lilic: Done, thanks!

lilic:
@openshift/openshift-team-monitoring PTAL, still want to test again before merging.

/hold

lilic:
Background on this -> https://github.com/openshift/cluster-monitoring-operator/pull/642#issuecomment-587242777

lilic:
/hold cancel
@simonpasquier PTAL, thanks!
I created a deployment that had overcommitted memory to our long lived cluster, LLAP, to test out and fine-tune this alert. I left it running overnight. After 4 containers kept OOMing over a period of approx 5 hours, iptables stale alert fired and networking went down. That caused other components to also become degraded because of the node being down and later not ready. No idea if this is related, can't confirm but networking team said it is possible, incident doc here -> https://docs.google.com/document/d/1PEd4XQREQQA9F7hMjLRCSP_tjO1WtjhdjBumeavAl50/edit?usp=sharing

The importance of the alert is clear, but while more containers can cause damage or signal something is wrong, we can assume problems could happen more frequently when containers OOM in the long run, so I changed this to use smaller amount of containers and extend the for to longer, as the duration here is key from my observations, we do not want the alerts to fire too soon if one container OOMs or if few OOM for 2 minutes. 

@smarterclayton as you requested the alert I assume you want this in 4.4 already not just 4.5?

PR #675, opened 2020-02-24
Bug 1802941: Fix more many to many errors

simonpasquier:
I've spotted those issues by running 2 kube-state-metrics instances.

cc @openshift/openshift-team-monitoring-maintainers


s-urbaniak:
nice! Does this effectively address https://bugzilla.redhat.com/show_bug.cgi?id=1802941? If so, I suggest to reference the PR title to that BZ and potentially backport to 4.4?

brancz:
> I've spotted those issues by running 2 kube-state-metrics instances.

I think it would be very worthwhile if in our e2e tests we could somehow run 2 instances of kube-state-metrics to tickle out these types of errors before we ship them to customers and get bugs opened.

simonpasquier:
@brancz totally agree. Let me create an item to track this.

simonpasquier:
> Does this effectively address https://bugzilla.redhat.com/show_bug.cgi?id=1802941? If so, I suggest to reference the PR title to that BZ and potentially backport to 4.4?

Technically these issues were not listed in BZ 1802941 so I can either a) create a new BZ or b) extend the original BZ.

s-urbaniak:
@simonpasquier let's then extend the existing one. Can we fix the existing issue in https://bugzilla.redhat.com/show_bug.cgi?id=1802941 here too?

simonpasquier:
@s-urbaniak 

> Can we fix the existing issue in https://bugzilla.redhat.com/show_bug.cgi?id=1802941 here too?

Some have already been addressed in #670. And there are other rules than need to be fixed upstream first (e.g. https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/361).

PR #676, opened 2020-02-25
Bug 1807140: jsonnet/rules.jsonnet: Add MultipleContainersOOMKilled alert 

lilic:

/hold
Lets see if we need this adjusted first or if its fine like this. So far in 4.4 there were no alerts firing for this, but then we only have CI and nightlies and the OOM bug was fixed.


jsonnet/rules.jsonnet
```diff
@@ -196,6 +196,17 @@
...
           },
+          {
+            expr: 'sum(max by(namespace, container, pod) (increase(kube_pod_container_status_restarts_total[12m])) and max by(namespace, container, pod) (kube_pod_container_status_last_terminated_reason{reason="OOMKilled"}) == 1) > 10',
+            alert: 'MultipleContainersOOMKilled',
+            'for': '15m',
+            annotations: {
+              message: 'Multiple containers were out of memory killed within the past 15 minutes.',
+            },
+            labels: {
+              severity: 'info',
```
paulfantom: Why `info`? OOM means we don't have enough memory in a cluster and it should be scaled, which in many cases is a manual action requiring human intervention and possibly a page. Can we raise it to `warning`?

lilic: No, because this alert is not exact we do not want to do warning level, decided in the PR that was merged and offline conversations with the team. They are adding a proper metric in crio directly in 4.5 that will actually have the info we need on how many containers are being OOM and we will remove this alert and make it warning then. For now as this is good enough alert and its mainly used for us to get a ballpark on how often we see OOMs in the wild.

See https://github.com/openshift/cluster-monitoring-operator/pull/668 and https://github.com/openshift/cluster-monitoring-operator/pull/642#discussion_r375254585

brancz: It's mostly because this is a heuristic, we assume that if a container restarted and it's last termination reason was an OOM, then all of those restarts were OOMs. If we truly knew that x amount of OOMs happened then I would agree to making this warning but with this heuristic I would be more careful.

lilic:
note: This lowers the limit, compared to master, to 3 containers OOMing. I tested and this is the rough guesstimate of my testing, we can always adjust later.
Screenshot 2020-02-26 at 12 58 27


Should be ready to be merged. @openshift/openshift-team-monitoring PTAL

/hold cancel

jsonnet/rules.jsonnet
```diff
@@ -196,6 +196,17 @@
               severity: 'warning',
             },
           },
+          {
+            expr: 'sum(max by(namespace, container, pod) (increase(kube_pod_container_status_restarts_total[12m])) and max by(namespace, container, pod) (kube_pod_container_status_last_terminated_reason{reason="OOMKilled"}) == 1) > 5',
```
brancz: iirc 12min comes from some kube behavior right? Do we have some docs about this to link to for future Frederic to understand?

lilic: ha :D yes this is the way we restart containers, no docs just a lot of testing with various numbers of timespans, so it eventually normalised the graph.

metalmatze:
Looking at this without extra context is there a reason this isn't added to the kubernetes-mixin project first and then in here?
Is it the urgency of the Bugzilla? Doesn't seem to Openshift specific. 🤔

lilic:
@metalmatze we will be adding the proper alert in the mixin project when the crio metric is available, see the metric in the PR linked in the previous PR that was merged into master, this PR is just backporting it :D This is a guess alert and not exact at all, we want to fully test it and just want to get an indication of how often these alerts fire, see discussion in this PR a bit above :) https://github.com/openshift/cluster-monitoring-operator/pull/676#discussion_r384398225 and its also detailed in the bugzilla linked and on slack. ;) I have created a jira ticket for that so we don't forget to remove this once the proper metric that tracks the number of OOM killed containers is added.

jsonnet/rules.jsonnet
brancz: ok, that works for me .. no need for comment then

s-urbaniak:
@lilic this needs a rebase now

lilic:
@smarterclayton to have this merged in 4.4 we need to have the dependant bugzilla in verified or modified state, you mentioned you will get it verified?

smarterclayton:
#668 was at 10, you need change master to 5 and then associate it with 1807139 and then once it merges this will merge (you don't need a verified PR in master, you just need a merged PR in master)

lilic:
PR opened https://github.com/openshift/cluster-monitoring-operator/pull/690

lilic:
Had to rebase so lgtm was removed @openshift/openshift-team-monitoring PTAL

PR #696, opened 2020-03-11
Connect thanos ruler with thanos querier

lilic:
/hold

This needs https://github.com/openshift/prometheus-operator/pull/57 to be merged first and the image to be built so we use it.

PR #697, opened 2020-03-11
update alertmanagersUrl field of thanos ruler

pgier:
Fix capitalization and convert to string slice


pgier:
Closing because this will be fixed in #696

PR #699, opened 2020-03-12
Bug 1812719: jsonnet/*: adapt resource requests

s-urbaniak:
This adapts resource requests to lower values.
Rationale: cpu is compressible and we have to make sure we stay within tight boundaries.

Values have been taken from actually used cpu time using the following query: 
```
sum by(pod) (rate(container_cpu_usage_seconds_total{pod=~"POD_NAME*",container=""}[15m])*1024)
```


assets/alertmanager/alertmanager.yaml
```diff
@@ -51,7 +51,7 @@ spec:
       name: web
     resources:
       requests:
-        cpu: 10m
+        cpu: 1m
```
lilic: wow now thats drastic :D

assets/prometheus-k8s/prometheus.yaml
```diff
@@ -154,7 +157,7 @@ spec:
     image: quay.io/thanos/thanos:v0.10.0
     resources:
       requests:
-        cpu: 50m
+        cpu: 1m
```
lilic: is this enough for thanos?

s-urbaniak: This is for the thanos sidecar next to prometheus. In the e2e run I analyzed ~1millicore CPU:
image

assets/alertmanager/alertmanager.yaml
s-urbaniak: this is the alertmanager-proxy (oauth-proxy) sidecar. In the e2e run I analzyed ~1millicore CPU:
image

smarterclayton: Unfortunately anything less than 5m has no effect, the scheduler has an implicit floor at 5m.

smarterclayton: Let me double check this, we have a minimum floor for memory.

smarterclayton: Nm, we don't.  This should be fine.

s-urbaniak:
/hold
master is currently broken in master due to work in user workload monitoring, i will open a PR against 4.4 directly.

lilic:
Can you rebase, this should be ready to go.

lilic:
upgrade test passed but something went wrong with prow:
> 2020/03/13 10:06:25 Build cluster-monitoring-operator succeeded after 1m56s
2020/03/13 10:06:26 Tagging cluster-monitoring-operator into stable

lilic:
/hold cancel

🤷‍♀

PR #701, opened 2020-03-12
Add kube-rbac-proxy and prom-label-proxy to Alertmanager

simonpasquier:
~~This needs e2e tests.~~


assets/alertmanager/alertmanager.yaml
```diff
@@ -59,6 +59,39 @@ spec:
...
+    - --logtostderr=true
+    - --v=10
+    image: quay.io/coreos/kube-rbac-proxy:v0.4.1
+    name: kube-rbac-proxy
+    ports:
+    - containerPort: 9092
+      name: tenancy
+    resources:
+      requests:
+        cpu: 10m
```
s-urbaniak: let's put this down to `1m`. Current measurements show this doesn't need more (at least for requests).

assets/alertmanager/alertmanager.yaml
```diff
@@ -59,6 +59,39 @@ spec:
...
+      name: secret-alertmanager-kube-rbac-proxy
+  - args:
+    - --insecure-listen-address=127.0.0.1:9096
+    - --upstream=http://127.0.0.1:9093
+    - --label=namespace
+    image: quay.io/coreos/prom-label-proxy:v0.1.0
+    name: prom-label-proxy
+    resources:
+      requests:
+        cpu: 10m
```
s-urbaniak: ditto

s-urbaniak:
just nits around resource requests, else lgtm module e2e tests :+1:

lilic:
Why do we need this PR? Is this for thanos rule integration with alertmanager? 🤔

s-urbaniak:
@lilic xref https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/user-workload-monitoring.md#alertmanager-silences

brancz:
lgtm as well after e2e tests

s-urbaniak:
@simonpasquier i believe it is unrelated but do you mind to peek at the e2e failure?
```
level=error msg="Cluster operator monitoring Degraded is True with UpdatingAlertmanagerFailed: Failed to rollout the stack. Error: running task Updating Alertmanager failed: waiting for Alertmanager object changes failed: waiting for Alertmanager: expected 3 replicas, updated 0 and available 0"
level=fatal msg="failed to initialize the cluster: Cluster operator monitoring is still updating"
```

pkg/manifests/manifests.go
```diff
@@ -350,6 +353,14 @@ func (f *Factory) AlertmanagerTrustedCABundle() (*v1.ConfigMap, error) {
 	return cm, nil
 }
 
+const (
+	// These constants refer to indices of alertmanager-main containers.
+	// They need to be in sync with jsonnet/alertmanager.jsonnet
+	ALERTMANAGER_CONTAINER_OAUTH_PROXY      = 0
+	ALERTMANAGER_CONTAINER_KUBE_RBAC_PROXY  = 1
+	ALERTMANAGER_CONTAINER_PROM_LABEL_PROXY = 2
```
brancz: in other places we instead started converting this to a container name based scheme, to avoid having to guarantee order in generated files, not a blocker, but since this is net new maybe worth doing

simonpasquier: ok I wasn't sure since I've at least 3 ways of doing it but happy to fix it in this PR!

manifests/0000_50_cluster_monitoring_operator_02-role.yaml
```diff
@@ -145,6 +147,25 @@ rules:
   - namespaces
   verbs:
   - get
+- apiGroups:
```
lilic: Why do we need this?

pkg/manifests/manifests.go
lilic: Yes, please, container name based approach is better as we have had few problems with the ordered approach as we do not know if those change, whereas name based will fail, and we should at least be migrating away from this IMHO.

pkg/tasks/clustermonitoringoperator.go
```diff
@@ -54,6 +54,26 @@ func (t *ClusterMonitoringOperatorTask) Run() error {
 		return errors.Wrap(err, "reconciling cluster-monitoring ClusterRole failed")
 	}
 
+	cr, err = t.factory.ClusterMonitoringRulesEditClusterRole()
```
lilic: This is the edit cluster rule not that we are editing the cluster role, correct? :)

test/e2e/alertmanager_test.go
```diff
@@ -161,3 +170,220 @@ func TestAlertmanagerTrustedCA(t *testing.T) {
...
+		t.Logf("Deleting silence as %q (denied)", sa)
+		_ = clients[sa].do("DELETE", fmt.Sprintf("/api/v2/silence/%s", silID), nil, http.StatusForbidden)
+	}
+
+	t.Log("Deleting silence as 'editor' (allowed)")
+	_ = clients["editor"].do("DELETE", fmt.Sprintf("/api/v2/silence/%s", silID), sil, http.StatusOK)
+}
+
+// The Alertmanager API should be protected by the OAuth proxy.
+func TestAlertmanagerOAuthProxy(t *testing.T) {
```
lilic: Can you explain how this is testing alertmanger oauth proxy, we just do a query against alertmanager.

test/e2e/alertmanager_test.go
```diff
@@ -161,3 +170,220 @@ func TestAlertmanagerTrustedCA(t *testing.T) {
...
+
+	t.Log("Creating silence as 'editor' (allowed)")
+	b := clients["editor"].do("POST", "/api/v2/silences", sil, http.StatusOK)
+
+	// Save silence ID for deletion.
+	parsed, err := gabs.ParseJSON(b)
+	if err != nil {
+		t.Fatalf("%v (data: %q)", err, string(b))
+	}
+	silID, ok := parsed.Path("silenceID").Data().(string)
```
brancz: let's additionally validate that the namespace label exists no?

manifests/0000_50_cluster_monitoring_operator_02-role.yaml
simonpasquier: I'm not sure, it was automatically generated by the script...

cc @s-urbaniak @paulfantom

pkg/tasks/clustermonitoringoperator.go
simonpasquier: correct

test/e2e/alertmanager_test.go
simonpasquier: We're querying Alertmanager via its OpenShift route which *should* be mapped to the OAuth proxy service. This is copied from https://github.com/openshift/cluster-monitoring-operator/blob/e597c67470412d5094d162f415c8b0d7cd3fe7f9/test/e2e/user_workload_monitoring_test.go#L376-L403

I should probably add an e2e test checking that a request without sufficient permissions is denied.

test/e2e/alertmanager_test.go
simonpasquier: done, the original silence is created with a `namespace="openshift-monitoring"` matcher and the test verifies that it is modified to  `namespace="test-kube-rbac-proxy"`.

simonpasquier:
Thanks Lili! cc @brancz

PR #702, opened 2020-03-12
Bug 1812565: Remove cpu limits for config reloader

pgier:
This affects the config-reloader containers for alertmanager and prometheus instances.


pgier:
cc @s-urbaniak

simonpasquier:
I think you need to update jsonnet-bundler to v0.3.1.

PR #705, opened 2020-03-12
Bug 1809353: Exclude Kubernetes control plane rules when running on IBM Cloud

csrwng:
If the current cluster's platform is `IBMCloud`, Kubernetes control plane rules are skipped.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -30,6 +30,7 @@ import (
 	"strings"
 
 	monv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	securityv1 "github.com/openshift/api/security/v1"
 	"github.com/openshift/cluster-monitoring-operator/pkg/promqlgen"
@@ -217,6 +218,10 @@ var (
 
 	PrometheusTrustedCABundleDir  = "/etc/pki/prometheus-ca-bundle/"
 	PrometheusTrustedCABundlePath = PrometheusTrustedCABundleDir + "ca-bundle.crt"
+)
+
+const (
+	IBMCloudPlatformType configv1.PlatformType = "IBMCloud"
 )
 
 func MustAssetReader(asset string) io.Reader {
@@ -798,6 +803,21 @@ func (f *Factory) PrometheusK8sRules() (*monv1.PrometheusRule, error) {
 		groups := []monv1.RuleGroup{}
 		for _, g := range r.Spec.Groups {
 			if g.Name != "etcd" {
+				groups = append(groups, g)
+			}
+		}
+		r.Spec.Groups = groups
+	}
+
+	if f.config.Platform == IBMCloudPlatformType {
+		groups := []monv1.RuleGroup{}
+		for _, g := range r.Spec.Groups {
+			switch g.Name {
+			case "kubernetes-system-apiserver",
+				"kubernetes-system-controller-manager",
+				"kubernetes-system-scheduler":
+				// skip
+			default:
 				groups = append(groups, g)
 			}
 		}
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -22,6 +22,8 @@ import (
 	"testing"
 
 	monv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	configv1 "github.com/openshift/api/config/v1"
+
 	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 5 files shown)


csrwng:
Alternate implementation to https://github.com/openshift/cluster-monitoring-operator/pull/687
@s-urbaniak ptal
/cc @derekwaynecarr

pkg/manifests/manifests_test.go
```diff
@@ -1176,6 +1177,63 @@ func TestOpenShiftStateMetrics(t *testing.T) {
...
+		r, err := f.PrometheusK8sRules()
+		if err != nil {
+			t.Fatal(err)
+		}
+		apiServerRulesFound := false
+		controllerManagerRulesFound := false
+		schedulerRulesFound := false
+		for _, g := range r.Spec.Groups {
+			switch g.Name {
+			case "kubernetes-system-apiserver":
```
brancz: what about etcd? and what about the servicemonitors that are provisioned to monitor these in the first place?

pkg/manifests/manifests.go
```diff
@@ -219,6 +220,10 @@ var (
 	PrometheusTrustedCABundlePath = PrometheusTrustedCABundleDir + "ca-bundle.crt"
 )
 
+var (
+	IBMCloudPlatformType configv1.PlatformType = "IBMCloud"
```
smarterclayton: Don't redefine constant as vars, that's horribly dangerous.

pkg/manifests/manifests_test.go
csrwng: Service monitors that are part of control plane operators are skipped via annotation: https://github.com/openshift/cluster-kube-apiserver-operator/blob/4de583dffa47c25adde871e7361386b0d55c674a/manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml#L7
Etcd rules are disabled because the `etcd-metric-client` secret is not placed in the `openshift-config` namespace: https://github.com/openshift/cluster-monitoring-operator/blob/207ec4ea710aa7397dd9cd578d09b4d4a8606074/pkg/operator/operator.go#L418-L421

pkg/manifests/manifests.go
csrwng: Thx! fixed

pkg/manifests/manifests_test.go
brancz: I see ok.

brancz:
There is mostly a long term maintenance question here. The team is already way beyond full capacity, and this adds yet another dimension of bugs that are inevitably going to occur. Who takes care of that?

brancz:
Can we have a ROKS component in bugzilla for example that these types of bugs would be assigned to?

s-urbaniak:
also cc'ing @bparees to consult if existing components/repositories like cluster-monitoring-operator can be assigned to this type of initiatives like ROKS and not our team. I didn't find a ROKS component in bugzilla.

bparees:
> Can we have a ROKS component in bugzilla for example that these types of bugs would be assigned to?

i don't know that we need a ROKS BZ component, but it sounds like we need a ROKS-enablement cross-team intiative with an identified owner and tracking jira stories.

imho this is similar to any other cloud provider enablement, though i realize there's significantly more structual differences in ROKS vs just another cloud provider.

I don't view this change as a bug fix.

derekwaynecarr:
@bparees see epic and jira tracking:

https://issues.redhat.com/browse/CO-716

an agreed upon solution is required this week and needs backport to 4.3.z per our discussion last week.  we are happy to improve testing in any form for any managed service as a follow-on to ensure regressions do not occur.  this could include managed service specific testing scenarios in existing openshift e2e.

bparees:
Turns out there is a jira epic for this: https://issues.redhat.com/browse/CO-716 and i think this work is specifically tied to https://issues.redhat.com/browse/CO-756

but i'd reiterate that this is basically another platform so having its own BZ component doesn't make sense.

lilic:
> but i'd reiterate that this is basically another platform so having its own BZ component doesn't make sense.

Where do you think it should go then? Not under monitoring :) 

Note that we are not against this feature or merging this PR. We only want to ensure that we do not support this feature as we go onwards, or the backports and there is a component we can reassign any bugzillas we might potentially get.

s-urbaniak:
The problem from our side is long term maintainability. If, at any point in time, a BZ is opened that something doesn't work in a ROKS environment, we are effectively flying blind to fix it.

Having a ROKS component owner/team would solve that specific concern. Someone who can verify and fix bugs in that environment.

bparees:
@lilic @s-urbaniak let's take this discussion offline

csrwng:
/retitle Bug 1809353: Exclude Kubernetes control plane rules when running on IBM Cloud

PR #707, opened 2020-03-13
Remove thanos ruler CRD

simonpasquier:
The CRD is already created automatically by the Prometheus operator running in `openshift-monitoring`. Also the asset isn't used by the CMO.


lilic:
there is a problem with CI right now.

PR #709, opened 2020-03-16
Bug 1812834: schedule Thanos Querier on worker nodes, make resources configurable

s-urbaniak:


pkg/manifests/manifests.go
```diff
@@ -2508,6 +2508,22 @@ func (f *Factory) ThanosQuerierDeployment(grpcTLS *v1.Secret, enableUserWorkload
 		d.Spec.Template.Spec.Volumes = append(d.Spec.Template.Spec.Volumes, volume)
 	}
 
+	if f.config.ThanosQuerierConfig.Resources != nil {
```
lilic: So we just default to worker nodes?

s-urbaniak:
/hold
@lilic indeed, i was just about to add a test

pkg/manifests/manifests.go
s-urbaniak: correct, that is what i synced with @brancz

s-urbaniak:
/hold cancel
@lilic ptal

PR #713, opened 2020-03-18
[release-4.3] Bug 1814736: Exclude Kubernetes control plane rules when running on IBM Cloud

csrwng:
Backport of #705


csrwng:
Added permission for infrastructures to cluster role

lilic:
The bug has not been verified yet it seems.

PR #714, opened 2020-03-18
prometheus rule validating webhook

pgier:
This introduces a new ValidatingWebhookConfiguration which adds
validation for new and updated PrometheusRule objects across all namespaces.


s-urbaniak:
I think we are missing some RBAC:
```
level=error msg="Cluster operator monitoring Degraded is True with UpdatinguserworkloadPrometheusOperatorFailed: Failed to rollout the stack. Error: running task Updating user workload Prometheus Operator failed: deleting Prometheus Rule ValidatingWebhook failed: validatingwebhookconfigurations.admissionregistration.k8s.io \"prometheusrules.openshift.io\" is forbidden: User \"system:serviceaccount:openshift-monitoring:cluster-monitoring-operator\" cannot delete resource \"validatingwebhookconfigurations\" in API group \"admissionregistration.k8s.io\" at the cluster scope"
level=fatal msg="failed to initialize the cluster: Cluster operator monitoring is still updating"
```

pgier:
/hold 
not sure if we want to switch the validation hook to run against the cluster prometheus operator instead of the user workload monitoring one.

lilic:
> not sure if we want to switch the validation hook to run against the cluster prometheus operator instead of the user workload monitoring one.

I thought I seen some conversations on slack about this that it was decided?

s-urbaniak:
yes, according to @brancz we indeed want to run the hook on the in-cluster prometheus-operator.

brancz:
> yes, according to @brancz we indeed want to run the hook on the in-cluster prometheus-operator.

the validations are exactly the same no matter which instance is used. Whichever one is more convenient, but it could be useful to always have for cluster-monitoring rules as well.

pgier:
The tests seem to be consistently failing with an authorization issue.
```
Error: running task Updating Prometheus Operator failed: 
reconciling Prometheus Rule Validating Webhook failed: 
creating ValidatingWebhookConfiguration object failed: 
validatingwebhookconfigurations.admissionregistration.k8s.io is forbidden: 
User \"system:serviceaccount:openshift-monitoring:cluster-monitoring-operator\" cannot create resource \"validatingwebhookconfigurations\" in API group \"admissionregistration.k8s.io\" at the cluster scope"
```

But I'm not sure why this is failing now, and didn't fail previously when the webhook was set to point to the user workload namespace.  AFAICT, the role used by the CMO service account should have the permissions to create the webhook.

assets/prometheus-operator/prometheus-rule-validating-webhook.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: prometheusrules.openshift.io
```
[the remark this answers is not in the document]
pgier: Yes, I think that would be good.  I'll add labels to match the other prometheus operator assets.

lilic:
/hold

to stop retesting

pgier:
@lilic You're right, we should have at least one test.  I'll try to figure out a way to test this.

pgier:
I think the permissions issue is fixed now, and I added some tests.  But the tests will fail until I switch this over to using TLS between k8s and the webhook.

lilic:
@pgier I forgot what we had left to do for this, do you mind just recapping it here as well, thanks!

pgier:
This is currently blocked by the lack of TLS config in prometheus-operator.  Kubernetes validatingwebhooks require an HTTPS connection, and currently the only way to do that with prometheus-operator would be with some proxy sidecar.  We discussed the different options for handling this, and it was decided to add TLS support directly to prometheus-operator, so I'm working on that currently.

pgier:
This depends on https://github.com/openshift/prometheus-operator/pull/65

s-urbaniak:
@pgier looks like a valid e2e failure:

```
level=error msg="Cluster operator monitoring Degraded is True with UpdatingPrometheusK8SFailed: Failed to rollout the stack. Error: running task Updating Prometheus-k8s failed: reconciling Prometheus rules PrometheusRule failed: creating PrometheusRule object failed: Internal error occurred: failed calling webhook \"prometheusrules.openshift.io\": expected webhook response of admission.k8s.io/v1, Kind=AdmissionReview, got /, Kind="
```

lilic:
I think it's a valid failure, the alerts that is firing is:

> {\"metric\":{\"alertname\":\"TargetDown\",\"alertstate\":\"firing\",\"job\":\"prometheus-operator\",\"namespace\":\"openshift-monitoring\",\"service\":\"prometheus-operator\",\"severity\":\"warning\"},\"value\":[1587448911.751,\"45\"]}

lilic:
Note that feature freeze is over so we should be aiming this towards the next release, as it’s not a bug. @s-urbaniak unless I missed something?

s-urbaniak:
yes, I agree this is too invasive to be brought in now, it's not a bug but a feature we worked on. let's stick to the regular cadence.

pgier:
@s-urbaniak @lilic :+1:  should I unlink and close the BZ since we already have a jira for this?

lilic:
Yes, sgtm, I think @s-urbaniak mentioned we should create a master-4.6 branch to aim new features against that. Thanks!

lilic:
@pgier can you aim this towards the master-4.6 branch and I assume its then ready for review?

assets/prometheus-operator/deployment.yaml
```diff
@@ -43,13 +44,18 @@ spec:
...
         securityContext: {}
         terminationMessagePolicy: FallbackToLogsOnError
+        volumeMounts:
+        - mountPath: /etc/tls/private
+          name: prometheus-operator-tls
+          readOnly: false
       - args:
         - --logtostderr
         - --secure-listen-address=:8443
         - --tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
```
lilic, line 54: Should this not be --web.tls-cipher-suites?

pgier: Hmm, I think you are right, I'm not sure why this wasn't caught by the tests.

pgier: Ok, that one is used by the kube-rbac-proxy and not by prometheus-operator, so it should be fine.  This PR currently doesn't set the cipher-suites for prometheus-operator, so it will just use the default ones providied by the golang API.  I guess this should be updated to match the ones used by kube-rbac-proxy?  Or is there another standard set we should be using for openshift?

pgier: I'll update the jsonnet to pull from the same ciphersuites field as the other assets.

pgier: I've added the ciphersuites to the prometheus-operator jsonnet config, so it should match the other deployments now.

lilic:
@s-urbaniak can you add the valid bugzilla label please, thanks! :)

lilic:
Or @brancz Can you add a valid-bugzilla label, thanks :)

lilic:
/retest

seems like a prow problem, the tests actually passed

lilic:
/hold

something off with the tests, also remembered we should update the CHANGELOG as this is a user facing feature @pgier do you mind just doing that, and by then AWS might be happier and pass. :)

pgier:
rebased and added a CHANGELOG entry

PR #716, opened 2020-03-20
Bug 1809353: Add infrastructures.config.openshift.io to cluster-monitoring-operator role

csrwng:
Follow up to #705 
The `cluster-monitoring-operator` service account needs access to infrastructure resources.


csrwng:
@lilic @s-urbaniak ptal

lilic:
Random failures in CI ranging from 
> 2020/03/20 10:11:24 wait for prometheus-k8s: error executing prometheus query: unexpected status code response, want 200, got 403: timed out waiting for the condition

to actual prow failures.

PR #720, opened 2020-03-23
update prometheus-operator to v0.38.0

pgier:


lilic:
I guess we have to merge prom-operator PR first and then this one. Judging by the failure.

pgier:
Looks like it was a flake, I think the prometheus rule test failure was just a coincidence.

PR #731, opened 2020-04-01
Bug 1807430: don't sync on invalid configuration

simonpasquier:

### DIFF
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -1,0 +1,120 @@
+// Copyright 2020 The Cluster Monitoring Operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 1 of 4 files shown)


simonpasquier:
@lilic yes, when an invalid configuration is provided, the operator will stop its synchronization job instead of using the default (empty) configuration. The operator will be marked as degraded so the admin is notified and able to fix the issue.

test/e2e/config_test.go
```diff
@@ -0,0 +1,120 @@
+// Copyright 2019 The Cluster Monitoring Operator Authors
```
lilic, line 1: 2020 sadly :)

test/e2e/config_test.go
```diff
@@ -0,0 +1,120 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
+)
+
+func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
+	// Enable user workload monitoring to assess that an invalid configuration
```
lilic, line 30: Do we need to enable this just to test configuration, we use same configmap to configure both workloads, or is it some specific edge case we want to cover?

simonpasquier: The goal here is to verify that a valid "custom" configuration isn't rolled back after we push an invalid config. I could have used another setting (like volume claim, resource requests, ...) but enabling user workload monitoring and checking that the `prometheus-user-workload` statefulset still exists was the "simplest" thing to do from my point of view.

test/e2e/config_test.go
simonpasquier: :eyeglasses:

test/e2e/config_test.go
lilic: Okay makes sense, thanks for explaining!

PR #734, opened 2020-04-01
pkg/manifests: remove Hostport from Thanos ruler config

simonpasquier:
This isn't used anywhere and most probably a copy/paste from the Prometheus/Alertmanager configuration.

cc @pgier @lilic who worked on this part before.


lilic:
That would be @pgier so leaving up to him to review, but lgtm.

simonpasquier:
I think we'll need this soon so holding for now :)

simonpasquier:
After more thinking, we shouldn't need this hence:

/unhold

PR #738, opened 2020-04-02
OCM: Add OpenShift Cluster Manager team

vkareh:
The OCM team relies on these queries for reporting metrics to customers
and would like to be notified when they change so that the cluster
service can adapt.

### DIFF
--- a/Documentation/data-collection.md
+++ b/Documentation/data-collection.md
@@ -32,20 +32,22 @@ data:
     # to identify when an update causes a service to begin to crash-loop or
     # flake.
     - '{__name__="count:up1"}'
-    # cluster_version reports what payload and version the cluster is being
-    # configured to and is used to identify what versions are on a cluster
-    # that is experiencing problems.
+    # (@openshift/openshift-team-cluster-manager) cluster_version reports what
+    # payload and version the cluster is being configured to and is used to
+    # identify what versions are on a cluster that is experiencing problems.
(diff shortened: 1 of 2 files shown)


vkareh:
@cben - please take a look. Does this look like a comprehensive list of the metrics we query in clusters service?

Documentation/data-collection.md
```diff
@@ -32,20 +32,20 @@ data:
     # to identify when an update causes a service to begin to crash-loop or
     # flake.
     - '{__name__="count:up1"}'
-    # cluster_version reports what payload and version the cluster is being
+    # (team-OCM) cluster_version reports what payload and version the cluster is being
```
lilic, line 35: Do you have a github openshift handle? so we just ping you on PRs via `@team-ocm `?

vkareh: I don't - how do I get one?

lilic: We are for example `@openshift/openshift-team-monitoring` I am guessing there is a mojo page for this? :D Would just make it easier to have folks in one github team to just ping you. Or if you put one person there who is responsible, its up to you.

vkareh: I've requested a team be created for OCM. Thanks

vkareh: We have a team now: @openshift/openshift-team-cluster-manager 

I've updated this patch with the new team handle. Thanks!

PR #741, opened 2020-04-06
Update kube-thanos

brancz:
This PR updates the kube-thanos dependency to latest `master` and consistently uses the Kubernetes recommended app labels for the Thanos query component.

@openshift/openshift-team-monitoring


lilic:
Your generate test is failing 😝 think you need run the make generate command.

brancz:
> Your generate test is failing stuck_out_tongue_closed_eyes think you need run the make generate command.

Also realized this, was already running :slightly_smiling_face: . Done.

assets/thanos-querier/deployment.yaml
```diff
@@ -37,7 +59,7 @@ spec:
             - if [ -x "$(command -v curl)" ]; then curl http://localhost:9090/-/healthy;
               elif [ -x "$(command -v wget)" ]; then wget --quiet --tries=1 --spider
               http://localhost:9090/-/healthy; else exit 1; fi
-        name: thanos-querier
+        name: thanos-query
```
lilic, line 62: Note this is used and should be changed here as well https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/manifests/manifests.go#L2605

brancz: Good catch! Updated.

lilic:
> 2020/04/06 11:22:05 wait for prometheus-k8s: error executing prometheus query: unexpected status code response, want 200, got 403: timed out waiting for the condition

seems like tests fail 🤔

lilic:
Note there is a flake around our e2e tests, which might be the failure is about.

simonpasquier:
These failures don't involve port-forwarding.

/retest

assets/thanos-querier/deployment.yaml
```diff
@@ -2,32 +2,54 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
-    app.kubernetes.io/name: thanos-querier
+    app.kubernetes.io/component: query-layer
+    app.kubernetes.io/instance: thanos-querier
+    app.kubernetes.io/name: thanos-query
```
lilic, line 7: deployment name is still thanos-querier, I expect the name to reflect that, wdyt?

pkg/manifests/manifests.go
```diff
@@ -2602,7 +2602,7 @@ func (f *Factory) ThanosQuerierDeployment(grpcTLS *v1.Secret, enableUserWorkload
 
 	if f.config.ThanosQuerierConfig.Resources != nil {
 		for i, c := range d.Spec.Template.Spec.Containers {
-			if c.Name == "thanos-querier" {
+			if c.Name == "thanos-query" {
```
lilic, line 2605: We have this in the unit tests as well. Can you also add we should error out if that container name is not there, thanks! Or we can do that in follow up as well. Lets just adjust that test, thanks!

https://github.com/openshift/cluster-monitoring-operator/blob/98bc99134498721360fdde87ca54f9233ae4bffa/pkg/manifests/manifests_test.go#L1393

assets/thanos-querier/deployment.yaml
brancz: name is the kubernetes app name, as kube-thanos is the defining entity here, I don't think we should change this, we should however eventually rename thanos-querier to thanos-query

pkg/manifests/manifests.go
brancz: good catch, fixed

assets/thanos-querier/deployment.yaml
lilic: > we should however eventually rename thanos-querier to thanos-query

Agreed, can you open jira for this? Thanks! :)

PR #742, opened 2020-04-06
Use DNS SRV record in Thanos Ruler configuration for querier endpoint

simonpasquier:


jsonnet/thanos-ruler.jsonnet
```diff
@@ -181,7 +181,7 @@ local thanosRulerRules =
                 server_name: 'thanos-querier.openshift-monitoring.svc',
               },
             },
-            static_configs: ['thanos-querier.openshift-monitoring.svc:9091'],
+            static_configs: ['dnssrv+_web._tcp.thanos-querier.openshift-monitoring.svc.cluster.local'],
```
lilic, line 184: hmm interesting, after discussing with @brancz and Bartek we said we didn't need this last time.

s-urbaniak: ping @bwplotka can you help in finding the correct setting?

simonpasquier:
After offline discussion, it's been agreed that it isn't worth the change. If we feel that using SRV records (`_svc._tcp.host`) is more readable than A records + port (`host:port`) then we should switch to SRV everywhere we can.

PR #751, opened 2020-04-10
test/e2e: forward port for every query

simonpasquier:
It should reduce the likelihood of failing the test when
port-forwarding encounters connection issues.

It also fixes a typo in the test which was running the same query
thrice.

### DIFF
--- a/test/e2e/user_workload_monitoring_test.go
+++ b/test/e2e/user_workload_monitoring_test.go
@@ -448,24 +448,25 @@ func assertTenancyForMetrics(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// The tenancy port (9092) is only exposed in-cluster so we need to use
-	// port forwarding to access kube-rbac-proxy.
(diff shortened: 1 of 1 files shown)


test/e2e/user_workload_monitoring_test.go
```diff
@@ -448,24 +448,25 @@ func assertTenancyForMetrics(t *testing.T) {
...
-		host,
-		token,
-		map[string][]string{"namespace": []string{userWorkloadTestNs}},
-	)
-
 	for _, q := range []string{"up", `up{namespace="should-be-overwritten"}`, fmt.Sprintf(`up{namespace="%s"}`, userWorkloadTestNs)} {
 		t.Logf("Running query %q", q)
+
 		err = framework.Poll(5*time.Second, time.Minute, func() error {
-			b, err := client.PrometheusQuery("up")
```
simonpasquier, line 468: :bug:

PR #754, opened 2020-04-14
Add monitoring-edit role

simonpasquier:


simonpasquier:

assets/cluster-monitoring-operator/monitoring-targets-edit-cluster-role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: monitoring-targets-edit
+rules:
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - servicemonitors
+  - podmonitors
```
brancz, line 10: maybe I'm missing context, but I would have expected a role for monitoring configurations, that is not just about targets but both scraping and alerting configuration

s-urbaniak: For configuration we specified a dedicated role, see https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/user-workload-monitoring.md#monitoring-targets-edit, and https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/user-workload-monitoring.md#monitoring-workload-config-edit:

`monitoring-targets-edit`: This is a cluster role which can be bound against a concrete namespace by the cluster admin. Embeds the get/create/edit/delete permissions for the following custom resources:

- ServiceMonitor
- PodMonitor
- PrometheusRule

It allows to create new scraping targets for services/pods and allows to create new recording or alerting rules.

`monitoring-workload-config-edit`:
This is a named role against the workload-monitoring-config configmap resource in the openshift-monitoring namespace embedding get/create/edit/delete permissions.

s-urbaniak: Or did we miss something else here? :thinking:

s-urbaniak, line 10: according to the OEP we need to have `PrometheusRule` here too.

s-urbaniak: on an after-thought: I believe we added PrometheusRule too much in the enhancement proposal, because as per https://issues.redhat.com/browse/MON-936?focusedCommentId=13977634&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13977634 this refers only to ServiceMonitor/PodMonitor custom resources. We should submit a PR against the enhancement.

brancz: This is what I meant. I think PrometheusRule _should_ be part of this, although then just "targets" doesn't make sense.

s-urbaniak: We can unify `monitoring-rules-edit` and `monitoring-targets-edit`, but that would imply that the user cannot be given distinct permissions to silence alerts.

brancz: People can still create those roles if they need but I don't see a standard use case where the distinction needs to be present. The basic use case is self service monitoring, and a user will always need both for that.

lilic: I agree, I don't see a use case for when often one would scrape their metrics but not allowed to add rules for those metrics. Admins can still create those if they need granularity. Or was this a specific requirement from some SRE folks?

s-urbaniak: i am fine with that, let's just make sure @sichvoge is on board too.

sichvoge: No strong feelings from my side. It's more a QoL feature anyways and I am ok to combine the two edit roles into one, and fall back to an admin to create more granularity.

simonpasquier: I've added a comment to the Jira ticket for Christian.

s-urbaniak: @simonpasquier we reached consensus that we need to add `prometheusrule` custom resources here. The OEP doesn't need an update as it specifies that already.

brancz: I think there is a small change that we should do though, which is change the naming, as this doesn't reflect just targets. How about just `monitoring-config-edit`?

simonpasquier: I agree about changing the name. The OEP already provisions a [monitoring-workload-config-edit](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/user-workload-monitoring.md#monitoring-workload-config-edit) role so having `monitoring-config-edit` might be confusing?
Maybe simply `monitoring-edit`?

brancz: I'm happy with `monitoring-edit` as well.

assets/thanos-ruler/thanos-ruler-prometheus-rule.yaml
```diff
@@ -37,7 +37,7 @@ spec:
         )
       for: 5m
       labels:
-        severity: critical
```
lilic, line 40: jsonnet file lock did not change? 🤔

lilic: Also no need to bump this, as I will do as part of the release checklist thing.

simonpasquier: this is what happens when you push code changes during a meeting :smile:

lilic: haha :D

PR #758, opened 2020-04-16
user-workload: configure prometheus replica count

s-urbaniak:

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -34,8 +34,8 @@ type Config struct {
 	PrometheusOperatorConfig             *PrometheusOperatorConfig `json:"prometheusOperator"`
 	PrometheusOperatorUserWorkloadConfig *PrometheusOperatorConfig `json:"prometheusOperatorUserWorkload"`
 
-	PrometheusK8sConfig          *PrometheusK8sConfig `json:"prometheusK8s"`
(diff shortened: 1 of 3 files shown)


pkg/manifests/config.go
```diff
@@ -89,6 +89,7 @@ type PrometheusK8sConfig struct {
 	Hostport            string                    `json:"hostport"`
 	RemoteWrite         []monv1.RemoteWriteSpec   `json:"remoteWrite"`
 	TelemetryMatches    []string                  `json:"-"`
+	Replicas            int32                     `json:"replicas"`
```
simonpasquier, line 92: Since it only applies to the user-workload Prometheus, shouldn't we maintain 2 config structs to avoid confusion? I see other fields (e.g. `TelemetryMatches` that only apply to the infra Prometheus.

s-urbaniak: sounds good :+1: I will split the struct.

s-urbaniak: fixed

s-urbaniak:
To persist the concern expressed by @lilic out of band: There are architectural concerns whether it makes sense to configure this setting at all for user-workload-prometheus which we should clarify before merging this

/cc @brancz @sichvoge

PR #761, opened 2020-04-22
Use Alertmanager API v2 for Thanos Ruler

simonpasquier:
I think it is the last bit that was using v1 of Alertmanager API.


lilic:
Lets aim this towards the master-4.6, as it's not a bug right?

PR #763, opened 2020-04-23
Use Alertmanager's API v2 for Thanos ruler

simonpasquier:


lilic:
@s-urbaniak can you put a valid-bug label on it? :D

PR #769, opened 2020-04-27
Bug 1813221: set remaining cpu requests

s-urbaniak:


simonpasquier:
/lgtm

you need to remove your local `go-bindata` and regenerate again :)

s-urbaniak:
@simonpasquier hmm i did `make generate-in-docker` :thinking:

lilic:
I think you are using the wrong version of the jsonnet-ci image, a while ago pinned to the specific release.

s-urbaniak:
deleted all local docker images too before `make generate-in-docker`'ing  :man_shrugging:

lilic:
> deleted all local docker images too before make generate-in-docker'ing 🤷‍♂️

Yes that is the problem, we are not using latest, but pinned release from what I last remember. cc @paulfantom we seem to not have updated the makefile? https://github.com/openshift/cluster-monitoring-operator/blob/master/Makefile#L30 

@s-urbaniak so I don't need a bugzilla do you mind setting this -> https://github.com/openshift/cluster-monitoring-operator/blob/master/Makefile#L30 to use `quay.io/coreos/jsonnet-ci:release-0.38` if that is what worked for you? Thanks!

s-urbaniak:
@lilic pinned, great catch :+1

s-urbaniak:
```
level=error msg="Cluster operator monitoring Degraded is True with UpdatingAlertmanagerFailed: Failed to rollout the stack. Error: running task Updating Alertmanager failed: waiting for Alertmanager object changes failed: waiting for Alertmanager: expected 3 replicas, updated 0 and available 0"
```

@lilic @pgier @simonpasquier  I believe we have to remove setting the cpu request of the alertmanager sidecar in 4.4, as this patch-merge in alertmanager custom resources was fixed in >=4.5. are you ok with that?

lilic:
> I believe we have to remove setting the cpu request of the alertmanager sidecar in 4.4, as this patch-merge in alertmanager custom resources was fixed in >=4.5. are you ok with that?

We can also bring in that patch that Paul did in the whatever the 4.4 release of prom-operator is?

s-urbaniak:
@lilic i believe this is fair, let me try to apply the cherry-pick :ok_hand:

s-urbaniak:
@lilic unfortunately this is not a clean cherry-pick, hence my judgement is this too risky unless @pgier wants to give it a try.

lilic:
Fair enough, is this for the config-reloader only?

s-urbaniak:
@lilic yes this is just for config-reloader which would have cpu requests simply unset.

s-urbaniak:

PR #771, opened 2020-04-27
[DNM] force fail CI

simonpasquier:


simonpasquier:
Closing since we've got a clear idea of why the CI wasn't failing while config-reloader can't reload rules in Thanos Ruler...

PR #773, opened 2020-04-28
Bug 1828702: wait for trusted CA bundle to be created

simonpasquier:
This change ensures that the operator waits for the trusted CA bundle
ConfigMap to be populated before proceeding with the deployment of the
monitoring components.

Note that it should fail the `e2e-aws-operator` until https://bugzilla.redhat.com/show_bug.cgi?id=1827530 is fixed.

Update May 6: https://bugzilla.redhat.com/show_bug.cgi?id=1827530 is now fixed and the CI passes.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -223,8 +223,7 @@ var (
 	AuthProxyCookieDomainFlag = "-cookie-domain="
 	AuthProxyRedirectURLFlag  = "-redirect-url="
 
-	PrometheusTrustedCABundleDir  = "/etc/pki/prometheus-ca-bundle/"
-	PrometheusTrustedCABundlePath = PrometheusTrustedCABundleDir + "ca-bundle.crt"
+	TrustedCABundleKey = "ca-bundle.crt"
 )
 
 const (
@@ -413,7 +412,7 @@ func (f *Factory) AlertmanagerMain(host string, trustedCABundleCM *v1.ConfigMap)
 				a.Spec.VolumeMounts = append(a.Spec.VolumeMounts, trustedCABundleVolumeMount(volumeName))
 				volume := trustedCABundleVolume(trustedCABundleCM.Name, volumeName)
 				volume.VolumeSource.ConfigMap.Items = append(volume.VolumeSource.ConfigMap.Items, v1.KeyToPath{
-					Key:  "ca-bundle.crt",
+					Key:  TrustedCABundleKey,
 					Path: "tls-ca-bundle.pem",
 				})
 				a.Spec.Volumes = append(a.Spec.Volumes, volume)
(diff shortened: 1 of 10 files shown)


simonpasquier:
Fails as expected:

```
    --- FAIL: TestUserWorkloadMonitoring/assert_user_workload_metrics (350.58s)
        user_workload_monitoring_test.go:410: timed out waiting for the condition: expected 1 fired VersionAlert, got 0 
```

lilic:
Nice!

simonpasquier:

pkg/tasks/trusted_ca.go
```diff
@@ -0,0 +1,86 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+type caBundleSyncer struct {
+	prefix  string
+	client  *client.Client
+	factory *manifests.Factory
+}
+
+func (cbs *caBundleSyncer) syncTrustedCABundle(trustedCA *v1.ConfigMap) (*v1.ConfigMap, error) {
```
lilic, line 33: To me in this package we only have very clear separated tasks per file, this seems like a helper function shared by the tasks wdyt if that should go in the client or manifests package instead? Or a new package like we have with strings package?

simonpasquier: It can't really go in `pkg/client` or `pkg/manifests` because it depends on both. I can create a new package but I felt that since it's only used in `pkg/tasks`, it could live there. Maybe it's easier to differentiate from the other files if I rename it `helpers.go`?

lilic: > Maybe it's easier to differentiate from the other files if I rename it helpers.go

That works for me as well, either helpers.go or utils.go.

simonpasquier: I've picked `helpers.go`.

pkg/manifests/manifests.go
```diff
@@ -3258,16 +3257,15 @@ func NewSecurityContextConstraints(manifest io.Reader) (*securityv1.SecurityCont
 // It adds "monitoring.openshift.io/name" and "monitoring.openshift.io/hash" labels.
 // Any other labels from the given configmap are discarded.
 //
-// It returns nil, if the given configmap does not contain the "ca-bundle.crt" the data key
+// It returns an error if the given configmap does not contain the "ca-bundle.crt" data key
 // or data is empty string.
-func (f *Factory) HashTrustedCA(caBundleCM *v1.ConfigMap, prefix string) *v1.ConfigMap {
-	caBundle, ok := caBundleCM.Data["ca-bundle.crt"]
-	if !ok || caBundle == "" {
-		// We return here instead of erroring out as we need
```
lilic, line 3266: hmm it seems we wrote this at some point, no clue who it was, will this not the be the case anymore?

simonpasquier: With this change, the operator calls `*caBundleSyncer.syncTrustedCABundle()` which ensures that the resulting config map is populated with the expected keys.

PR #774, opened 2020-04-29
[WIP] Add an API performance grafana dashboard

tkashem:


brancz:
Can you run `make pkg/manifests/bindata.go`?

brancz:
Can you share what about this is specific to openshift and maybe a screenshot of the dashboard?

tkashem:
The dashboard is still work in progress.

/hold

tkashem:
> openshift and maybe a screenshot of the dashbo

image

We have a metric relabeling here - https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml#L91

I am using it to filter by `kube-apiserver` or `openshift-apiserver`

tkashem:
Suggestions from @brancz 
- from experience with our customers you're going to want to write out everything, and not use abbreviations
- recommend using a more descriptive label for to select the apiserver .. instance being IPs confuses people

tkashem:
image

rps by instance is broken. but if i install the dashboard independently  this works fine. I wonder what the root cause is, any idea?

metalmatze:
I don't really see how this improves the user experiences for a _users_ of the APIServer. Sure, this looks like a helpful dashboard for low-level debugging of the APIServer. I don't think that this dashboard should be part of the dashboard shipped to customers by default though.
What specifically are you missing in the current APIServer dashboard, that we maybe could improve on that one?!

tkashem:
> What specifically are you missing in the current APIServer dashboard, that we maybe could improve on that one?!

@metalmatze can you point me to the "current APIServer dashboard" that ships with OpenShift? I am afraid I did not find any. I also looked at the dashboards in "coreos/kube-prometheus" that are disabled and they are not comprehensive.



> I don't think that this dashboard should be part of the dashboard shipped to customers by default though

@metalmatze can you please explain a bit more on why it should not be part of the dashboard shipped? We think it will be useful in troubleshooting customers cases.

paulfantom:
@tkashem there is an apiserver dashboard in kubernetes-mixins project (jsonnet code available [here](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/apiserver.libsonnet), generated code available [here](https://github.com/cloudalchemy/mixins/blob/master/assets/kubernetes/dashboards/apiserver.json)). We would prefer to not lock-down knowledge in OpenShift repository but to propagate it to a broader community. That's why it may be more beneficial to improve the upstream dashboard than develop a new one internally.

paulfantom:
I am closing this as it seems to be abandoned. Please reopen if needed, but preferably open issue in kubernetes-mixins project as per https://github.com/openshift/cluster-monitoring-operator/pull/774#issuecomment-635901352

/close

PR #776, opened 2020-04-29
Remove thanos ruler CRD

simonpasquier:
Left over from #707.

/hold until we figure out the workflow to manage 4.5 and 4.6 releases simultaneously.


s-urbaniak:
good catch :+1:

PR #784, opened 2020-05-12
Remove thanos ruler CRD

simonpasquier:
Left over from #707


lilic:
> Pending — Not mergeable. Needs bugzilla/valid-bug label.

But it's 4.6 branch 🤔 did we bring this up with testplatform team?

PR #793, opened 2020-05-22
Bug 1800489: Patch kube pv storage alert release 4.2

pgier:
This is kind of a hack to the jsonnet so that we can backport a fix to a prometheus rule without bringing in a full update of kube-prometheus.


pgier:
@lilic I wasn't sure how we could update this in kube-prometheus without updating the import of kubernetes-mixin.  Maybe we should just update the imported kubernetes-mixin?  I was worried that might bring in some breaking changes.

Also this would need to be backported to 3.11 and I don't think we have a kube-prometheus branch for that version yet.

pgier:
Closing this one since we will just fix this in 3.11 and not in 4.1 or 4.2

lilic:
Yeah 4.2 support cycle is over I think? I could be wrong but if thats the case sgtm

PR #804, opened 2020-06-09
Allow user workload monitoring configuration ConfigMap to be created in UWM ns

lilic:

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -155,7 +155,8 @@ func Main() int {
 		return 1
 	}
 
-	o, err := cmo.New(config, *releaseVersion, *namespace, *namespaceUserWorkload, *namespaceSelector, *configMapName, *remoteWrite, images.asMap(), telemetryConfig.Matches)
+	userWorkloadConfigMapName := "user-workload-monitoring-config"
+	o, err := cmo.New(config, *releaseVersion, *namespace, *namespaceUserWorkload, *namespaceSelector, *configMapName, userWorkloadConfigMapName, *remoteWrite, images.asMap(), telemetryConfig.Matches)
 	if err != nil {
 		fmt.Fprint(os.Stderr, err)
 		return 1
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -31,23 +31,28 @@ type Config struct {
 	RemoteWrite bool                  `json:"-"`
 	Platform    configv1.PlatformType `json:"-"`
 
-	PrometheusOperatorConfig             *PrometheusOperatorConfig `json:"prometheusOperator"`
+	ClusterMonitoringConfiguration *ClusterMonitoringConfiguration `json:"-"`
+	UserWorkloadConfiguration      *UserWorkloadConfiguration      `json:"-"`
+}
+
+type ClusterMonitoringConfiguration struct {
+	PrometheusOperatorConfig *PrometheusOperatorConfig    `json:"prometheusOperator"`
+	PrometheusK8sConfig      *PrometheusK8sConfig         `json:"prometheusK8s"`
+	AlertmanagerMainConfig   *AlertmanagerMainConfig      `json:"alertmanagerMain"`
+	KubeStateMetricsConfig   *KubeStateMetricsConfig      `json:"kubeStateMetrics"`
+	OpenShiftMetricsConfig   *OpenShiftStateMetricsConfig `json:"openshiftStateMetrics"`
+	GrafanaConfig            *GrafanaConfig               `json:"grafana"`
+	EtcdConfig               *EtcdConfig                  `json:"-"`
+	HTTPConfig               *HTTPConfig                  `json:"http"`
+	TelemeterClientConfig    *TelemeterClientConfig       `json:"telemeterClient"`
+	K8sPrometheusAdapter     *K8sPrometheusAdapter        `json:"k8sPrometheusAdapter"`
+	ThanosQuerierConfig      *ThanosQuerierConfig         `json:"thanosQuerier"`
+	UserWorkloadEnabled      *bool                        `json:"enableUserWorkload"`
+	// TODO: Remove in 4.7 release.
+	PrometheusUserWorkloadConfig         *PrometheusK8sConfig      `json:"prometheusUserWorkload"`
 	PrometheusOperatorUserWorkloadConfig *PrometheusOperatorConfig `json:"prometheusOperatorUserWorkload"`
-
-	PrometheusK8sConfig          *PrometheusK8sConfig `json:"prometheusK8s"`
-	PrometheusUserWorkloadConfig *PrometheusK8sConfig `json:"prometheusUserWorkload"`
-
-	AlertmanagerMainConfig *AlertmanagerMainConfig      `json:"alertmanagerMain"`
-	KubeStateMetricsConfig *KubeStateMetricsConfig      `json:"kubeStateMetrics"`
-	OpenShiftMetricsConfig *OpenShiftStateMetricsConfig `json:"openshiftStateMetrics"`
-	GrafanaConfig          *GrafanaConfig               `json:"grafana"`
-	EtcdConfig             *EtcdConfig                  `json:"-"`
-	HTTPConfig             *HTTPConfig                  `json:"http"`
-	TelemeterClientConfig  *TelemeterClientConfig       `json:"telemeterClient"`
-	K8sPrometheusAdapter   *K8sPrometheusAdapter        `json:"k8sPrometheusAdapter"`
-	UserWorkloadConfig     *UserWorkloadConfig          `json:"techPreviewUserWorkload"`
-	ThanosRulerConfig      *ThanosRulerConfig           `json:"thanosRuler"`
-	ThanosQuerierConfig    *ThanosQuerierConfig         `json:"thanosQuerier"`
+	ThanosRulerConfig                    *ThanosRulerConfig        `json:"thanosRuler"`
+	UserWorkloadConfig                   *UserWorkloadConfig       `json:"techPreviewUserWorkload"`
 }
 
 type Images struct {
@@ -75,11 +80,13 @@ type HTTPConfig struct {
 }
 
 type PrometheusOperatorConfig struct {
+	LogLevel     string            `json:"logLevel"`
 	NodeSelector map[string]string `json:"nodeSelector"`
 	Tolerations  []v1.Toleration   `json:"tolerations"`
 }
 
 type PrometheusK8sConfig struct {
+	LogLevel            string                    `json:"logLevel"`
 	Retention           string                    `json:"retention"`
 	NodeSelector        map[string]string         `json:"nodeSelector"`
 	Tolerations         []v1.Toleration           `json:"tolerations"`
@@ -100,6 +107,7 @@ type AlertmanagerMainConfig struct {
 }
 
 type ThanosRulerConfig struct {
+	LogLevel            string                    `json:"logLevel"`
 	NodeSelector        map[string]string         `json:"nodeSelector"`
 	Tolerations         []v1.Toleration           `json:"tolerations"`
 	Resources           *v1.ResourceRequirements  `json:"resources"`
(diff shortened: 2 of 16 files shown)


cmd/operator/main.go
```diff
@@ -89,6 +89,7 @@ func Main() int {
 	namespaceUserWorkload := flagset.String("namespace-user-workload", "openshift-user-workload-monitoring", "Namespace to deploy and manage user workload monitoring stack in.")
 	namespaceSelector := flagset.String("namespace-selector", "openshift.io/cluster-monitoring=true", "Selector for namespaces to monitor.")
 	configMapName := flagset.String("configmap", "cluster-monitoring-config", "ConfigMap name to configure the cluster monitoring stack.")
+	userWorkloadConfigMapName := flagset.String("userWorkloadConfigmap", "user-workload-monitoring-config", "ConfigMap name to configure the user workload monitoring stack.")
```
s-urbaniak, line 92: why does this need to be configurable?

pkg/manifests/config.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 The Cluster Monitoring Operator Authors
+/// Copyright 2018 The Cluster Monitoring Operator Authors
```
s-urbaniak, line 1: The triple slash is unneeded i believe.

lilic: I have no idea what happened here :D lets see if its my weird vim configuration, nice catch

cmd/operator/main.go
lilic: Just followed pattern we had, see line above here, honestly happy to leave it out as well!

pkg/operator/operator.go
```diff
@@ -355,6 +360,32 @@ func (o *Operator) sync(key string) error {
...
 
+func (o *Operator) loadUserWorkloadConfig() (*manifests.UserWorkloadConfiguration, error) {
+	userCM, err := o.client.GetConfigmap(o.namespaceUserWorkload, o.userWorkloadConfigMapName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Warning("No User Workload Monitoring ConfigMap was found. Using defaults.")
+			return manifests.NewDefaultUserWorkloadMonitoringConfig(), nil
+		} else {
+			klog.Warningf("Error loading User Workload Monitoring ConfigMap. Error: %v", err)
+			return nil, err
```
s-urbaniak, line 371: let's wrap the returning error to have context.

pkg/operator/operator.go
```diff
@@ -355,6 +360,32 @@ func (o *Operator) sync(key string) error {
...
+		}
+	}
+
+	configContent, found := userCM.Data["config.yaml"]
+	if found {
+		uwc, err := manifests.NewUserConfigFromString(configContent)
+		if err != nil {
+			klog.Warningf("Error creating User Workload Configuration from ConfigMap. Error: %v", err)
+			return nil, err
```
s-urbaniak, line 380: dito

pkg/operator/operator.go
```diff
@@ -387,8 +419,19 @@ func (o *Operator) Config(key string) (*manifests.Config, error) {
...
 	}
 
+	// Only use User Workload Monitoring ConfigMap from user ns and populate if
+	// its enabled by admin via Cluster Monitoring ConfigMap.  The above
+	// loadConfig() already initializes the structs with nil values for
+	// UserWorkloadConfiguration struct.
+	if *c.ClusterMonitoringConfiguration.UserWorkloadEnabled {
+		c.UserWorkloadConfiguration, err = o.loadUserWorkloadConfig()
+		if err != nil {
+
```
s-urbaniak, line 429: i believe this is WIP

s-urbaniak: or isn't it? :-) do we really want to swallow the error?

s-urbaniak:
looking great so far! :tada:

pkg/operator/operator.go
lilic: nope, just me forgetting to return error :( fixed

cmd/operator/main.go
s-urbaniak: my intuition is we should limit exposure here as we want to move to a CRD based setup :thinking: 

This cli flag increases the public API surfaces.

lilic: good point, removed it!

lilic:
@s-urbaniak PTAL changes since last time:
- I removed the flag, good call!
- added example configmap and used it in one of the config unit tests
- added CHANGELOG entry.

Thanks!

examples/user-workload/1q
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
```
simonpasquier, line 1: I'm confused about this file's name. Is it needed?

lilic: I believe that was committed wrongly, possibly outcome of some generation or just my mistake, will remove.

pkg/manifests/config.go
```diff
@@ -31,23 +31,28 @@ type Config struct {
 	RemoteWrite bool                  `json:"-"`
 	Platform    configv1.PlatformType `json:"-"`
 
-	PrometheusOperatorConfig             *PrometheusOperatorConfig `json:"prometheusOperator"`
-	PrometheusOperatorUserWorkloadConfig *PrometheusOperatorConfig `json:"prometheusOperatorUserWorkload"`
-
-	PrometheusK8sConfig          *PrometheusK8sConfig `json:"prometheusK8s"`
-	PrometheusUserWorkloadConfig *PrometheusK8sConfig `json:"prometheusUserWorkload"`
+	ClusterMonitoringConfiguration *ClusterMonitoringConfiguration `json:"-"`
+	UserWorkloadConfiguration      *UserWorkloadConfiguration      `json:"-"`
```
simonpasquier, line 35: Having both `UserWorkloadConfiguration` and `UserWorkloadConfig` is confusing.

pkg/manifests/config.go
```diff
@@ -31,23 +31,28 @@ type Config struct {
...
+	OpenShiftMetricsConfig   *OpenShiftStateMetricsConfig `json:"openshiftStateMetrics"`
+	GrafanaConfig            *GrafanaConfig               `json:"grafana"`
+	EtcdConfig               *EtcdConfig                  `json:"-"`
+	HTTPConfig               *HTTPConfig                  `json:"http"`
+	TelemeterClientConfig    *TelemeterClientConfig       `json:"telemeterClient"`
+	K8sPrometheusAdapter     *K8sPrometheusAdapter        `json:"k8sPrometheusAdapter"`
+	ThanosQuerierConfig      *ThanosQuerierConfig         `json:"thanosQuerier"`
+	UserWorkloadEnabled      *bool                        `json:"enableUserWorkload"`
+	// TODO: Remove in 4.7 release.
+	PrometheusUserWorkloadConfig         *PrometheusK8sConfig      `json:"prometheusUserWorkload"`
```
simonpasquier, line 52: Instead of duplicating `PrometheusUserWorkloadConfig`, `PrometheusOperatorUserWorkloadConfig` and `ThanosRulerConfig` here and in `UserWorkloadConfiguration` maybe it would be simpler to use only `UserWorkloadConfiguration`? IOW handle the "legacy" fields only when loading the config map but don't expose them here.

pkg/manifests/manifests.go
```diff
@@ -1404,39 +1408,77 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret) (*monv1.Prometheus,
...
 
-	if len(f.config.PrometheusUserWorkloadConfig.RemoteWrite) > 0 {
-		p.Spec.RemoteWrite = f.config.PrometheusUserWorkloadConfig.RemoteWrite
+	if len(f.config.UserWorkloadConfiguration.Prometheus.RemoteWrite) > 0 {
+		p.Spec.RemoteWrite = f.config.UserWorkloadConfiguration.Prometheus.RemoteWrite
+	}
+
+	// TODO: remove after 4.7
+
+	if f.config.ClusterMonitoringConfiguration.PrometheusUserWorkloadConfig.LogLevel != "" {
```
simonpasquier, line 1449: When you have config fields defined both in `f.config.ClusterMonitoringConfiguration`  and `f.config.UserWorkloadConfiguration` then the former will take precedence. Couldn't this cause confusion?

lilic: techPreview is what we have to support right now primarily hence why it takes precedence here, but that is a good point to document.

pkg/manifests/config.go
lilic: The point was to be able to easily remove the current logic, hence the TODO comments, as we will stop supporting this in 4.7. And to separate out the two clear tenants we have right now cluster and user workload.

pkg/manifests/config.go
lilic: UserWorkloadConfig?

lilic:
/retest 
🤔

pkg/manifests/config.go
simonpasquier: We have a `UserWorkloadConfig` struct:
https://github.com/openshift/cluster-monitoring-operator/blob/882dc1b7310edd7981fab2aff76fdc30672a470c/pkg/manifests/config.go#L157-L159

And another `UserWorkloadConfiguration` struct:
https://github.com/openshift/cluster-monitoring-operator/blob/882dc1b7310edd7981fab2aff76fdc30672a470c/pkg/manifests/config.go#L381-L385

The names are close which makes it confusing for me when reviewing the code.

lilic: I see, makes sense! Any suggestions for new or old name?

pkg/manifests/config.go
simonpasquier: An alternative approach would be to remove the user workload monitoring fields from the `ClusterMonitoringConfiguration` struct and only use `Config.UserWorkloadConfiguration`.  `NewConfig()` would try to load the uwm fields until we decide that it shouldn't.

Something like this: https://github.com/simonpasquier/cluster-monitoring-operator/commit/44bd3ddbed3f2ee07ef99d239e15f1b599ccb581

The side effect is that uwm configuration will be taken either from the "legacy" fields or from the new configmap in `openshift-user-workload-monitoring` but it won't be a merge of both. In the current state, `PrometheusOperatorConfig` can be defined in the `openshift-montoring` configmap and `PrometheusConfig` in the `openshift-user-workload-monitoring` configmap, it can be confusing or desired, depending on how you see things :)

(IOW: in other words)

brancz: hmm .. especially with configuration I much prefer duplication structs. This is a lesson learned the hard way in Kubernetes. Only because things happen to be identical at this point in time, doesn't mean they may not very well develop in different direction in the future, and they almost always do unless the concept is literally the same (eg. TLS configuration in Prometheus). In this case we even know it will be deleted in the future, so I actually prefer @lilic 's currently proposed way.

pkg/operator/operator.go
```diff
@@ -355,6 +360,32 @@ func (o *Operator) sync(key string) error {
 	return nil
 }
 
+func (o *Operator) loadUserWorkloadConfig() (*manifests.UserWorkloadConfiguration, error) {
+	userCM, err := o.client.GetConfigmap(o.namespaceUserWorkload, o.userWorkloadConfigMapName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Warning("No User Workload Monitoring ConfigMap was found. Using defaults.")
```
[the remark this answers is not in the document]
lilic: We use warning level for the cluster monitoring configmap, I just followed that pattern. It makes sense to me in a way, as you are warning user they did not configure their stack at all. But don't have too strong opinion so can change, but then we should do both?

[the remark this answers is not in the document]
simonpasquier: I'd rather have the message at the info level because for me a warning is something I need at least to investigate and not customizing the monitoring isn't something suspicious. At info, the message would still logged so it would be available for support requests.
And +1000 that if we change it here, it should be consistent across the board.

PR #808, opened 2020-06-10
Bug 1846001: Add rhmi_status metric to telemeter - backport 4.5

david-martin:
backport of #795 


david-martin:
@s-urbaniak Can you help with the necessary approval of this backport?

lilic:
/hold

We are still resolving the discussion around RHMI and the namespace where these metrics come from, so would prefer if that conversation is resolved before we backport this. In case we need to remove it in master.

david-martin:
> the namespace where these metrics come from

@lilic  I don't see how this has any impact on the inclusion of the metric in the allow list.
The metric name will be the same no matter which namespace the pod that exposes it runs in.

cc @alanmoran for awareness of this as it affects ocm integration.

lilic:
@david-martin 
The point what it effects is that it is not tested ever, I understood when doing lgtm on the initial PR is that you are using OLM to install this or integrating with openshift directly, hence my questions about if its part of cluster monitoring openshift namespaces. From my perspective it would be better to have this tested against existing telemetry tests (e.g. number of series sent, etc.) same as its done for OLM cc @brancz as telemetry team lead and architect to make final decision for this.

david-martin:
Thanks for the explanation @lilic.
I understand where you are coming from now.

Can you & @brancz help me figure out what's needed to get this change landed?
I think it would be best to make an incremental change that unblocks downstream, while lining up work to ensure the right gates, testing & processes are put in place for future metrics from *any* source (not just rhmi)

brancz:
OLM already sets precedence here, we also didn't allow any OLM metrics that are not part of that test run. I think the same should apply to RHMI.

david-martin:
> OLM already sets precedence here

@brancz Could you share what has been done here in terms of testing, or any other material for what is being asked of layered products?

lilic:
@david-martin they have tests in origin that are periodic jobs that install all the red hat operators and run a set of our origin monitoring tests against that cluster, notifications for failures of those go to slack so we they are informed of failures. These ensure the limit of series sent is under 500 series per cluster for example. https://github.com/openshift/release/blob/master/ci-operator/jobs/operator-framework/operator-lifecycle-manager/operator-framework-operator-lifecycle-manager-master-periodics.yaml

sdodson:
@lilic Should this still be held, if not it needs a rebase.

eparis:
please feel free to reopen if this is still relevant (and rebased)

PR #810, opened 2020-06-11
Bug 1845443: KubeApiLatency alert firing even though not all conditions are matched

RiRa12621:


lilic:
/assign @squat 
:p

RiRa12621:
OCP 4.3 is leaving the full support and approaching EOL, therefore backporting to it does not seem like a viable option

PR #820, opened 2020-06-24
Bug 1846397: Fix AlertmanagerConfigInconsistent alert

simonpasquier:


assets/node-exporter/daemonset.yaml
```diff
@@ -10,7 +10,6 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/name: node-exporter
-      app.kubernetes.io/version: v0.18.1
```
lilic, line 13: Lets see if upgrade tests complain about this, as we are changing labels.

PR #825, opened 2020-06-25
Bug 1850717: Add namespace selector to CPUThrottlingHigh

lilic:
@openshift/openshift-team-monitoring PTAL

I was not sure if we want these changes to be in the CHANGELOG? Thoughts?


simonpasquier:
> I was not sure if we want these changes to be in the CHANGELOG? Thoughts?

I consider they are already covered by the fact that there's a BZ attached to the PR.

/lgtm

PR #838, opened 2020-07-02
Add scheduler feature usage metrics

damemi:
Adds `cluster_legacy_scheduler_policy` (https://github.com/openshift/cluster-kube-scheduler-operator/pull/250) and `cluster_master_schedulable` (https://github.com/openshift/cluster-kube-scheduler-operator/pull/168) metrics

This is important for us because the upstream scheduler is migrating from Policy configs (setting predicates/priorities) to a new Profile config, and is planning to remove Policy support entirely in the coming releases.

This requires us to react (see https://github.com/openshift/origin/pull/25203, https://github.com/openshift/cluster-kube-scheduler-operator/pull/255, https://github.com/openshift/api/pull/677) if we wish to continue supporting the ability to configure the scheduler algorithm through KSO. 

The current usage % of Policy configs from `cluster_legacy_scheduler_policy` will help us determine whether it is worth continuing to support. `cluster_master_schedulable`, assumed being a reasonably used setting, will help set the benchmark threshold against which Policy config is compared (as well as provide usage information for that feature).


Documentation/data-collection.md
```diff
@@ -209,6 +209,12 @@ data:
     # Possible values are bootstrap|cloud-resources|monitoring|authentication|products|solution-explorer|deletion|complete.
     # This metric is used by OCM to detect when an RHMI installation is complete & ready to use i.e. rhmi_status{stage='complete'}
     - '{__name__="rhmi_status"}'
+    # (workloads-team, @openshift/openshift-group-b) cluster_legacy_scheduler_policy reports whether the scheduler operator is
```
smarterclayton, line 212: Does the metric itself have a description that explains the labels and values?  If so, copy it here (along with why).  Not knowing possible values (and really should just be a copy of the core metric) might cause a customer / user to ask "but what is it reporting" and we want that answer to be easy.

smarterclayton:
In description add how many series you expect.

Also, in general, your commit message and title should roughly contain some of this info (so someone doesn't have to look at the PR later)

Documentation/data-collection.md
damemi: Both are "boolean" gauges that just report 0 or 1 (https://github.com/openshift/cluster-kube-scheduler-operator/blob/master/pkg/operator/configmetrics/configmetrics.go#L15-L21). Updated the description with that

damemi:
> In description add how many series you expect.
@smarterclayton both of these metrics just report 0 or 1 for the cluster. Is that 1 or 2 series?

metalmatze:
That would be one series with a value of `0` or `1`, I suppose. :blush:

metalmatze:
Just out of curiosity: Do you want to later on count the number of cluster that have this enabled vs the ones that have it disabled? Trying to understand the use case :wink:

damemi:
> Just out of curiosity: Do you want to later on count the number of cluster that have this enabled vs the ones that have it disabled? Trying to understand the use case 😉

@metalmatze that's exactly what we want to gather, a percent of clusters that have these features enabled to gauge against our effort required to continue supporting them

s-urbaniak:
- [x] approval from @smarterclayton given in https://coreos.slack.com/archives/C0VMT03S5/p1593703067141800?thread_ts=1593701332.131000&cid=C0VMT03S5
- [x] the metrics here are shipped as core OpenShift and are already part of the product
- [x] Total number of series: 2 (`cluster_legacy_scheduler_policy`: 1, `cluster_master_schedulable`: 1)
- [ ] Follows naming best practices: `cluster_master_schedulable` implies a bool value, `cluster_legacy_scheduler_policy` seems a bit off for a bool value

s-urbaniak:
@openshift/openshift-team-monitoring (cc @brancz ) from my perspective lgtm modulo the naming, asking for another set of eyes.

brancz:
As these are gauges, I'm fine with the naming. Leaving final lgtm to @smarterclayton though to give a chance to verify if his comment was addressed in a way he expected.

damemi:
bump @smarterclayton does this look good to you?

smarterclayton:
This LGTM, feel free to tag.

PR #845, opened 2020-07-08
Bug 1852767: set user workload prometheus retention

pgier:
Allows the user workload prometheus retention setting to override
the cluster monitoring settings only if the cluster monitoring
setting is not set to the default.  Before this change the user workload
setting is never used because the cluster monitoring setting is
compared to an empty string instead of the default value.

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -196,6 +196,10 @@ func NewConfig(content io.Reader) (*Config, error) {
 	return res, nil
 }
 
+const (
+	defaultPrometheusRetention = "15d"
+)
+
 func (c *Config) applyDefaults() {
 	if c.Images == nil {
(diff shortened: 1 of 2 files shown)


pgier:
cc @lilic 
I also noticed that the user workload setting will only take effect when `enableUserWorkload == true` and not when `techPreviewUserWorkload.enabled == true`, but looks like this is intentional based on the comments in the source https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/operator/operator.go#L421

lilic:
Hey, I have a PR for this already :) https://github.com/openshift/cluster-monitoring-operator/pull/839

pgier:
Ah sorry, I didn't see any linked PR in the bug, I'll close this one.

PR #853, opened 2020-07-15
*: Bump mixins for: Tweak etcdMembersDown to reduce false negatives

ironcladlou:
This is to pull in https://github.com/etcd-io/etcd/pull/12177.


ironcladlou:
cc @hexfusion @retroflexer 

I'm following @wking's example with https://github.com/openshift/cluster-monitoring-operator/pull/844 — this is my first time trying to do the bump.

ironcladlou:
If there's some way to restrict the scope of the bump to just https://github.com/etcd-io/etcd/pull/12121 please let me know

jsonnet/jsonnetfile.lock.json
```diff
@@ -4,7 +4,7 @@
     {
       "source": {
         "git": {
-          "remote": "https://github.com/brancz/kubernetes-grafana",
+          "remote": "https://github.com/brancz/kubernetes-grafana.git",
           "subdir": "grafana"
```
hexfusion, line 8: is the git extension required? I am surprised to see it diff now or is this the result of some automation?

ironcladlou: Yeah no idea, I just ran the steps from https://github.com/openshift/cluster-monitoring-operator/pull/844#issue-446390661

jsonnet/jsonnetfile.lock.json
```diff
@@ -4,7 +4,7 @@
     {
       "source": {
         "git": {
-          "remote": "https://github.com/brancz/kubernetes-grafana",
```
lilic, line 7: Can you not commit the entire jsonnetfile.lock.json, just the ones related to etcd. We will bump everything in the next few weeks. 

Can you just commit the lines around etcd and then regenerate again, thanks! :)

ironcladlou: If I remove everything but the etcd changes in the lock file and then re-run `make generate-in-docker` I get no updates to the rules, and if I run `jb update` I get all the changes you see in this PR — what is the procedure for updating just etcd?

assets/prometheus-k8s/rules.yaml
```diff
@@ -1873,7 +1889,7 @@ spec:
           sum by (job) (up{job=~".*etcd.*"} == bool 0)
         or
           count by (job,endpoint) (
-            sum by (job,endpoint,To) (rate(etcd_network_peer_sent_failures_total{job=~".*etcd.*"}[3m])) > 0.01
+            sum by (job,endpoint,To) (rate(etcd_network_peer_sent_failures_total{job=~".*etcd.*"}[1m])) > 0.01
```
paulfantom, line 1892: This is doing rate over 1m, but etcd scrape interval is set to 30s. Such setup may result in incorrect data as you need at least 3 data points to do `rate()`. Either etcd scrape interval needs to be fixed to 15s or rate needs to be done over at least 2m.

ironcladlou: Good catch, thanks. My gut reaction is to shorten the scrape interval in this case, but I'd like some feedback from others. cc @hexfusion @wking @smarterclayton 

Makes me wonder if my testing was somehow flawed...

ironcladlou: Are there any benefits to a more granular scrape interval for etcd specifically? Having a higher resolution seems to be useful in this specific case (and I'm going to do more testing to see if the same results can be achieved with 30s+2m window), but I'm curious what other reasons there might be to change the interval.

ironcladlou: Here's what a reboot looks like with 30s and 15s etcd scrape intervals

Screen Shot 2020-07-16 at 10 32 55 AM

paulfantom: I don't see any immediate benefit of having higher resolution metrics for etcd as all alerts (apart from this one) use longer range selector and additionally have `for` statements.

As for this alert, what about shortening `for` statement and increasing range selector to `2m`?

ironcladlou: Revisiting this, @wking made the point that when this alert accurately fires and someone gets paged, an extra 5 or 10 minutes before being notified is unlikely to be a significant factor. And the alert today isn't actually considered actionable. Further, I'm also observing a ton of variance during reboots with regards to `up` and `etcd_network_peer_sent_failures_total` leveling off — anywhere from 1 to 6 minutes.

With those considerations, what if we used a 2m window and increased the `for` clause to something far more conservative like 10m? It seems like that could yield the accuracy improvements at the expense of a (hopefully insignificant) delay in alerting.

ironcladlou: https://github.com/etcd-io/etcd/pull/12177

ironcladlou:
Well, https://github.com/etcd-io/etcd/pull/12177 merged but the `make generate` tooling on the master branch is broken for me on both OSX and linux. After incorporating [this fix|https://github.com/openshift/cluster-monitoring-operator/pull/887] I was able to at least get the etcd change incorporated in, but ultimately bindata couldn't be regenerated because the tooling trips over this:

```
[ -f "assets/telemeter-client/serving-certs-c-a-bundle.yaml" ] || exit 1
make: *** [Makefile:85: assets/telemeter-client/serving-certs-c-a-bundle.yaml] Error 1
```

The containerized generation tooling I used a week or so ago when I first opened this PR has been removed and replaced with native tooling, so I can't go back and use the thing I know worked before. I can keep digging in to try and understand this toolchain's internals but I'm not hopeful I'll solve the issue today.

ironcladlou:
Looks like https://github.com/openshift/cluster-monitoring-operator/pull/887 might solve my problem after all, trying again

ironcladlou:
Updated to include https://github.com/etcd-io/etcd/pull/12177.

lilic:
Hey, do you mind rebasing some of these changes might have already landed in master?

@paulfantom you said there was a bugzilla for etcd mixin, does this fix that?

wking:
Bug is [rbhz#1824986](https://bugzilla.redhat.com/show_bug.cgi?id=1824986).

ironcladlou:
this already got picked up by some other update.

PR #861, opened 2020-07-17
[release-3.11] Added Prometheus K8S parameters remote write

stempher:
Added Prometheus K8S parameters remote write, updated documentation


stempher:
> 
> 
> Hello! 3.11 is closed for major feature work and this will most likely require a bump of Prometheus operator. Can you explain the reasoning of this PR, thanks! :) Does it fix a bug you have?

Hello
I was missing this functionality in 3.11 :)
Prometheus operator v23.2 supports these parameters

lilic:
Thanks for explanation, but we cannot backport major features if they were not shipped with the release. See docs on this https://access.redhat.com/support/policy/updates/openshift

PR #865, opened 2020-07-20
pkg/manifests: Allow users to configure EnforcedSampleLimit for user

lilic:

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1441,6 +1441,10 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret) (*monv1.Prometheus,
 		p.Spec.RemoteWrite = f.config.UserWorkloadConfiguration.Prometheus.RemoteWrite
 	}
 
+	if f.config.UserWorkloadConfiguration.Prometheus.EnforcedSampleLimit != nil {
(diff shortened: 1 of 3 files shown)


lilic:
This also needs bump of prometheus operator dep, which another PR introduces, hence the failed tests.

lilic:
@openshift/openshift-team-monitoring ready for review! I tested this on a cluster and checked the limit was applied to the scrape config, but will be a bit harder to add an e2e test for this as the data ends up in a Secret gzipped. 

Side note tests passed on first try :O 🎉

pkg/manifests/manifests.go
```diff
@@ -1441,6 +1441,10 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret) (*monv1.Prometheus,
 		p.Spec.RemoteWrite = f.config.UserWorkloadConfiguration.Prometheus.RemoteWrite
 	}
 
+	if f.config.UserWorkloadConfiguration.Prometheus.EnforcedSampleLimit != nil {
```
lilic, line 1444: This is a new configuration field so we do not want to add this to the old ConfigMap as a way to encourage users to migrate away from the old configmap as it will be deprecated in 4.7.

brancz: precisely why having separate structs for configs is a really good idea :+1: :100: 

very nice!

PR #866, opened 2020-07-20
Bug 1853250: pkg/manifests: update notBefore on cert renewal

simonpasquier:
This change makes sure that the renewed certificate isn't valid for
longer than the default lifetime of 2 years.

### DIFF
--- a/pkg/manifests/tls.go
+++ b/pkg/manifests/tls.go
@@ -30,6 +30,8 @@ import (
 	"k8s.io/apiserver/pkg/authentication/user"
 )
 
+const certificateLifetime = time.Duration(crypto.DefaultCertificateLifetimeInDays) * 24 * time.Hour
+
 // Taken from
 // https://github.com/openshift/library-go/blob/08c2fd1b452520da35ad210930ea9d100545589a/pkg/operator/certrotation/signer.go#L68-L86
 // without refresh time handling. We just take care of rotation if we reach 1/5 of the validity timespan before expiration.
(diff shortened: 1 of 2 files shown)


pkg/manifests/tls_test.go
```diff
@@ -111,11 +112,36 @@ func TestRotateGrpcTLSSecret(t *testing.T) {
 			name: "force rotation",
 			setup: func(s *v1.Secret) {
 				s.Annotations["monitoring.openshift.io/grpc-tls-forced-rotate"] = "true"
+				// Introduce a delay to make sure that the initial and rotated
+				// certificates aren't created within the same second.
+				time.Sleep(1 * time.Second)
```
lilic, line 117: I am guessing the below fails, can we not instead retry vs sleeping for one second? I guess since its unit tests its okay but still curious if we can approach a different way?

pkg/manifests/tls.go
```diff
@@ -110,7 +112,9 @@ func RotateGRPCSecret(s *v1.Secret) error {
 		}
 	} else {
 		template := curCA.Config.Certs[0]
-		template.NotAfter = time.Now().Add(time.Duration(crypto.DefaultCertificateLifetimeInDays) * 24 * time.Hour)
+		now := time.Now()
+		template.NotBefore = now.Add(-1 * time.Second)
```
lilic, line 116: Why can't we just use `noq`, does one second make so much difference?

simonpasquier: I've copied what's done in `github.com/openshift/library-go/pkg/crypto`:
https://github.com/openshift/cluster-monitoring-operator/blob/25720e264449ad870c1570b0a6674c545d7eac73/vendor/github.com/openshift/library-go/pkg/crypto/crypto.go#L911-L912

AFAICT it's very unlikely that we ever have a race by using `now`.

lilic: Thanks! I am fine with either, was just curious about the reasoning.

pkg/manifests/tls_test.go
simonpasquier: I've added the delay because the new certificate might have exactly the same start/end as the original certificate so it isn't 100% guaranteed that `NotBefore` has actually been updated. I don't think that retrying would be easier to reason about.
I've tried to inject a custom `time.Now()` function in the tests but without success because we're mostly relying on `github.com/openshift/library-go/pkg/crypto` for which we can't replace `time.Now`.

PR #867, opened 2020-07-20
Bug 1855275: jsonnet: update prometheus-operator cluster role

simonpasquier:

The prometheus-operator service account doesn't manage the custom
resource definitions (it's handled by CMO directly). This change cleans
up the permissions associated to that account.


simonpasquier:

PR #876, opened 2020-07-22
test/e2e: wait until the token for e2e tests is ready

simonpasquier:
There might be a race between when we retrieve the token of the service
account used for end-to-end testing and when the secret containing the
token is effectively created.

This might solve the issue we're seeing in https://github.com/openshift/release/pull/10377


simonpasquier:
cc @openshift/openshift-team-monitoring, this is ready for review.

PR #888, opened 2020-07-28
Bug 1857229: do not deploy leaf rules redundantly

s-urbaniak:


test/e2e/user_workload_monitoring_test.go
```diff
@@ -532,6 +558,18 @@ func assertUserWorkloadMetrics(t *testing.T) {
...
+	// via user workload prometheus.
+	f.ThanosQuerierClient.WaitForQueryReturn(
+		t, 10*time.Minute, `version:blah:leaf:count{prometheus_replica="prometheus-user-workload-0"}`,
+		func(i int) error {
+			if i == 1 {
+				return nil
+			}
+			return fmt.Errorf("expected count of recording rule from user application to be equal 1 but got %v", i)
+		},
+	)
```
s-urbaniak, line 572: we should also add a test that asserts that `version:blah:leaf:count{prometheus_replica="prometheus-user-workload-0"}` does not return any value.

s-urbaniak:
e2e-aws-operator tests should be more stable once https://github.com/openshift/cluster-monitoring-operator/pull/899 merges.

PR #893, opened 2020-07-30
Bug 1824996: Wait 15 minutes before alerting on KubeNodeUnreachable

joelsmith:
This brings it inline with KubeNodeNotReady

### DIFF
--- a/assets/prometheus-k8s/rules.yaml
+++ b/assets/prometheus-k8s/rules.yaml
@@ -1633,7 +1633,7 @@ spec:
         message: '{{ $labels.node }} is unreachable and some workloads may be rescheduled.'
       expr: |
         kube_node_spec_taint{job="kube-state-metrics",key="node.kubernetes.io/unreachable",effect="NoSchedule"} == 1
-      for: 2m
+      for: 15m
       labels:
         severity: warning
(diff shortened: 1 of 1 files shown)


lilic:
We do all changes in the https://github.com/kubernetes-monitoring/kubernetes-mixin hence why generate is failing. But I am curious why this change, 15 minutes seems like a long time for KubeNodeUnreachable? Can you explain the rationale a bit on the upstream PR, thanks!

joelsmith:
@lilic now that https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/491 has merged upstream, how do I get the change in this repo?

lilic:
FYI Frederic left a comment https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/491#issuecomment-686299216 saying @smarterclayton did this on purpose.

@joelsmith for openshift, do you need this to land in 4.6? If yes then you need to port the fix into the https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/release-0.5 and update it in this repo once its merged, our guide is here https://github.com/openshift/cluster-monitoring-operator/blob/master/CONTRIBUTING.md feel free to ping either here or on slack if you need more help! :)

PR #894, opened 2020-07-31
Bump jsonnet

s-urbaniak:

WIP: CHANGELOG entries for the jsonnet changes


assets/prometheus-adapter/service-monitor.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: monitoring.coreos.com/v1
```
lilic, line 1: If we want to actually create this we need to add it to pkg/manifests otherwise lets just mask it out for now?

assets/prometheus-k8s/service-monitor-kubelet.yaml
```diff
@@ -70,6 +70,19 @@ spec:
     tlsConfig:
       caFile: /etc/prometheus/configmaps/kubelet-serving-ca-bundle/ca-bundle.crt
       insecureSkipVerify: false
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    honorLabels: true
+    interval: 30s
+    path: /metrics/probes
```
brancz, line 76: let's double check with a cluster-bot that this actually works

assets/prometheus-adapter/service-monitor.yaml
s-urbaniak: sounds good, let's mask it out.

assets/prometheus-k8s/service-monitor-kubelet.yaml
s-urbaniak: seems to be working fine
image

s-urbaniak:
@openshift/openshift-team-monitoring added changelog entries PTAL

PR #898, opened 2020-08-03
Bug 1846805: Remove kube-mixin direct dependancy 

lilic:
https://github.com/kubernetes-monitoring/kubernetes-mixin should be pinned via kube-promethus https://github.com/coreos/kube-prometheus/blob/release-0.6/jsonnet/kube-prometheus/jsonnetfile.json#L48


RiRa12621:
Doesn't this make it significantly harder to get changes from kubernetes-mixin into this project as one would have to make an extra adjustment in kube-prometheus instead of bumping the version here and regenerating?

s-urbaniak:
@RiRa12621 if kubernetes-mixin has to be bumped, it should be bumped in kube-prometheus in the first place.

That was actually the motivation for this PR in the first place. kubernetes-mixin was bumped in kube-prometheus, but when vendoring kube-prometheus changes we overlooked the bump and things started breaking.

PR #902, opened 2020-08-07
Bug 1867603: ds/node-exporter: Set maxUnavailable 10%

sdodson:
Daemonsets which are not critical to availability should have higher
maxUnavailable, this will potentially speed up upgrades in large clusters.


sdodson:
Current upgrade rate is about 4m30s to upgrade 10 nodes in a 250 node cluster.

s-urbaniak:
@sdodson you'd need a BZ to merge this into current master, since we're in the QA cycle of 4.6.

assets/node-exporter/daemonset.yaml
```diff
@@ -10,6 +10,10 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/name: node-exporter
+  updateStrategy:
+    rollingUpdate:
+      maxUnavailable: 10%
```
s-urbaniak, line 15: Is unavailability of daemonsets during rollouts something that is expected? Why would daemonsets be unavailable systemically during a rolling update?

s-urbaniak:

assets/node-exporter/daemonset.yaml
sdodson: The default for daemonsets is rollingUpdate with maxUnavailable 1, so there's already one unavailable in the current scheme, you can confirm by getting the daemonset from a cluster. This just changes it so that 10% of them may be unavailable at once which parallelizes the rollout, on a 250 node cluster that decreases rollout time from around 100 minutes down to under 10 minutes.

Your operator shouldn't change its behavior around marking itself upgraded / progressing. I guess if it's monitoring the number of unavailable that may need to change

sdodson:
/retitle Bug 1867603: ds/node-exporter: Set maxUnavailable 10%

assets/node-exporter/daemonset.yaml
s-urbaniak: got it, so it is about parallelization

assets/node-exporter/daemonset.yaml
```diff
@@ -10,6 +10,10 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/name: node-exporter
+  updateStrategy:
```
s-urbaniak, line 13: @sdodson changing assets directly won't work. You'll have to update the underlying jsonnet as the library for the node_exporter daemonset is an open source upstream project.

Please contribute the change upstream in https://github.com/prometheus-operator/kube-prometheus/blob/master/jsonnet/kube-prometheus/node-exporter/node-exporter.libsonnet.

Please create an issue upstream in https://github.com/prometheus-operator/kube-prometheus and let's discuss with the community there. We can then, once merged into master, cherry-pick into releases and then vendor downstream.

sdodson: ok, will work on that over the coming days, thanks

s-urbaniak: @sdodson let me know if you need help in the upstream contribution, and thank you! :tada:

assets/node-exporter/daemonset.yaml
brancz: 10% seems like a good default

s-urbaniak:
@sdodson ping, is there anything we can help in moving this upstream?

assets/node-exporter/daemonset.yaml
sdodson: Opened https://github.com/prometheus-operator/kube-prometheus/pull/661

sdodson:
@s-urbaniak I've tried to update things after the change merged upstream.  Am I correct that there's no way to have picked those commits, the implementation in this repo seems quite different, just wanted to make sure I wasn't missing something.

s-urbaniak:
closing in favor of #918

PR #909, opened 2020-08-17
Bug 1868304: remove ValidatingWebhookConfiguration for downgrades

simonpasquier:
When downgrading from 4.6 to 4.5, the operator needs to remove the
ValidatingWebhookConfiguration resource used to validate rules in 4.6.


s-urbaniak:
ping @bparees as discussed OOB to override the BZ rules here :-)

bparees:
https://bugzilla.redhat.com/show_bug.cgi?id=1868304 really should target 4.5.z since the change is to 4.5.z.

i'm going to fix its target.

i will also set https://bugzilla.redhat.com/show_bug.cgi?id=1869301 as a 4.6.0 target (which you've already closed as fixed, which is fine since it's not actually relevant to 4.6.0).

s-urbaniak:
@simonpasquier i believe we need to add the delete permission to CMO:

```
is forbidden: User \"system:serviceaccount:openshift-monitoring:cluster-monitoring-operator\" cannot delete resource \"validatingwebhookconfigurations\" in API group \"admissionregistration.k8s.io\" at the cluster scope"
```

PR #910, opened 2020-08-17
Bug 1872309: Rename "cluster_monitoring_operator" -> "cluster-monitoring-operator"

wking:
The regular expression for the component portion is [`[a-z0-9\-]+`][1] or [`[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*?)`][2].  Either way, it cannot include underscores, so manifests like `0000_50_cluster_monitoring_operator_01-namespace.yaml` end up in the `cluster` component.  With this change, they will end up in the `cluster-monitoring-operator` component.

Generated with:

```console
$ rename cluster_monitoring_operator cluster-monitoring-operator manifests/*.yaml
$ sed -i 's/cluster_monitoring_operator/cluster-monitoring-operator/g' Documentation/data-collection.md Makefile README.md hack/build-jsonnet.sh hack/build-jsonnet.sh hack/local-cmo.sh hack/telemeter_query.go
```

[1]: https://github.com/openshift/cluster-version-operator/blob/71aef74480d199fe96a590f2f1e4e8056a9cb687/docs/dev/operators.md#what-is-the-order-that-resources-get-createdupdated-in
[2]: https://github.com/openshift/cluster-version-operator/blob/71aef74480d199fe96a590f2f1e4e8056a9cb687/pkg/payload/task_graph.go#L21


wking:
And I've filed openshift/cluster-version-operator#439 to remove the oversimplified regexp from the CVO's docs.

wking:
The outgoing, underscore form landed in 9fd35c2f59 (#388, new in 4.2), so we could backport this to any of our not-yet-end-of-life release streams if folks wanted to.  Personally, I doubt it needs backporting at all, since the only risk of the old naming is that a separate operator made the same mistake and set `cluster` as their component and 50 or 90 as their runlevel.  And even then, the outcome would be that the two operators' manifests would be interleaved in a single [manifest block][1], instead of applied in separate, parallel manifests blocks.

[1]: https://github.com/openshift/cluster-version-operator/blob/master/docs/user/reconciliation.md#manifest-graph

s-urbaniak:
this generally lgtm, but i assume we need a BZ for this as we are in QA phase?!

s-urbaniak:
Does this impact down/upgrades in any way? I don't believe so but just want to make sure we are not running in any edge case.

wking:
> this generally lgtm, but i assume we need a BZ for this as we are in QA phase?!

I asked about that a few weeks ago, but nobody has landed a bug requirement for master yet (see Tide saying `Not mergeable. Needs approved, lgtm labels.` here.  Still, I can file a bug if you like.  But...

> Does this impact down/upgrades in any way?

There's no relationship between manifest filenames between releases, so this rename has about the same pact there as the earlier #388.  The only impacts I can see now are the unlikely overlap-fix from my previous comment.  And I haven't audited a recent release, but I would be very surprised if we had an overlap like that today.  This is just cosmetic, and a guard against low-risk, future overlap.

s-urbaniak:
@wking according to tide this needs now a bugzilla :man_shrugging:

simonpasquier:
/retitle Bug 1872309: Rename "cluster_monitoring_operator" -> "cluster-monitoring-operator"

wking:
hold

GitHub claims a conflict.

wking:
Oops, looks like I fumbled the hold above.  Anyway, rebased onto master with with 886d9982 -> f11007bf

PR #913, opened 2020-08-21
Bug 1860889: decrease CMO log verbosity from 3 to 2

simonpasquier:
With `-v=3`, the logs are spammed by client-go telling that client requests have been throttled. Looking at the Kubernetes code, we won't lose anything useful by lowering the verbosity level.
    
The change also increases the QPS and burst settings (respectively 5->20 and 10->40) because the default values aren't suited for CMO that runs lots of tasks in parallel. With QPS=100 and Burst=200, the throttling logs disappeared but we felt that these values were too high for the API server.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -155,6 +155,10 @@ func Main() int {
 		return 1
 	}
 
+	// CMO runs many tasks in parallel and the default values for rate limiting are too low.
+	config.QPS = 20
+	config.Burst = 40
+
 	userWorkloadConfigMapName := "user-workload-monitoring-config"
 	o, err := cmo.New(config, *releaseVersion, *namespace, *namespaceUserWorkload, *namespaceSelector, *configMapName, userWorkloadConfigMapName, *remoteWrite, images.asMap(), telemetryConfig.Matches)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/client/client.go
```diff
@@ -716,7 +716,7 @@ func (c *Client) WaitForDeploymentRollout(dep *appsv1.Deployment) error {
 		}
 		if d.Status.UnavailableReplicas != 0 {
 			lastErr = errors.Errorf("got %d unavailable replicas",
-				d.Status.UpdatedReplicas)
+				d.Status.UnavailableReplicas)
```
s-urbaniak, line 719: good catch :+1:

cmd/operator/main.go
```diff
@@ -155,6 +155,10 @@ func Main() int {
 		return 1
 	}
 
+	// CMO runs many tasks in parallel and the default values for rate limiting are too low.
+	config.QPS = 50
+	config.Burst = 100
```
s-urbaniak, line 160: Can we comment on how we came up with these numbers?

simonpasquier: I'm still experimenting :)

simonpasquier: Still seeing throttling logs with `QPS = 50` and `Burst = 100` though less frequently.

simonpasquier:
With `QPS=100` and `Burst=200`, the throttling logs disappeared but it would be too demanding for the API server. Instead of optimizing these settings, lets decrease the log verbosity of CMO from `3` to `2`: we shouldn't lose any useful information in the logs.
QPS and burst settings still get an increase because the default values aren't suited in the context of CMO.

PR #919, opened 2020-09-03
Bug 1875470: Include OS, arch labels for node, core count

ravisantoshgudimetla:
As OpenShift becomes heterogeneous, it is ideal to get a feel of how many nodes
belong to a certain OS type, arch etc via telemetry. Especially with OpenShift
supporting Windows nodes in 4.6, the Windows Containers team would like to get
- Total number of Windows nodes used by all customers
- Total number of cores on the Windows nodes

This commit adds OS and arch labels to following rules:
- cluster:capacity_cpu_cores:sum
- cluster:node_instance_type_count:sum

which will help us in telemetry


s-urbaniak:
@ravisantoshgudimetla please rebase, the e2e tests should be fixed now.

Anandnatraj:
We will need to capture the core count for licensing purposes.

PR #921, opened 2020-09-08
Bug 1862432: unmask prometheus-adapter service monitor

dgrisonnet:

cc @s-urbaniak

### DIFF
--- a/assets/prometheus-adapter/service-monitor.yaml
+++ b/assets/prometheus-adapter/service-monitor.yaml
@@ -1,0 +1,20 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    name: prometheus-adapter
+  name: prometheus-adapter
+  namespace: openshift-monitoring
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 30s
+    port: https
+    scheme: https
+    tlsConfig:
+      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+      insecureSkipVerify: false
+      serverName: server-name-replaced-at-runtime
+  selector:
+    matchLabels:
+      name: prometheus-adapter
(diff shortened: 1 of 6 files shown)


assets/prometheus-adapter/service-monitor.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: prometheus-adapter
+  namespace: openshift-monitoring
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 30s
+    port: https
+    scheme: https
+    tlsConfig:
+      insecureSkipVerify: true
```
dgrisonnet, line 15: Should I add the certs in a follow-up PR?

s-urbaniak: We have to do it here :-) look for a reference i.e. here: https://github.com/openshift/cluster-monitoring-operator/blob/4b907eac67450edad388fe04f51343f8d6582d6d/jsonnet/prometheus-operator.jsonnet#L104-L119

s-urbaniak:
good initial start :+1 you'll have to deploy the service monitor in CMO's reconcile code, right after the deployment of prometheus-adapter: https://github.com/openshift/cluster-monitoring-operator/blob/f4b5b9b15b2691a1d6499e9587d32c260d5dbcbd/pkg/tasks/prometheusadapter.go#L179-L181

Injecting the server name can be inspired from here: https://github.com/openshift/cluster-monitoring-operator/blob/004b4975992e84460a09cb935a3a6b1ffb009aaf/pkg/manifests/manifests.go#L1764-L1774

dgrisonnet:
@s-urbaniak I added the cert and the service monitor reconciliation. PTAL

pkg/manifests/manifests.go
```diff
@@ -1711,6 +1712,18 @@ func (f *Factory) PrometheusAdapterService() (*v1.Service, error) {
 	return s, nil
 }
 
+func (f *Factory) PrometheusAdapterServiceMonitor() (*monv1.ServiceMonitor, error) {
```
s-urbaniak, line 1715: let's add a unit test in `manifests_test.go` too

s-urbaniak:
one nit, else LGTM :+1:

s-urbaniak:
Please rebase as we fixed the failing e2e test.

PR #927, opened 2020-09-15
Bug 1878155: Remove deprecated baseImage

dgrisonnet:


lilic (commented):
All the examples still include baseImage, we should modify those as well so users don't copy paste those. https://github.com/openshift/cluster-monitoring-operator/blob/047d127d78f7bc55af58b31973c7b999549aee72/Documentation/user-guides/configuring-cluster-monitoring.md#configuring-custom-images

dgrisonnet:
The examples refer to CMO's configmap configuration and not the prometheus-operator one.
The baseImage from the configmap is loaded into the [.spec.Image](https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/manifests/manifests.go#L1238) field of the custom resource which is not deprecated so we shouldn't have any problem here.

That being said, you mentioned that the new naming might be confusing to users coming from the prometheus-operator community. However, as all the objects of the configmap contains a baseImage and no Image field, I don't think it will be that confusing. It is just that the configmap spec is slightly different from the targeted object one.

lilic:
>That being said, you mentioned that the new naming might be confusing to users coming from the prometheus-operator community. However, as all the objects of the configmap contains a baseImage and no Image field, I don't think it will be that confusing. It is just that the configmap spec is slightly different from the targeted object one.

Yes that is my fear, a bit of a nit but wonder if we should deprecate this and add image field as well? @s-urbaniak WDYT?

dgrisonnet:
Since all the baseImage from the configmap gets translated to Image fields, I would agree that replacing by Image would be more appropriate if we want to reflect the original spec.

s-urbaniak:
> Yes that is my fear, a bit of a nit but wonder if we should deprecate this and add image field as well? @s-urbaniak WDYT?

I believe this is a good idea :+1: My hypothesis is that very few customers use that field anyways as it is dubious from a support perspective anyways.

I am fine with adding an explicit `image` field, and deprecating `baseImage`. My intuition is that both fields should be functional and target just to `image` fields of the corresponding specs as described above.

lilic:
> I am fine with adding an explicit image field, and deprecating baseImage. My intuition is that both fields should be functional and target just to image fields of the corresponding specs as described above.

Yes agreed, I would say we can do this in 4.6 onwards (no need to backport) or do you think its too late?

s-urbaniak:
@lilic i would not change the API during feature freeze, we'd probably need more coördination here with @bparees for instance :thinking:

bparees:
> @lilic i would not change the API during feature freeze, we'd probably need more coördination here with @bparees for instance

i need a little more background on the api (who uses it, when was it introduced, and are you removing it or just deprecating it?)

lilic:
> i need a little more background on the api (who uses it, when was it introduced, and are you removing it or just deprecating it?)

Our end users use it to configure our monitoring stack (we are planning on moving to CRs in the future), this field is used to override the base image, lets say if you want to run it from a different registry or just run a different version I guess for some reason. Upstream Prometheus-operator has deprecated the name `baseImage` to configure in favour of `image`.

We have couple of options:
1. Deprecate it in the docs, add a new field called image for new users to configure that and existing to hopefully migrate from that one and logs of cluster-monitoring-operator, continue using baseImage by setting image instead. e.g. support both baseImage and add new image field.
2. Do not deprecate in docs but add new field and continue supporting both.
3. Keep as is and hide from the user that anything changed.
4. Remove in the future and use new field called image.

I think either is fine for us, just curious what is the process in openshift around user facing APIs, since we use ConfigMaps we don't really say this is stable, e.g. v1 for example so maybe we have the right to change our API. I think this question is good to answer also as we want to move to CustomResource and away from ConfigMaps.

Thanks @bparees! :)

bparees:
option (1) is closest to what we'd normally do in openshift.  I don't quite understand how option 3 would work.  And option 4 is a non-starter (for better or worse, the content of your configmap is your api, you need to keep supporting the content that used to work in that configmap)


but @s-urbaniak is correct that this isn't something we'd normally want to merge post-FF.  

that said, it sounds low risk with no real impact to existing customers unless they chose to use the new field, so if it's going to reduce confusion in 4.6 i'm ok w/ you moving forward with it.

dgrisonnet:
Sounds good to me, should we create a new ticket for this API change? This PR is only meant to fix the prometheus-operator arguments, so I don't think we should update the configmap here. Also, I will need to backport this fix down to 4.4 which we will not for the API changes.

lilic:
Thanks Ben, my thinking as well! Jus to clarify the option 3 is this PR which keeps baseImage but just fixes the "bug".

@s-urbaniak which way do we want to go here?

bparees:
>  Just to clarify the option 3 is this PR which keeps baseImage but just fixes the "bug".

so the "bug" is that the baseImage field takes precedence and there is an older/invalid value being put into that field, so you're removing the use of that field in your default config?

s-urbaniak:
Option 1 sounds the least intrusive for me now and using that I am fine with merging. Once we migrate to CRDs we can have a more elaborate migration logic.

dgrisonnet:
I looked a bit more into the ability to update images from CMO's configmap and could not understand how this was supposed to work in practice. After digging a bit more, it seemed like it was a leftover from CoreOS Tectonic that is not needed and usable anymore.
The only way to override CMO components' images should be via the "--images" flag, and we should stop advertising the old mechanism in the documentation.

simonpasquier:
Agreed. I guess we don't need the change in `pkg/manifests/manifests_test.go` for now.

simonpasquier:
leaving up to @s-urbaniak and/or @lilic to unhold.

lilic:
👍 lgtm as well

PR #934, opened 2020-09-21
Bug 1881061: Add UWM namespace to the operator's related objects

simonpasquier:
Previously CMO was only referencing the "openshift-monitoring" namespace
as a related object.

This change adds "openshift-user-workload-monitoring" namespace as well
as all the ThanosRuler and PodMonitor objects.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -57,19 +57,20 @@ const (
 )
 
 type Client struct {
-	version           string
-	namespace         string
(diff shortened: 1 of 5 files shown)


s-urbaniak:
@simonpasquier i believe e2e tests need a fix

simonpasquier:
@s-urbaniak correct, e2e tests updated too.

PR #936, opened 2020-09-23
Bug 1881072: Bump dependencies including prometheus-operator to 0.42.1

lilic:
promethus-operator is now at 0.42.1, this just finalizes the merge, I also updated all our deps which brings in a few new alerts but that would have happened eventually with the next bump.


lilic:
/retest

gcp...

PR #938, opened 2020-09-24
Bug 1845561: disable (temporarily) resource requests for config reloaders

simonpasquier:
* [ ] I added CHANGELOG entry for this change.
* [X] No user facing changes, so no entry in CHANGELOG was needed.

PR #943, opened 2020-09-29
Bug 1845561: enable resource requests for config reloaders

simonpasquier:
Follow-up of #938. Needs https://github.com/openshift/prometheus-operator/pull/94 to be merged first.


jsonnet/prometheus-user-workload.jsonnet
```diff
@@ -256,21 +256,15 @@ local alertmanagerRole =
...
-            //      memory: '5Mi',
-            //      cpu: '1m',
-            //    },
-            //  },
-            //},
+            {
+              name: 'config-reloader',
+              resources: {
+                requests: {
+                  memory: '5Mi',
```
simonpasquier, line 263: @s-urbaniak not sure if we still want to set memory requests here. If yes, which value should we pick?
`5Mi` is the previous value when we used `jimmidyson/configmap-reload`. Looking at the numbers from the `e2e-aws-operator` job, `prometheus-config-reloader` is a bit more demanding.

image

s-urbaniak: looking at the graph `10Mi` seems like a reasonable value to me.

simonpasquier: :+1:

second question: should we apply the same memory request to all config requests (Alertmanager, prometheus-k8s and Thanos ruler)? I'll also gather data from a running cluster to corroborate.

s-urbaniak: if they expose the same memory consumption, I would say so.

simonpasquier: `10Mi` seems about right.

image

simonpasquier:
PR updated setting memory request to `10Mi` for all config-reloader containers.

jsonnet/prometheus-user-workload.jsonnet
s-urbaniak: perfect

s-urbaniak:
@simonpasquier does it make sense to bump the resource requests in kube-prometheus too?

PR #947, opened 2020-10-05
Allowlist cco_credentials_mode metric from cloud credential operator.

dgoodwin:
This metric reports the mode the operator is configured to use, which we
would like to see across the entire fleet to understand real world
customer use.


dgoodwin:
Results from this metric currently include what modes are disabled:

```

{mode="credsremoved",service="cco-metrics"} | 0
{mode="degraded",service="cco-metrics"} | 0
{mode="manual",service="cco-metrics"} | 0
{mode="mint",service="cco-metrics"} | 1
{mode="passthrough",service="cco-metrics"} | 0
{mode="unknown",service="cco-metrics"} | 0
```

This may need changing, please let me know. These 6 modes will likely be 8 or 10 within a few releases.

Documentation/data-collection.md
```diff
@@ -236,6 +236,8 @@ data:
     - '{__name__="che_workspace_start_time_seconds_sum"}'
     # (codeready workspaces, @ibuziuk) The overall number of attempts for starting all the workspaces.
     - '{__name__="che_workspace_start_time_seconds_count"}'
+    # (cloud credential operator, @dgoodwin) Track current mode the cloud-credentials-operator is functioning under.
```
s-urbaniak, line 239: please add a github team, not a single person. We have precedences, but they were exceptions which need to be resolved.

s-urbaniak:
cc'ing @smarterclayton for final approval

Documentation/data-collection.md
dgoodwin: My mistake, addressed now.

dgoodwin:
Based on discussion on slack I think we need to modify this in the code to only report the 1's. Can I pursue this separately or does that have to be in place before this can merge?

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -228,6 +228,8 @@ data:
     - '{__name__="che_workspace_start_time_seconds_sum"}'
     # (codeready workspaces, @ibuziuk) The overall number of attempts for starting all the workspaces.
     - '{__name__="che_workspace_start_time_seconds_count"}'
+    # (cloud credential operator, @openshift/openshift-team-hive) Track current mode the cloud-credentials-operator is functioning under.
```
lilic, line 231: Is this operator installed by default or is an optional component?

dgoodwin: Installed by default CVO operator.

dgoodwin:
Removing the 0 metrics for 4.7: https://github.com/openshift/cloud-credential-operator/pull/256

dgoodwin:
Bump, would like to get this wrapped up this sprint. Can anyone approve?

PR #951, opened 2020-10-07
jsonnet: Collect all resource counts for telemetry

smarterclayton:
The data about how many instances of a CRD are in use on the cluster
is a key question for analyzing the success and health of a cluster
as well as for understanding how a cluster can be more successful.
From a configuration perspective, it indicates use of core platform
features. From an operator perspective, it shows adoption and usage
in a way that installation stats alone do not capture. From a support
perspective it highlights interactions between operators that may
help problem resolution.

On an average cluster this increases the series gathered by ~130
series (18 gathered today, 155 after this change, + 280 on average).
While this is a 50% increase in gathered series, it drives significant
value in all of the domains mentioned above and we believe we have
the budget to perform this increase and that using the budget on these
series specifically will deliver significant value.


assets/prometheus-k8s/rules.yaml
```diff
@@ -927,8 +927,7 @@ spec:
       record: cluster:kubelet_volume_stats_used_bytes:provisioner:sum
     - expr: sum(etcd_object_counts) BY (instance)
       record: instance:etcd_object_counts:sum
-    - expr: max(etcd_object_counts{resource=~"builds\\.build\\.openshift\\.io|deploymentconfigs\\.apps\\.openshift\\.io|images\\.image\\.openshift\\.io|statefulsets\\.apps|deployments\\.apps|cronjobs\\.batch|jobs\\.batch|pods|persistentvolumeclaims|persistentvolumes|services|namespaces|routes\\.route\\.openshift\\.io|ingresses\\.networking\\.k8s\\.io|horizontalpodautoscalers\\.autoscaling|users\\.user\\.openshift\\.io|machinesets\\.machine\\.openshift\\.io|resourcequotas"})
-        by (resource)
+    - expr: max(etcd_object_counts) by (resource)
```
lilic, line 930: Seems like with this change we edging close to the limit we set in our origin test (its at 473 series we sent from this CI run), do you mind adjusting that limit to a reasonable one as well, so we don't get flaky or failing tests. Thanks!
https://github.com/openshift/origin/blob/712d7c621fe4353077f889bb754fb55827dbabcf/test/extended/prometheus/prometheus.go#L103-L106

lilic (commented):
/lgtm
/hold

I assume someone from observatorium team wants to approve this as it adds a huge change @bwplotka @squat  PTAL and remove hold if its okay for you.

bwplotka:
As mentioned offline, given the value it brings and some spare room we have still on our nodes we are happy with this from the telemeter side 👍 /lgtm

bwplotka:
/hold cancel 

🔥

assets/prometheus-k8s/rules.yaml
smarterclayton: https://github.com/openshift/origin/pull/25592

smarterclayton:
I'm debating changing this to `topk(250, etcd_object_counts)` to prevent potentially unbounded cardinality from people doing CRD insanity.  Objections / opinions?

bwplotka:
I feel it's an amazing idea @smarterclayton  since we drop the allow list approach. 

However, this will drop metrics quietly. I wonder if we can actually add something in Prometheus upstream (ideas welcome) to make it a rule config field etc that will limit series etc.

smarterclayton:
I think we could live with the "drop" for the short term but I completely agree the rate limit on both series count per metric and overall series would be something that a tenant / service owner would prefer (with some notification of such)

smarterclayton:
Capped to 500 in https://github.com/openshift/cluster-monitoring-operator/pull/957

PR #957, opened 2020-10-12
jsonnet: Cap total resources:sum series

smarterclayton:
In order to defend telemetry against clusters that may have thousands
of resource types, cap the cluster:usage:resources:sum series to 500.
Most have 150, and of those a large number are single or zero instance
counts, so we anticipate for the vast majority of clusters 500 is
enough to get reasonable workload and prevent a full cardinality
explosion.

* [x]  No user facing changes, so no entry in CHANGELOG was needed.


bwplotka (approved):
/lgtm 

:+1:

PR #970, opened 2020-10-27
Bug 1885244: bump prometheus operator to v0.43.0

simonpasquier:


jsonnet/main.jsonnet
```diff
@@ -106,6 +106,30 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
...
+                                 ),
+                               }
+                             else
+                               c,
+                           super.containers,
+                         ),
+                     },
+                   },
+                 },
+               },
```
simonpasquier, line 132: @s-urbaniak for awareness: rather than modifying the "upstream" [mixin](https://github.com/openshift/telemeter/blob/master/jsonnet/telemeter/client/kubernetes.libsonnet), I've decided to override because it avoids opening a third PR for the bump and it would be easier to backport to 4.6 if needed.

simonpasquier:
xref https://github.com/openshift/prometheus-operator/pull/98

simonpasquier:
@s-urbaniak this needs another lgtm :smile:

jsonnet/main.jsonnet
s-urbaniak: sounds good, thank you for the heads-up!

side note: the above code snippet looks very scary :see_no_evil:

PR #973, opened 2020-10-28
make role manifest sort stable

dhellmann:
The sort command produces different output on my system than in the CI
job, reversing the order of
`assets/prometheus-operator-user-workload/cluster-role.yaml` and
`assets/prometheus-operator/cluster-role.yaml`. This seems to be
caused by a difference in evaluating the `-` and `/`
characters. Moving the filename sort into python produces stable
output.

See early versions of #971 for an example of the difference in output
caused by the unstable sort.


dgrisonnet:
Thank you for making this system independent.
Just for the record, this sounds like a locale issue, setting `LC_COLLATE=C` should fix it.

/lgtm

simonpasquier:
/lgtm
Thanks a lot @dhellmann!

PR #980, opened 2020-11-16
Bug 1872786: jsonnet: remove apiserver_request:availability30d

dgrisonnet:
The apiserver_request:availability30d recording rule evaluation fails in
big clusters. As this rule is currently neither used by customers nor by
the monitoring stack, it is safe to remove it for now until the rule is
fixed upstream.


dgrisonnet:
Holding this PR until we run the experiment with the potential upstream fix.

/hold

simonpasquier:
/hold
waiting for #985

simonpasquier:
you'd have to rebase because of bindata removal :)

RiRa12621:
@simonpasquier I think tests pass, can we LGTM this so it gets to QA? We're getting a significant amount of those for OSD apparently

RiRa12621:
/hold
depends on #1010

PR #981, opened 2020-11-16
Bug 1868976: jsonnet: configure SCCs

s-urbaniak:


lilic:
/retest

doesn't seem related.

s-urbaniak:

side note: i am in parallel working out where to place a sensible upgrade test for this in origin/e2e. I guess we cannot do an upgrade test here.

s-urbaniak:
fwiw we already have an e2e test that at least tests Prometheus PVC: https://github.com/openshift/cluster-monitoring-operator/blob/f4b5b9b15b2691a1d6499e9587d32c260d5dbcbd/test/e2e/prometheus_test.go#L30-L88

do you think it makes sense to add more PVC tests here, i.e. for alertmanager and user-workload-prometheus?

jsonnet/alertmanager.jsonnet
```diff
@@ -21,6 +21,16 @@ local authorizationRole = policyRule.new() +
                           ]) +
                           policyRule.withVerbs(['create']);
 
+local sccRole = policyRule.new() +
```
simonpasquier, line 24: It would make sense to have a comment IMO.

jsonnet/prometheus-user-workload.jsonnet
```diff
@@ -54,6 +54,18 @@ local alertmanagerRole =
   ]) +
   policyRule.withVerbs(['get']);
 
+local sccRole =
```
simonpasquier, line 57: ditto

simonpasquier (commented):
Agree that additional tests for Alertmanager and UWM would be good to have.

jsonnet/alertmanager.jsonnet
s-urbaniak: agreed, this is a good point.

jsonnet/prometheus-user-workload.jsonnet
s-urbaniak: agreed

s-urbaniak:
@simonpasquier thanks for the suggestion! ptal if the place/wording lgty. I would submit the additional e2e tests as a follow-up if possible.

s-urbaniak:
looking into the e2e failure, it seems to be related.
```
    --- FAIL: TestUserWorkloadMonitoring/enable_user_workload_monitoring,_assert_prometheus_rollout#01 (0.01s)
```

s-urbaniak:
failure does seem to be unrelated, the failing test function fails at creating the configmap to turn on user workload monitoring:
```
    user_workload_monitoring_test.go:139: retrieving ConfigMap object failed: Get "https://api.ci-op-w6vf2c6m-a317e.origin-ci-int-aws.dev.rhcloud.com:6443/api/v1/namespaces/openshift-monitoring/configmaps/cluster-monitoring-config": dial tcp 54.243.39.215:6443: connect: connection refused
=== CONT  TestUserWorkloadMonitoring
    user_workload_monitoring_test.go:96: scenario "enable user workload monitoring, assert prometheus rollout" failed
...
```

s-urbaniak:
thanos ruler is missing, thank you @simonpasquier for the OOB catch :+1:

s-urbaniak:
the more i think about it though, the more i believe we should add those e2e tests here :thinking: we are adding quite some settings to those statefulsets here, so let me amend the tests in this very PR.

s-urbaniak:
I just found we already have PVC tests for alertmanager :sweat_smile:

s-urbaniak:
And we have also PVC tests for UWM Prometheus, adding a dedicated one for Thanos Ruler.

simonpasquier (commented):
/lgtm

PR #983, opened 2020-11-18
Remove deprecated user workload configuration

lilic:
With 4.7 onwards we do not support techPreview configuration, so code that handles that can be removed. Users had one release to migrate, as agreed beforehand. I added a log warning users if they still use that that it will simply not work. We will also mention this in the CHANGELOG as well as in the docs for 4.7, in 4.6 we do not mention the old techPreview configuration anyways.


lilic:
@openshift/openshift-team-monitoring please have a look, thanks!

lilic:
Tested on the cluster launched from this PR and used old tech preview field which nicely resulted in non broken stack and the following in the logs:

> I1118 13:24:33.034750       1 tasks.go:46] ran task 3 of 15: Updating Cluster Monitoring Operator
I1118 13:24:34.098157       1 tasks.go:46] ran task 1 of 15: Updating Prometheus Operator
I1118 13:24:34.911699       1 tasks.go:46] ran task 12 of 15: Updating Telemeter client
W1118 13:24:36.345776       1 config.go:431] DEPRECATED: Migrate to new user workload monitoring configuration, this tech preview was removed.

PR #984, opened 2020-11-18
pkg/operator/operator.go: Adjust log format of resyncPeriod

lilic:
@paulfantom found that now we log the following by :

> ConfigMap to configure stack does not exist. Reconciling with default config every 900000000000 minutes.

This is incorrect, this adjusts it to include minutes and seconds format, see preview [here](https://play.golang.org/p/NAV03uZYEo4).

New format should be:

> ConfigMap to configure stack does not exist. Reconciling with default config every 15m0s.

This should work for all further changes to the resyncPeriod, if we change it to seconds or hours. :) 
 

lilic:
Output from logs:

> I1118 14:17:28.242341       1 operator.go:216] ConfigMap to configure stack does not exist. Reconciling with default config every 15m0s.

PR #986, opened 2020-11-19
IBM Cloud manifest profile patch

csrwng:
Part 2 of cluster profiles support for IBM Cloud.
Introduces a directory that will contain profile-specific patches.
Profile-specific manifests are generated with 'make profile-manifests'
(which is called by 'make generate'). The generate test should verify
that the profiles are up to date.

On IBM Cloud the monitoring operator needs to run on worker nodes, thus the
master node selector needs to be removed from the operator deployment
manifest.

### DIFF
--- a/jsonnet/ibm-cloud-managed-profile.libsonnet
+++ b/jsonnet/ibm-cloud-managed-profile.libsonnet
@@ -1,0 +1,42 @@
+{
+  // Generates manifests specific to the ibm-cloud-managed profile for use
+  // in ROKS clusters.
+  // Specifically, it generates a new cluster monitoring operator deployment by
+  // modifiying the original deployment manifest:
+  // - adds the ibm-cloud-managed profile annotation
+  // - removes the master node selector
+  // Contact: cewong@redhat.com
+
+  local removeMasterNodeSelector(sel) = {
+      [if x != "node-role.kubernetes.io/master" then x]: sel[x]
+      for x in std.objectFields(sel)
+    },
+
+  local setProfileAnnotation(metadata,profile) =
+    local annotations = if "annotations" in metadata then metadata.annotations else {};
+    {
+      [if !std.startsWith(x, "include.release.openshift.io/") then x]: annotations[x]
+      for x in std.objectFields(annotations)
+    } +
+    {
+      ["include.release.openshift.io/" + profile]: "true"
+    },
+
+
+  manifests+:: {
+    local operatorDeployment = import 'manifests/0000_50_cluster-monitoring-operator_05-deployment.json',
+    "0000_50_cluster-monitoring-operator_05-deployment-ibm-cloud-managed": operatorDeployment {
+      local originalMetadata = super.metadata,
+      metadata+: {
+        annotations: setProfileAnnotation(originalMetadata,"ibm-cloud-managed"),
+      },
+      spec+: {
+        template+: {
+          spec+: {
+            nodeSelector: removeMasterNodeSelector(super.nodeSelector),
+          },
+        },
+      },
+    },
+  },
+}
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,11 @@ JSONNETFMT_BIN=$(BIN_DIR)/jsonnetfmt
 PROMTOOL_BIN=$(BIN_DIR)/promtool
 TOOLING=$(EMBEDMD_BIN) $(GOBINDATA_BIN) $(JB_BIN) $(GOJSONTOYAML) $(JSONNET_BIN) $(JSONNETFMT_BIN) $(PROMTOOL_BIN)
 
+MANIFESTS_DIR ?= $(shell pwd)/manifests
+JSON_MANIFESTS_DIR ?= $(shell pwd)/tmp/json-manifests/manifests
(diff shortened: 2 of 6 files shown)


csrwng:
The profile-specific manifest should handle the deployment of the cluster monitoring operator, however the prometheus operator applied by the cluster monitoring operator still contains a node selector for master nodes. Open to ideas on how this could be handled by the operator.

lilic:
Hey, thanks for the PR, can you explain a bit about profiles, was there an OpenShift enhancements for this you can share will help provide more context when reviewing. Also help our QE do testing. Thanks!

lilic:
> The profile-specific manifest should handle the deployment of the cluster monitoring operator, however the prometheus operator applied by the cluster monitoring operator still contains a node selector for master nodes. 

Yes we have `nodeSelector` field in Prometheus operator, which you can specify in a ConfigMap, so I would suggest to just create a configmap that is ready to be picked up by cluster-monitoring-operator. https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/manifests/config.go#L85
Docs -> https://docs.openshift.com/container-platform/4.1/monitoring/cluster_monitoring/configuring-the-monitoring-stack.html#moving-monitoring-components-to-different-nodes_configuring-monitoring

s-urbaniak:
I agree with what @lilic mentioned. It seems we have the necessary API exposed for prometheus-operator.

When it comes to cluster-monitoring-operator I see the `include.release.openshift.io/ibm-cloud-managed` annotation which resembles to me the discussion we had around "profiles" earlier. Is this going to be consistent across other cloud implementations?

Generally I am very hesitant to add this here as we have to maintain a second, redundant manifest that is not covered by e2e tests.

/cc @derekwaynecarr @smarterclayton

s-urbaniak:
cc'ing @bbrowning for awareness of additional stakeholders.

s-urbaniak:
also cc'ing @sichvoge for PM perspective

simonpasquier:
IIUC this is the corresponding enhancement proposal: https://github.com/openshift/enhancements/blob/master/enhancements/update/ibm-public-cloud-support.md

> Generally I am very hesitant to add this here as we have to maintain a second, redundant manifest that is not covered by e2e tests.

My very concern too. We need to clearly identify who owns this code.

csrwng:
Hi folks, thank you so much for taking a look. The other relevant enhancement is https://github.com/openshift/enhancements/blob/master/enhancements/update/cluster-profiles.md
particularly the section around node selectors. 

It may be useful to schedule some time to discuss issues around maintainability/support. I'll try to set that up.

csrwng:
> When it comes to cluster-monitoring-operator I see the `include.release.openshift.io/ibm-cloud-managed` annotation which resembles to me the discussion we had around "profiles" earlier. Is this going to be consistent across other cloud implementations?

@s-urbaniak not cloud implementations, but profiles. Clouds like amazon, azure, google do not require a different OCP profile. However, IBM Cloud managed uses a completely different deployment model where the control plane is hosted on a separate cluster. This is why we have a separate profile for it.

s-urbaniak:
@csrwng referring this PR: the consensus from the team so far is that the patching mechanism should be also solved using jsonnet as we do elsewhere in this project. As the `manifests` has been static historical this needs some thoughts though. A suggestion in this PR is very welcome!

csrwng:
@s-urbaniak please take a look at the recent change. It leaves existing manifests static as is. However, generates json from them so they can be imported and modified in jsonnet. Then an IBM cloud specific libsonnet module creates a new deployment manifest based on the existing one.

jsonnet/ibm-cloud-managed-profile.libsonnet
```diff
@@ -0,0 +1,34 @@
+{
```
simonpasquier, line 1: Can we add a comment saying the intent + a point of contact?

simonpasquier (commented):
Very nice work!

jsonnet/ibm-cloud-managed-profile.libsonnet
csrwng: done, thx!

s-urbaniak:
@csrwng is this still WIP? If not, please remove it so we can do a final pass and thank you for the jsonnet integration! :tada:

csrwng:
@s-urbaniak yes, it needs to remain in WIP (or hold) until cluster profile support is merged in to the CVO (https://github.com/openshift/cluster-version-operator/pull/404). Otherwise the current CVO will try to apply all manifests and ignore profile-specific annotations. Thank you for reviewing!

lilic:
@csrwng I was under the impression that https://github.com/openshift/cluster-monitoring-operator/pull/986#issuecomment-730047373 is still needed for Prometheus operator?

csrwng:
@lilic yes we still need to change the selector for the prometheus operator, but that can be done by a manifest deployed by the ROKS toolkit on cluster creation. We don't need to include it in the release manifests.

csrwng:
@simonpasquier @s-urbaniak this should be ready now. Cluster profiles are now supported by the CVO and it should be safe to include in the release.

simonpasquier (commented):
/lgtm

PR #987, opened 2020-11-23
Create pod priority class for user workload monitoring

lilic:
This introduces a new PriorityClass named `openshift-user-critical`, and applies that policy to all user workload components. We want this because we have seen user workload pods get scheduled in favor of platform monitoring pods. We always want platform monitoring to have higher priority scheduling as it monitors platform critical workloads, user workload monitoring stack is now second priority class before all rest of user workloads.

See background on this https://github.com/openshift/enhancements/pull/369.


lilic:
cc @openshift/openshift-team-monitoring PTAL, I tested this on a cluster and the class is applied, not sure we need a test for it? If we were to test any disruption to the workloads, that would be testing the priority class k8s feature. Any other test suggestions, welcome!

s-urbaniak:
My intuition is also that testing if the priority is applied is out of scope here (i.e. disruption).

What we _could_ test is if the actual priority value of `system-cluster-critical` is higher than `openshift-user-critical`. That could be done as an e2e tests which reads out the values in a concrete cluster.

lilic:
@s-urbaniak  Sounds good, this test would be there to determine the new priority class is lower priority than the system one, correct?

test/e2e/main_test.go
```diff
@@ -142,3 +143,29 @@ func TestMemoryUsageRecordingRule(t *testing.T) {
 		"count(namespace:container_memory_usage_bytes:sum)",
 	)
 }
+
+// This makes sure that the priority class we create is
+// present and lower than the system priority classes. Example:
+// openshift-user-critical   1000000000   false            66m
+// system-cluster-critical   2000000000   false            114m
+// system-node-critical      2000001000   false            114m
+func TestToEnsureUserPriorityClassIsPresentAndLower(t *testing.T) {
```
lilic, line 152: @s-urbaniak wdyt of this test case, is this something like what you had in mind? Was not sure where to place it, as it does relate to user workload monitoring but since the class is created before it can be part of the main as well? Suggestions welcome!

lilic:
> level=error msg=Cluster operator authentication Degraded is True with IngressStateEndpoints_MissingSubsets::OAuthServerConfigObservation_Error::OAuthServiceCheckEndpointAccessibleController_SyncError::OAuthServiceEndpointsCheckEndpointAccessibleController_SyncError::RouterCerts_NoRouterCertSecret: OAuthServiceCheckEndpointAccessibleControllerDegraded: Get "https://172.30.152.80:443/healthz": context deadline exceeded (Client.Timeout exceeded while awaiting headers)

unrelated it seems
/retest

test/e2e/main_test.go
s-urbaniak: suggestion: maybe let's place it in a separate `test/e2e/priority_class_test.go`?

test/e2e/main_test.go
```diff
@@ -142,3 +143,29 @@ func TestMemoryUsageRecordingRule(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Get our user priority class value.
+	userPriorityClass, err := f.SchedulingClient.PriorityClasses().Get(context.TODO(), "openshift-user-critical", v1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Ensure our is a lower value than the system class' ones.
+	if userPriorityClass.Value >= systemClusterPriorityClass.Value && userPriorityClass.Value >= systemNodePriorityClass.Value {
```
s-urbaniak, line 168: logical question: shouldn't this be an or `||`? This would pass if i.e. the user priority is higher than system cluster priority but lower than the system node priority, aka "in between".

s-urbaniak:
@lilic the test looks spot on (at least to me) :+1: I have just one logical question.

test/e2e/main_test.go
lilic: Sounds good as well!

test/e2e/main_test.go
lilic: Indeed good catch!

lilic:
@s-urbaniak PTAL, adjusted as per feedback, thanks!

simonpasquier:
/lgtm

should we get an official ack from OpenShift architects?

lilic:
@simonpasquier thanks!

> should we get an official ack from OpenShift architects?

We got an 👍  from Derek already, before that on forum-arch folks said yes to it, on both the name and this approach and it can be used by other user workloads in the future. Just location was never agreed upon, so for now we create it, and if other components want to consume it we can move it. SGTY?

simonpasquier:
@lilic this is fine by me then! I wasn't sure reading the enhancement proposal and comments that everything was already agreed :)

lilic:
@simonpasquier Yep was lots of comments :) https://github.com/openshift/enhancements/pull/369#discussion_r504037208 and https://github.com/openshift/enhancements/pull/369#discussion_r508504250 are the main comments, everything else is specific to the priority class being applied to other components or its location. There was also a lot of out of band discussions on slack.

PR #989, opened 2020-11-23
Bug 1900792: Collect all resource counts for telemetry

smarterclayton:
Transcribe all etcd_object_counts (number of all objects) into cluster resource usage for reporting in cluster telemetry.

The data about how many instances of a CRD are in use on the cluster is a key question for analyzing the success and health of a cluster as well as for understanding how a cluster can be more successful. From a configuration perspective, it indicates use of core platform features. From an operator perspective, it shows adoption and usage in a way that installation stats alone do not capture. From a support perspective it highlights interactions between operators that may help problem resolution.

On an average cluster this increases the series gathered by ~130 series (18 gathered today, 155 after this change, + 280 on average). While this is a 50% increase in gathered series, it drives significant value in all of the domains mentioned above and we believe we have the budget to perform this increase and that using the budget on these series specifically will deliver significant value.

In order to defend telemetry against clusters that may have thousands of resource types, cap the cluster:usage:resources:sum series to 500. Most have 150, and of those a large number are single or zero instance counts, so we anticipate for the vast majority of clusters 500 is enough to get reasonable workload and prevent a full cardinality explosion.


lilic:
Sounds good to me, but seems like bugzilla is not valid.

/retest

lilic:
@smarterclayton As either 150 or 500 seems like a large amount, I assume you informed the observatorium team about it?

smarterclayton:
/hold

We will chat about and verify early next week (this just accelerates the timeframe, but it will pop load).

bwplotka (approved):
LGTM from my side.

Discussion: https://coreos.slack.com/archives/CEG5ZJQ1G/p1606749054374200?thread_ts=1606748500.374000&cid=CEG5ZJQ1G

/lgtm

smarterclayton:
/hold cancel

Just waiting for cherry-pick approval

juzhao:
tested with the cluster launched by cluster-bot
`launch openshift/cluster-monitoring-operator#989 aws`, "cluster:usage:resources:sum" record rule is as below and the metric works well
# oc -n openshift-monitoring get cm prometheus-k8s-rulefiles-0 -oyaml 
      - expr: topk(500, max(etcd_object_counts) by (resource))
        record: cluster:usage:resources:sum

/lgtm

sdodson:
Putting a temporary hold on things to prioritize items that fix test flakes while we're hitting AWS quota limits. I'll clear this later today.
/hold

sdodson:
Needs a rebase and we're dumping the patch manager queue to accelerate fixes for a recently discovered regression.

sdodson:
/lgtm cancel
Needs rebase anyway

smarterclayton:
Second prometheus didn't come up (looks like a node was down?).

/retest

markmc:
> LGTM from my side.
> 
> Discussion: https://coreos.slack.com/archives/CEG5ZJQ1G/p1606749054374200?thread_ts=1606748500.374000&cid=CEG5ZJQ1G
> 
> /lgtm

Summary: the Telemetry team are comfortable with the additional load this introduces

markmc:
(patch manager) backport of #951 and #957, everyone seems on board with the plan

PR #995, opened 2020-11-30
Bug 1897252: Add Thanos query log level

lilic:
This adds the ability to modify verbosity of log level for thanos query. We need this in CI to debug a thanos query bug (linked as a bugzilla), so might be useful for our customers as well.

Also noticed we now always logged the msg about techPreview being DEPRECATED, so now we only shout at users in logs that we have DEPRECATED the techPreview, is they set Enabled field.


lilic:

pkg/manifests/config.go
```diff
@@ -428,6 +429,8 @@ func (c *Config) IsUserWorkloadEnabled() bool {
 // and depracted from 4.7 onwards.
 // Instead warn user in the logs.
 func (c *UserWorkloadConfig) isEnabled() bool {
-	klog.Warning("DEPRECATED: Migrate to new user workload monitoring configuration, this tech preview was removed.")
+	if c.Enabled != nil {
```
simonpasquier, line 432: nice!

PR #996, opened 2020-11-30
test/e2e: Add Thanos Query Watchdog Alert test

lilic:

### DIFF
--- a/test/e2e/thanos_querier_test.go
+++ b/test/e2e/thanos_querier_test.go
@@ -108,3 +108,15 @@ func TestThanosQuerierTrustedCA(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestThanosQueryCanQueryWatchdogAlert(t *testing.T) {
+	// The 2 minute timeout is what console CI tests set.
+	// If this test is flaky, we should increase until
+	// we can fix the possible DNS resolve issues.
+	f.ThanosQuerierClient.WaitForRulesReturn(
+		t, 2*time.Minute,
+		func(body []byte) error {
+			return getThanosRules(body, "general.rules", "Watchdog")
+		},
+	)
+}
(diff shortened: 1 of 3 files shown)


lilic:
/retest

some gke issues...

lilic:
Irony of there being DNS errors in the logs of cluster setup 🤣 

/retest

lilic:
Strange Prometheus discover problems 🤔 

/retest

lilic:
/retest

...

test/e2e/framework/framework.go
```diff
@@ -138,6 +138,12 @@ func New(kubeConfigPath string) (*Framework, cleanUpFunc, error) {
 		return nil, nil, err
 	}
 
+	// Setup debug log level in ThanosQuery.
+	err = f.setupDebug()
```
simonpasquier, line 142: The setting won't be permanent because other tests override the CMO configuration (`TestPrometheusVolumeClaim` for instance). Is it really needed?

lilic: I can move it closer to where I would like to see debug logs in this case where we query watchdog alert, but yes true it would be overridden.

Not sure I understand what you mean by "Is it really needed?"?

simonpasquier: I meant to ask if we always need to run with debug log level. I guess it's for troubleshooting when the test fails?

lilic: Yes, exactly for that test only. I haven't checked if they are run in parallel as to one to override this configmap? If not will just create configmap in this test and wait for thanos query pod to be rolled out. My only fear is, this might affect the test, as restarting the pods is not the same as happens on a regular cluster and we might not catch addr resolve issues as it would not be running for long enough. 🤔 Any ideas/suggestions welcome!

simonpasquier: yeah, not sure what is best... Running always with debug log level might hide issues (I've seen so many bugs that don't manifest when you enable debug logs).

lilic: Okay will remove the debug enabling, if we see the failures more often we can add it later.

lilic:
@simonpasquier PTAL thanks!

simonpasquier (commented):
/lgtm

PR #999, opened 2020-12-01
Bug 1899582: Increase rest config burst and QPS rate limits

dgrisonnet:
Increase rest config QPS value to reflect what is set in upstream
prometheus-operator.
Also, it was reported that CMO is producing a lot of throttling error
logs when it is run in an environnement with a lot CRDs. To fix that, we
can align with the value set in other component that encountered the
same issue. For instance, in `oc` the burst value was set to 200.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -155,9 +155,13 @@ func Main() int {
 		return 1
 	}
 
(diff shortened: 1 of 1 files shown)


simonpasquier:
/hold
waiting for #985

simonpasquier (commented):
/lgtm

PR #1005, opened 2020-12-04
Bug 1892706: fix cluster:cpu_core_hyperthreading rule for s390x

simonpasquier:

Backport of #953 (it couldn't be cherry-picked automatically by openshift-robot because of `pkg/manifests/bindata.go`)


lilic (commented):
/lgtm

PR #1008, opened 2020-12-07
Bug 1904985: fix TLS secrets for Thanos sidecars

simonpasquier:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ export GOPROXY
 
 PKGS=$(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
 GOLANG_FILES:=$(shell find . -name \*.go -print)
-ASSETS=$(shell grep -oh 'assets/.*\.yaml' pkg/manifests/manifests.go)
+ASSETS=$(shell grep -oh '[^"]*/.*\.yaml' pkg/manifests/manifests.go | sed 's/^/assets\//')
 
 BIN_DIR ?= $(shell pwd)/tmp/bin
 
--- a/assets/prometheus-k8s/prometheus.yaml
+++ b/assets/prometheus-k8s/prometheus.yaml
@@ -136,7 +136,7 @@ spec:
     terminationMessagePolicy: FallbackToLogsOnError
     volumeMounts:
     - mountPath: /etc/tls/private
-      name: secret-prometheus-k8s-tls
+      name: secret-prometheus-k8s-thanos-sidecar-tls
   - args:
     - sidecar
     - --prometheus.url=http://localhost:9090/
@@ -180,6 +180,7 @@ spec:
   - prometheus-k8s-tls
   - prometheus-k8s-proxy
   - prometheus-k8s-htpasswd
+  - prometheus-k8s-thanos-sidecar-tls
   - kube-rbac-proxy
   securityContext:
     fsGroup: 65534
--- a/assets/prometheus-k8s/service-monitor-thanos-sidecar.yaml
+++ b/assets/prometheus-k8s/service-monitor-thanos-sidecar.yaml
@@ -13,7 +13,7 @@ spec:
     scheme: https
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
-      serverName: prometheus-k8s
+      serverName: prometheus-k8s-thanos-sidecar
   selector:
     matchLabels:
       app: thanos-sidecar
--- a/assets/prometheus-k8s/service-thanos-sidecar.yaml
+++ b/assets/prometheus-k8s/service-thanos-sidecar.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   annotations:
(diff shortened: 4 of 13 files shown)


simonpasquier:
/hold

I'm looking why it wasn't caught in tests.

Makefile
```diff
@@ -12,7 +12,7 @@ export GOPROXY
 
 PKGS=$(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
 GOLANG_FILES:=$(shell find . -name \*.go -print)
-ASSETS=$(shell grep -oh 'assets/.*\.yaml' pkg/manifests/manifests.go)
+ASSETS=$(shell grep -oh '[^"]*/.*\.yaml' pkg/manifests/manifests.go | sed -s 's/^/assets\//')
```
simonpasquier, line 15: @paulfantom it seems that the Makefile doesn't detect properly when to regenerate asset files now. I'm trying this but open to better ideas :)

simonpasquier:
The root cause is that 2 services (prometheus-k8s and prometheus-k8s-thanos-sidecar) used the same value for the `service.beta.openshift.io/serving-cert-secret-name` annotation (it was set to `prometheus-k8s-tls`) but the 2 associated service monitors always reference `prometheus-k8s` as the server's name.
Depending on which service was picked up by the certificate generation service, the resulting TLS certificate would either be valid for:
* `prometheus-k8s.openshift-monitoring.svc` (which would be ok) 
* `prometheus-k8s-thanos-sidecar.openshift-monitoring.svc` (which would fail because Prometheus detected a server name mismatch).

Digging into CI history for #993, I've found one origin e2e failure because `TargetDown` fired for Thanos sidecards and Prometheus:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/993/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic/1334787256627499008

But openshift-robot retried and it eventually passed.

Makefile
lilic: Let's see if this work on non linux env, sed tends to be different from the gnu sed on Linux. Will test and report back.

lilic (commented):
/retest

s-urbaniak:
(/)lgtm but leaving the final pass to @lilic to verify if the BSD sed toolchain is compatible with the changes here.

Makefile
lilic: Yes as I suspected, this does not work on mac:
```
 ➜  cluster-monitoring-operator git:(86d11142) ✗ make generate
sed: illegal option -- s
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
sed: illegal option -- s
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
/Users/lili/go/src/github.com/openshift/cluster-monitoring-operator/tmp/bin/embedmd -w `find Documentation -name "*.md"
````

I am fine with merging it now as it fixes a bug, but we should follow up with a fix to make it work in all env.

**Comment by lilic** [comment]
/retest

/lgtm
/hold

Up to you @simonpasquier if you want to adjust now, or in a follow up, both sound good to me!

File: Makefile (hunk 1 above)
Comment (simonpasquier) [reply Makefile]: Ho I don't need the `-s` option (here you can hint that I've copied the first answer from Stackoverflow...). I've updated the change and it should be compatible with all seds hopefully.

File: Makefile [hunk 2]
```diff
@@ -12,7 +12,7 @@ export GOPROXY
 
 PKGS=$(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
 GOLANG_FILES:=$(shell find . -name \*.go -print)
-ASSETS=$(shell grep -oh 'assets/.*\.yaml' pkg/manifests/manifests.go)
+ASSETS=$(shell grep -oh '[^"]*/.*\.yaml' pkg/manifests/manifests.go | sed 's/^/assets\//')
```
lilic, line 15: Nice worked for me now 🎉

lilic (commented):
/lgtm

Thanks!

Makefile
s-urbaniak: I _knew_ BSD and GNU can peacefully coexist together :blush:

sjenning:
@simonpasquier can we remove the hold on this while you investigate the tests situation?

We are seeing consistent failures on the following 4.7 CI tests that this PR will hopefully fix (some or all)

```
[sig-instrumentation] Prometheus when installed on the cluster should report telemetry if a cloud.openshift.com token is present [Late] [Suite:openshift/conformance/parallel]

[sig-instrumentation] Prometheus when installed on the cluster should start and expose a secured proxy and unsecured metrics [Suite:openshift/conformance/parallel] 

[sig-instrumentation] Prometheus when installed on the cluster shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured [Early] [Suite:openshift/conformance/parallel]
```

sjenning:
talking to @lilic I think we can remove the hold.  We need to get this in ASAP for 4.7 CI.
/hold cancel

PR #1009, opened 2020-12-07
MON-1302: Bump prometheus-operator and k8s golang dependencies

dgrisonnet:
Bump prometheus-operator golang dependency to v0.44.0 and Kubernetes dependencies to v0.19.4.
This effort is part of the 4.7 feature freeze checklist: #1007 


go.mod
```diff
@@ -1,34 +1,34 @@
 module github.com/openshift/cluster-monitoring-operator
 
-go 1.14
+go 1.15
```
lilic, line 3: Why do we need this, does something require 1.15 go version?

simonpasquier:
/hold
needs #1010 first

go.mod
```diff
@@ -16,19 +16,19 @@ require (
...
 )
 
 replace (
-	k8s.io/api => k8s.io/api v0.19.2
-	k8s.io/apimachinery => k8s.io/apimachinery v0.19.2
-	k8s.io/client-go => k8s.io/client-go v0.19.2
+	k8s.io/api => k8s.io/api v0.19.4
+	k8s.io/apimachinery => k8s.io/apimachinery v0.19.4
+	k8s.io/client-go => k8s.io/client-go v0.19.4
 	k8s.io/code-generator => k8s.io/code-generator v0.0.0-20190912054826-cd179ad6a269
```
simonpasquier, line 33: not directly related to this PR but I'm wondering why we have this replace directive?

```
$ go mod why -m k8s.io/code-generator
# k8s.io/code-generator
(main module does not need module k8s.io/code-generator)
```

go.mod
dgrisonnet: AFAIK, we don't particularly need it, I just bumped to 1.15  to keep the version up to date. Also, upstream projects are already using 1.15, so I don't think CMO should lack behind.
Do you have any concerns with this upgrade? I can remember past discussions about this topic but after double checking, 1.15 should be available starting from 4.6.

go.mod
dgrisonnet: I am not too sure either, I'll investigate

dgrisonnet: good catch! I removed it for the following reason:

The k8s.io/code-generator replace directive was inherited from
prometheus-operator. This directive has now been removed upstream and
the dependency was moved to the tooling module of prometheus-operator.
Since we don't use it in CMO, we can safely remove it.

go.mod
lilic: This does not bump the version, just states the minimum required version this project needs to compile its deps. If there is no need to require the bump to 1.15, I would say lets use 1.14 instead.

dgrisonnet:
@lilic @simonpasquier I updated this PR, please take another look at it.

simonpasquier (commented):
/lgtm

PR #1011, opened 2020-12-07
Update release checklist template

dgrisonnet:

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -8,6 +8,8 @@ about: Create release checklist
 ## Golang projects and backports
 
 - [ ] update downstream backport of [openshift/node_exporter](https://github.com/openshift/node_exporter)
+  - [ ] bump [openshift/procfs](https://github.com/openshift/procfs) to the version used in the node_exporter's upstream version we want to bump to, including any downstream patch we may have
+  - [ ] bump [openshift/node_exporter](https://github.com/openshift/node_exporter) to the upstream version replacing the procfs dependency by the downstream repo
 - [ ] update downstream backport of [openshift/prometheus](https://github.com/openshift/prometheus)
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/release.md
```diff
@@ -16,6 +16,8 @@ about: Create release checklist
 - [ ] update downstream backport of [openshift/kube-rbac-proxy](https://github.com/openshift/kube-rbac-proxy)
 - [ ] update downstream backport of [openshift/k8s-prometheus-adapter](https://github.com/openshift/k8s-prometheus-adapter)
 - [ ] update downstream backport of [openshift/grafana](https://github.com/openshift/grafana)
+- [ ] update downstream backport of [openshift/prom-label-proxy](https://github.com/openshift/prom-label-proxy)
+- [ ] update downstream backport of [openshift/procfs](https://github.com/openshift/procfs)
```
s-urbaniak, line 20: good points :+1:

dgrisonnet:
@openshift/openshift-team-monitoring PTAL

.github/ISSUE_TEMPLATE/release.md
simonpasquier, line 20: This repository is a dependency of `github.com/openshift/node_exporter` so it should probably go with the `node_exporter` instructions. The workflow would be:
1. Bump `openshift/procfs` to the version used in the node_exporter's upstream version we want to bump to, including any downstream patch we may have.
2. Bump `openshift/node_exporter` to the upstream version replacing the `procfs` dependency by the downstream repo.

simonpasquier:
we need also to bump versions in `hack/tools/go.mod` (especially prometheus version should be synced with the main `go.mod`). Noticed while looking at #1020 which fails because it uses the `group` aggregator that isn't supported by the `promtool` version pulled from `hack/tools/go.mod`.

simonpasquier (commented):
/lgtm

PR #1017, opened 2020-12-10
Bug 1906570: Mount /var/log/wtmp into node_exporter init container

smarterclayton:
The init container can now view wtmp so the new node_boots metric
can be read via a text collector. This exposes no additional security
risk except potentially allowing the node_exporter init processes
to see the users who have been logged into the system during startup.

Used by https://github.com/openshift/node_exporter/pull/74


simonpasquier (commented):
/lgtm

PR #1019, opened 2020-12-11
Bug 1907475: Add recording rules for ingress traffic and error rate

smarterclayton:
The overall rates of ingress traffic to the cluster are relevant in
assessing health and usage - these rules add calculations for the
rate of requests that error (5xx and other types of errors) across
the frontends, the total requests passing across the frontends, the
bandwidth used in both directions, and the number of open connections.
Error rate is split for openshift and workloads (like CPU is) in order
to better categorize trends before, during, and after upgrades.

This data is relevant for both admins and engineering teams to classify
and categorize the overall health and traffic of the cluster. This adds
seven new series to telemetry via the cluster:usage: prefix.

### DIFF
--- a/assets/prometheus-k8s/rules.yaml
+++ b/assets/prometheus-k8s/rules.yaml
@@ -924,12 +924,26 @@ spec:
     rules:
     - expr: sum by (code) (rate(haproxy_server_http_responses_total[5m]) > 0)
       record: code:cluster:ingress_http_request_count:rate5m:sum
-    - expr: sum by (frontend) (rate(haproxy_frontend_bytes_in_total[5m]))
-      record: frontend:cluster:ingress_frontend_bytes_in:rate5m:sum
-    - expr: sum by (frontend) (rate(haproxy_frontend_bytes_out_total[5m]))
-      record: frontend:cluster:ingress_frontend_bytes_out:rate5m:sum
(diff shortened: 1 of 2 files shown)


smarterclayton:
@Miciah can you or someone else on the team take a look?  I want to quantify "happy" and "non-disrupted" ingress (this won't catch the nodeport / local traffic problem, but it will catch lower level problems like ovs)

Miciah:
@sgreene570, could you take a look?  @RiRa12621, I figure you might be interested in well.

RiRa12621:
Thanks for letting me know @Miciah .
Had a chat with Clayton about it before already and it's *only* recording rules for now so that's fine from my perspective.

smarterclayton:
Can one of the network edge reviewers confirm these make sense to capture given the reasoning in the bug and PR?  Just want to make sure before I get final review on the rule structure.

Miciah:
> Can one of the network edge reviewers confirm these make sense to capture given the reasoning in the bug and PR? Just want to make sure before I get final review on the rule structure.

Yeah, it makes sense to me.

/lgtm

quarterpin:
/lgtm

Verified via pre-merge verification workflow, more reference related to the test can be found in:
https://bugzilla.redhat.com/show_bug.cgi?id=1907475#c1

PR #1020, opened 2020-12-11
Bug 1908562: Add a recording rule measuring pod readiness

smarterclayton:
For all pending, running, or unknown pods report a 0 if the pod is
unready and a 1 if a pod ready. Use that to build two aggregate
average readiness fractions for openshift-* and !openshift-*
namespace pods, which we will use to measure impact of upgrades
on pod availability before, during, and after the change.

The fundamental rule can be used in the future for alerting when
the aggregate readiness drops below a threshold after we identify
a minimum bar.

This adds 2 series to telemetry for monitoring the impact of
upgrades.

### DIFF
--- a/assets/prometheus-k8s/rules.yaml
+++ b/assets/prometheus-k8s/rules.yaml
@@ -920,6 +920,15 @@ spec:
       for: 15m
       labels:
         severity: info
+    - expr: (max without (condition,container,endpoint,instance,job,service) (((kube_pod_status_ready{condition="false"}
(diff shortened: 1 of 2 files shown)


simonpasquier:
lgtm but the unit tests fail because we're using an outdated version of prometheus in `hack/tools/go.mod`. Since @dgrisonnet is bumping versions for all the things in CMO, he agreed to take care of it.

dgrisonnet:
I opened #1021 to bump the outdated prometheus.

lilic (commented):
/lgtm

PR #1023, opened 2020-12-15
Bug 1899839: jsonnet/thanos-ruler.jsonnet: Set resources to the native spec field

lilic:
The user configuration was not propagated correctly as we were modifying the container resources via merging directly, but thanos ruler CR exposes this configuration so we should use that instead.

Opening PR to test easier, but plan on adding a test for this as well in our e2e UWM stack.


lilic:
@openshift/openshift-team-monitoring PTAL thanks!

PR #1025, opened 2020-12-18
WIP: assets/node-exporter: SCC should be selinux MustRunAs

smarterclayton:
In Kubernetes an empty dir shared between two containers running at
different privilege levels (like the init container and regular node
exporter container do) without an explicit selinux label will end
up with two different labels, and content cannot be automatically
shared between the volumes. This may be a bug with the Kubelet (some
work is ongoing to address the issue) but the immediate and general
workaround is to have the node-exporter SCC use the `MustRunAs` SCC
policy so the default selinux context of all pods defaults to the
namespace policy. This results in the pod having an explicit selinux
label and both init container and regular container having the same
labels, which means the regular container can read files created by
the init container in the emptydir.

* [x]  No user facing changes, so no entry in CHANGELOG was needed.

### DIFF
--- a/assets/node-exporter/security-context-constraints.yaml
+++ b/assets/node-exporter/security-context-constraints.yaml
@@ -13,5 +13,5 @@ readOnlyRootFilesystem: false
 runAsUser:
   type: RunAsAny
 seLinuxContext:
-  type: RunAsAny
+  type: MustRunAs
 users: []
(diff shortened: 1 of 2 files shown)


s-urbaniak:
@smarterclayton it seems e2e tests pass green, is there anything else left to do here?

smarterclayton:
/hold

Even though this may still be right, Im not sure about all of the implications on upgrade.  I was able to workaround the issue that blocked node boots from being counted without this (using /tmp between the init container and container) so it’s not critical.  Will hold this for now and come back if we hit this again

PR #1032, opened 2021-01-11
Bug 1915003: Add a rule for calculating rolling node readiness

smarterclayton:
The number of schedulable nodes reporting ready is a key measurement
of the disruption upgrades incur as well as being an effective way
to assess whether our software is available on the hardware users
provide. Report a rolling average of node readiness measured across
all schedulable nodes and one across all nodes, using both readiness
and reachability. Also report a more general node readiness average
which matches the API view and should be higher than reachability.

This info will help assess more accurately how often we are ready
and identify future areas to improve in both reporting and practice.

### DIFF
--- a/jsonnet/rules.jsonnet
+++ b/jsonnet/rules.jsonnet
@@ -317,6 +317,19 @@ local droppedKsmLabels = 'endpoint, instance, job, pod, service';
             },
           },
           {
+            expr: 'avg_over_time((((count((max by (node) (up{job="kubelet",metrics_path="/metrics"} == 1) and max by (node) (kube_node_status_condition{condition="Ready",status="true"} == 1) and min by (node) (kube_node_spec_unschedulable == 0))) / scalar(count(min by (node) (kube_node_spec_unschedulable == 0))))))[5m:1s])',
+            record: 'cluster:usage:kube_schedulable_node_ready_reachable:avg5m',
+            # Report a 5m rolling average of the number of schedulable nodes that are ready and reachable to be scraped by metrics. This is
+            # used to estimate the disruption imposed on a cluster during upgrades, excluding any machines administrators may be performing
+            # maintenance on.
+          },
+          {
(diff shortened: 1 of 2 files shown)


jsonnet/rules.jsonnet
```diff
@@ -316,6 +316,19 @@ local droppedKsmLabels = 'endpoint, instance, job, pod, service';
               severity: 'info',
             },
           },
+          {
+            expr: 'avg_over_time((((count((max by (node) (up{job="kubelet",metrics_path="/metrics"} == 1) and max by (node) (kube_node_status_condition{condition="Ready",status="true"} == 1) and min by (node) (kube_node_spec_unschedulable == 0))) / scalar(count(min by (node) (kube_node_spec_unschedulable == 0))))))[5m:1s])',
```
lilic, line 320: Why is an average the important key here? This query gives me result 1 on a cluster that has 10 nodes with 1 unready, is that correct?

Just curious about it that's all :D

lilic (commented):
One question, but not blocking, just curious!

/lgmt
/hold

Feel free to remove hold if you think that is the correct result of that query.

jsonnet/rules.jsonnet
smarterclayton: So in the last 5m if you had 1 node out of 10 unready for 1m, you'd get 0.9 + 0.2 (1 node unready for 1m out of 5).  If you scaled down and the node didn't go unready, you'd have 1.  We can estimate over time what the percent downtime was, but also work backward from node count to estimate how many nodes were down.

Until we move to direct write, the average also gives us a better rolling window of downtime with the current scrape-every-5m model - when I modelled it as instantaneous average we overreported the real average.

smarterclayton:
/hold cancel

(rolling average ensures we don't miss downtime, using average to get a precise rate when there are no samples - 0 nodes = 0 uptime)

PR #1035, opened 2021-01-12
Bug 1916056: manifest allowlist: add ACM metric

mark-nc:
Add Red Hat Advanced Cluster Management for Kubernetes metric visual_web_terminal_sessions_total to the manifest allowlist


Documentation/data-collection.md
```diff
@@ -245,6 +245,9 @@ data:
     # (@openshift/storage, OpenShift Storage team) Persistent Volume usage metrics: this is the number of volumes per plugin
     # and per volume type (filesystem/block)
     - '{__name__="cluster:kube_persistentvolume_plugin_type_counts:sum"}'
+    # (Advanced Cluster Management) visual_web_terminal_sessions_total is the count of Visual Web Terminal sessions created
```
lilic, line 248: Do you not have a team under openshift org we can ping in case we need to change/remove anything?

Documentation/data-collection.md
```diff
@@ -245,6 +245,9 @@ data:
     # (@openshift/storage, OpenShift Storage team) Persistent Volume usage metrics: this is the number of volumes per plugin
     # and per volume type (filesystem/block)
     - '{__name__="cluster:kube_persistentvolume_plugin_type_counts:sum"}'
+    # (Advanced Cluster Management) visual_web_terminal_sessions_total is the count of Visual Web Terminal sessions created 
+    # on the hub cluster.
+    - '{__name__="visual_web_terminal_sessions_total"}'
```
lilic, line 250: What is the cardinality for this in total for this metric, I could not find the metric `visual_web_terminal_sessions_total` via search, do you mind sharing the repo where its created?

cc @smarterclayton I was told they have your 👍 already

lilic (commented):
Thanks couple of questions from my side

Documentation/data-collection.md
mark-nc: I will have to check.  I think all of our Github teams (if that is what you mean) are under our `open-cluster-management` org.

Documentation/data-collection.md
mark-nc: This metric has no labels so it will be a singleton per hub OCP cluster where ACM is installed.  So from a telemetry standpoint, there will be one occurrence of this counter for every ACM hub cluster participating in telemetry.

The metric is created in this repo here https://github.com/open-cluster-management/kui-web-terminal/blob/master/proxy/app/lib/metricsUtils.js

Documentation/data-collection.md
lilic: Yes we prefer a GitHub team if possible.

mark-nc: OK, let me get the right one and I will update comment in the yaml with that.

lilic (commented):
/lgtm

Clayton already approved on a google doc.

PR #1037, opened 2021-01-13
Bug 1905141: Add vsphere-problem-detector to telemetry

jsafrane:
*This is not 4.7 blocker, still we'd like to have it in 4.7 either now or  when 4.7.z opens.*

vsphere-problem-detector regularly checks configuration of vSphere / vCenter where the OCP cluster runs and reports its issues. [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/storage/vsphere-problem-detector.md)

We want in telemetry:

* vCenter version, `cluster:vsphere_vcenter_info:sum`
  * Cardinality: always 1
  * Sample:
    ```
    {version="7.0.0"} 1
    ```
 
* ESXi (=host) versions, `cluster:vsphere_esxi_version_total:sum`
  * Future OCP / vSphere CSI driver versions may require newer ESXi versions and we want to know if it's an issue.
  * Cardinality: one series for each ESXi host version in the cluster
    * Typically 1.
    * It may grow in heterogeneous vSphere clusters (when 6.5, 6.7 and 7.0 hosts or various patch versions are mixed in one cluster).
    * Expected to double during vSphere cluster upgrades (e.g. from 7.0.0 to 7.0.1 - both version will be observed until all hosts are upgraded).
  * Sample:
    ```
    {version="7.0.0"} 2
    ```
 
 * VM HW versions, `cluster:vsphere_node_hw_version_total:sum`
   * HW version = (very roughly) version of vmdk file of a VM.
   * vSphere CSI driver requires newer VM HW versions than we use today and we want to know what users use.
   * Cardinality: one series for each VM HW version in the cluster.
     * Typically one.
     * Expected to be 2 in some clusters (OCP creates VMs with HW version 13 right now, users may update them to 15 to use upstream CSI driver).
     * It may grow in crazy UPI environments, but very unlikely. Minimum HW version supported by OCP is 13, max supported by vSphere is 18, so 6 series max.
   * Sample:
     ```
     {hw_version="vmx-13"} 6
     ```
All samples were collected on 6 node cluster in our vSphere test environment (VMC, i.e. hosted vSphere 7.0.0 on AWS)

cc @smarterclayton for new telemetry approval


smarterclayton:
Do any of these trigger alerts?  Why or why not?

jsafrane:
Alerts are on the todo list: https://bugzilla.redhat.com/show_bug.cgi?id=1904503

Hmm, should they be enough for the "List of failed {node,cluster}-level checks"?

smarterclayton:
If the alert is actionable with docs, I might recommend reporting them as alerts.  If you're not sure they're actionable yet, then returning them via telemetry only for now is ok if you plan to assess that first.  Are the checks exposed elsewhere yet (UI or cluster operator status)?

jsafrane:
Checks are exposed only as events + metrics (+ documented). Alerts are in progress, but unlikely to catch 4.7.0.

smarterclayton:
/approve

I approve the cardinality increase of ~10 on vsphere clusters, with the understanding in the future we may wish to fold some of these into alerts as we gain confidence or other constructs like cluster operators.  The impact to overall telemetry load should be minimal given the fleet composition.

jsafrane:
Update: removed list of failed cluster or node checks, we will get them via alerts. https://github.com/openshift/cluster-storage-operator/pull/126

jsafrane:
@openshift/openshift-team-monitoring, PTAL

lilic (commented):
/lgtm

PR #1046, opened 2021-02-02
Bug 1913386: allow only /metrics for kube-rbac-proxy in front of UWM prometheu

simonpasquier:


lilic (commented):
/lgtm

PR #1052, opened 2021-02-15
[WIP] Remove the "instance" label for kube-state-metrics metrics

simonpasquier:


s-urbaniak:
/lgtm
(just in case)

jsonnet/kube-state-metrics.jsonnet
```diff
@@ -33,6 +33,25 @@ local tlsVolumeName = 'kube-state-metrics-tls';
...
                 serverName: 'server-name-replaced-at-runtime',
               },
+              // Drop the "instance" and "pod" labels since we're runinng only
+              // one instance of kube-state-metrics. The "instance" label can't
+              // be dropped at the discovery stage because otherwise Prometheus
+              // will default its value to the address being scraped.
+              // This is to avoid excessive series churn when
+              // kube-state-metrics is redeployed because of node reboot or
+              // upgrade.
+              metricRelabelings: [
```
lilic, line 43: This also is CPU expensive as it needs to process every single series in every single scrape, do we think we are getting enough out of it, is something I feel we should consider.

jsonnet/kube-state-metrics.jsonnet
```diff
@@ -33,6 +33,25 @@ local tlsVolumeName = 'kube-state-metrics-tls';
                 caFile: '/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt',
                 serverName: 'server-name-replaced-at-runtime',
               },
+              // Drop the "instance" and "pod" labels since we're runinng only
+              // one instance of kube-state-metrics. The "instance" label can't
+              // be dropped at the discovery stage because otherwise Prometheus
+              // will default its value to the address being scraped.
+              // This is to avoid excessive series churn when
```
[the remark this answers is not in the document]
simonpasquier: Yes it may happen when the kube-state-metrics pod's IP address changes (e.g. on restart/redeploy) but it should be rare. The normal path should be:
1. Prometheus detects that the old pod is gone, stops the scrape loop and waits for (2.1 x scrape interval) before inserting the stale markers.
2.  Prometheus creates a new scrape loop for the new kube-state-metrics pod's address.
3. At the next scrape interval, Prometheus scrapes the new target before the stale markers get inserted.
4. Prometheus tries to insert stale markers for the old loop but fails with an out-of-order error because series have already been inserted at stage 3.

If the new scrape loop doesn't complete before step 4 then it will fail because of out-of-order samples.

jsonnet/kube-state-metrics.jsonnet
simonpasquier: I'm going to check locally but I don't expect that it would cost us much CPU-wise.

simonpasquier:
The PR avoids creating about 70,000 series based on the data gathered from e2e-upgrade job.

Graph of `prometheus_tsdb_head_series` for this PR
image

Same metric for de7bf906764d16f2a73975f2d657b6810b884b9b
image

simonpasquier:
I've checked with a fresh 4.8 cluster and compared the CPU before/after the PR. There is no significant difference between them CPU wise. I've taken CPU profiles of Prometheus around the time when it scrapes kube-state-metrics but I couldn't find traces of the relabeling function.

image

The PR was tested between 15:50 and 16:20

2 other side effects with this PR:
* it seems that we do less memory allocations in [mutateSampleLabels()](https://github.com/openshift/prometheus/blob/66f4fc6b2ec60fe577de9caab1e5bde264bc8b88/scrape/scrape.go#L539-L567) (more specifically in labels.Labels()) which means less work for GC and less memory pressure. My explanation would be that Prometheus doesn't have to add the "exported" label for `pod` so the [pre-allocated slice](https://github.com/openshift/prometheus/blob/66f4fc6b2ec60fe577de9caab1e5bde264bc8b88/pkg/labels/labels.go#L453-L455) for the output labels has always the right capacity and it doesn't have to be reassigned.
* the scrape duration is a bit shorter: 100ms vs 120-120ms on average
image

paulfantom:
/hold

holding to allow merging #1044 first

paulfantom:
/lgtm

maybe we should consider moving this to kube-prometheus :thinking:

PR #1054, opened 2021-02-16
pkg/manifests: fix data race when accessing assets

simonpasquier:

cc @paulfantom I've got a `fatal error: concurrent map read and map write` panic while testing the operator locally.


paulfantom (approved):
/lgtm

PR #1057, opened 2021-02-16
Bug 1929354: [4.6] Adjust priority class of core platform prometheus

bparees:


lilic (commented):
/lgtm

Thank you 🙇‍♀️

lilic:
You need to run the generate command in a different way in 4.6 sadly and commit bindata changes as well.

bparees:
/hold

needs to be aligned w/ the changes to 4.7 and it sounds like we may not need this urgently for 4.7 anyway since the component upgrades happen before the node reboots.

PR #1064, opened 2021-02-18
Bug 1929944: etcdInsufficientMembers is wrong when etcd is in a pod

smarterclayton:
The upstream etcd alert is incorrect because it only excludes instance labels, but OpenShift runs etcd in a pod and therefore the pod label must be excluded.

Exclude the upstream alert, improve the resiliency of the alert expression, target the alert to the expected job for the cluster etcd
(job="etcd"), update the description and health text to include a clearer description of what insufficient members means and consequences and some impact actions, and separate the alert into its own rule group to prepare (in the future) of moving the alert into the cluster-etcd-operator repo.

### DIFF
--- a/assets/prometheus-k8s/rules.yaml
+++ b/assets/prometheus-k8s/rules.yaml
@@ -961,6 +961,24 @@ spec:
         offset 25s) or (absent(cluster:usage:workload:capacity_physical_cpu_core_seconds
         offset 25s)*0))
       record: cluster:usage:workload:capacity_physical_cpu_core_seconds
+  - name: openshift-etcd.rules
+    rules:
+    - alert: etcdInsufficientMembers
+      annotations:
+        message: etcd is reporting fewer instances are available than are needed ({{
+          $value }}). When etcd does not have a majority of instances available the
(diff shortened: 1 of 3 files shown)


s-urbaniak:
/approve
/cc @hexfusion 
for another pair of eyes

hexfusion:
thanks running a few quick tests

hexfusion:
I like this so we are comparing members reporting have leader

```
sum(up{job="etcd"} == bool 1 and etcd_server_has_leader{job="etcd"} == bool 1) without (instance,pod)
```
against the majority
```
((count(up{job="etcd"}) without (instance,pod) + 1) / 2)
```
so that if a quorum has been lost for more than 3 minutes we fire a critical alert.

I guess my only concern here is 3m is a really long time to lose quorum. We probably want to know if we ever lose quorum for more than 30s as a warn? Leader elections are very fast and if we have functional issues with actions the cluster is taking that result in quorum loss they are bugs. What do you think?

hexfusion:
I think this is progress though I will try to get a grasp on how often we might see quorum loss in the wild and can followup with warn thresholds.

/lgtm

hexfusion:
I will grab the changelog

PR #1065, opened 2021-02-19
jsonnet/rules: Mask KubeDeploymentReplicasMismatch alert for upgrade

smarterclayton:
The KubeDeploymentReplicasMismatch alert fires after 15m, but deployments that run on all three control plane nodes may see disruption that lasts for 3 x 20m = 1h during upgrades of slow to reboot metal hardware. Instead, prevent the alert rule from firing when a control plane nodes are unready (which should be covered by other alerts). While this may mask failures that occur on non control plane nodes, once the control plane is recovered that alert will begin to fire. Provide the alert with a more accurate description.

A future alert may be able to leverage a better metric such as "1 if and only if the control plane is not being upgraded and has been progressing for less than X budget per machine", but that would only be slightly less accurate and simpler may be better.

Found while testing https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/1064/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-upgrade/1362412232272515072 - this is the only alert still firing during this upgrade and will allow https://github.com/openshift/origin/pull/25904 to pass.

### DIFF
--- a/assets/prometheus-k8s/rules.yaml
+++ b/assets/prometheus-k8s/rules.yaml
@@ -908,6 +908,10 @@ spec:
       record: cluster:vsphere_esxi_version_total:sum
     - expr: sum by(hw_version)(vsphere_node_hw_version_total)
       record: cluster:vsphere_node_hw_version_total:sum
+    - expr: absent(count(max by (node) (kube_node_role{role="master"}) and (min by
+        (node) (kube_node_status_condition{condition="Ready",status="true"} == 0)))
+        > 0)
+      record: cluster:control_plane:all_nodes_ready
     - alert: ClusterMonitoringOperatorReconciliationErrors
       annotations:
         message: Cluster Monitoring Operator is experiencing reconciliation error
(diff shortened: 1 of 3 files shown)


assets/prometheus-k8s/rules.yaml
```diff
@@ -927,6 +931,31 @@ spec:
...
+        (
+          kube_deployment_spec_replicas{namespace=~"(openshift-.*|kube-.*|default|logging)",job="kube-state-metrics"}
+            !=
+          kube_deployment_status_replicas_available{namespace=~"(openshift-.*|kube-.*|default|logging)",job="kube-state-metrics"}
+        ) and (
+          changes(kube_deployment_status_replicas_updated{namespace=~"(openshift-.*|kube-.*|default|logging)",job="kube-state-metrics"}[5m])
+            ==
+          0
+        ) and cluster:control_plane:all_nodes_ready
+      for: 15m
```
michaelgugino, line 956: We should set 60m to allow for MHC to remediate.

PR #1068, opened 2021-02-25
Bug 1929056: Remove kube-apiserver-availability.rules

dgrisonnet:
The recording rules coming from kubernetes-mixin `kube-apiserver-availability.rules` are only used for the apiserver availability Grafana dashboard that is disabled downstream. Previously, we only disabled the problematic `apiserver_request:availability30d` because it was the only rule for which evaluation errors were reported, but now that we were reported the same issues with the underlying rules, we should just remove all of them.

On a side note, it might be a good idea at some point to start iterating over disabled Grafana dashboards and remove unused recording rules. 


simonpasquier (commented):
/lgtm

PR #1084, opened 2021-03-18
jsonnet: fix etcd mixin import

dgrisonnet:


dgrisonnet:
We still import the etcd mixin here: https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/control-plane/etcd-prometheus-rule.yaml

It might be worth keeping the import in case we need to manually pin the version of the mixin.

/hold

PR #1090, opened 2021-03-24
Bug 1921335: Fix and adjust ThanosSidecarUnhealthy alert

dgrisonnet:
Bring the following upstream fix https://github.com/thanos-io/thanos/commit/95b2fe908b29977a20f695afac33f6257bb2d3ea that should prevent the ThanosSidecarUnhealthy alert from firing in CI.

Also readjust the Thanos sidecar alerts to `warning` severity and their duration to `1h` as per the recent discussions around alerting in OCP.

### DIFF
--- a/assets/prometheus-k8s/prometheus-rule.yaml
+++ b/assets/prometheus-k8s/prometheus-rule.yaml
@@ -239,24 +239,34 @@ spec:
       for: 15m
       labels:
         severity: critical
-  - name: thanos-sidecar.rules
+  - name: thanos-sidecar
     rules:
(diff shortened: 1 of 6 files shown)


dgrisonnet:
Note that this PR should fulfill all the requirements of the Bugzilla, except making the Thanos sidecar alerts more resilient to WAL replays as this is still in progress. Should I open a separate BZ just for this effort?

s-urbaniak:
> Should I open a separate BZ just for this effort?
Yes i think this is a good idea! :+1: 
/lgtm

PR #1094, opened 2021-03-29
Bug 1943667: increase for duration of KubeDaemonSetRolloutStuck

simonpasquier:


smarterclayton:
/retest

Rough day

PR #1096, opened 2021-03-29
WIP: pkg/manifests: enforce alert_source external label

dgrisonnet:
WIP PR to experiment around enforcing an external label that could be used to route alerts based on where they are coming from.


pkg/manifests/manifests.go
```diff
@@ -3227,6 +3231,8 @@ func (f *Factory) ThanosRulerCustomResource(queryURL string, trustedCA *v1.Confi
 
 	t.Namespace = f.namespaceUserWorkload
 
+	t.Spec.Labels = map[string]string{"alert_source": "user"}
```
simonpasquier, line 3234: Looking at the prometheus-operator code [1], we need to inject the `thanos_ruler_replica` label as well as the label drop directive for it. 

[1] https://github.com/prometheus-operator/prometheus-operator/blob/1f735d7089b41e6763180b6c4029191989433518/pkg/thanos/statefulset.go#L186-L193

dgrisonnet: good catch :+1:

dgrisonnet:
Holding until we give more thought to this feature.

/hold

PR #1102, opened 2021-04-01
Bug 1945677: manifest allowlist: add acm_managed_cluster_info

bjoydeep:
Add Red Hat Advanced Cluster Management for Kubernetes metric acm_managed_cluster_info to the manifest allowlist. This provides Subscription watch and other information for the managed clusters for an ACM Hub cluster. Detailed description of this metric is here: https://docs.google.com/document/d/1uvCk73lWQ-LGqFNla6IMRuur5Fj6q59Y2UMpGK5AaWs/edit?usp=sharing

 


simonpasquier:
cc @smarterclayton for approval. The metric is only present and collected for hub clusters, not managed clusters.

According @bjoydeep to:
 
> cardinality will be at least once for every managed cluster. (can be more than 1 - if versions etc change - but dont see it exploding)

/assign smarterclayton

smarterclayton:
I approve the metric

PR #1105, opened 2021-04-06
Bug 1946597: Backport Remove the "instance" and "pod" labels for kube-state-metrics metrics

jan--f:
Backport of #1052


simonpasquier:
The `generate` job is failing because the generation isn't deterministic for `manifests/0000_50_cluster-monitoring-operator_02-role.yaml`. We have a task to migrate from the hacky Python script to full jsonnet eventually but for now, you'd have to amend the last commit and rollback the changes on `manifests/0000_50_cluster-monitoring-operator_02-role.yaml`.

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 1.00

PR #1108, opened 2021-04-07
Refactor proxy config

simonpasquier:
Refactoring a bit the code to have the same behavior for managing infrastructure and proxy configurations. Especially CMO now remembers the last know proxy configuration so it can't get the latest version from the API, it will default to the last known version. This makes the operator more resilient to transient blips from the API server.

This change also avoids putting the operator into a Degraded status if it can't retrieve infrastructure or proxy configurations.


pkg/operator/operator.go
```diff
@@ -424,20 +489,47 @@ func (o *Operator) sync(key string) error {
...
+}
+
+func (o *Operator) loadProxyConfig() (*ProxyConfig, error) {
+	var proxyConfig *ProxyConfig
+
+	proxy, err := o.client.GetProxy(clusterResourceName)
+	if err != nil {
+		klog.Warningf("Error getting cluster proxy configuration: %v", err)
+
+		if proxyConfig == nil {
```
dgrisonnet, line 522: this should be `o.lastKnowProxyConfig`

pkg/operator/operator.go
```diff
@@ -424,20 +489,47 @@ func (o *Operator) sync(key string) error {
...
+	infrastructure, err := o.client.GetInfrastructure(clusterResourceName)
 	if err != nil {
-		err = errors.Wrap(err, "error getting cluster infrastructure")
-		klog.Info(err)
-		reportErr := o.client.StatusReporter().SetFailed(err, "FailedInfrastructureConfig")
-		if reportErr != nil {
-			klog.Errorf("error occurred while setting status to failed: %v", reportErr)
+		klog.Warningf("Error getting cluster infrastructure: %v", err)
+
+		if infrastructureConfig == nil {
```
dgrisonnet, line 499: this should be `o.lastKnowInfrastructureConfig`

pkg/operator/operator.go
simonpasquier: thanks!

PR #1109, opened 2021-04-07
Add management workload annotations

lack:
In support of the workload partitioning feature
(https://github.com/openshift/enhancements/pull/703), we need to add
annotations to all management pods and namespaces so they can be
properly identified and assigned to segregated management cores on
clusters configured to do so.

### DIFF
--- a/assets/prometheus-user-workload/prometheus.yaml
+++ b/assets/prometheus-user-workload/prometheus.yaml
@@ -119,6 +119,8 @@ spec:
   overrideHonorLabels: true
   overrideHonorTimestamps: true
   podMetadata:
+    annotations:
+      workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
     labels:
       app.kubernetes.io/component: prometheus
       app.kubernetes.io/managed-by: cluster-monitoring-operator
--- a/assets/alertmanager/alertmanager.yaml
+++ b/assets/alertmanager/alertmanager.yaml
@@ -109,6 +109,8 @@ spec:
   nodeSelector:
     kubernetes.io/os: linux
   podMetadata:
+    annotations:
+      workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
     labels:
       app.kubernetes.io/component: alert-router
       app.kubernetes.io/managed-by: cluster-monitoring-operator
--- a/assets/grafana/deployment.yaml
+++ b/assets/grafana/deployment.yaml
@@ -22,6 +22,7 @@ spec:
       annotations:
         checksum/grafana-config: 4edd2a7c41f827b93497c2d118a926e7
         checksum/grafana-datasources: 0fc2251d80a7fdcb28e58d9944addd91
+        workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
         app.kubernetes.io/component: grafana
         app.kubernetes.io/managed-by: cluster-monitoring-operator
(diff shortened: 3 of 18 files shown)


simonpasquier:
Thanks for the PR! To make sure that we don't forget to set the annotation if we add new pods in the future, it would be better to write a jsonnet function that would automatically add the annotation to all deployments and other custom resources. It would be something similar to the [addReleaseAnnotation()](https://github.com/openshift/cluster-monitoring-operator/blob/a227f52b202b322bcf3dea7230c6cb0a86b9865b/jsonnet/add-release-annotation.libsonnet#L1-L27) function and invoked from [jsonnet/main.jsonnet](https://github.com/openshift/cluster-monitoring-operator/blob/a227f52b202b322bcf3dea7230c6cb0a86b9865b/jsonnet/main.jsonnet#L376):

```
{
  addWorkloadAnnotation(o): {
    local annotation = {
      'workload.openshift.io/management': '{"effect": "PreferredDuringScheduling"}',
    },
    local addAnnotation(o) = o {
      [if std.setMember(o.kind, ['DaemonSet', 'Deployment', 'ReplicaSet']) then 'spec']+: {
        template+: {
          metadata+: {
            annotations+: annotation,
          },
        },
      },
      [if std.setMember(o.kind, ['Alertmanager', 'Prometheus', 'ThanosRuler']) then 'spec']+:
        {
          podMetadata+: {
            annotations+: annotation,
          },
        },
    },
    [k]: addAnnotation(o[k])
    for k in std.objectFields(o)
  },
}
```

lack:
> Thanks for the PR! To make sure that we don't forget to set the annotation if we add new pods in the future, it would be better to write a jsonnet function that would automatically add the annotation to all deployments and other custom resources. It would be something similar to the [addReleaseAnnotation()](https://github.com/openshift/cluster-monitoring-operator/blob/a227f52b202b322bcf3dea7230c6cb0a86b9865b/jsonnet/add-release-annotation.libsonnet#L1-L27) function and invoked from [jsonnet/main.jsonnet](https://github.com/openshift/cluster-monitoring-operator/blob/a227f52b202b322bcf3dea7230c6cb0a86b9865b/jsonnet/main.jsonnet#L376):

This is so much better.  It even caught one case I'd missed in my original run through.  Thanks for the great pointer!  I'm pretty new to this jsonnet stuff :)

assets/prometheus-user-workload/prometheus.yaml
```diff
@@ -119,6 +119,8 @@ spec:
   overrideHonorLabels: true
   overrideHonorTimestamps: true
   podMetadata:
+    annotations:
+      workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
```
jan--f, line 123: Do we actually want the user-workload bits treated as a management workload here? Iiuc the user-workload bits are provided as best-effort, so is this maybe a reason to not mark them as management workloads?

lack: I think management is appropriate, even if they're best-effort.  We want them on the dedicated management cores and not taking CPU that should be instead guaranteed to user workloads.

simonpasquier: I agree with @lack: everything that is managed by CMO falls into the "management" bucket.

simonpasquier:
@lack could you rebase on top of master. I suspect that https://github.com/openshift/cluster-monitoring-operator/pull/986 will require to regenerate the manifests.

lack:
> @lack could you rebase on top of master? I suspect that #986 will require to regenerate the manifests.

It did indeed.  Thanks for the heads-up!

simonpasquier:
/lgtm

You'd need a BZ (or an explicit approval) but you're probably aware of it :)

PR #1112, opened 2021-04-12
Bug 1948082: Set unavailable message

jan--f:

### DIFF
--- a/pkg/client/status_reporter.go
+++ b/pkg/client/status_reporter.go
@@ -26,6 +26,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/client/status_reporter.go
```diff
@@ -147,16 +147,13 @@ func (r *StatusReporter) SetFailed(statusErr error, reason string) error {
 	time := metav1.Now()
 	// The Reason should be upper case camelCase (PascalCase) according to the API docs.
 	reason = strings.ToPascalCase(reason)
+	unavailableMessage := "Rollout of the monitoring stack failed and is degraded. Please investigate the degraded status error."
```
dgrisonnet, line 150: nit: we could put this message in a const at the top of the file

simonpasquier: I agree with Damien, otherwise lgtm :)

PR #1115, opened 2021-04-12
BUG 1949145: Delete 0000_50_cluster-monitoring-operator_00_0user-priority-class.yaml

ravisantoshgudimetla:
Moving this priority class to [CCO](https://github.com/openshift/cluster-config-operator/pull/191) - as it is beneficial enough to be used by multiple components.


simonpasquier:
/lgtm
/hold
it needs to wait for https://github.com/openshift/cluster-config-operator/pull/191 to be merged.

ravisantoshgudimetla:
https://github.com/openshift/cluster-config-operator/pull/191  merged

/retest

ravisantoshgudimetla:
/hold

We don't have a CI build for openshift/cluster-config-operator#191, will hold till then

ravisantoshgudimetla:
/hold

@wking mentioned that cluster-profile annotations are mandatory for CVO to load manifests if we have opted into the cluster-profile we're currently running. So once https://github.com/openshift/cluster-config-operator/pull/193 merges, we should see the priority class getting created.

lihongyan1:
test with payload 4.8.0-0.nightly-2021-04-15-152737, the PR can be merged now

```
#oc get PriorityClass openshift-user-critical -oyaml
apiVersion: scheduling.k8s.io/v1
description: This priority class should be used for user facing OpenShift workload pods only.
kind: PriorityClass
metadata:
  annotations:
    include.release.openshift.io/self-managed-high-availability: "true"
    include.release.openshift.io/single-node-developer: "true"
  creationTimestamp: "2021-04-15T23:10:07Z"
  - apiVersion: scheduling.k8s.io/v1
  name: openshift-user-critical
  resourceVersion: "1589"
  uid: 9a0fe9cb-7e3b-42e9-a799-f073ec1a89ba
preemptionPolicy: PreemptLowerPriority
value: 1000000000
```

PR #1116, opened 2021-04-13
Provide an explicit list of allowed labels for the kube_node_labels metric

simonpasquier:


simonpasquier:
We should rather wait until we can configure a wildcard list for the `kube_node_labels` metric in our downstream kube-state-metrics (https://github.com/kubernetes/kube-state-metrics/pull/1403).

dgrisonnet:
Agreed, but we should still allow the labels used in Telemetry, otherwise they won't be shown.

PR #1118, opened 2021-04-14
Bug 1949123: [3.11]: rules: fix instance:node_filesystem_usage

dgrisonnet:
The instance:node_filesystem_usage was only considering the `/`
mountpoint instead of accounting for all the filesystem.
Also replace node_filesystem_free_bytes by node_filesystem_avail_bytes
to take into account reserved blocks.

Previously, the version of the go-jsonnet package that we were
installing in the docker image used to generate manifests wasn't fixed
to a particular version. This resulted in golang errors with go-jsonnet
v0.17.0 as it required golang 1.10, but we are using a 1.9 image.

Also, this required to get the package from the jsonnet repository
instead of go-jsonnet as the assets weren't available for go-jsonnet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,8 @@ $(JB_BIN):
 	chmod +x $(GOPATH)/bin/jb
 
 $(JSONNET_BIN):
-	go get -u github.com/google/go-jsonnet/cmd/jsonnet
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
/hold 
not sure the bug severity warrants the change.

dgrisonnet:
I also asked wondered about that, but I found this doc https://access.redhat.com/support/policy/updates/openshift_noncurrent that state that we still have to fully support 3.11 until June.

PR #1120, opened 2021-04-15
Bug 1940933: Revert "jsonnet: make AggregatedAPIDown more resilient"

dgrisonnet:
This reverts commit 0746d65ac8ce639f769d88ed623fb2931e355eab.

Revert the previous attempt to fix `AggregatedAPIDown` alerts during OCP upgrades. It turned out that the alert wasn't at cause, but rather prometheus-adapter not being highly-available. There is a separate on-going effort to make it HA in #1119.


dgrisonnet:
Holding until https://github.com/openshift/cluster-monitoring-operator/pull/1119 is merged and we can verify that it fixed the issue with the AggregatedAPIDown alert firing.

/hold

simonpasquier:
https://github.com/openshift/cluster-monitoring-operator/pull/1119 got reverted so this needs to wait a bit longer.

dgrisonnet:
#1124 has been merged so we can move forward with this revert.

/unhold

PR #1123, opened 2021-04-19
Rename workload annotations

lack:
As per https://github.com/openshift/enhancements/pull/739, the workload
annotations names are changing.

/hold
Hold until after openshift/kubernetes#632 is merged please

### DIFF
--- a/assets/alertmanager/alertmanager.yaml
+++ b/assets/alertmanager/alertmanager.yaml
@@ -110,7 +110,7 @@ spec:
     kubernetes.io/os: linux
   podMetadata:
     annotations:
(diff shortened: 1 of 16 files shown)


simonpasquier:
/lgtm
Feel free to ignore the single-node job, it's optional and needs (at least) #1127 to pass.

dgrisonnet:
@lack you can ignore the `e2e-aws-single-node` job, it is optional. The reasons for the failures are known, but we prefer keeping it in order to avoid any regressions.

imiller0:
/unhold
openshift/kubernetes#632 has now merged.

PR #1126, opened 2021-04-20
Bug 1951505: Remove deprecated techPreviewUserWorkload

dgrisonnet:
Remove `techPreviewUserWorkload` from CMO's configmap as it is deprecated in 4.8.


pkg/manifests/config.go
```diff
@@ -404,21 +394,7 @@ func NewDefaultUserWorkloadMonitoringConfig() *UserWorkloadConfiguration {
 }
 
 // IsUserWorkloadEnabled checks if user workload monitoring is
-// enabled. If old techPreview configuration exists, we warn user in logs.
+// enabled.
```
squat, line 397: Can we entirely remove this func as it is simply the lookup of one value and in go we normally avoid writing getters/setters?

simonpasquier (commented):
/lgtm

PR #1130, opened 2021-04-21
Bug 1952149: oc adm top reporting unknown status for Windows node

mansikulkarni96:
This PR overrides the nodeQueries in the prometheus-adapter config-map to include windows_exporter queries in order to retrieve resource metrics: CPU and memory for Windows nodes. 
This fix has been merged upstream in [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus/pull/1058)
Ran commands:
`make generate`

This PR has been opened in an effort to merge this fix into release-4.7 since it is not included in release-0.7 of kube-prometheus.

- [x] No user facing changes, so no entry in CHANGELOG was needed.

### DIFF
--- a/jsonnet/main.jsonnet
+++ b/jsonnet/main.jsonnet
@@ -96,6 +96,16 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
                },
                prometheusAdapter+:: {
                  prometheusURL: 'https://prometheus-k8s.openshift-monitoring.svc:9091',
+                 config+: {
+                   resourceRules+: {
+                     cpu+: {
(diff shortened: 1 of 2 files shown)


jsonnet/main.jsonnet
```diff
@@ -96,6 +96,35 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') +
                },
                prometheusAdapter+:: {
                  prometheusURL: 'https://prometheus-k8s.openshift-monitoring.svc:9091',
+                 config: {
```
dgrisonnet, line 99: In jsonnet you can actually override deeply nested fields with `+:`. In your case since you only want to modify the node queries, you can do something like that:

```
config+: {
  resourceRules+: {
    cpu+: {
      nodeQuery: 'your change',
    },
  },
},
```

mansikulkarni96: thanks for the tip @dgrisonnet , that definitely reduces the code duplication here, I have pushed the changes PTAL.

dgrisonnet:
/lgtm
/approve

Thank you for taking care of this @mansikulkarni96 :slightly_smiling_face:

PR #1131, opened 2021-04-26
Bug 1955457: drop container_memory_failures_total metric

simonpasquier:
The `container_memory_failures_total` metric isn't used anywhere and it generates at least 4 series per container.

### DIFF
--- a/assets/control-plane/service-monitor-kubelet.yaml
+++ b/assets/control-plane/service-monitor-kubelet.yaml
@@ -62,6 +62,10 @@ spec:
       regex: container_(network_tcp_usage_total|network_udp_usage_total|tasks_state|cpu_load_average_10s)
       sourceLabels:
       - __name__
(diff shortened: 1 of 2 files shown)


dgrisonnet:
@simonpasquier since we are removing metrics from containers, would it be meaningful to ask for reviews from the node team to better understand why these metrics exist and assess whether any alerts or dashboards are missing? Moreover since we plan to give them ownership of the kubelet monitoring, it might be worth putting them in the loop.

simonpasquier:
/hold
until we've got agreement that it's ok to drop the metric.

rphillips:
For reference, this metric tracks container page faults. We can remove it for now.

simonpasquier:
/hold
until the CI is fixed

simonpasquier:
/retitle Bug 1955457: drop container_memory_failures_total metric

PR #1132, opened 2021-04-27
Extended PrometheusK8sConfig and PrometheusRestrictedConfig with AdditionalAlertManagerConfigs

dislbenn:
* [ ] No user-facing changes, so no entry in CHANGELOG was needed.


assets/prometheus-k8s/prometheus.yaml
```diff
@@ -39,6 +39,7 @@ spec:
   configMaps:
   - serving-certs-ca-bundle
   - kubelet-serving-ca-bundle
+  - hub-router-ca
```
clyang82, line 42: I do not think you should add `hub-router-ca` here. I think you can make the `cluster-monitoring-config` configmap to support configuring configmap so that it can be able to update with customized configmap externally. I am not sure if it is correct approach. @simonpasquier can you give us recommendations? Thanks.

pkg/manifests/config.go
```diff
@@ -83,15 +83,16 @@ type PrometheusOperatorConfig struct {
...
+	LogLevel                      string                               `json:"logLevel"`
+	Retention                     string                               `json:"retention"`
+	NodeSelector                  map[string]string                    `json:"nodeSelector"`
+	Tolerations                   []v1.Toleration                      `json:"tolerations"`
+	Resources                     *v1.ResourceRequirements             `json:"resources"`
+	ExternalLabels                map[string]string                    `json:"externalLabels"`
+	VolumeClaimTemplate           *monv1.EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate"`
+	RemoteWrite                   []monv1.RemoteWriteSpec              `json:"remoteWrite"`
+	TelemetryMatches              []string                             `json:"-"`
+	AdditionalAlertManagerConfigs *v1.SecretKeySelector                `json:"additionalAlertManagerConfigs"`
```
simonpasquier, line 95: Thinking more about the user experience here, I'm not sure that the configuration should just be a reference to an uncontrolled secret. The risk is high that the provided configuration is invalid and breaks Prometheus. Also not all fields of the [`alertmanager_config`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alertmanager_config) section are worth exposing (for instance configuring an external Alertmanager endpoint only needs the `static_configs` service discovery) and some might be not be usable in practice (for instance any field like `ca_file` that reference a file on disk).

It might be more appropriate to expose the main fields that are required to configure external Alertmanager endpoints. From there, CMO could generate the secret to be referenced in the Prometheus resource.

Here is the rough idea:

```go
type PrometheusK8sConfig struct {
    ...
    AlertmanagerConfigs []AlertmanagerConfig `json:"additionalAlertManagerConfigs"`
}

type AlertmanagerConfig struct {
    # URL to the Alertmanager server.
    URL net.URL `json:"url"`
    Timeout string `json:"timeout"`
    # Reference to a key in a Secret containing the certificate authority to verify the server's certificate. 
    CA *v1.SecretKeySelector
    # Authn parameters (to be defined)
    ...
}
```

bjoydeep:
@simonpasquier @dislbenn when the Prometheus on cluster A will post the alert to Alertmanager on the Hub via this, how will the user know from which cluster this came? in ACM metric collector, when we take in data from Prometheus on cluster A and send it to the Thanos API Gateway on the Hub, we do append the clusterid and clustername as labels. Is this happening already - may be I missed it in the code?

clyang82:
> @simonpasquier @dislbenn when the Prometheus on cluster A will post the alert to Alertmanager on the Hub via this, how will the user know from which cluster this came? in ACM metric collector, when we take in data from Prometheus on cluster A and send it to the Thanos API Gateway on the Hub, we do append the clusterid and clustername as labels. Is this happening already - may be I missed it in the code?

@bjoydeep @morvencao is trying to use `relabel_configs` to add clusterid and clustername in managed cluster before send alerts to hub alertmanager.

morvencao:
I also found another configuration `AdditionalAlertRelabelConfigs` that may resolve the relabel issue for alerts.
@simonpasquier may know more details.
see: https://github.com/openshift/prometheus-operator/blob/34a5b31452901e7d88240420f372969a01128691/pkg/apis/monitoring/v1/types.go#L299
Will try that and give feedback here.

simonpasquier:
Regarding alert relabeling, I would expect that using external labels is enough for your use case?

clyang82:
> Regarding alert relabeling, I would expect that using external labels is enough for your use case?

Yes. external labels is enough for our cases. Do you want to reduce the risk to not support alert relabeling right now?

simonpasquier:
> Do you want to reduce the risk to not support alert relabeling right now?
yes

clyang82:
> > Do you want to reduce the risk to not support alert relabeling right now?
> > yes

Can we keep using `[]*monv1.ProbeTargetStaticConfig` and just ignore `RelabelConfigs`? or do you suggest to define a new object in CMO? Thanks.

simonpasquier:
No need to retest the `e2e-aws-single-node` job, it is optional.

pkg/manifests/config.go
```diff
@@ -92,6 +92,36 @@ type PrometheusK8sConfig struct {
...
+	// The timeout used when sending alerts.
+	Timeout *string `json:"timeout,omitempty"`
+	// The api version of Alertmanager.
+	APIVersion string `json:"apiVersion"`
+	// TLS Config to use for alertmanager connection.
+	TLSConfig TLSConfig `json:"tlsConfig,omitempty"`
+	// Bearer token to use when authenticating to Alertmanager.
+	BearerToken string `json:"bearerToken,omitempty"`
+	// List of labeled statically configured Alertmanagers.
+	StaticConfigs []*monv1.ProbeTargetStaticConfig `json:"staticConfigs,omitempty"`
```
simonpasquier, line 112: I don't think we need to reuse `monv1.ProbeTargetStaticConfig` here. A `[]string` slice is enough IMHO.

pkg/manifests/config.go
```diff
@@ -92,6 +92,36 @@ type PrometheusK8sConfig struct {
...
+	// Path prefix to add in front of the push endpoint path.
+	PathPrefix string `json:"pathPrefix,omitempty"`
+	// The timeout used when sending alerts.
+	Timeout *string `json:"timeout,omitempty"`
+	// The api version of Alertmanager.
+	APIVersion string `json:"apiVersion"`
+	// TLS Config to use for alertmanager connection.
+	TLSConfig TLSConfig `json:"tlsConfig,omitempty"`
+	// Bearer token to use when authenticating to Alertmanager.
+	BearerToken string `json:"bearerToken,omitempty"`
```
simonpasquier, line 110: It would be better to have a reference to a secret's key rather than the token stored as plain-text in the configmap.

pkg/manifests/config.go
```diff
@@ -92,6 +92,36 @@ type PrometheusK8sConfig struct {
...
+// TLSConfig configures the options for TLS connections.
+type TLSConfig struct {
+	// Path to the CA cert in the Prometheus container to use for the targets.
+	CAFile string `json:"caFile,omitempty"`
+	// Path to the client cert file in the Prometheus container for the targets.
+	CertFile string `json:"certFile,omitempty"`
+	// Path to the client key file in the Prometheus container for the targets.
+	KeyFile string `json:"keyFile,omitempty"`
+	// Disable target certificate validation.
+	InsecureSkipVerify bool `json:"insecureSkipVerify"`
```
simonpasquier, line 124: Any reason why we don't expose `ServerName`?

```suggestion
// TLSConfig configures the options for TLS connections.
type TLSConfig struct {
	// Path to the CA cert in the Prometheus container to use for the targets.
	CAFile string `json:"caFile,omitempty"`
	// Path to the client cert file in the Prometheus container for the targets.
	CertFile string `json:"certFile,omitempty"`
	// Path to the client key file in the Prometheus container for the targets.
	KeyFile string `json:"keyFile,omitempty"`
	// Used to verify the hostname for the targets.
	ServerName string `yaml:"server_name,omitempty"`
	// Disable target certificate validation.
	InsecureSkipVerify bool `json:"insecureSkipVerify"`
```

pkg/manifests/config.go
```diff
@@ -349,15 +379,16 @@ type UserWorkloadConfiguration struct {
...
+	LogLevel                      string                               `json:"logLevel"`
+	Retention                     string                               `json:"retention"`
+	NodeSelector                  map[string]string                    `json:"nodeSelector"`
+	Tolerations                   []v1.Toleration                      `json:"tolerations"`
+	Resources                     *v1.ResourceRequirements             `json:"resources"`
+	ExternalLabels                map[string]string                    `json:"externalLabels"`
+	VolumeClaimTemplate           *monv1.EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate"`
+	RemoteWrite                   []monv1.RemoteWriteSpec              `json:"remoteWrite"`
+	EnforcedSampleLimit           *uint64                              `json:"enforcedSampleLimit"`
+	AdditionalAlertManagerConfigs *v1.SecretKeySelector                `json:"additionalAlertManagerConfigs"`
```
simonpasquier, line 391: this is still using a secret key selector.

pkg/manifests/manifests.go
```diff
@@ -243,6 +244,9 @@ var (
 	AuthProxyRedirectURLFlag  = "-redirect-url="
 
 	TrustedCABundleKey = "ca-bundle.crt"
+
+	AdditionalAlertmanagerConfigKey  = "alertmanager-configs.yaml"
+	AdditionalAlertmanagerConfigName = "prometheus-k8s-additional-alertmanager-configs"
```
simonpasquier, line 249:
```suggestion
	AdditionalAlertmanagerConfigSecretKey  = "alertmanager-configs.yaml"
	AdditionalAlertmanagerConfigSecretName = "prometheus-k8s-additional-alertmanager-configs"
```

pkg/manifests/config.go
```diff
@@ -92,6 +92,36 @@ type PrometheusK8sConfig struct {
...
+}
+
+// TLSConfig configures the options for TLS connections.
+type TLSConfig struct {
+	// Path to the CA cert in the Prometheus container to use for the targets.
+	CAFile string `json:"caFile,omitempty"`
+	// Path to the client cert file in the Prometheus container for the targets.
+	CertFile string `json:"certFile,omitempty"`
+	// Path to the client key file in the Prometheus container for the targets.
+	KeyFile string `json:"keyFile,omitempty"`
```
simonpasquier, line 122: Same concern than for `BearerToken`

pkg/manifests/manifests.go
```diff
@@ -1214,6 +1218,16 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
...
 	}
 
+	if f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs != nil {
+		p.Spec.AdditionalAlertManagerConfigs = &v1.SecretKeySelector{
+			Key: AdditionalAlertmanagerConfigKey,
+			LocalObjectReference: v1.LocalObjectReference{
+				Name: AdditionalAlertmanagerConfigName,
+			},
+		}
+		p.Spec.ConfigMaps = append(p.Spec.ConfigMaps, getConfigMapName(f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs)...)
```
simonpasquier, line 1228: it's not clear to me why we need to mount an additional configmap?

morvencao: The additional configmap is used to store CA for the Alertmanager.

simonpasquier: Hmm so the idea is that a user has to 
1. Create a configmap `foo` in the `openshift-monitoring` namespace with key `ca.crt` that would hold the CA certificate.
2. Configure `CAFile: /etc/prometheus/configmaps/foo/ca.crt` in the cluster-monitoring-operator configmap?

Now the user has to know about where the file is mounted in the Prometheus pod. I'd rather see that the CA certificate content is stored directly in the CMO configmap (there's no security concern here) or `CAFile` is a configmap key selector in the CMO configmap. In both cases, CMO can generate the final configmap/secret, pass the generated configmap/secret to the Prometheus resource and inject the proper `ca_file` path in the additional Alertmanagers secret.

pkg/manifests/config.go
clyang82: Yes. we do not have case to assign the labels to all metrics scraped from target. Will update to `[]string`

pkg/manifests/config.go
clyang82: Good catch 👍 . We should expose `ServerName`.

pkg/manifests/config.go
clyang82: Can we support both? I mean that I can have another PR to support a secret key as `BearerTokenFile`.

pkg/manifests/manifests.go
clyang82: Agreed that the user should not have to know about where the file is mounted in the Prometheus pod. How about have `CertsConfigmapName` in `TLSConfig`? if so, CMO can inject the proper `ca_file` `cert_file` `key_file` in the additional Alertmanagers secret. I think maybe MCO is not responsible for cert rotation. it should be owned by external process who generate the cert configmap. it is simple and easy to update the cert configmap by external process rather than `cluster-monitoring-config`. WDYT？

pkg/manifests/config.go
simonpasquier: My concern is that the token value is exposed in clear text in the CMO configmap. I'm fine with the final Prometheus configuration using `bearer_token`.

clyang82: I have changed the bearerToken to be a secret key selector. and use `authorization` in Promtheus configuration since `bearer_token` and `bearer_token_file` marked as deprecated.

pkg/manifests/config.go
simonpasquier: @clyang82 do we really have to support UWM prometheus? Note that it doesn't cover Thanos ruler which is the main producer for UWM alerts (alerting rules can be deployed to UWM prometheus but by default they land on Thanos ruler).

pkg/manifests/manifests.go
```diff
@@ -1353,9 +1357,137 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
...
+				Name: AdditionalAlertmanagerConfigSecretName,
+			},
+		}
+		p.Spec.Secrets = append(p.Spec.Secrets, getAlertmanagerTLSSecretsName(f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs)...)
+	}
+
 	return p, nil
 }
 
+func (f *Factory) AdditionalAlertManagerConfigs() (*v1.Secret, error) {
```
simonpasquier, line 1373:
```suggestion
func (f *Factory) AdditionalAlertManagerConfigsSecret() (*v1.Secret, error) {
```

pkg/manifests/manifests.go
```diff
@@ -3581,3 +3713,44 @@ func trustedCABundleVolume(configMapName, volumeName string) v1.Volume {
...
+		if alertmanagerConfig.BearerToken != nil {
+			secretsName = append(secretsName, alertmanagerConfig.BearerToken.Name)
+		}
+	}
+	return removeEmptyDuplicates(secretsName)
+}
+
+func removeEmptyDuplicates(elements []string) []string {
+	// Use map to record duplicates as we find them.
+	encountered := map[string]bool{}
```
simonpasquier, line 3739:
```suggestion
	encountered := map[string]struct{}
```

pkg/manifests/manifests.go
```diff
@@ -3581,3 +3713,44 @@ func trustedCABundleVolume(configMapName, volumeName string) v1.Volume {
...
+	result := []string{}
+
+	for v := range elements {
+		if encountered[elements[v]] == true {
+			// Do not add duplicate.
+		} else {
```
simonpasquier, line 3745:
```suggestion
		if _, found := encountered[v]; found {
			continue
		}
```

pkg/manifests/manifests.go
```diff
@@ -3581,3 +3713,44 @@ func trustedCABundleVolume(configMapName, volumeName string) v1.Volume {
...
+	}
+	return removeEmptyDuplicates(secretsName)
+}
+
+func removeEmptyDuplicates(elements []string) []string {
+	// Use map to record duplicates as we find them.
+	encountered := map[string]bool{}
+	result := []string{}
+
+	for v := range elements {
```
simonpasquier, line 3742: you can work on values instead of indices
```suggestion
	for _, v := range elements {
```

pkg/manifests/manifests.go
```diff
@@ -3581,3 +3713,44 @@ func trustedCABundleVolume(configMapName, volumeName string) v1.Volume {
...
+		if encountered[elements[v]] == true {
+			// Do not add duplicate.
+		} else {
+			// Record this element as an encountered element.
+			encountered[elements[v]] = true
+			if elements[v] != "" {
```
simonpasquier, line 3748: not sure why the value would be empty?

pkg/manifests/manifests.go
```diff
@@ -1353,9 +1357,137 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
...
+func (f *Factory) AdditionalAlertManagerConfigs() (*v1.Secret, error) {
+	alertmanagerConfigs := f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs
+	if len(alertmanagerConfigs) == 0 {
+		return nil, nil
+	}
+
+	cfgs := []yaml2.MapSlice{}
+	for _, alertmanagerConfig := range alertmanagerConfigs {
+		cfg := yaml2.MapSlice{}
+		cfg = append(cfg, yaml2.MapSlice{
```
simonpasquier, line 1382: you probably want to skip injecting the fields into the YAML if the values are empty

pkg/manifests/manifests_test.go
```diff
@@ -953,9 +954,29 @@ func TestPrometheusK8sConfiguration(t *testing.T) {
...
+      key: token
+    scheme: https
+    tlsConfig:
+      ca: 
+        name: cert
+        key: root-ca.crt
+      cert: 
+        name: cert
+        key: cert.crt
+      file:
```
simonpasquier, line 970: key?

pkg/manifests/manifests_test.go
```diff
@@ -1043,6 +1064,21 @@ ingress:
 	if p.Spec.ExternalLabels["datacenter"] != "eu-west" {
 		t.Fatal("Prometheus external labels are not configured correctly")
 	}
+
+	if p.Spec.AdditionalAlertManagerConfigs.Name != AdditionalAlertmanagerConfigSecretName {
```
simonpasquier, line 1068: could you create a new Test function that configures AdditionalAlertmanagerConfigs and leave this function without AdditionalAlertmanagerConfigs (+ check that `p.Spec.AdditionalAlertManagerConfigs` is nil)?

pkg/manifests/manifests_test.go
```diff
@@ -1054,6 +1090,57 @@ ingress:
...
+
+	f := NewFactory("openshift-monitoring", "openshift-user-workload-monitoring", c, defaultInfrastructureReader(), &fakeProxyReader{}, NewAssets(assetsPath))
+
+	s, err := f.AdditionalAlertManagerConfigs()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	d := []AdditionalAlertmanagerConfig{}
+	err = yaml.Unmarshal(s.Data[AdditionalAlertmanagerConfigSecretKey], &d)
```
simonpasquier, line 1138: hmm not sure why you deserialiaze the secret's data into `[]AdditionalAlertmanagerConfig`? I'd be ok with verifying that the resulting byte slice is the expected one.

simonpasquier (commented):
this is still missing the update to the sync loop so the operator reconciles the Prometheus configuration when the secrets change.

simonpasquier:
> this is still missing the update to the sync loop so the operator reconciles the Prometheus configuration when the secrets change.

sorry I missed that since the secrets are mounted directly in the Prometheus container, we don't need to reconcile in CMO...

pkg/manifests/manifests_test.go
clyang82: Yes.

clyang82: Yes.

pkg/manifests/manifests.go
clyang82: I did not add logic to judge whether the `name` is empty or not. so it has chance to be empty value.

pkg/manifests/manifests.go
clyang82: why you suggest to be `struct`? I think `bool` is sufficient here.

pkg/manifests/manifests_test.go
clyang82: It is designed to deserialiaze the secret's data into `[]AlertmanagerConfig`. agreed with just verify that the resulting byte slice is the expected one. Updated.

pkg/manifests/manifests_test.go
clyang82: Done

pkg/manifests/manifests.go
clyang82: 💯 so rigorously. great suggestion. Thank you

pkg/manifests/config.go
simonpasquier: @clyang82 what about this line? see previous comment

clyang82: We do not need to support UWM prometheus. Removed this field. Thanks.

pkg/manifests/config.go
```diff
@@ -92,6 +92,38 @@ type PrometheusK8sConfig struct {
...
+	PathPrefix string `json:"pathPrefix,omitempty"`
+	// The timeout used when sending alerts.
+	Timeout *string `json:"timeout,omitempty"`
+	// The api version of Alertmanager.
+	APIVersion string `json:"apiVersion"`
+	// TLS Config to use for alertmanager connection.
+	TLSConfig TLSConfig `json:"tlsConfig,omitempty"`
+	// Bearer token to use when authenticating to Alertmanager.
+	BearerToken *v1.SecretKeySelector `json:"bearerToken,omitempty"`
+	// List of labeled statically configured Alertmanagers.
```
simonpasquier, line 111:
```suggestion
	// List of statically configured Alertmanagers.
```

pkg/manifests/manifests.go
```diff
@@ -3586,3 +3737,42 @@ func trustedCABundleVolume(configMapName, volumeName string) v1.Volume {
 		},
 	}
 }
+
+func getAlertmanagerTLSSecretsName(alertmanagerConfigs []AdditionalAlertmanagerConfig) []string {
```
simonpasquier, line 3741: (nit) secrets also include bearer tokens.

```suggestion
func getAdditionalAlertmanagerSecrets(alertmanagerConfigs []AdditionalAlertmanagerConfig) []string {
```

pkg/manifests/manifests.go
```diff
@@ -3586,3 +3737,42 @@ func trustedCABundleVolume(configMapName, volumeName string) v1.Volume {
...
+	secretsName := []string{}
+	for _, alertmanagerConfig := range alertmanagerConfigs {
+		if alertmanagerConfig.TLSConfig.CA != nil {
+			secretsName = append(secretsName, alertmanagerConfig.TLSConfig.CA.Name)
+		}
+		if alertmanagerConfig.TLSConfig.Cert != nil {
+			secretsName = append(secretsName, alertmanagerConfig.TLSConfig.Cert.Name)
+		}
+		if alertmanagerConfig.TLSConfig.Key != nil {
+			secretsName = append(secretsName, alertmanagerConfig.TLSConfig.Cert.Name)
```
simonpasquier, line 3752:
```suggestion
			secretsName = append(secretsName, alertmanagerConfig.TLSConfig.Key.Name)
```

pkg/manifests/manifests_test.go
```diff
@@ -1046,6 +1052,92 @@ ingress:
...
+    timeout: 60s
+    tlsConfig:
+      ca: 
+        name: alertmanager2-cert
+        key: root-ca.crt
+      cert:
+        name: alertmanager2-cert
+        key: cert.crt
+      key:
+        name: alertmanager2-cert
```
simonpasquier, line 1082: can you use a different secret name? it would detect the typo in `getAlertmanagerTLSSecretsName`.

pkg/manifests/manifests_test.go
```diff
@@ -1046,6 +1052,92 @@ ingress:
...
+		"prometheus-k8s.openshift-monitoring.svc",
+		&v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+		&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+	)
+
+	if !strings.Contains(strings.Join(p.Spec.Secrets, ","), "alertmanager2-cert") ||
+		!strings.Contains(strings.Join(p.Spec.Secrets, ","), "alertmanager1-ca") ||
+		!strings.Contains(strings.Join(p.Spec.Secrets, ","), "alertmanager1-bearer-token") {
+		t.Fatal("Prometheus secrets are not generated correctly, expected to have alertmanager2-cert/alertmanager1-ca/alertmanager1-bearer-token secret, but found", p.Spec.Secrets)
+	}
```
simonpasquier, line 1129: Could you iterate over `p.Spec.Secrets` and check that all expected secrets are present?

```suggestion
	secrets := make(map[string]struct{})
	for _, s := range p.Spec.Secrets {
		secrets[s] = struct{}{}
	}

	for _, exp := range []string{"alertmanager2-cert",  "alertmanager1-ca", "alertmanager1-bearer-token"} {
		if _, found := secrets[exp]; found {
			continue
		}
		t.Fatal("Prometheus secrets are not generated correctly, expected to have %s but got none", exp)
	}
```

pkg/manifests/manifests.go
```diff
@@ -1358,9 +1362,156 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
...
+		if alertmanagerConfig.Timeout != nil {
+			cfg = append(cfg, yaml2.MapItem{
+				Key: "timeout", Value: alertmanagerConfig.Timeout,
+			})
+		}
+
+		auth := yaml2.MapSlice{}
+		if alertmanagerConfig.BearerToken != nil {
+			if alertmanagerConfig.BearerToken.Name == "" {
+				return nil, errors.Errorf("bearer token secret name not found")
```
simonpasquier, line 1412: please return the name of the secret (it would help when troubleshooting misconfiguration)
```suggestion
				return nil, errors.Errorf("secret %q for bearer token not found", alertmanagerConfig.BearerToken.Name)
```

pkg/manifests/manifests.go
```diff
@@ -1358,9 +1362,156 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
...
+			cfg = append(cfg, yaml2.MapItem{
+				Key:   "authorization",
+				Value: auth,
+			})
+		}
+
+		tlsConfig := yaml2.MapSlice{}
+		if alertmanagerConfig.TLSConfig.CA != nil {
+			if alertmanagerConfig.TLSConfig.CA.Name == "" {
+				return nil, errors.Errorf("ca secret name not found")
```
simonpasquier, line 1432: same remark about adding the secret name in the error message (also valid for all other returns)

pkg/manifests/manifests.go
```diff
@@ -1358,9 +1362,156 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
...
+			cfg = append(cfg, yaml2.MapItem{
+				Key:   "static_configs",
+				Value: []yaml2.MapSlice{sc},
+			})
+		}
+
+		cfgs = append(cfgs, cfg)
+	}
+
+	if len(cfgs) == 0 {
```
simonpasquier, line 1493: can it happen since we already return early if `f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs` is zero-length?

pkg/manifests/manifests.go
```diff
@@ -1358,9 +1362,156 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
...
+	amConfigYamlMap := map[string][]byte{}
+	amConfigYamlMap[AdditionalAlertmanagerConfigSecretKey] = amConfigYaml
+
+	additionalPromToAmConfigSecret := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      AdditionalAlertmanagerConfigSecretName,
+			Namespace: f.namespace,
+		},
+		Data: amConfigYamlMap,
+	}
```
simonpasquier, line 1510: (nit)
```suggestion
	additionalPromToAmConfigSecret := &v1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Name:      AdditionalAlertmanagerConfigSecretName,
			Namespace: f.namespace,
		},
		Data: map[string][]byte{
			AdditionalAlertmanagerConfigSecretKey: amConfigYaml,
		},
	}
```

pkg/manifests/manifests.go
```diff
@@ -1358,9 +1362,156 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
...
+		tlsConfig = append(tlsConfig, yaml2.MapSlice{
+			{Key: "insecure_skip_verify", Value: alertmanagerConfig.TLSConfig.InsecureSkipVerify},
+		}...)
+
+		cfg = append(cfg, yaml2.MapItem{
+			Key:   "tls_config",
+			Value: tlsConfig,
+		})
+
+		if len(alertmanagerConfig.StaticConfigs) > 0 {
```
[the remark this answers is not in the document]
clyang82: just skip to append the current `cfg` to `cfgs`. cannot return an error which leads to skip the correct configurations.

pkg/manifests/manifests_test.go
clyang82: Here I would like to test the case that `ca` `cert` `key` are in the same secret. Let me add another case to have different secret for `ca` `cert` `key`

pkg/manifests/manifests.go
clyang82: Good catch. Yes. it can be removed now.

pkg/manifests/manifests.go
```diff
@@ -3586,3 +3737,42 @@ func trustedCABundleVolume(configMapName, volumeName string) v1.Volume {
...
+	// Use map to record duplicates as we find them.
+	encountered := map[string]struct{}{}
+	result := []string{}
+
+	for _, v := range elements {
+		if _, found := encountered[v]; found {
+			continue
+		}
+		encountered[v] = struct{}{}
+		if v != "" {
```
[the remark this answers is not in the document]
clyang82: In this case, it may not happen.  But `removeEmptyDuplicates` can be reused. I suggest to keep as it is.

pkg/manifests/manifests.go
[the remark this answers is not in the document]
simonpasquier: It would be easier to read by having this at the top of the for loop

```golang
for _, alertmanagerConfig := range alertmanagerConfigs {
   if len(alertmanagerConfig.StaticConfigs) == 0 {
       continue
   }

   ...
```

pkg/tasks/prometheus.go
```diff
@@ -299,6 +299,18 @@ func (t *PrometheusTask) Run() error {
...
+		if err != nil {
+			return errors.Wrap(err, "initializing Prometheus additionalAlertManagerConfigs secret failed")
+		}
+		if secret != nil {
+			klog.V(4).Info("initializing Prometheus additionalAlertManagerConfigs secret")
+			err = t.client.CreateOrUpdateSecret(secret)
+			if err != nil {
+				return errors.Wrap(err, "reconciling Prometheus additionalAlertManagerConfigs secret failed")
+			}
+		}
```
simonpasquier, line 312: Late realization for me (sorry!) but we need to delete the secret in case the admin had configured additional Alertmanagers and then remove them all.

pkg/manifests/config.go
```diff
@@ -92,6 +92,38 @@ type PrometheusK8sConfig struct {
 	VolumeClaimTemplate *monv1.EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate"`
 	RemoteWrite         []monv1.RemoteWriteSpec              `json:"remoteWrite"`
 	TelemetryMatches    []string                             `json:"-"`
+	AlertmanagerConfigs []AdditionalAlertmanagerConfig       `json:"additionalAlertManagerConfigs"`
```
simonpasquier, line 95:
```suggestion
	// EXPERIMENTAL: this configuration field may change in future releases.
	AlertmanagerConfigs []AdditionalAlertmanagerConfig       `json:"additionalAlertManagerConfigs"`
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - [#1090](https://github.com/openshift/cluster-monitoring-operator/pull/1090) Increase "for" duration to 1 hour for all Thanos sidecar alerts.
 - [#1093](https://github.com/openshift/cluster-monitoring-operator/pull/1093) Bump kube-state-metrics to major new release v2.0.0-rc.1. This changes a lot of metrics and flags, see kube-state-metrics CHANGELOG for full changes. 
 - [#1126](https://github.com/openshift/cluster-monitoring-operator/pull/1126) Remove deprecated techPreviewUserWorkload field from CMO's configmap.
+- [#1132](https://github.com/openshift/cluster-monitoring-operator/pull/1132) Extended PrometheusK8sConfig to support AdditionalAlertManagerConfigs field.
```
simonpasquier, line 12: I wouldn't mention it in the changelog because we don't intend to support it formally in 4.8.

```suggestion
```

simonpasquier (commented):
/lgtm

This looks good from my end. The code changes are well isolated meaning that I'm confident that users that don't configure additional Alertmanagers have no risk of being impacted by the change.

eparis:
/lgtm
looks like just a rebase

simonpasquier:
/lgtm

ping @eparis for overriding the `bugzilla/valid-bug` label

PR #1136, opened 2021-04-27
Bug 1958367: Add recording rule for builds by strategy

adambkaplan:
Aggregate the number of builds run on a cluster by strategy. This will
help the Build API team determine which build strategies are actively
being used on a cluster. This will produce a time series with the
following strategy labels:

- docker
- source
- jenkinspipeline
- custom


adambkaplan:
If possible, we'd like this rule backported to 4.6 so we can measure build usage in the active fleet.

simonpasquier:
/hold
until the CI is fixed

siamaksade (approved):

adambkaplan:
Reviewer note - recording rules with `cluster:usage` prefix are automatically exported to Telemeter.

@simonpasquier should I open a BZ so we can backport the recording rule?

jsonnet/rules.libsonnet
```diff
@@ -485,6 +485,10 @@ local droppedKsmLabels = 'endpoint, instance, job, pod, service';
             expr: 'sum(openshift_build_total{job="kubernetes-apiservers",phase="Error"})/(sum(openshift_build_total{job="kubernetes-apiservers",phase=~"Failed|Complete|Error"}))',
             record: 'build_error_rate',
           },
+          {
+            expr: 'sum by (strategy) (openshift_build_status_phase_total)',
+            record: 'cluster:usage:openshift:build_by_strategy:sum',
```
sbose78, line 490: 👍

dgrisonnet:
@adambkaplan, we will need a BZ for 4.8, 4.7 and 4.6.

Since this PR is adding a new recording rule to telemeter, you'll need to add a comment about which team owns this rule as in https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster-monitoring-operator_04-config.yaml. The rule will also need to be validated by @smarterclayton before we can proceed.

On a side note about allowing `'{__name__=~"cluster:usage:.*"}` in the telemeter config. I am not aware of the exact context behind this, but wouldn't it be better to allow each rule manually instead of grouping them with this regex? The benefit would be to improve the tractability of the ownership of these rules which is currently disregarded. cc @simonpasquier 

Unholding since the CI has been fixed.
/unhold

adambkaplan:
/assign @smarterclayton 

@dgrisonnet given that I'll need to add a comment to the telemeter config, I've renamed the rule so that we explicitly need to export it.

adambkaplan:
@simonpasquier ran into this issue trying to get `make generate` happy. https://github.com/openshift/cluster-monitoring-operator/issues/1153

xiuwang:
@adambkaplan Add a case for this feature [OCP-41751](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-41751) .

Test this feature on cluster launched from pr, the metrics works well.

/label qe-approved

adambkaplan:

adambkaplan:
Can we skip the single-node test for this PR?

dgrisonnet:
@adambkaplan the single-node job is optional

dgrisonnet:
/lgtm

Leaving approval to @smarterclayton.

assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -368,6 +368,8 @@ spec:
     rules:
     - expr: sum(openshift_build_total{job="kubernetes-apiservers",phase="Error"})/(sum(openshift_build_total{job="kubernetes-apiservers",phase=~"Failed|Complete|Error"}))
       record: build_error_rate
+    - expr: sum by (strategy) (openshift_build_status_phase_total)
```
simonpasquier, line 371: one issue with summing counters is that the numbers would be off when a counter resets. It is recommended to rate/increase then sum rather.

```suggestion
    - expr: sum by (strategy) (rate(openshift_build_status_phase_total[5m]))
```

smarterclayton: Is phase_total a counter?  I thought it was a gauge?  Agree if it's not a gauge this is a bit wierd.

smarterclayton:
Cardinality and appropriateness to be sent to telemetry approved, consider my part of approval granted.

assets/cluster-monitoring-operator/prometheus-rule.yaml
dgrisonnet: from what I can tell from the prometheus target metadata, it is a gauge. @adambkaplan usually the naming convention for Prometheus says that the `_total` suffix is reserved to counters. The correct naming for your metric should be `openshift_build_status_phases` since it's a gauge.
https://prometheus.io/docs/practices/naming/

adambkaplan: Metric is collected via `openshift-state-metrics` - it's been around since 3.x and is marked STABLE in our GitHub docs. I'm fairly certain this name was chosen before the naming conventions were codified upstream.

https://github.com/openshift/openshift-state-metrics/blob/master/docs/build-metrics.md

PR #1139, opened 2021-04-28
Bug 1955467: remove node_mountstats_nfs_* metrics

simonpasquier:
The node_exporter's mountstats collector gathers low-level metrics from
NFS mountstats (such as operation timings, transport and event
statistics). When available, these metrics can have high-cardinality
(for one cluster, we've seen that it can account for half of *all* the
series) though they aren't used by any rule or dashboard.

The mountstats collector had been enabled in #409 following a customer
request for enhancement. But looking at the history, the customer was
asking for the kubelet_volume_* metrics which weren't supported by their
storage provider at this time (it's been fixed since then NetApp/trident#134). The
mountstats metrics don't fill the same need and are superfluous.


simonpasquier:
cc @paulfantom

paulfantom (approved):
/lgtm

PR #1140, opened 2021-04-30
Bug 1955051: Update kube node status capacity metric

slashpai:
`kube_node_status_capacity` is updated in [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics/blob/master/CHANGELOG.md) 2.0. New change need us to use unit label to retrieve the metric

### DIFF
--- a/jsonnet/rules.libsonnet
+++ b/jsonnet/rules.libsonnet
@@ -111,7 +111,7 @@ local droppedKsmLabels = 'endpoint, instance, job, pod, service';
                   cluster:master_nodes
                   * on(node) group_left() max by(node)
                   (
-                    kube_node_status_capacity_cpu_cores
+                    kube_node_status_capacity{resource="cpu",unit="core"}
                   )
                 )
                 or on(node) (
@@ -120,7 +120,7 @@ local droppedKsmLabels = 'endpoint, instance, job, pod, service';
                     kube_node_labels
                   ) * on(node) group_left() max by(node)
                   (
-                    kube_node_status_capacity_cpu_cores
+                    kube_node_status_capacity{resource="cpu",unit="core"}
                   )
                 )
               )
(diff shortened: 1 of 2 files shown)


jsonnet/rules.libsonnet
```diff
@@ -111,7 +111,7 @@ local droppedKsmLabels = 'endpoint, instance, job, pod, service';
                   cluster:master_nodes
                   * on(node) group_left() max by(node)
                   (
-                    kube_node_status_capacity_cpu_cores
+                    kube_node_status_capacity{unit="core"}
```
simonpasquier, line 114: it might be more explicit to scope down to the "cpu" resource label too.
 
```suggestion
                    kube_node_status_capacity{resource="cpu",unit="core"}
```

jsonnet/rules.libsonnet
```diff
@@ -120,7 +120,7 @@ local droppedKsmLabels = 'endpoint, instance, job, pod, service';
                     kube_node_labels
                   ) * on(node) group_left() max by(node)
                   (
-                    kube_node_status_capacity_cpu_cores
+                    kube_node_status_capacity{unit="core"}
```
simonpasquier, line 123: ditto

jsonnet/rules.libsonnet
```diff
@@ -180,7 +180,7 @@ local droppedKsmLabels = 'endpoint, instance, job, pod, service';
                   cluster:master_nodes
                   * on(node) group_left() max by(node)
                   (
-                    kube_node_status_capacity_memory_bytes
+                    kube_node_status_capacity{unit="byte"}
```
simonpasquier, line 183: same here but with memory

```suggestion
                    kube_node_status_capacity{resource="memory", unit="byte"}
```

slashpai:
@simonpasquier Updated with changes suggested

simonpasquier (commented):
/lgtm
/hold
Putting on hold until we have #1127 sorted out and merged.

PR #1143, opened 2021-05-03
Bug 1957832: remove dependency of e2e tests on AWS

simonpasquier:
The end-to-end tests validating persistent storage configuration
explicitly required the 'gp2' storage class which is AWS specific. By
removing any requirement on storage class, we ensure that the end-to-end
tests can be truly agnostic.


dgrisonnet:
cc @Danil-Grigorev fyi we will be removing the AWS specific storage classes to make our e2e tests platform agnostic. From our past discussion, I am wondering if you would still need cloud integration in CI after this change.

simonpasquier:
/hold 
I missed one occurence.

simonpasquier:
/retitle Bug 1957832: remove dependency of e2e tests on AWS

PR #1147, opened 2021-05-05
Bug 1955471: [4.6] Disable collection of node_mountstats_nfs metrics in node_exporter

arajkumar:
The node_exporter's mountstats collector gathers low-level metrics from
NFS mountstats (such as operation timings, transport and event
statistics). When available, these metrics can have high-cardinality
(for one cluster, we've seen that it can account for half of all the
series) though they aren't used by any rule or dashboard.

The mountstats collector had been enabled in #409 following a customer
request for enhancement. But looking at the history, the customer was
asking for the kubelet_volume_* metrics which weren't supported by their
storage provider at this time (it's been fixed since then NetApp/trident#134). The
mountstats metrics don't fill the same need and are superfluous.

This PR Cherry-picks https://github.com/openshift/cluster-monitoring-operator/pull/1139.


simonpasquier:
/lgtm

Thanks! it needs Bugzilla 1955469 to be fixed and verified in the release-4.7 branch first but the PR itself is ok.

arajkumar:
> /lgtm
> 
> Thanks! it needs Bugzilla 1955469 to be fixed and verified in the release-4.7 branch first but the PR itself is ok.

Thanks @simonpasquier for reviewing this. Shall I pull the 4.7 backport into my radar as well?

juzhao:
create nfs sc and create pvc based on it, checked no metric name with prefix node_mountstats_nfs
/label qe-approved

PR #1158, opened 2021-05-12
Consolidate memory requests for all resources

fpetkovski:
The memory requests for resources created by CMO are slightly misaligned with the actual usage in OpenShift clusters.
This change adjusts the requests based on data gathered from CI runs and from the daily OpenShift cluster.

Please refer to the data from a single CI run. A negative number means the container uses less than it requests. A positive number means the container uses more than it requests.
CMO resource usage


jsonnet/openshift-state-metrics.libsonnet
```diff
@@ -1,4 +1,20 @@
 function(params) {
+  // override_memory sets the memory request for a container to the given value.
```
fpetkovski, line 2: I am just starting with jsonnet. If there's a better way to accomplish this please make a suggestion :)

simonpasquier: you should be able to modify the limit using `$._config.baseMemory`:

https://github.com/openshift/openshift-state-metrics/blob/101429149266e2c86a41b82baf47a033c9a93b02/jsonnet/openshift-state-metrics.libsonnet#L12-L15

fpetkovski: Good point, should be there now

dgrisonnet:
@fpetkovski I think it is worth having this improvement in 4.8, would you mind opening a BZ and link it to this PR?

fpetkovski:
> @fpetkovski I think it is worth having this improvement in 4.8, would you mind opening a BZ and link it to this PR?

Sure thing, here is the BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1962261.
Let me know if more information is needed

dgrisonnet:
/retitle Bug 1962261: jsonnet: consolidate memory requests for all resources

PR #1160, opened 2021-05-14
Bug 1955462: drop container_memory_failures_total metric backport

slashpai:
The container_memory_failures_total metric isn't used anywhere and it generates at least 4 series per container.


jsonnet/prometheus.jsonnet
```diff
@@ -233,7 +233,22 @@ local metrics = import 'telemeter-client/metrics.jsonnet';
...
+                  metricRelabelings+: [
+                    {
+                      sourceLabels: ['__name__'],
+                      action: 'drop',
+                      regex: 'container_memory_failures_total',
+                    },
+                  ],
+                }
+              else
+                {}
```
dgrisonnet, line 250: this block needs to be indented once

dgrisonnet (commented):
Just a small nit, otherwise it looks good to me

PR #1164, opened 2021-05-17
Bug 1953798: Remove the upstream KubeAPIErrorBudgetBurn alerts

ihcsim:
This PR removes the upstream `KubeAPIErrorBudgetBurn` alerts and their
recording rules. These rules have been moved to the
openshift/cluster-kube-apiserver-operator repository. See
https://github.com/openshift/cluster-kube-apiserver-operator/pull/1126.


ihcsim:
@openshift/openshift-team-monitoring PTAL. Thanks!

cc @s-urbaniak

s-urbaniak:
lgtm from my side :+1: 
/approve

s-urbaniak:
leaving final pair of eyes to monitoring team.

simonpasquier:
/retitle Bug 1953798: Remove the upstream KubeAPIErrorBudgetBurn alerts

simonpasquier:
/lgtm
thanks!

pkg/manifests/manifests.go
```diff
@@ -2294,7 +2294,9 @@ func (f *Factory) ControlPlanePrometheusRule() (*monv1.PrometheusRule, error) {
 			switch g.Name {
 			case "kubernetes-system-apiserver",
 				"kubernetes-system-controller-manager",
-				"kubernetes-system-scheduler":
+				"kubernetes-system-scheduler",
+				"kubernetes-apiserver-slos",
+				"kubernetes-apiserver.rules":
```
dgrisonnet, line 2299: Since these rules were removed from the manifest, there is no need to skip them since they don't exist anymore.
I'm pretty sure it is the same for the two others.

s-urbaniak: good catch! :+1:

ihcsim: got it! will fix. yeah, i wasn't sure whether changing this file is necessary.

ihcsim: to confirm, the changes in this file isn't necessary, but `jsonnet/patch-rules.libsonnet` still needs to be updated, right?

dgrisonnet: yes, the changes in `jsonnet/patch-rules.libsonnet` will allow you to prevent the rules from being generated in the manifests. Meanwhile the changes here were only applied when reading the manifest. Since you updated it in a way that it removed the rules, they won't be seen at this level anymore.

wking:
Tide seems to have gotten lost.

/retest

simonpasquier:
@ihcsim can you check with DPTP what happens with the tests? They seem to be stuck...

PR #1168, opened 2021-05-18
Bug 1957190: use NamespaceSelector instead of explicit allow/deny list

simonpasquier:
Instead of the `--namespaces` and `deny-namespaces` flags to specify
which namespaces the Prometheus operator should watch/not watch, we
configure the Prometheus resources with the proper namespace selectors.
In practice, the Prometheus `k8s` resource looks for all namespaces with
the `openshift.io/cluster-monitoring=true` label and The Prometheus
`user-workload` resource for all namespaces without the
`openshift.io/cluster-monitoring=true` label.
    
The goal is to decrease the number of listers/watchers that the operator
needs to maintain which should reduce the load on the API server.
Concretely with the `--namespaces` flag, the operator would create
40 listers/watchers (1 per namespace) per watched object type (secret,
prometheus, statefulset, ...). With namespace selectors, the
operator needs only 1 lister/watcher per object type.

### DIFF
--- a/jsonnet/prometheus-operator-user-workload.libsonnet
+++ b/jsonnet/prometheus-operator-user-workload.libsonnet
@@ -57,7 +57,6 @@ function(params)
                         function(arg) !std.startsWith(arg, '--kubelet-service'),
                         super.args,
                       ) + [
-                        '--deny-namespaces=' + cfg.denyNamespace,
                         '--prometheus-instance-namespaces=' + cfg.namespace,
                         '--alertmanager-instance-namespaces=' + cfg.namespace,
                         '--thanos-ruler-instance-namespaces=' + cfg.namespace,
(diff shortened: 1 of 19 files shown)


jsonnet/prometheus-operator-user-workload.libsonnet
```diff
@@ -57,7 +57,6 @@ function(params)
                         function(arg) !std.startsWith(arg, '--kubelet-service'),
                         super.args,
                       ) + [
-                        '--deny-namespaces=' + cfg.denyNamespace,
```
simonpasquier, line 60: we could keep the `--deny-namespaces` flag.

jan--f:
/retitle [WIP] Bug 1957190: use NamespaceSelector instead of explicit allow/deny list

jan--f:
The changes look great, love the code deletion. It would be awesome to note _why_ this reduces the watcher counts in the commit message, just to be nice to future us.

simonpasquier:
@jan-f I've updated the commit message in https://github.com/openshift/cluster-monitoring-operator/pull/1168/commits/721d74d769332cb7d478d58c59402097bfb4425b

PR #1170, opened 2021-05-19
Bug 1949519: Add DELETECOLLECTION to long-running apiserver verbs

arajkumar:
This PR takes https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/375 the changes and apply it directly into 3.11 downstream branch.


Makefile
```diff
@@ -45,8 +45,8 @@ push: container
 
 clean:
 	rm -f $(BIN)
-	go clean -r $(MAIN_PKG)
-	docker images -q $(REPO) | xargs --no-run-if-empty docker rmi --force
+	-go clean -r $(MAIN_PKG)
+	-docker images -q $(REPO) | xargs --no-run-if-empty docker rmi --force
```
arajkumar, line 49: Non zero return code from these 2 commands are ignored as it would always fail on the fresh build.

paulfantom (approved):
lgtm

PR #1173, opened 2021-05-20
Bug 1955478: fix regex patterns in kube-state-metrics deny list

simonpasquier:

### DIFF
--- a/assets/kube-state-metrics/deployment.yaml
+++ b/assets/kube-state-metrics/deployment.yaml
@@ -36,8 +36,8 @@ spec:
         - --metric-labels-allowlist=pods=[*],node=[*]
         - |
           --metric-denylist=
-          kube_*_created,
-          kube_*_metadata_resource_version,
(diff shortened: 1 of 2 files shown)


lihongyan1:
/label qe approved

`
$token=`oc sa get-token prometheus-k8s -n openshift-monitoring`
$ oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- curl -k -H "Authorization: Bearer $token" 'https://prometheus-k8s.openshift-monitoring.svc:9091/api/v1/label/__name__/values' | jq | grep -e kube_|grep created 
return no result
$ oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- curl -k -H "Authorization: Bearer $token" 'https://prometheus-k8s.openshift-monitoring.svc:9091/api/v1/label/__name__/values' | jq | grep -e kube_|grep _metadata_resource_version
return no result
`

PR #1175, opened 2021-05-21
Bug 1962172: increase scrape timeouts for kubelet endpoints

simonpasquier:

### DIFF
--- a/assets/control-plane/service-monitor-kubelet.yaml
+++ b/assets/control-plane/service-monitor-kubelet.yaml
@@ -50,6 +50,7 @@ spec:
       - __metrics_path__
       targetLabel: metrics_path
     scheme: https
+    scrapeTimeout: 30s
     tlsConfig:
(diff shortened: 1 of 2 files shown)


simonpasquier:
/hold cancel

@paulfantom given that it's blocking SNO, I took the option of patching in CMO but it might make sense to bring this change upstream, wdyt?

On a small cluster with 3 worker nodes and no load (60 pods max per node), this is how the scrape duration for kubelet looks like

image

And this is how it looks like in the SNO case (about 200 pods on a single node)

image

omertuc:
Something prevented the tests from running on that run, but looks like the timeout was increased successfully:
image

/retest

paulfantom:
#1176 was fixing segfaults

/unhold
/lgtm

simonpasquier:
From the latest `e2e-aws-single-node` run, the scrape duration hovers between 5 seconds and 30 seconds:

image

@rphillips I would still recommend to investigate why it takes so long for the kubelet to serve metrics...

PR #1179, opened 2021-05-25
Bug 1961081: update PodDisruptionBudget api version to policy/v1

fpetkovski:
The policy/v1 `apiVersion` is supported starting from Kubernetes 1.21.
As a consequence, we cannot make the change directly in kube-prometheus
since it would break backward compatibility.


simonpasquier (commented):
/lgtm
/hold
Leaving @paulfantom the final decision

PR #1183, opened 2021-05-27
Bug 1964334: guard monitoring related /metrics endpoint from unauthorised access

arajkumar:
Currently the /metrics endpoint of alertmanager/prometheus/thanos-querier are accessible without any authentication. This must be fixed to avoid any leakage of any sensitive metrics details.

### DIFF
--- a/assets/alertmanager/alertmanager.yaml
+++ b/assets/alertmanager/alertmanager.yaml
@@ -39,7 +39,6 @@ spec:
     - -openshift-service-account=alertmanager-main
     - -openshift-ca=/etc/pki/tls/cert.pem
     - -openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
-    - -skip-auth-regex=^/metrics
     env:
     - name: HTTP_PROXY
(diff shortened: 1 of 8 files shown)


simonpasquier:
This e2e origin test [1] would need to be changed because it assumes that `/metrics` doesn't require authentication.
 
[1] https://github.com/openshift/origin/blob/a05d969809318bed5a98c5c19773a7e97ba4151c/test/extended/prometheus/prometheus.go#L268-L294

arajkumar:
> This e2e origin test [1] would need to be changed because it assumes that `/metrics` doesn't require authentication.
> 
> [1] https://github.com/openshift/origin/blob/a05d969809318bed5a98c5c19773a7e97ba4151c/test/extended/prometheus/prometheus.go#L268-L294

Thanks @simonpasquier. I raised a PR to fix the test.
https://github.com/openshift/origin/pull/26187

simonpasquier:
/lgtm
@arajkumar no need to worry about `e2e-aws-single-node` it is optional :)

PR #1176, opened 2021-05-24
Bug 1963833: don't attempt to delete nil PodDisruptionBudget object

omertuc:
`NewPodDisruptionBudget` may return nil PodDisruptionBudget when `!f.infrastructure.HighlyAvailableInfrastructure()`:

https://github.com/openshift/cluster-monitoring-operator/blob/cddf199b164c64b365867cdcc449e6c75825337a/pkg/manifests/manifests.go#L2430-L2432

this would cause a segfault later when `DeletePodDisruptionBudget` was
called with that nil `PodDisruptionBudget`, causing Single Node clusters
to fail to install.

* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

PR #1184, opened 2021-05-27
Bug 1961081: update pdb apiversion to v1

fpetkovski:
Pull request #1176 updated the assets managed by the operator
to use the policy/v1 version for all pod disruption budgets.
This was not sufficient to make the operator use the v1 version.

This commit updates the operator code to use the v1 client-go APIs
to make sure no deprecation warnings show in the logs.


simonpasquier (commented):
/lgtm

PR #1185, opened 2021-05-27
Bug 1965367: pkg/operator: Fix typo in etcd-metric-serving-ca resource name

bison:
It looks like when the watch for the etcd-metric-serving-ca ConfigMap
was added in 632cddad, there was a typo in the name used for the
watch, but not elsewhere, e.g. when actually retrieving the object.


dgrisonnet:
Good catch :+1:! Would you mind opening a BZ and linking it to this PR so that we can get this fix in 4.8?

/lgtm

simonpasquier:
Nice catch!

bison:
I actually wonder if this needs to be back-ported at all. It's possible this causes missed updates and gaps in metrics for etcd.

PR #1193, opened 2021-06-03
Bug 1949840: Improve update and status reporting

jan--f:
This PR adds Three commits, that aim to improve our reconcile loop:

1. Update `prometheus-operator` first and wait for it to finish before updating the other operands concurrently.
2. Allow the reconcile loop three attempts before calling `SetFailed` and thus reporting `available=false`.
3. Back-off more aggressively in case of failures. The `DefaultControllerRateLimiter`s default back-off is a bit slow and it also adds per bucket back-off tracking, which we don't use due to only ever queuing one event. Instead this introduces an instance of `ItemExponentialFailureRateLimiter`, which is also used by `DefaultControllerRateLimiter`, but with more aggressive rate limiting settings.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -443,24 +443,29 @@ func (o *Operator) sync(key string) error {
 
 	tl := tasks.NewTaskRunner(
 		o.client,
-		[]*tasks.TaskSpec{
-			tasks.NewTaskSpec("Updating Prometheus Operator", tasks.NewPrometheusOperatorTask(o.client, factory)),
-			tasks.NewTaskSpec("Updating user workload Prometheus Operator", tasks.NewPrometheusOperatorUserWorkloadTask(o.client, factory, config)),
-			tasks.NewTaskSpec("Updating Cluster Monitoring Operator", tasks.NewClusterMonitoringOperatorTask(o.client, factory)),
-			tasks.NewTaskSpec("Updating Grafana", tasks.NewGrafanaTask(o.client, factory, config)),
-			tasks.NewTaskSpec("Updating Prometheus-k8s", tasks.NewPrometheusTask(o.client, factory, config)),
-			tasks.NewTaskSpec("Updating Prometheus-user-workload", tasks.NewPrometheusUserWorkloadTask(o.client, factory, config)),
-			tasks.NewTaskSpec("Updating Alertmanager", tasks.NewAlertmanagerTask(o.client, factory)),
-			tasks.NewTaskSpec("Updating node-exporter", tasks.NewNodeExporterTask(o.client, factory)),
-			tasks.NewTaskSpec("Updating kube-state-metrics", tasks.NewKubeStateMetricsTask(o.client, factory)),
-			tasks.NewTaskSpec("Updating openshift-state-metrics", tasks.NewOpenShiftStateMetricsTask(o.client, factory)),
-			tasks.NewTaskSpec("Updating prometheus-adapter", tasks.NewPrometheusAdapterTaks(o.namespace, o.client, factory)),
-			tasks.NewTaskSpec("Updating Telemeter client", tasks.NewTelemeterClientTask(o.client, factory, config)),
-			tasks.NewTaskSpec("Updating configuration sharing", tasks.NewConfigSharingTask(o.client, factory, config)),
-			tasks.NewTaskSpec("Updating Thanos Querier", tasks.NewThanosQuerierTask(o.client, factory, config)),
-			tasks.NewTaskSpec("Updating User Workload Thanos Ruler", tasks.NewThanosRulerUserWorkloadTask(o.client, factory, config)),
-			tasks.NewTaskSpec("Updating Control Plane components", tasks.NewControlPlaneTask(o.client, factory, config)),
-		},
+		// update prometheus-operator before anything else because it is responsible for managing many other resources (e.g. Prometheus, Alertmanager, Thanos Ruler, ...).
+		tasks.NewTaskGroup(
+			[]*tasks.TaskSpec{
+				tasks.NewTaskSpec("Updating Prometheus Operator", tasks.NewPrometheusOperatorTask(o.client, factory)),
+			}),
+		tasks.NewTaskGroup(
+			[]*tasks.TaskSpec{
+				tasks.NewTaskSpec("Updating user workload Prometheus Operator", tasks.NewPrometheusOperatorUserWorkloadTask(o.client, factory, config)),
+				tasks.NewTaskSpec("Updating Cluster Monitoring Operator", tasks.NewClusterMonitoringOperatorTask(o.client, factory)),
+				tasks.NewTaskSpec("Updating Grafana", tasks.NewGrafanaTask(o.client, factory, config)),
+				tasks.NewTaskSpec("Updating Prometheus-k8s", tasks.NewPrometheusTask(o.client, factory, config)),
+				tasks.NewTaskSpec("Updating Prometheus-user-workload", tasks.NewPrometheusUserWorkloadTask(o.client, factory, config)),
+				tasks.NewTaskSpec("Updating Alertmanager", tasks.NewAlertmanagerTask(o.client, factory)),
+				tasks.NewTaskSpec("Updating node-exporter", tasks.NewNodeExporterTask(o.client, factory)),
+				tasks.NewTaskSpec("Updating kube-state-metrics", tasks.NewKubeStateMetricsTask(o.client, factory)),
+				tasks.NewTaskSpec("Updating openshift-state-metrics", tasks.NewOpenShiftStateMetricsTask(o.client, factory)),
+				tasks.NewTaskSpec("Updating prometheus-adapter", tasks.NewPrometheusAdapterTaks(o.namespace, o.client, factory)),
+				tasks.NewTaskSpec("Updating Telemeter client", tasks.NewTelemeterClientTask(o.client, factory, config)),
+				tasks.NewTaskSpec("Updating configuration sharing", tasks.NewConfigSharingTask(o.client, factory, config)),
+				tasks.NewTaskSpec("Updating Thanos Querier", tasks.NewThanosQuerierTask(o.client, factory, config)),
+				tasks.NewTaskSpec("Updating User Workload Thanos Ruler", tasks.NewThanosRulerUserWorkloadTask(o.client, factory, config)),
+				tasks.NewTaskSpec("Updating Control Plane components", tasks.NewControlPlaneTask(o.client, factory, config)),
+			}),
 	)
 
 	klog.Info("Updating ClusterOperator status to in progress.")
@@ -471,9 +476,7 @@ func (o *Operator) sync(key string) error {
 
 	taskName, err := tl.RunAll()
 	if err != nil {
-		klog.Infof("Updating ClusterOperator status to failed. Err: %v", err)
-		failedTaskReason := strings.Join(strings.Fields(taskName+"Failed"), "")
-		reportErr := o.client.StatusReporter().SetFailed(err, failedTaskReason)
+		reportErr := o.reportError(err, taskName)
 		if reportErr != nil {
 			klog.Errorf("error occurred while setting status to failed: %v", reportErr)
 		}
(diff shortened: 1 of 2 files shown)


pkg/operator/operator.go
```diff
@@ -469,18 +469,30 @@ func (o *Operator) sync(key string) error {
 		klog.Errorf("error occurred while setting status to in progress: %v", err)
 	}
 
+	promOpTaskName := "Updating Prometheus Operator"
```
simonpasquier, line 472: what if `TaskRunner` can take an ordered list of tasks? e.g. 

```golang
tl := NewTaskRunner(client *client.Client)

# first group of tasks to execute.
tl.AddTasks(
    tasks.NewTaskSpec("Updating Prometheus Operator", tasks.NewPrometheusOperatorTask(o.client, factory)),
)
# second group of tasks to execute.
tl.AddTasks(
    tasks.NewTaskSpec("Updating user workload Prometheus Operator", tasks.NewPrometheusOperatorUserWorkloadTask(o.client, factory, config)),
    ...
)

taskName, err := tl.RunAll()
```

pkg/operator/operator.go
```diff
@@ -489,6 +501,16 @@ func (o *Operator) sync(key string) error {
 	return nil
 }
 
+func (o *Operator) reportError(err error, taskName string) error {
+	if o.failedReconcileAttempts < 3 {
+		o.failedReconcileAttempts++
+		return nil
+	}
```
simonpasquier, line 508: should we reset `o.failedReconcileAttempts`?

jan--f: No, I think we only want to do reset once we were successful. This happens [here](https://github.com/openshift/cluster-monitoring-operator/pull/1193/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR495)

pkg/operator/operator.go
jan--f: Thought about it, but figured as long as we only have one task to run before, I might as well use whats already there.

simonpasquier: Thinking about it more, I'd rather see us only reuse TaskRunner (even for only one task):

```golang
# first group of tasks to execute
tl := NewTaskRunner(o.client,[]*tasks.TaskSpec{
    tasks.NewTaskSpec("Updating Prometheus Operator", tasks.NewPrometheusOperatorTask(o.client, factory)),
})

taskName, err := tl.RunAll()
...

# second group of tasks to execute.
tl = NewTaskRunner(o.client,[]*tasks.TaskSpec{
    tasks.NewTaskSpec("Updating user workload Prometheus Operator", tasks.NewPrometheusOperatorUserWorkloadTask(o.client, factory, config)),
    ...
)

taskName, err = tl.RunAll()
...
```

pkg/operator/operator.go
simonpasquier: So the operator would retry up to 3 times before reporting available=false/degraded=true and the next sync() operations wouldn't retry?

```
Updating ClusterOperator status to failed after retrying 3 times. Err: xxx
...
Updating ClusterOperator status to failed after retrying 4 times. Err: xxx
...
Updating ClusterOperator status to failed after retrying 5 times. Err: xxx
```

pkg/operator/operator.go
jan--f: Opted for the first idea of a list of task lists. This way we don't need to add error handling if the first task list has a failed task.

pkg/operator/operator.go
jan--f: The intention is that it allows for 3 failed attempts before reporting `available=false/degraded=true`. All error cause a retry to sync the event in question though. `o.failedReconcileAttempts` is really only affecting if/when the operator status changes...unless I'm missing something.

The status and `o.failedReconcileAttempts` is reset to `0` and `available=true/degraded=false` respectively once a call to sync completes without error

pkg/operator/operator.go
```diff
@@ -170,7 +172,7 @@ func New(
 		namespace:                 namespace,
 		namespaceUserWorkload:     namespaceUserWorkload,
 		client:                    c,
-		queue:                     workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "cluster-monitoring"),
+		queue:                     workqueue.NewNamedRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(50*time.Millisecond, 3*time.Minute), "cluster-monitoring"),
```
ianbillett, line 175: this change doesn't appear to be related to the stated goals of the PR? This reads like a different back-off mechanism?
> * Update prometheus-operator first and wait for it to finish before updating the other operands concurrently.
> * Allow the reconcile loop three attempts before calling SetFailed and thus reporting available=false.

pkg/tasks/tasks.go
```diff
@@ -21,43 +21,53 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// TaskRunner executes groups of tasks, the groups sequentially, each group concurrently
```
ianbillett, line 24: nit: semantically, the struct itself does not execute the group of tasks -I would guess the `RunAll` function does that?

ianbillett (commented):
I don't have much context so a couple of points about readability :)

pkg/operator/operator.go
jan--f: Ah yes, I'll update the PR description. The respective [commit](https://github.com/openshift/cluster-monitoring-operator/pull/1193/commits/62e2d9dcd878389fb7d8e368e2b8727bf845da9f) has details.

Thanks, good catch.

pkg/tasks/tasks.go
jan--f: Fair point, add more comments and shuffled stuff around.

pkg/tasks/tasks.go
```diff
@@ -21,43 +21,60 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// TaskRunner manages lists of tasks. Through the RunAll method tasks are
+// executed, the lists sequentially, each list of tasks concurrently.
 type TaskRunner struct {
-	client *client.Client
-	tasks  []*TaskSpec
+	client    *client.Client
+	taskLists [][]*TaskSpec
```
fpetkovski, line 28: What is the reason for changing the type here?

pkg/tasks/tasks.go
```diff
@@ -21,43 +21,60 @@ import (
...
 }
 
+// NewTaskRunner returns a task runner. tasks is the first task list that will
```
fpetkovski, line 31: What do you mean by `tasks is the first task`?

jan--f: `tasks` is referring to the function argument. And it is the first task list that gets processed.

pkg/tasks/tasks.go
jan--f: Does the [commit message for this change](https://github.com/openshift/cluster-monitoring-operator/pull/1193/commits/d6d1510c5253e1ec2cdeedf3b2d5bbbb8345697f) make things clearer?

Part of this PR is to extend `TaskRunner` to process multiple groups of tasks as a way to enforce some order.

fpetkovski: Ah yes, that makes sense. To make sure this is reflected in the codebase, can we introduce a new abstraction (e.g. `TaskGroup`) where we can document this as part of the godoc?

fpetkovski: This is definitely just optional and if it is too big of a change it's fine to keep the existing solution.

simonpasquier: I agree with @fpetkovski that it might be good to introduce another struct like `TaskGroup` for clarity.

pkg/operator/operator.go
simonpasquier: :+1:

pkg/operator/operator.go
```diff
@@ -489,6 +489,18 @@ func (o *Operator) sync(key string) error {
 	return nil
 }
 
+func (o *Operator) reportError(err error, taskName string) {
+	if o.failedReconcileAttempts > 3 {
+		klog.Infof("Updating ClusterOperator status to failed after retrying %v times. Err: %v", o.failedReconcileAttempts, err)
```
simonpasquier, line 494: I might be worth logging the error message at every step? IIUC the first 3 errors would be missing currently.

pkg/tasks/tasks.go
```diff
@@ -21,35 +21,64 @@ import (
...
-	client *client.Client
-	tasks  []*TaskSpec
+	client     *client.Client
+	taskGroups []*TaskGroup
 }
 
-func NewTaskRunner(client *client.Client, tasks []*TaskSpec) *TaskRunner {
+// NewTaskRunner returns a task runner. tasks is the first task group that will
+// be executed, before any list added via AppendTaskGroup.
+func NewTaskRunner(client *client.Client, tasks *TaskGroup) *TaskRunner {
```
simonpasquier, line 33: We could make it a variadic function and get rid of `AppendTaskGroup`?

```suggestion
func NewTaskRunner(client *client.Client, taskGroups ...*TaskGroup) *TaskRunner {
```

pkg/operator/operator.go
```diff
@@ -489,6 +491,20 @@ func (o *Operator) sync(key string) error {
 	return nil
 }
 
+func (o *Operator) reportError(err error, taskName string) {
```
simonpasquier, line 494: (nit) small suggestion to make it easier to read IMHO

```golang
func (o *Operator) reportError(err error, taskName string) {
    defer o.failedReconcileAttempts++

    klog.Infof("Operator reconciliation failed, retrying. Err: %v", err)
    if o.failedReconcileAttempts < 3 {
        // Don't update the ClusterOperator status until 3 retries have been attempted.
        return
    }

    ...
```

pkg/tasks/tasks.go
```diff
@@ -21,35 +21,64 @@ import (
...
 		// shadow vars due to concurrency
 		ts := ts
 		i := i
 
 		g.Go(func() error {
-			klog.V(2).Infof("running task %d of %d: %v", i+1, len(tl.tasks), ts.Name)
-			err := tl.ExecuteTask(ts)
-			klog.V(2).Infof("ran task %d of %d: %v", i+1, len(tl.tasks), ts.Name)
+			klog.V(2).Infof("running task %d of %d: %v", i+1, tgLength, ts.Name)
+			err := ts.Task.Run()
```
simonpasquier, line 73: I suppose that `ExecuteTask()` can be removed then?

pkg/tasks/tasks.go
jan--f: :+1: Indeed we could...and probably should too.

pkg/tasks/tasks.go
jan--f: Good catch, yes that was meant to be removed.

pkg/operator/operator.go
jan--f: Hm only that `defer` doesn't work. It expects a function call, whereas we have a statement. Not sure wrapping it in a function helps readability here.

pkg/operator/operator.go
```diff
@@ -441,26 +439,31 @@ func (o *Operator) sync(key string) error {
 	}
 	factory := manifests.NewFactory(o.namespace, o.namespaceUserWorkload, config, o.loadInfrastructureConfig(), proxyConfig, o.assets)
 
+	// update prometheus-operator before anything else
```
simonpasquier, line 442: Maybe move this right before the first `tasks.NewTaskGroup`?

```suggestion
	// Update prometheus-operator before anything else because it is responsible for managing many other resources (e.g. Prometheus, Alertmanager, Thanos Ruler, ...).
```

pkg/operator/operator.go
simonpasquier: my bad I didn't verify that my pseudo code worked :) maybe we could still split the logs (1 for logging the error and the other telling that we update the cluster operator status)?

```golang
    klog.Infof("Reconciliation failed (%d retries). Err: %v", o.failedReconcileAttempts, err)
    if o.failedReconcileAttempts > 3 {
        // Only update the ClusterOperator status after 3 retries have been attempted to avoid flapping status.
        klog.Info("Updating ClusterOperator status to failed ")
        ...
    }
```

jan--f: Changes the logging as per your comment. I also changed the condition to 2 (we start counting at 0 so `> 3` was incorrect and to an equality check. No need to set `degraded` repeatedly I figure.

PR #1194, opened 2021-06-03
pkg/client: Retry on API errors in polling methods

bison:
The client package has several methods that poll to wait for certain
conditions to become true, but they all fail immediately when the
initial request(s) to the API server return an error.  This changes
that behavior to ensure the API requests are retried as well, and
simply logs the error returned by the API server at debug level.

The main goal here is to help with e2e test flakiness. We have a lot of test failures that seem to be the result of transient API server or network issues. The tests that use the methods that poll often fail needlessly when they could just retry. Having said that, I'm a bit concerned this could have unintended consequences. For instance, things now taking multiple minutes to retry when there is a legitimate configuration issue.


simonpasquier:

paulfantom (commented):
/lgtm
/retest

PR #1150, opened 2021-05-06
Bug 1957703: [4.7]: Add hard anti-affinity constraints to Prometheuses

dgrisonnet:
Backport of https://github.com/openshift/cluster-monitoring-operator/pull/1135 to release-4.7.

I couldn't cherry-pick the commit because of the changes made to kube-prometheus in 4.8.

* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

PR #1200, opened 2021-06-04
Bug 1967966: Switch back anti-affinity to soft from hard for Prometheus

simonpasquier:
Revert #1150 because it can cause regressions for clusters using persistent storage. More details in https://bugzilla.redhat.com/show_bug.cgi?id=1967614


derekwaynecarr:
(Patch-Manager) manually removing the bugzilla/invalid-bug label and applying bugzilla/valid-bug label so this PR can land in the release branch without blocking on the upstream bug as this is needed to ship.

PR #1202, opened 2021-06-07
Bug 1955589: add PodDisruptionBudget to ThanosQuerier to meet HA requ…

raptorsun:
…irements

To guarantee high availability during upgrade of ThanosQuerier, we set a PodDisruptionBudget to prevent all its pods go offline at the same time.

### DIFF
--- a/assets/thanos-querier/pod-disruption-budget.yaml
+++ b/assets/thanos-querier/pod-disruption-budget.yaml
@@ -1,0 +1,17 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  labels:
+    app.kubernetes.io/component: query-layer
+    app.kubernetes.io/instance: thanos-querier
+    app.kubernetes.io/name: thanos-query
+    app.kubernetes.io/version: 0.20.2
+  name: thanos-querier-pdb
+  namespace: openshift-monitoring
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: query-layer
+      app.kubernetes.io/instance: thanos-querier
+      app.kubernetes.io/name: thanos-query
--- a/jsonnet/thanos-querier.libsonnet
+++ b/jsonnet/thanos-querier.libsonnet
@@ -517,4 +517,21 @@ function(params)
         },
       },
     },
+
+    podDisruptionBudget: {
+      apiVersion: 'policy/v1',
+      kind: 'PodDisruptionBudget',
+      metadata: {
+        name: 'thanos-querier-pdb',
+        namespace: cfg.namespace,
+        labels: tq.config.commonLabels,
+      },
+      spec: {
+        minAvailable: 1,
+        selector: {
+          matchLabels: tq.config.podLabelSelector,
+        },
+
+      },
+    },
   }
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -193,6 +193,7 @@ var (
 	TelemeterClientServingCertsCABundle   = "telemeter-client/serving-certs-ca-bundle.yaml"
 
 	ThanosQuerierDeployment           = "thanos-querier/deployment.yaml"
+	ThanosQuerierPodDisruptionBudget  = "thanos-querier/pod-disruption-budget.yaml"
 	ThanosQuerierService              = "thanos-querier/service.yaml"
 	ThanosQuerierServiceMonitor       = "thanos-querier/service-monitor.yaml"
 	ThanosQuerierPrometheusRule       = "thanos-querier/prometheus-rule.yaml"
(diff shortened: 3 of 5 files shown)


simonpasquier:
/hold cancel

scratch that, I confused Thanos rule with Thanos querier :)

simonpasquier:
Hmm not sure why but `e2e-agnostic` failed with 

> alert PodDisruptionBudgetAtLimit fired for 1357 seconds with labels: {namespace="openshift-monitoring", poddisruptionbudget="thanos-querier-pdb", severity="warning"}

raptorsun:
> Hmm not sure why but `e2e-agnostic` failed with
> 
> > alert PodDisruptionBudgetAtLimit fired for 1357 seconds with labels: {namespace="openshift-monitoring", poddisruptionbudget="thanos-querier-pdb", severity="warning"}

This alert is generated by having the number of healthy pod of thanos-querier equals to the minimal number of healthy pod. In this case max unavailable is set to 1, so 2 replica - 1 offline = 1 the minimal number of healthy pod. ([link to alert condition](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml#L26))

I am going to set replica number of thanos-querier to 3 so it is resistant to one additional failure pod during upgrade, though it will consume a bit more resources.

raptorsun:
updating e2e test with 3 replicas for ThanosQuerier.

jsonnet/thanos-querier.libsonnet
```diff
@@ -517,4 +517,21 @@ function(params)
...
+    podDisruptionBudget: {
+      apiVersion: 'policy/v1',
+      kind: 'PodDisruptionBudget',
+      metadata: {
+        name: 'thanos-querier-pdb',
+        namespace: cfg.namespace,
+        labels: tq.config.commonLabels,
+      },
+      spec: {
+        maxUnavailable: 1,
```
dgrisonnet, line 530: we want `minAvailable: 1` to make sure that thanos querier is always available

jsonnet/main.jsonnet
```diff
@@ -294,7 +294,7 @@ local inCluster =
       thanosQuerier: $.values.thanos {
         name: 'thanos-querier',
         namespace: $.values.common.namespace,
-        replicas: 2,
+        replicas: 3,
```
dgrisonnet, line 297: with `minAvailable: 1` we don't need to increase the number of replicas

pkg/manifests/manifests.go
```diff
@@ -2740,6 +2741,10 @@ func (f *Factory) NewValidatingWebhook(manifest io.Reader) (*admissionv1.Validat
 	return NewValidatingWebhook(manifest)
 }
 
+func (f *Factory) ThanosQuerierPodDisruptionBudget() (*policyv1.PodDisruptionBudget, error) {
+	return f.NewPodDisruptionBudget(f.assets.MustNewAssetReader(ThanosQuerierPodDisruptionBudget))
```
dgrisonnet, line 2745: we also need to make sure the PDB namespace is the same as the one from the factory. You can look at https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/manifests/manifests.go#L1622-L1633 for an example.

jsonnet/main.jsonnet
raptorsun: Thanks, I will update both parts accordingly 👍

raptorsun:
The alert PodDisruptionBudgetLimit still fires up during upgrade with minAvailable=1 and replica=2. Shall we increase replica to 3 or silence that alert during upgrade (more risk to have a total outage on this service though)? @dgrisonnet  @simonpasquier

assets/thanos-querier/pod-disruption-budget.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  labels:
+    app.kubernetes.io/component: query-layer
+    app.kubernetes.io/instance: thanos-querier
+    app.kubernetes.io/name: thanos-query
+    app.kubernetes.io/version: 0.20.2
+  name: thanos-querier-pdb
+  namespace: openshift-monitoring
+spec:
+  labelSelector:
```
dgrisonnet, line 12: From the CI must-gather we get that the PDB isn't able to find the thanos-querier pods and we can also see that it doesn't have any selector:
```
- apiVersion: policy/v1
  kind: PodDisruptionBudget                                                                                                                                                                                          
  metadata:
    creationTimestamp: "2021-06-09T10:24:02Z"
    generation: 1
    labels:
      app.kubernetes.io/component: query-layer
      app.kubernetes.io/instance: thanos-querier
      app.kubernetes.io/name: thanos-query
      app.kubernetes.io/version: 0.20.2
    name: thanos-querier-pdb
    resourceVersion: "47099"
    uid: c35f9eba-0950-48bf-a3e4-fe54901dbf1f
  spec:
    minAvailable: 1
  status:
    conditions:
    - lastTransitionTime: "2021-06-09T10:24:02Z"
      message: ""
      observedGeneration: 1
      reason: InsufficientPods
      status: "False"
      type: DisruptionAllowed
    currentHealthy: 0
    desiredHealthy: 1
    disruptionsAllowed: 0
    expectedPods: 0
    observedGeneration: 1
```

```suggestion
  selector:
```
This should fix the CI failures since the pod disruption budget will now be able to find the thanos-querier pods.

PR #1198, opened 2021-06-04
Bug 1967614: Revert anti-affinity to soft

slashpai:
* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

Reverts https://github.com/openshift/cluster-monitoring-operator/pull/1135/files

**Why revert?**

CMO fails repeatedly because the prometheus-k8s statefulset never converges to the desired state
https://bugzilla.redhat.com/show_bug.cgi?id=1967614#c0

PR #1204, opened 2021-06-08
Bug 1967614: Remove PDB for prometheus and alertmanager

slashpai:
We had to remove hard pod anti-affinity in https://github.com/openshift/cluster-monitoring-operator/pull/1198 but since PDB was set for prometheus it won't let reboot node during upgrade hence we cannot have PDB when pod hard anti-affinity rule is not set. Removing it for alertmanager as well since issue can happen for that as well since it can PVC too which causes volume affinity issues


simonpasquier:
I think we need to remove PodDisruptionBudget from `pkg/tasks/prometheus_user_workload.go` too.

simonpasquier (commented):
/lgtm

Thanks a lot @slashpai!

PR #1211, opened 2021-06-11
Bug 1970624: jsonnet: reduce threshold of AggregatedAPIDown

dgrisonnet:
Reduce threshold of AggregatedAPIDown alert from 85 to 70 to be more
resilient to DNS rollouts and CA rotations. A time window of 1m30 was
too tight when facing these, so we want to increase the window to 3m.


simonpasquier (commented):
/lgtm

s-urbaniak:
thank you! :tada:

PR #1212, opened 2021-06-11
allowlist: Add partners operator CPU usage metric

machacekondra:
This PR add new partners:restricted_use_cores:sum metric.

This metric is used by our partners to report the CPU usage of their
operators.


Documentation/data-collection.md
```diff
@@ -332,6 +332,9 @@ data:
     # (@openshift/team-build-api, OpenShift Build API Team) openshift:build_by_strategy:sum measures total number of builds on a cluster, aggregated by build strategy.
     - '{__name__="openshift:build_by_strategy:sum"}'
     #
+    # (partners metrics) The number of the cores used by the partner operator.
```
simonpasquier, line 335: we ask for a GitHub team handle + name here (see examples above).

machacekondra: This is special case, where the metric is emitted by ISV partner operator, so there is no team responsible for the metric. What do you suggest to put here?

bwplotka: Let's talk about consequences. If this metric will start exposing millions series suddenly, who we should talk to fix this or to notify that we are removing this metric due to abuse?

PR #1213, opened 2021-06-11
manifests: Add openshift-partners-metrics namespace

machacekondra:
This PR add a new namespace called openshift-partners-metrics namespace
to the manifest directory.

This namespace will be used by the partners who could define special
metrics to be sent by telemeter.

### DIFF
--- a/manifests/0000_50_cluster-monitoring-operator_01-namespace.yaml
+++ b/manifests/0000_50_cluster-monitoring-operator_01-namespace.yaml
@@ -27,3 +27,18 @@ metadata:
     openshift.io/cluster-monitoring: "true"
     name: openshift-user-workload-monitoring
     network.openshift.io/policy-group: monitoring
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-partners-metrics
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    openshift.io/node-selector: ""
+    include.release.openshift.io/single-node-developer: "true"
+    workload.openshift.io/allowed: "management"
+  labels:
+    openshift.io/cluster-monitoring: "true"
+    name: openshift-partners-metrics
+    network.openshift.io/policy-group: monitoring


bwplotka (approved):
This makes sense to me from high level POV, but let's wait for @jan--f @simonpasquier opinion too 🤗 

Context: https://docs.google.com/document/d/1OsPD05cGpcyqjl89TTOCEVbGWnHqLpNwx-Jee5W_srw (internal doc)

simonpasquier:
I have concerns that we extend metrics collection to "random" services without digging more in the requirements. In particular we haven't defined if:
* "partner" endpoints should only expose metrics over HTTPS?
* Do they require authentication?

machacekondra:
> I have concerns that we extend metrics collection to "random" services without digging more in the requirements. In particular we haven't defined if:
> 
>     * "partner" endpoints should only expose metrics over HTTPS?

I think this should be recommended, and default, but I don't think we should require. Or do you see any reason we should require?

> 
>     * Do they require authentication?

It's not currently as part of the specification, so not currently. Maybe in future.

pgodowski:
@machacekondra @fpetkovski
Representing here IBM as the partner willing to provide the subject metrics. Our joint customers will object with IBM's code accessing namespaces where IBM workload does not run. Specifically, namespace `openshift-partners-metrics` will not be allowed IBM to create any resources there.

Perhaps alternate approach would be via labelling `ServiceMonitor` resources with some label, in conjuction with the allowlist, https://github.com/openshift/cluster-monitoring-operator/pull/1212 ?

machacekondra:
@pgodowski  Currently prometheus-operator is deployed using [serviceMonitorNamespaceSelector](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/prometheus-k8s/prometheus.yaml#L210), so using `serviceMonitorSelector` would change the behavior and [ServiceMonitor](https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/prometheus/operator.go#L1715) would have to be in namespace labeled by `openshift.io/cluster-monitoring: "true"` anyways.

@simonpasquier @bwplotka  @pkliczewski  Do you think we should close this PR and go back to original approach by suggesting to label the specific namespace with label `openshift.io/cluster-monitoring: "true"`` where `ServiceMonitor` object would be created?

@pgodowski Would it be ok to just label IBM namespace with  `openshift.io/cluster-monitoring: "true"`` and create ServiceMonitor there?

pgodowski:
@machacekondra 
Our customers create namespaces upfront and then deploy IBM software there. IBM software do not manage namespaces, at least in the long term we believe apps should not manage namespaces where apps are running in. Labeling might not be ok, as IBM apps do not guarantee such label is not removed by the customer / label is not added upfront and there is a bigger stake on implications where such configuration is not in place (e.g. license compliance challenges).

What our apps are in control are the resources in the namespace, esp. `ServiceMonitor` resources.

Have you considered using [`serviceMonitorSelector`](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#prometheusspec) and what it really takes, from the backwards compatibility / behavior change perspective, given your comment implies it was rejected?

machacekondra:
> @machacekondra
> Our customers create namespaces upfront and then deploy IBM software there. IBM software do not manage namespaces, at least in the long term we believe apps should not manage namespaces where apps are running in. Labeling might not be ok, as IBM apps do not guarantee such label is not removed by the customer / label is not added upfront and there is a bigger stake on implications where such configuration is not in place (e.g. license compliance challenges).

I understand. We do this for the OpenShift layered products like OCS[1], so we thought it could work also for this use case.

https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.2/html-single/deploying_openshift_container_storage/index#installing-openshift-container-storage-operator-using-the-operator-hub_rhocs

> 
> What our apps are in control are the resources in the namespace, esp. `ServiceMonitor` resources.
> 
> Have you considered using [`serviceMonitorSelector`](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#prometheusspec) and what it really takes, from the backwards compatibility / behavior change perspective, given your comment implies it was rejected?

We can't use it because it would break backward compatibility. As stated above we do label namespaces of our layered products. What we could do is to introduce new parameter called something like `useAnyNamespace` which would list all `ServiceMonitors` in all namespaces and combine the list with `ServiceMonitors` from the namespace with specific `label`, we wouldn't break backward compatibility then. @simonpasquier @bwplotka Do you think that is feasible?

pgodowski:
> What we could do is to introduce new parameter called something like `useAnyNamespace` which would list all `ServiceMonitors` in all namespaces and combine the list with `ServiceMonitors` from the namespace with specific label, we wouldn't break backward compatibility then.

Thanks for considering alternatives, @machacekondra

machacekondra:
> > We can't use it because it would break backward compatibility.
> 
> Was it a public interface for non-OCP internal components to rely on namespace label `openshift.io/cluster-monitoring: "true"` ? Or you are considered about OCP components using this namespace label?

Non-default applications(like altready mentioned OCS or OpenShift Virtualization) must label their namepsace where they are installed by label  `openshift.io/cluster-monitoring: "true"` (if it was not done by OpenShift console), so they are included in OpenShift cluster monitoring stack, and you can query them via Prometheus. So yes it is a public interface of the Monitoring stack to include the application's metrics.

bwplotka:
👋🏽 @simonpasquier is OOO until Monday, but from my perspective, this feels very free form. Anyone including customers, all partners can just start adding any metric to platform monitoring, breaking all recording rules / queries or causing DOS or platform prometheus itself with the scraped data.

I wonder if it's not better to pre define ServiceMonitors that takes data from outside to exact name of operator resource and its /metrics page with exactly relabelling controlled by CMO. It's less flexible, but this is the constraint of using the Telemeter part at the current moment. Platform has to decide where the data is pulled from, not any individual user on the cluster. 

Some naive solution is:
* Partners work with us to define correct Service Monitors that will land in  `openshift partners namespace` through CMO directly (maybe under flag). It will target specific operator (operators)
* Platform admin can enable this flag to deploy those service monitors or not.

What's missing in this flow? 🤔

pgodowski:
@machacekondra @bwplotka What is the latest on this one?
My position is still that Partner's (like IBM's) workload should not access 'system' namespace like `openshift-partners-metrics` with write access - and also understand the solution must not be free-form, allowing any app to push metrics to back-end, without any vetting upfront.

I like the proposal of pre-defined `ServiceMonitor`s which are shipped with OCP, based on agreed content with each Partner, mentioned in https://github.com/openshift/cluster-monitoring-operator/pull/1213#issuecomment-867705631

pgodowski:
@machacekondra @bwplotka Where the discussion about the parner metrics ended up with?
What's the best way to have it restarted?

PR #1222, opened 2021-06-16
Bug 1941634: allow only /metrics for kube-rbac-proxy in front of UWM

slashpai:


dgrisonnet:
Looks like I wasn't an approver in 4.6.

/cc @paulfantom @simonpasquier

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 0.40

approved

PR #1223, opened 2021-06-16
Remove upstream UI URLs

PhilipGough:

This is related to [console/#9276](https://github.com/openshift/console/pull/9276) and see [epic 1591](https://issues.redhat.com/browse/MON-1591) for context


simonpasquier:
/lgtm
/hold
this needs https://github.com/openshift/console/pull/9276 (at least) to be merged

simonpasquier:
Ah I realized that we should probably keep the `ConfigSharingTask` task that would simply remove the `openshift-config-managed/monitoring-shared-config` configmap in 4.9 and then we can kill the task in 4.10.

PhilipGough:
@simonpasquier I've addressed you comment (I think) in a second commit just so it is easier to review the change but lets squash them prior to merge.

That change also led me to question how/if we will handle the cleanup of routes or are we going to continue to create those?

PhilipGough:
Closing as obsolete - "the current plan is to keep the routes (because of the multi-cluster console effort which is going to be dev preview in 4.10)"

PR #1227, opened 2021-06-17
Improve monitoring telemetry rules

simonpasquier:
When the Prometheus pods are redeployed and persistent storage is used,
the recording rules used to report telemetry about monitoring would
double count for 5 minutes (at most). It happens because of how
Prometheus handles staleness: the individual metrics would be present
for both the old and new pods if their `instance` labels have changed
(e.g. their IP addresses are different). To avoid this artifact, the
rules aggregate away the `instance` label.

Original `openshift:prometheus_tsdb_head_series:sum` recording rule
image

The modified query on the same dataset
image

### DIFF
--- a/assets/cluster-monitoring-operator/prometheus-rule.yaml
+++ b/assets/cluster-monitoring-operator/prometheus-rule.yaml
@@ -353,12 +353,12 @@ spec:
       record: openshift:build_by_strategy:sum
   - name: openshift-monitoring.rules
     rules:
-    - expr: sum by (job,namespace) (prometheus_tsdb_head_series{namespace=~"openshift-monitoring|openshift-user-workload-monitoring"})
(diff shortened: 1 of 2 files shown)


jan--f:
Is `max` the correct aggregation in all cases? I understand why this is needed and with counters this will work correctly. With `rate` and gauges however this could pick the old value, couldn't it? I don't have a better solution and this is probably splitting hairs, but maybe we should at least leave a comment for future readers?

simonpasquier:
> Is max the correct aggregation in all cases?

Worst case, `max` will overestimate the memory usage until the new Prometheus pod has caught up with all targets and rule evaluations but at least it won't double account. We could use `min()` but it may not be more accurate (if the new pod starts consuming more memory than the old). Happy to add a comment though :)

jan--f (approved):

PR #1229, opened 2021-06-17
Bug 1972076: jsonnet: Disable cpufreq collector in node_exporter

bison:
The node_exporter cpufreq collector seems to be causing high load on
some machines with lots of cores.  This disables it temporarily as a
workaround until the issue can be addressed upstream.  We don't seem
to rely on these metrics anywhere.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1972076


simonpasquier:
/lgtm

even though we're not sure that node_exporter is the root cause for the kernel soft locks, it might amplify the issue.

simonpasquier:
/retitle Bug 1972076: jsonnet: Disable cpufreq collector in node_exporter

PR #1237, opened 2021-06-22
AggregatedAPIDown alert threshold set back to 85%

raptorsun:

This pull request reverts [a patch](https://github.com/openshift/cluster-monitoring-operator/pull/1211) to version 4.8 that mitigates a problem when upgrading monitoring components from 4.7 to 4.8: AggregatedAPIDown is triggered due to a long period of unavailability of prometheus adapter. Both deployments are offline at the same time during the upgrade.

After integrating these 2 PRs([PR1](https://github.com/openshift/cluster-monitoring-operator/pull/1149), [PR2](https://github.com/openshift/cluster-monitoring-operator/pull/1124)) into 4.8 release, Prometheus Adapter is able to keep itself always available during the upgrade process. We can set back the alert threshold to its default value 85%.

### DIFF
--- a/assets/control-plane/prometheus-rule.yaml
+++ b/assets/control-plane/prometheus-rule.yaml
@@ -398,7 +398,7 @@ spec:
           has been only {{ $value | humanize }}% available over the last 10m.
         summary: An aggregated API is down.
       expr: |
-        (1 - max by(name, namespace)(avg_over_time(aggregator_unavailable_apiservice[10m]))) * 100 < 70
+        (1 - max by(name, namespace)(avg_over_time(aggregator_unavailable_apiservice[10m]))) * 100 < 85
       for: 5m
       labels:
         severity: warning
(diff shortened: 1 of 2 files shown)


simonpasquier:
/lgtm
/hold
@dgrisonnet can you have a second look? Once it merges, we need to carefully monitor the [CI logs](https://search.ci.openshift.org/?search=alert+AggregatedAPIDown+fired+for.%2Bmetrics.k8s.io.%2B&maxAge=168h&context=1&type=junit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job) to verify that the `AggregatedAPIDown` alert doesn't start firing again for 4.9 jobs.

dgrisonnet:
Looks good from my side.
I will monitor the CI, but the alert shouldn't fire anymore because of non highly available aggregated APIs since they should all have been made HA in 4.8, so we shouldn't have any disruption during a 4.8 to 4.9 upgrade.

/unhold

PR #1239, opened 2021-06-23
Bug 1974651: Remove :apiserver_v1_image_imports:sum

dmage:
The metric `:apiserver_v1_image_imports:sum` is going to be removed by https://github.com/openshift/cluster-image-registry-operator/pull/700/files


simonpasquier (commented):
/lgtm

PR #1241, opened 2021-06-23
operator: Allow disabling Grafana deployment

bison:
This exposes a config option in the ConfigMap that allows disabling the Grafana deployment.

Summary of the changes:
- Added `Enabled` field to `GrafanaConfig` struct.
- The Grafana task is split into `create()` and `destroy()` methods.
- The htpasswd related bits (volume and container argument) are removed from jsonnet.
- The operator injects these at runtime if Grafana is enabled.
- The shared config task doesn't inject URLs if they are nil.
- Add some basic unit and e2e tests.


bison:
Quick note: There was some offline discussion about the fact that we generally try to keep the jsonnet / assets as close as possible to the default deployment. In this case that would require either needlessly enabling the static password auth when Grafana is disabled, or inverting the logic in `manifests.go` to *remove* things when Grafana is disabled instead of adding them when it is enabled. That is tricky in this case because these things are often buried in nested arrays. IMHO with the comments in the jsonnet, the original approach of adding the things at runtime is better.

pkg/manifests/config_test.go
```diff
@@ -186,3 +186,27 @@ func TestEtcdDefaultsToDisabled(t *testing.T) {
 		t.Error("an empty etcd configuration should have etcd disabled")
 	}
 }
+
+func TestGrafanaDefaultsToEnabled(t *testing.T) {
```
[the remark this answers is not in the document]
bison: ACK. I also prefer that. Honestly, I just copied the extremely similar test above this and changed it to test Grafana. I can try to find time to make it table driven.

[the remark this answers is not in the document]
bison: Done.

pkg/manifests/manifests.go
```diff
@@ -1333,6 +1346,32 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
...
+			Name: "secret-prometheus-k8s-htpasswd",
+			VolumeSource: v1.VolumeSource{
+				Secret: &v1.SecretVolumeSource{
+					SecretName: "prometheus-k8s-htpasswd",
+				},
+			},
+		})
+
+		// Add the htpasswd arg and volume mount to the proxy container.
+		for i := range p.Spec.Containers {
```
simonpasquier, line 1361: it might be clearer to move it to the range loop above where we alreayd tweak the prometheus-proxy container?

pkg/manifests/manifests.go
```diff
@@ -2942,6 +2981,32 @@ func (f *Factory) ThanosQuerierDeployment(grpcTLS *v1.Secret, enableUserWorkload
...
+			Name: "secret-thanos-querier-oauth-htpasswd",
+			VolumeSource: v1.VolumeSource{
+				Secret: &v1.SecretVolumeSource{
+					SecretName: "thanos-querier-oauth-htpasswd",
+				},
+			},
+		})
+
+		// Add the htpasswd arg and volume mount to the proxy container.
+		for i := range d.Spec.Template.Spec.Containers {
```
simonpasquier, line 2996: same remark here

simonpasquier (commented):
Just 2 minor comments but nothing blocking. Very nice job!

test/e2e/config_test.go
```diff
@@ -92,6 +94,50 @@ func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Log("asserting that CMO is healthy after disabling Grafana")
+	assertOperatorCondition(t, configv1.OperatorDegraded, configv1.ConditionFalse)
+	assertOperatorCondition(t, configv1.OperatorAvailable, configv1.ConditionTrue)
+
+	// Push a default configuration that re-enables Grafana.
+	validCM.Data["config.yaml"] = "enableUserWorkload: true"
```
fpetkovski, line 123: How does enabling user workload monitoring also enable grafana?

test/e2e/config_test.go
```diff
@@ -92,6 +94,50 @@ func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
 	t.Log("asserting that CMO goes back healthy after the configuration is fixed")
 	assertOperatorCondition(t, configv1.OperatorDegraded, configv1.ConditionFalse)
 	assertOperatorCondition(t, configv1.OperatorAvailable, configv1.ConditionTrue)
+
+	// Push a configuration that disables Grafana.
```
fpetkovski, line 98: I would extract this addition into a new independent test

bison: Done.

test/e2e/config_test.go
bison: It doesn't. That was just setting the config back to what it was originally, i.e. removing the previously added Grafana bit. It was definitely confusing though. The new test is different.

pkg/manifests/manifests.go
bison: Done.

pkg/manifests/manifests.go
bison: Done here as well.

pkg/manifests/manifests.go
```diff
@@ -1152,19 +1154,32 @@ func (f *Factory) ThanosQuerierRoute() (*routev1.Route, error) {
...
+	}
+
+	if grafanaHost != nil {
+		data["grafanaPublicURL"] = grafanaHost.String()
+	}
+
+	if thanosHost != nil {
+		data["thanosPublicURL"] = thanosHost.String()
+	}
+
```
paulfantom, line 1176: Note: This will be removed in https://github.com/openshift/cluster-monitoring-operator/pull/1223

simonpasquier (commented):
/lgtm
I've checked offline with @sichvoge who's fine with the field as exposed in the configmap.

bison:
@simonpasquier @fpetkovski: I had to rebase this. Can you have another look when you have a chance?

simonpasquier (commented):
/lgtm

PR #1243, opened 2021-06-23
jsonnet: remove unused remove-runbook script

dgrisonnet:
Remove unused remove-runbook script, we are using [remove-runbook-urls.libsonnet](https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/remove-runbook-urls.libsonnet) instead.


PhilipGough (approved):
/lgtm

PR #1245, opened 2021-06-23
alert:KubeDeploymentReplicasMismatch: only fire if cluster is in read…

PhilipGough:
…y state

The original change as per https://github.com/openshift/cluster-monitoring-operator/pull/1065
meant that this alert would never fire since there was no commonality between label sets.

This change uses a scalar to ensure the alert will fire if and only if the
'cluster:control_plane:all_nodes_ready' record evaluates as 1.

See linked PR for context on why this alert differs from upstream.

Thanks @paulfantom for help with the query

### DIFF
--- a/assets/cluster-monitoring-operator/prometheus-rule.yaml
+++ b/assets/cluster-monitoring-operator/prometheus-rule.yaml
@@ -275,7 +275,7 @@ spec:
           healthy nodes and then contact support.
         summary: Deployment has not matched the expected number of replicas
       expr: |
-        (
+        (((
           kube_deployment_spec_replicas{namespace=~"(openshift-.*|kube-.*|default|logging)",job="kube-state-metrics"}
             !=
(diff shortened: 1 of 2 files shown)


PhilipGough:
Verification: Create a deployment that won't ever rollout, eg due to anti affinity and ensure the alert fires when one of the Pods remains in pending state

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 2
  template:
    metadata:
      labels:
        app: nginx
        some: value
    spec:
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
             matchExpressions:
               - key: some
                 operator: In
                 values:
                   - value
      containers:
        - name: nginx
          image: nginx:1.14.2
          ports:
            - containerPort: 80
```

PR #1246, opened 2021-06-24
Makefile: add check to verify runbook urls

dgrisonnet:
Add a Makefile rule called `check-runbooks` to verify that the links
specified in the runbook_url annotation of alerts shipped in OpenShift
aren't broken.

/cc @paulfantom 


paulfantom (approved):
/lgtm

PR #1250, opened 2021-06-28
Bug 1975281: Remove resource limit from kube-rbac-proxy

PhilipGough:


juzhao:
/lgtm
resource limit for kube-rbac-proxy is removed,  influented pods: kube-state-metrics/node-exporter pods

PhilipGough:
```bash
 Kernel Version:             3.10.0-1062.el7.x86_64
 OS Image:                   Red Hat Enterprise Linux Server 7.9 (Maipo)

oc exec node-exporter-4gskm -c kube-rbac-proxy -n openshift-monitoring -- cat /sys/fs/cgroup/cpu/cpu.stat
nr_periods 0
nr_throttled 0
throttled_time 0
```

PR #1252, opened 2021-06-28
WIP: generate client key and certificates

s-urbaniak:
This is a work-in-progress PR to wire the creation of client certificates for metrics scraping.

I am setting this up early so we can discuss if this is good for CMO inclusion. It uses library-go idiomatic controller logic which is not used yet in CMO.

Kindly asking for feedback. If the general direction is accepted, I will continue with wiring those certificates to Prometheus.

Note: this requires https://github.com/openshift/cluster-policy-controller/pull/61 to land first.


pkg/operator/operator.go
```diff
@@ -240,6 +249,35 @@ func New(
 	})
 	o.informers = append(o.informers, informer)
 
+	nsInformerFactory := informers.NewSharedInformerFactoryWithOptions(
```
stlaz, line 252: `s/nsInformerFactory/kubeInformersOperatorNS` sounds more descriptive

s-urbaniak: renamed :+1:

jsonnet/prometheus-adapter.libsonnet
```diff
@@ -59,6 +59,8 @@ function(params)
               caFile: '/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt',
               serverName: 'server-name-replaced-at-runtime',
               insecureSkipVerify: false,
+              // TODO: prometheus-adapter currently is a stock upstream aggregated api server.
+              // It does not support static authorization.
```
dgrisonnet, line 63: Do we have an upstream issue against kube-apiserver to support static authorization for the aggregation layer? As far as I can tell from the [official doc](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/#authentication-flow) detailing the authentication flow for aggregated APIs, the authorization between the apiserver and the aggregated API can currently only be made by submitting a `SubjectAccessReview` call.

stlaz: I believe sig-auth has already discussed static authorization in the past, although I am not sure it was based upon a specific issue.

I am not aware of this being in the upstream works today, but in OpenShift we've got this implemented in the operators that are based on library-go: https://github.com/openshift/library-go/pull/998

dgrisonnet: Awesome, that's exactly what I was looking for, thanks @stlaz.

If I understand things correctly, https://github.com/openshift/kubernetes/pull/588 added static authorization to the apiserver, so theoretically we should be able to use it in prometheus-adapter. But, thinking about it again, authorizing custom resources such as PodMetrics and NodeMetrics might be out of scope since I don't see them in the goals of the OEP.

On another note, this discussion made me realize that prometheus-adapter isn't protected by a kube-rbac-proxy even though we are scraping its metrics. I will open a separate PR to fix that.

dgrisonnet: I encountered an issue while adding authz to prometheus-adapter metrics endpoints. Since metrics serving isn't decoupled from the generic apiserver code, scrapers need to be authorized by the apiserver in order to access the server. So if we add a kube-rbac-proxy in front of the /metrics endpoint of prometheus-adapter, the proxy would need to forward the bearer token in order for the scrape to succeed. But, kube-rbac-proxy wasn't designed to allow that and is only supposed to forward requests to applications that aren't kubernetes aware.

I opened https://github.com/kubernetes-sigs/prometheus-adapter/issues/425 upstream.

stlaz: The decoupling would be one way to do it, another would be to have the static authorizer be a part of the prometheus-adapter.

This PR attempts to make it possible to use client certificate authentication for the scraping. For that, the side receiving the scraping request only needs to have the correct CA bundle and would not have to send any requests to the kube-apiserver.

Please know that we're not aiming for a 100% coverage of static authn/z from the beginning. For now it would be enough if we could get this PR merged, and then we'll have to update the ServiceMonitors for most of our components so that the client cert authentication is taken into effect.

I'll check whether the e2e test failures are related to this PR.

stlaz:
/retest
I believe the failed tests might be a result of https://github.com/openshift/cluster-policy-controller/pull/61 not being present in the test runs

jsonnet/prometheus-adapter.libsonnet
dgrisonnet: Yeah, it is definitely not a blocker for me, I just wanted to make sure that we at least track this effort.

pkg/manifests/tls.go
```diff
@@ -54,6 +54,39 @@ func (f *Factory) GRPCSecret() (*v1.Secret, error) {
...
+	}
+
+	s.Namespace = f.namespace
+	s.Data = make(map[string][]byte)
+	s.Annotations = make(map[string]string)
+
+	return s, nil
+}
+
+func (f *Factory) MetricsClientCASecret(apiAuthConfigmap *v1.ConfigMap) (*v1.Secret, error) {
```
simonpasquier, line 70: (nit) I would expect the function to accept []byte rather than configmap. Any reason why we pass a configmap here?

```suggestion
func (f *Factory) MetricsClientCASecret(clientCA []byte) (*v1.Secret, error) {
```

pkg/tasks/clustermonitoringoperator.go
```diff
@@ -120,5 +120,20 @@ func (t *ClusterMonitoringOperatorTask) Run() error {
...
+	}
+
+	s, err = t.factory.MetricsClientCASecret(apiAuthConfigmap)
+	if err != nil {
+		return errors.Wrap(err, "initializing Metrics Client CA failed")
+	}
+
+	err = t.client.CreateOrUpdateSecret(s)
+	if err != nil {
+		return errors.Wrap(err, "reconciling PrometheusAdapter Secret failed")
```
simonpasquier, line 135:
```suggestion
		return errors.Wrap(err, "reconciling Metrics Client CA failed")
```

simonpasquier (commented):
This is very good! I've got only one question (there's also one small typo to fix).

simonpasquier:
/retest
to see if the CI errors clear out.

simonpasquier:
hmm the failures can be explained by kube-rbac-proxy hitting a panic...

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/1252/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic/1413495736489742336/artifacts/e2e-agnostic/gather-extra/artifacts/pods/openshift-monitoring_prometheus-k8s-0_kube-rbac-proxy-thanos.log

fpetkovski:
@stlaz could we squash all the commits into two, one for non-vendor files and the other for vendor files, so that the PR is easier to review?

stlaz:
@simonpasquier @fpetkovski thank you kindly for your reviews. I opened a new PR that I can edit - #1282. I included a few fixes that should get this thing off the ground and have the tests pass.

I squashed the commits from this PR and specifically added my fixes in 3 on top of it in order to highlight the differences, but I can squash it even further. Please check the other PR at your convenience :slightly_smiling_face:

pkg/manifests/tls.go
stlaz: this function more or less synchronizes the configmap from one namespace to another (the type is kept to be a configmap in my PR #1282).

I would keep the type of the argument in case we wanted to synchronize more fields in the future, given that it's a pointer, the size of the structure being passed should not differ that much.

pkg/tasks/clustermonitoringoperator.go
stlaz: fixed in #1282 :+1:

stlaz:
Also, the kube-rbac-panics look like a bug. It might be that the proxy expected some certificates to be present but none appeared, nevertheless, this shouldn't happen.

simonpasquier:
/close
superseded by #1282

PR #1257, opened 2021-06-30
Bug 1956308: Fix deployment update with retry option

slashpai:
**Why this change?**

The operator goes Degraded and Unavailable for a short period of time, the reason being '... Deployment failed: creating Deployment object failed after update failed: object is being deleted: deployments.apps "xxx" already exists'. Trying to fix conflict in updating deployment with this change. 

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -739,9 +739,16 @@ func (c *Client) CreateOrUpdateDeployment(dep *appsv1.Deployment) error {
 			if err != nil {
 				return errors.Wrap(err, "deleting Deployment object failed")
 			}
-			err = c.CreateDeployment(required)
-			if err != nil {
-				return errors.Wrap(err, "creating Deployment object failed after update failed")
+			retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				err = c.CreateDeployment(required)
+				if err != nil {
+					return err
+				}
+				return nil
+			})
+
+			if retryErr != nil {
+				return errors.Wrap(retryErr, "creating Deployment object failed after update failed")
 			}
 		}
 		return errors.Wrap(err, "updating Deployment object failed")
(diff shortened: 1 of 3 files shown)


simonpasquier (commented):
nice finding about the retry package! IIRC the issue happens in this code path and the returned error is 422.

https://github.com/openshift/cluster-monitoring-operator/blob/7f4925a7203622d70b3007fbddfb6bc5cce6c1d9/pkg/client/client.go#L716-L731

So I would say that we need more something like this in `CreateOrUpdateDeployment()`

```
    err = c.UpdateDeployment(required)
    if err != nil { 
        err = retry.OnError(retry.DefaultRetry, errors.IsInvalid, func() error {
            err := c.DeleteDeployment(existing)
            ...
        })
```

raptorsun (approved):
PR looks good 👍.

pkg/client/client.go
```diff
@@ -734,9 +735,16 @@ func (c *Client) CreateOrUpdateDeployment(dep *appsv1.Deployment) error {
 		uErr, ok := err.(*apierrors.StatusError)
 		if ok && uErr.ErrStatus.Code == 422 && uErr.ErrStatus.Reason == metav1.StatusReasonInvalid {
 			// try to delete Deployment
-			err = c.DeleteDeployment(existing)
-			if err != nil {
-				return errors.Wrap(err, "deleting Deployment object failed")
+			retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
simonpasquier, line 738: I'd expect the retry to be on `CreateDeployment()` since it might fail if the existing deployment resource hasn't been fully deleted yet.

pkg/client/client.go
```diff
@@ -16,6 +16,7 @@ package client
 
 import (
 	"context"
+	"k8s.io/client-go/util/retry"
```
dgrisonnet, line 19: this should be reordered

pkg/client/client.go
```diff
@@ -738,9 +739,16 @@ func (c *Client) CreateOrUpdateDeployment(dep *appsv1.Deployment) error {
...
 			}
-			err = c.CreateDeployment(required)
-			if err != nil {
-				return errors.Wrap(err, "creating Deployment object failed after update failed")
+			retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				err = c.CreateDeployment(required)
+				if err != nil {
+					return err
+				}
+				return nil
```
dgrisonnet, line 747:
```suggestion
				return c.CreateDeployment(required)
```

dgrisonnet (commented):
I think we should also update [CreateOrUpdateDaemonset](https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/client/client.go#L904-L934). The code is very similar to the one for deployments, so it should have the same issue. I am fine with either doing that in this PR or in a follow-up one.

slashpai:
> I think we should also update [CreateOrUpdateDaemonset](https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/client/client.go#L904-L934). The code is very similar to the one for deployments, so it should have the same issue. I am fine with either doing that in this PR or in a follow-up one.

I can add that in another PR

dgrisonnet:
Fine by me, thank you @slashpai.

/lgtm

PR #1262, opened 2021-07-02
Bug 1974832: Improve HighlyAvailableWorkloadIncorrectlySpread to detect single point of failure

dgrisonnet:
Improve HighlyAvailableWorkloadIncorrectlySpread alert to only fire when
all instances of a workload are on the same node. Ultimately what we
want to catch with this alert are single points of failures and even if
multiple instances end up on the same node, if one of them is correctly
spread, then the workload is highly available.

/cc @simonpasquier 

### DIFF
--- a/assets/cluster-monitoring-operator/prometheus-rule.yaml
+++ b/assets/cluster-monitoring-operator/prometheus-rule.yaml
@@ -32,15 +32,16 @@ spec:
       annotations:
         description: Workload {{ $labels.namespace }}/{{ $labels.workload }} is incorrectly
           spread across multiple nodes which breaks high-availability requirements.
-          There are {{ $value }} pods on node {{ $labels.node }}, where there should
-          only be one. Since the workload is using persistent volumes, manual intervention
-          is needed. Please follow the guidelines provided in the runbook of this
-          alert to fix this issue.
+          Since the workload is using persistent volumes, manual intervention is needed.
+          Please follow the guidelines provided in the runbook of this alert to fix
+          this issue.
         runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/HighlyAvailableWorkloadIncorrectlySpread.md
         summary: Highly-available workload is incorrectly spread across multiple nodes
           and manual intervention is needed.
       expr: |
-        count by(node,workload,namespace)
+        count without (node)
+        (
+          group by (node, workload, namespace)
           (
             kube_pod_info{node!=""}
             * on(namespace,pod) group_left(workload)
@@ -52,11 +53,11 @@ spec:
                 * on(namespace,workload,workload_type) group_left()
                 (
                   count without(pod) (namespace_workload_pod:kube_pod_owner:relabel{namespace=~"(openshift-.*|kube-.*|default)"}) > 1
-                  <= on() group_left count(kube_node_role{role="worker"})
                 )
               )
             )
-          ) > 1
+          )
+        ) == 1
       for: 1h
       labels:
         severity: warning
--- a/jsonnet/rules.libsonnet
+++ b/jsonnet/rules.libsonnet
@@ -24,7 +24,9 @@ function(params) {
         },
         {
           expr: |||
-            count by(node,workload,namespace)
+            count without (node)
+            (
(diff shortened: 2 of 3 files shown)


jsonnet/rules.libsonnet
```diff
@@ -24,7 +24,7 @@ function(params) {
         },
         {
           expr: |||
-            count by(node,workload,namespace)
+            count by(workload,namespace)
```
dgrisonnet, line 27: I will update the runbook according to this change in the aggregation since the `node` label will not be exposed anymore in the alert.

assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -40,7 +40,9 @@ spec:
         summary: Highly-available workload is incorrectly spread across multiple nodes
           and manual intervention is needed.
       expr: |
-        count by(node,workload,namespace)
+        count without (node)
```
dgrisonnet, line 43: This will require updating the runbook since we lost the `node` information in the aggregation so it won't be visible in the alert anymore.

assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -52,11 +54,11 @@ spec:
                 * on(namespace,workload,workload_type) group_left()
                 (
                   count without(pod) (namespace_workload_pod:kube_pod_owner:relabel{namespace=~"(openshift-.*|kube-.*|default|logging)"}) > 1
```
juzhao, line 56: `logging` could be removed, since default logging project is `openshift-logging`

dgrisonnet: Make sense, but I will do that in a separate PR since this value comes from the namespace selector used for all the mixin. https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/main.jsonnet#L51

assets/cluster-monitoring-operator/prometheus-rule.yaml
simonpasquier: the description annotations would break since it's using `$labels.node`

simonpasquier (commented):
since the alert gets quite complex, it would be good to add rule tests.

assets/cluster-monitoring-operator/prometheus-rule.yaml
dgrisonnet: Oh indeed, I forgot about that one

assets/cluster-monitoring-operator/prometheus-rule.yaml
dgrisonnet: I opened https://github.com/openshift/cluster-monitoring-operator/pull/1266

dgrisonnet:
Added some test cases to the alert. PTAL

assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -32,15 +32,16 @@ spec:
...
+          Please follow the guidelines provided in the runbook of this alert to fix
+          this issue.
         runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/HighlyAvailableWorkloadIncorrectlySpread.md
         summary: Highly-available workload is incorrectly spread across multiple nodes
           and manual intervention is needed.
       expr: |
-        count by(node,workload,namespace)
+        count without (node)
+        (
+          group by (node, workload, namespace)
```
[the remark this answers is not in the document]
dgrisonnet: Syntax-wise the aggregation is different.

```
count without (node) (
  group by (node, workload, namespace) ( ... )
)
```
The inner aggregation is here to aggregate away all the labels that don't matter so that they won't impact the final count. And the outer aggregation removes the `node` since we want to focus the count solely on the number of unique nodes that each namespace/workload pair is scheduled on.

If we were to use only one aggregation, we would be accounting for the number of unique label values that each namespace/workload pair has. For instance, if two workloads are scheduled on the same node, but one of their labels is different, let's say the pod name, they would count as 2 different series in the aggregation where we would want them to be one.

PR #1244, opened 2021-06-23
jsonnet: pull latest deps

fpetkovski:
* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

PR #1265, opened 2021-07-05
Bug 1943565: ThanosSidecarUnhealthy will never fire if the sidecar is never healthy

arajkumar:
This is a back port of https://github.com/thanos-io/thanos/pull/4342. Originally got into 4.9 as part of https://github.com/openshift/cluster-monitoring-operator/pull/1244/files#diff-9f32475601d9c279cd63600ea04308adf63ac5809e2b483526ae4105e8c3670cL289

### DIFF
--- a/assets/prometheus-k8s/prometheus-rule.yaml
+++ b/assets/prometheus-k8s/prometheus-rule.yaml
@@ -266,7 +266,7 @@ spec:
           more than {{ $value }} seconds.
         summary: Thanos Sidecar is unhealthy.
       expr: |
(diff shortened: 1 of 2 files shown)


dgrisonnet:
I think we can wait for 4.8 GA before merging this bug considering that it doesn't have a huge impact on the product. So targeting a patch release of 4.8 should be fine.

/lgtm

sdodson:
Re-validating bugzilla status after prow config updates.
/bugzilla refresh

PR #1266, opened 2021-07-06
Remove logging from the mixins namespace selectors

dgrisonnet:
Logging metrics have been moved from the logging namespace to
openshift-logging, so we can now remove logging from the namespace
selector used by all the mixins.

### DIFF
--- a/assets/cluster-monitoring-operator/prometheus-rule.yaml
+++ b/assets/cluster-monitoring-operator/prometheus-rule.yaml
@@ -51,7 +51,7 @@ spec:
                 namespace_workload_pod:kube_pod_owner:relabel
                 * on(namespace,workload,workload_type) group_left()
                 (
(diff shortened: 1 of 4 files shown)


simonpasquier:
in the same idea, I'm not sure that `kube-*` makes sense either (though the namespaces are present).

dgrisonnet:
We can't do that yet since `kube-system` holds metrics from cri-o and kubelet. Also, there are two other namespaces that are created by default in OCP, `kube-public` and `kube-node-lease` which holds a couple of secrets so we have metrics from them. That said, I wonder if we really need to match all the `kube-.*` namespaces. IMO `kube-system` should be enough, let me know what do you think about that.

image

simonpasquier:
Yeah I've realized the same thing after I wrote my comment. Lets keep it as-is.

PR #1270, opened 2021-07-08
Bug 1978662: Set a degraded message when persistent storage is not configured

fpetkovski:
When persistent storage is not configured for prometheus, upgrades and cluster disruptions can lead to data loss. 
In order to make cluster admins aware of this problem, this PR adds a message to the degraded condition indicating this problem. The value of the condition is kept as `degraded=false` since there is no real degradation with the cluster monitoring.

### DIFF
--- a/pkg/client/status_reporter.go
+++ b/pkg/client/status_reporter.go
@@ -29,7 +29,7 @@ import (
 const (
 	unavailableMessage string = "Rollout of the monitoring stack failed and is degraded. Please investigate the degraded status error."
 	asExpectedReason   string = "AsExpected"
-	StorageNotConfiguredMessage = "Prometheus is running without persistent storage and upgrades and cluster disruptions can lead to data loss. Please refer to the official documentation to see how to configure storage for Prometheus: https://docs.openshift.com/container-platform/4.7/monitoring/configuring-the-monitoring-stack.html"
+	StorageNotConfiguredMessage = "Prometheus is running without persistent storage which can lead to data loss during upgrades and cluster disruptions. Please refer to the official documentation to see how to configure storage for Prometheus: https://docs.openshift.com/container-platform/4.8/monitoring/configuring-the-monitoring-stack.html"
 	StorageNotConfiguredReason  = "PrometheusDataPersistenceNotConfigured"
 )
 
--- a/pkg/client/status_reporter.go
+++ b/pkg/client/status_reporter.go
@@ -29,6 +29,8 @@ import (
 const (
 	unavailableMessage string = "Rollout of the monitoring stack failed and is degraded. Please investigate the degraded status error."
 	asExpectedReason   string = "AsExpected"
+	StorageNotConfiguredMessage = "Prometheus is running without persistent storage and upgrades and cluster disruptions can lead to data loss. Please refer to the official documentation to see how to configure storage for Prometheus: https://docs.openshift.com/container-platform/4.7/monitoring/configuring-the-monitoring-stack.html"
+	StorageNotConfiguredReason  = "PrometheusDataPersistenceNotConfigured"
 )
 
 type StatusReporter struct {
@@ -67,7 +69,7 @@ func (r *StatusReporter) relatedObjects() []v1.ObjectReference {
 	}
 }
 
-func (r *StatusReporter) SetDone() error {
+func (r *StatusReporter) SetDone(degradedConditionMessage string, degradedConditionReason string) error {
 	co, err := r.client.Get(r.ctx, r.clusterOperatorName, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		co = r.newClusterOperator()
@@ -82,7 +84,7 @@ func (r *StatusReporter) SetDone() error {
 	conditions := newConditions(co.Status, r.version, time)
 	conditions.setCondition(v1.OperatorAvailable, v1.ConditionTrue, "Successfully rolled out the stack.", "RollOutDone", time)
 	conditions.setCondition(v1.OperatorProgressing, v1.ConditionFalse, "", "", time)
-	conditions.setCondition(v1.OperatorDegraded, v1.ConditionFalse, "", "", time)
+	conditions.setCondition(v1.OperatorDegraded, v1.ConditionFalse, degradedConditionMessage, degradedConditionReason, time)
 	conditions.setCondition(v1.OperatorUpgradeable, v1.ConditionTrue, "", asExpectedReason, time)
 	co.Status.Conditions = conditions.entries()
 
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -38,6 +38,19 @@ type Config struct {
 	UserWorkloadConfiguration      *UserWorkloadConfiguration      `json:"-"`
 }
 
+func (c Config) IsStorageConfigured() bool {
+	if c.ClusterMonitoringConfiguration == nil {
+		return false
+	}
+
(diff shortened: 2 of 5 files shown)


pkg/operator/operator.go
```diff
@@ -480,8 +482,12 @@ func (o *Operator) sync(key string) error {
 		return err
 	}
 
+	var degradedConditionMessage string
+	if !config.IsStorageConfigured() {
+		degradedConditionMessage = StorageNotConfiguredMessage
+	}
 	klog.Info("Updating ClusterOperator status to done.")
```
arajkumar, line 489: nit: Can you update this log to reflect the status instead of done?

pkg/manifests/config.go
```diff
@@ -38,6 +38,19 @@ type Config struct {
...
+	if c.ClusterMonitoringConfiguration == nil {
+		return false
+	}
+
+	prometheusK8sConfig := c.ClusterMonitoringConfiguration.PrometheusK8sConfig
+	if prometheusK8sConfig == nil {
+		return false
+	}
+
+	return prometheusK8sConfig.VolumeClaimTemplate != nil
```
arajkumar, line 51: What if PersistentVolumeClaim itself fails? I'm not sure whether that has been already handled.

arajkumar (commented):
Added a nit and minor comment, rest lgtm :)

pkg/operator/operator.go
fpetkovski: The status will be done, it's just that we will add another message to the degraded condition. We can output an log to indicate this, but I would not remove this line.

pkg/manifests/config.go
fpetkovski: I think this is something to be monitored with an alert and is not really in the scope of what this PR tries to achieve.

pkg/client/status_reporter.go
```diff
@@ -77,7 +77,7 @@ func (r *StatusReporter) SetDone() error {
 	conditions := newConditions(co.Status, r.version, time)
 	conditions.setCondition(v1.OperatorAvailable, v1.ConditionTrue, "Successfully rolled out the stack.", "RollOutDone", time)
 	conditions.setCondition(v1.OperatorProgressing, v1.ConditionFalse, "", "", time)
-	conditions.setCondition(v1.OperatorDegraded, v1.ConditionFalse, "", "", time)
+	conditions.setCondition(v1.OperatorDegraded, v1.ConditionFalse, degradedConditionMessage, "", time)
```
jan--f, line 80: We probably should set the actual condition here depending on the message passed?

jan--f: Also this should add a reason. Maybe `PrometheusNoDataPersistence` or something like it

jan--f (changes requested):
One issue for this patch inline.

I'm wondering if we maybe should introduce a `func (r *StatusReporter) SetDegraded` function for this and set it explicitly where appropriate. We talked before about setting degraded on other conditions and generally improve the status reporting. Not needed for this PR but also wouldn't hurt I think.

pkg/client/status_reporter.go
fpetkovski: > We probably should set the actual condition here depending on the message passed?

Do you mean we should set the condition to `true` when a message other than `""` is passed?
I probably should have provided context in the PR description, the requirement is to actually keep the condition to `degraded=false` but set a message for cluster admins to be aware of the state. I update the PR description to contain that information.

fpetkovski: > Also this should add a reason. Maybe PrometheusNoDataPersistence or something like it

Good point, added the reason as well.

fpetkovski:
@jan--f I agree that the workaround is not ideal, but also I think reworking the status reporting is a bit outside of the scope of what is requested from us in this BZ. I don't think introducing a `func (r *StatusReporter) SetDegraded` is really appropriate in this case because we need to set degraded to `False` and at the same time display message for the degraded condition.

Do we already have a ticket for reworking the status reporting? I would be more than happy to pair on that one and see how we can improve the current situation.

pkg/client/status_reporter.go
jan--f: > Do you mean we should set the condition to `true` when a message other than `""` is passed?

Yes that is the convention afaiu, at least in the other direction (don't set the condition without also setting a reason). To me it makes sense too, but I'll try to find some according docs.

jan--f:
> @jan--f I agree that the workaround is not ideal, but also I think reworking the status reporting is a bit outside of the scope of what is requested from us in this BZ. I don't think introducing a `func (r *StatusReporter) SetDegraded` is really appropriate in this case because we need to set degraded to `False` and at the same time display message for the degraded condition.
> 
> What we might need is some sort of a status builder, or some other mechanism to decouple the status reporter from the status itself. Right now both the status state and the status update is handled through the same abstraction.
> 
> Do you know if we already have a ticket for reworking the status reporting? I would be more than happy to pair on that one and see how we can improve the current situation.

I don't think we have an overarching ticket. But agreed, this is really not in scope for this ticket.

simonpasquier:
I don't find any ticket to rewrite the status reporter part but I agree that it can be improved (I presume that the code originates from times when the ClusterOperator semantics weren't well defined).

pkg/client/status_reporter.go
fpetkovski: @wking we could not find any documentation on this particular topic. Is it okay for an operator to set a message for the `degraded` condition but keep the value to `false`. This is what we discussed for CMO when there is no persistent storage enabled to indicate to the admin that metrics might be lost.

arajkumar (approved):
lgtm

arajkumar (approved):
lgtm

fpetkovski:
Per Trevor, setting a message for the `degraded` condition while keeping the value to `false` is a valid state for the operator status. A good practice in these situations is to add a link to the documentation on how to fix the underlying problem.

pkg/client/status_reporter.go
fpetkovski: Based on Trevor's reply in slack, setting a message for the degraded condition while keeping the value to false is a valid state for the operator status. A good practice in these situations is to add a link to the documentation on how to fix the underlying problem.

pkg/client/status_reporter.go
```diff
@@ -29,6 +29,8 @@ import (
 const (
 	unavailableMessage string = "Rollout of the monitoring stack failed and is degraded. Please investigate the degraded status error."
 	asExpectedReason   string = "AsExpected"
+	StorageNotConfiguredMessage = "Prometheus is running without persistent storage and upgrades and cluster disruptions can lead to data loss. Please refer to the official documentation to see how to configure storage for Prometheus: https://docs.openshift.com/container-platform/4.7/monitoring/configuring-the-monitoring-stack.html"
```
simonpasquier, line 32: small suggestion to avoid the "and" repetitions.

```suggestion
	StorageNotConfiguredMessage = "Prometheus is running without persistent storage which can lead to data loss during upgrades and cluster disruptions. Please refer to the official documentation to see how to configure storage for Prometheus: https://docs.openshift.com/container-platform/4.8/monitoring/configuring-the-monitoring-stack.html"
```

pkg/manifests/config.go
simonpasquier: If the PVC can't be fulfilled (wrong storageclass for instance), the prometheus pods would never start and we should get an alert from CMO.

pkg/client/status_reporter.go
fpetkovski: Good catch, added

simonpasquier (commented):
/lgtm

PR #1271, opened 2021-07-09
Allow configuring additional alertmanagers for UWM Prometheus and Thanos

fpetkovski:
This commit is a follow up on the work done in https://github.com/openshift/cluster-monitoring-operator/pull/1132 and allows users to configure additional Alertmanagers for the Prometheus instance and the Thanos Ruler in User Workload Monitoring.

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -83,7 +83,7 @@ func (c Config) GetThanosRulerAlertmanagerConfigs() []AdditionalAlertmanagerConf
 		return nil
 	}
 
-	alertmanagerConfigs := c.UserWorkloadConfiguration.ThanosRuler.AlertManagersConfigs
+	alertmanagerConfigs := c.UserWorkloadConfiguration.ThanosRuler.AlertmanagersConfigs
 	if len(alertmanagerConfigs) == 0 {
 		return nil
 	}
@@ -193,7 +193,7 @@ type ThanosRulerConfig struct {
 	Tolerations          []v1.Toleration                      `json:"tolerations"`
 	Resources            *v1.ResourceRequirements             `json:"resources"`
 	VolumeClaimTemplate  *monv1.EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate"`
-	AlertManagersConfigs []AdditionalAlertmanagerConfig       `json:"additionalAlertmanagerConfigs"`
+	AlertmanagersConfigs []AdditionalAlertmanagerConfig       `json:"additionalAlertmanagerConfigs"`
 }
 
 type ThanosQuerierConfig struct {
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -145,8 +145,7 @@ type PrometheusK8sConfig struct {
 	VolumeClaimTemplate *monv1.EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate"`
 	RemoteWrite         []monv1.RemoteWriteSpec              `json:"remoteWrite"`
 	TelemetryMatches    []string                             `json:"-"`
-	// EXPERIMENTAL: this configuration field may change in future releases.
-	AlertmanagerConfigs []AdditionalAlertmanagerConfig `json:"additionalAlertmanagerConfigs"`
+	AlertmanagerConfigs []AdditionalAlertmanagerConfig       `json:"additionalAlertmanagerConfigs"`
 }
 
 type AdditionalAlertmanagerConfig struct {
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -53,19 +53,39 @@ func (c Config) IsStorageConfigured() bool {
 
 // GetPrometheusUWAdditionalAlertmanagerConfigs returns the alertmanager configurations for
 // the User Workload Monitoring Prometheus instance.
-// If no additional configurations are specified, GetPrometheusUWAdditionalAlertmanagerConfigs returns an empty array.
+// If no additional configurations are specified, GetPrometheusUWAdditionalAlertmanagerConfigs returns nil.
 func (c Config) GetPrometheusUWAdditionalAlertmanagerConfigs() []AdditionalAlertmanagerConfig {
 	if c.UserWorkloadConfiguration == nil {
-		return []AdditionalAlertmanagerConfig{}
+		return nil
 	}
 
 	if c.UserWorkloadConfiguration.Prometheus == nil {
-		return []AdditionalAlertmanagerConfig{}
+		return nil
 	}
 
 	alertmanagerConfigs := c.UserWorkloadConfiguration.Prometheus.AlertmanagerConfigs
-	if alertmanagerConfigs == nil {
-		return []AdditionalAlertmanagerConfig{}
+	if len(alertmanagerConfigs) == 0 {
+		return nil
+	}
+
+	return alertmanagerConfigs
+}
+
+// GetThanosRulerAlertmanagerConfigs returns the alertmanager configurations for
+// the User Workload Monitoring Thanos Ruler instance.
+// If no additional configurations are specified, GetThanosRulerAlertmanagerConfigs returns nil.
+func (c Config) GetThanosRulerAlertmanagerConfigs() []AdditionalAlertmanagerConfig {
+	if c.UserWorkloadConfiguration == nil {
+		return nil
+	}
+
+	if c.UserWorkloadConfiguration.ThanosRuler == nil {
+		return nil
+	}
+
+	alertmanagerConfigs := c.UserWorkloadConfiguration.ThanosRuler.AlertManagersConfigs
+	if len(alertmanagerConfigs) == 0 {
+		return nil
 	}
 
 	return alertmanagerConfigs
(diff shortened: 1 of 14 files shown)


pkg/manifests/manifests.go
```diff
@@ -3458,6 +3631,58 @@ func (f *Factory) ThanosRulerCustomResource(queryURL string, trustedCA *v1.Confi
...
+	if alertmanagerConfig == nil {
+		return
+	}
+	digestBytes := md5.Sum([]byte(alertmanagerConfig.StringData["alertmanagers.yaml"]))
+	digest = fmt.Sprintf("%x", digestBytes)
+	for i, _ := range t.Spec.Containers {
+		containerName := t.Spec.Containers[i].Name
+		if containerName == "thanos-ruler" {
+			t.Spec.Containers[i].Env = append(t.Spec.Containers[i].Env, v1.EnvVar{
+				Name:  "ALERTMANAGER_CONFIG_SECRET_VERSION",
```
fpetkovski, line 3679: Thanos ruler does not refresh its config when the alertmanagers secret changes. Because of this, we need to redeploy the statefulset whenever there is a change in the data of the secret.

pkg/manifests/manifests.go
```diff
@@ -1416,8 +1436,199 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
...
+			Name:      PrometheusUWAdditionalAlertmanagerConfigSecretName,
+			Namespace: f.namespaceUserWorkload,
+		},
+		Data: map[string][]byte{
+			AdditionalAlertmanagerConfigSecretKey: config,
+		},
+	}, nil
+}
+
+func (f *Factory) prometheusAlertmanagerConfigHttpSection(alertmanagerConfig AdditionalAlertmanagerConfig) ([]yaml2.MapItem, error) {
```
dgrisonnet, line 1476: I had quite a hard time understanding the difference between `prometheusAlertmanagerConfigHttpSection` and `thanosAlertmanagerConfigHttpSection` at first. To make it clearer, what do you think about refactoring the part where we get the bearerToken and TLSConfig and have custom marshal functions to translate into the different config based on the upstream types?

pkg/manifests/config.go
```diff
@@ -38,6 +38,46 @@ type Config struct {
...
+	if c.UserWorkloadConfiguration == nil {
+		return nil
+	}
+
+	if c.UserWorkloadConfiguration.Prometheus == nil {
+		return nil
+	}
+
+	alertmanagerConfigs := c.UserWorkloadConfiguration.Prometheus.AlertmanagerConfigs
+	if alertmanagerConfigs == nil || len(alertmanagerConfigs) == 0 {
```
dgrisonnet, line 54: checking for nil is redundant here since even without the check, we would have return nil

pkg/manifests/config.go
```diff
@@ -38,6 +38,46 @@ type Config struct {
...
+	if c.UserWorkloadConfiguration == nil {
+		return nil
+	}
+
+	if c.UserWorkloadConfiguration.ThanosRuler == nil {
+		return nil
+	}
+
+	alertmanagerConfigs := c.UserWorkloadConfiguration.ThanosRuler.AlertManagersConfigs
+	if alertmanagerConfigs == nil || len(alertmanagerConfigs) == 0 {
```
dgrisonnet, line 74: ditto

pkg/manifests/config.go
```diff
@@ -136,11 +176,12 @@ type AlertmanagerMainConfig struct {
...
-	NodeSelector        map[string]string                    `json:"nodeSelector"`
-	Tolerations         []v1.Toleration                      `json:"tolerations"`
-	Resources           *v1.ResourceRequirements             `json:"resources"`
-	VolumeClaimTemplate *monv1.EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate"`
+	LogLevel             string                               `json:"logLevel"`
+	NodeSelector         map[string]string                    `json:"nodeSelector"`
+	Tolerations          []v1.Toleration                      `json:"tolerations"`
+	Resources            *v1.ResourceRequirements             `json:"resources"`
+	VolumeClaimTemplate  *monv1.EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate"`
+	AlertManagersConfigs []AdditionalAlertmanagerConfig       `json:"additionalAlertManagerConfigs"`
```
dgrisonnet, line 184: I noticed that this typo was also present in the precedent PR and IMO we should fix it since `AlertManager` is counterintuitive.
```suggestion
	AlertManagersConfigs []AdditionalAlertmanagerConfig       `json:"additionalAlertmanagerConfigs"`
```

pkg/manifests/config.go
```diff
@@ -404,6 +445,7 @@ type PrometheusRestrictedConfig struct {
 	RemoteWrite         []monv1.RemoteWriteSpec              `json:"remoteWrite"`
 	EnforcedSampleLimit *uint64                              `json:"enforcedSampleLimit"`
 	EnforcedTargetLimit *uint64                              `json:"enforcedTargetLimit"`
+	AlertmanagerConfigs []AdditionalAlertmanagerConfig       `json:"additionalAlertManagerConfigs"`
```
dgrisonnet, line 448: ditto

pkg/tasks/prometheus_user_workload.go
```diff
@@ -187,6 +189,28 @@ func (t *PrometheusUserWorkloadTask) create() error {
...
+		return errors.Wrap(err, "initializing UserWorkload Prometheus additionalAlertManagerConfigs secret failed")
+	}
+	if secret != nil {
+		klog.V(4).Info("initializing UserWorkload Prometheus additionalAlertManagerConfigs secret")
+		err = t.client.CreateOrUpdateSecret(secret)
+		if err != nil {
+			return errors.Wrap(err, "reconciling UserWorkload Prometheus additionalAlertManagerConfigs secret failed")
+		}
+	} else {
+		err = t.client.DeleteSecret(&v1.Secret{
```
dgrisonnet, line 203: If we never set the additional alertmanager configuration, this will always fail since we will delete an inexistant secret.

test/e2e/framework/framework.go
```diff
@@ -16,8 +16,14 @@ package framework
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"fmt"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
```
dgrisonnet, line 24: reorder

test/e2e/user_workload_monitoring_test.go
```diff
@@ -15,8 +15,12 @@
 package e2e
 
 import (
+	"context"
 	"fmt"
+	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
 	"io/ioutil"
+	"k8s.io/client-go/util/cert"
```
dgrisonnet, line 22: reorder

dgrisonnet (commented):
A first pass

pkg/manifests/manifests.go
fpetkovski: Ack, will put some effort into refactoring the code a bit.

pkg/manifests/config.go
fpetkovski: On second look, many of those typos are actually introduced in this PR. I added a new commit renaming `additionalAlertManagerConfigs` to `additionalAlertmanagerConfigs`

pkg/manifests/manifests.go
fpetkovski: I extracted the yaml rendering part here: https://github.com/openshift/cluster-monitoring-operator/pull/1271/files as part of a new commit.

Please take a look again and let me know what you think.

pkg/manifests/manifests.go
fpetkovski: Added an inline comment :+1:

pkg/tasks/prometheus.go
```diff
@@ -306,25 +306,25 @@ func (t *PrometheusTask) Run() error {
...
 		if secret != nil {
-			klog.V(4).Info("initializing Prometheus additionalAlertManagerConfigs secret")
+			klog.V(4).Info("initializing Prometheus additionalAlertmanagerConfigs secret")
 			err = t.client.CreateOrUpdateSecret(secret)
 			if err != nil {
-				return errors.Wrap(err, "reconciling Prometheus additionalAlertManagerConfigs secret failed")
+				return errors.Wrap(err, "reconciling Prometheus additionalAlertmanagerConfigs secret failed")
 			}
 		} else {
 			err = t.client.DeleteSecret(&v1.Secret{
```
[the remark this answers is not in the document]
fpetkovski: This is a good point. I changed the code to always create the additional alertmanagers config secret. Then depending on whether external alertmanagers are configured, we either attach it or not.

simonpasquier:
this needs a (small) rebase otherwise lgtm!

test/e2e/framework/framework.go
dgrisonnet: yes, we should add a linter, but not as a github action since we are already using prow downstream.

pkg/manifests/manifests.go
dgrisonnet: awesome! it is way easier to understand now :100:

PR #1274, opened 2021-07-09
jsonnet: Enable federated targets in thanos-querier

bison:
This adds the argument to thanos-querier to enable the new federated
targets functionality.  When user workload monitoring is enabled,
targets will be fetched from that Prometheus instance as well as from
the platform Prometheus.


simonpasquier (commented):
/lgtm

PR #1275, opened 2021-07-09
client: Always set OperatorUpgradeable reason to AsExpected

bison:
We always set the OperatorUpgradeable status condition on the
ClusterOperator resource to True, but we sometimes change the message
to indiciate errors.  This has proved confusing for users because the
errors are also indicated in the other status condition messages and
the Upgradeable condition has not actually changed values.  This
always sets the Upgradeable condition to true with an "AsExpcted"
reason.


simonpasquier (commented):
/lgtm

PR #1277, opened 2021-07-12
BUG 1980888: jsonnet: Favour http probes for thanos querier

PhilipGough:
The exec probe is timing out intermittently. This coud be caused in part
    by the timing on the probe not being sufficient, or there may be an underlying
    issue with the (exec, cri-o) combo.
    
    This change takes the config for httpGet probes from upstream thanos and proxies
    to thanos via the oauth-gateway where we bypass the health check endpoints.
    
    The HTTP probe cannot directly ping thanos since it listens on localhost
    and is unreachable from kubelet. This is by design.

I have not increased the timeout here because if there is an issue with query after setting the rest of the config to match what is upstream, I would prefer to observe that as an issue and report it upstream to find out when/why it might occur. We will of course, see some additional latency by going through the proxy but I would still expect a health endpoint to return sub 1 second over http

cc @simonpasquier - would you be able to take a look at this potential solution?

### DIFF
--- a/jsonnet/thanos-querier.libsonnet
+++ b/jsonnet/thanos-querier.libsonnet
@@ -336,18 +336,8 @@ function(params)
             securityContext:: {},
             containers: [
               super.containers[0] {
-                livenessProbe: {
-                  httpGet:: {},
-                  exec: {
-                    command: ['sh', '-c', 'if [ -x "$(command -v curl)" ]; then exec curl http://localhost:9090/-/healthy; elif [ -x "$(command -v wget)" ]; then exec wget --quiet --tries=1 --spider http://localhost:9090/-/healthy; else exit 1; fi'],
-                  },
-                },
-                readinessProbe: {
-                  httpGet:: {},
-                  exec: {
-                    command: ['sh', '-c', 'if [ -x "$(command -v curl)" ]; then exec curl http://localhost:9090/-/ready; elif [ -x "$(command -v wget)" ]; then exec wget --quiet --tries=1 --spider http://localhost:9090/-/ready; else exit 1; fi'],
-                  },
-                },
+                livenessProbe:: {},
+                readinessProbe:: {},
                 args: std.map(
                   function(a)
                     if std.startsWith(a, '--grpc-address=') then '--grpc-address=127.0.0.1:10901'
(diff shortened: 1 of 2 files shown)


PhilipGough:
The inital commit will result in the following events because of probe failures, hence the deployment will never roll out:

```
24m       Warning   Unhealthy    pod/thanos-querier-757dc7b8b4-xxmjc   Liveness probe failed: Get "http://10.131.0.24:9090/-/healthy": dial tcp 10.131.0.24:9090: connect: connection refused
25m       Warning   ProbeError   pod/thanos-querier-757dc7b8b4-xxmjc   Readiness probe error: Get "http://10.131.0.24:9090/-/ready": dial tcp 10.131.0.24:9090: connect: connection refused
body:
25m       Warning   Unhealthy           pod/thanos-querier-757dc7b8b4-xxmjc    Readiness probe failed: Get "http://10.131.0.24:9090/-/ready": dial tcp 10.131.0.24:9090: connect: connection refused
```

The second commit addresses the failure with `httpGet` probes by alering the listen address for the http-server. I'm not sure if there are any consequences to doing so but the deployment rolls out.

We probably have a couple of path forward:

1. Alter the existing `exec` probes and add an intial wait time and increase the thresholds to make it more reliable
2. Go with the solution in this PR assuming that it does not cause any issues to other components or querier itself.
3. We have removed the probes in some other components, https://github.com/prometheus-operator/prometheus-operator/pull/3502. We could follow suit here but perhaps fixing the underlying issue(s) is preferable.

PhilipGough:
The upstream repo includes the change of ip address to the http server. It also sets some values which perhaps we could sync with https://github.com/thanos-io/kube-thanos/blob/main/manifests/thanos-query-deployment.yaml#L42-L63

jsonnet/thanos-querier.libsonnet
```diff
@@ -418,6 +428,7 @@ function(params)
                   '-cookie-secret-file=/etc/proxy/secrets/session_secret',
                   '-openshift-ca=/etc/pki/tls/cert.pem',
                   '-openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt',
+                  '-bypass-auth-for=/-/(healthy|ready)',
```
simonpasquier, line 431: Shouldn't we add beginning/end of text markers?
```suggestion
                  '-bypass-auth-for=^/-/(healthy|ready)$',
```

PR #1278, opened 2021-07-12
pkg/manifests: Add EnforcedTargetLimit for user-workload monitoring

bison:
This adds support for configuring the EnforcedTargetLimit for the
user-workload Prometheus instance.  This lets admins set an overall
limit on the number of targets scraped.


simonpasquier (commented):
/lgtm

PR #1279, opened 2021-07-13
Bug 1986375: adding check for node exporter daemon set

prashbnair:
removing operator log

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -974,9 +974,20 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
 				d.Status.DesiredNumberScheduled, d.Status.UpdatedNumberScheduled)
 			return false, nil
 		}
-		if d.Status.NumberUnavailable != 0 {
-			lastErr = errors.Errorf("got %d unavailable nodes",
-				d.Status.NumberUnavailable)
+
+		var nodeReadyCount int32
+		nodeList, err := c.kclient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+		for _, node := range nodeList.Items {
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == v1.NodeReady && condition.Status == v1.ConditionTrue {
+					nodeReadyCount++
+				}
+			}
+		}
+
+		if d.Status.NumberAvailable != nodeReadyCount {
+			lastErr = errors.Errorf("expected %d ready pods for %q daemonset, got %d ",
+				nodeReadyCount, d.GetName(), d.Status.NumberAvailable)
 			return false, nil
 		}
 		return true, nil


pkg/client/client.go
```diff
@@ -974,7 +974,21 @@ func (c *Client) WaitForDaemonSetRollout(ds *appsv1.DaemonSet) error {
...
 		}
-		if d.Status.NumberUnavailable != 0 {
+
+		var nodeNotReadyCount int32
+		nodeList, err := c.kclient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+		nodes := nodeList.Items
+		for _, node := range nodes {
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == v1.NodeReady {
+					if condition.Status != v1.ConditionTrue {
```
arajkumar, line 984: nit
```suggestion
				if condition.Type == v1.NodeReady && condition.Status != v1.ConditionTrue {
```

pkg/client/client.go
```diff
@@ -974,7 +974,21 @@ func (c *Client) WaitForDaemonSetRollout(ds *appsv1.DaemonSet) error {
 				d.Status.DesiredNumberScheduled, d.Status.UpdatedNumberScheduled)
 			return false, nil
 		}
-		if d.Status.NumberUnavailable != 0 {
+
+		var nodeNotReadyCount int32
+		nodeList, err := c.kclient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+		nodes := nodeList.Items
+		for _, node := range nodes {
```
simonpasquier, line 981: (nit) saving one line

```suggestion
		for _, node := range nodeList.Items {
```

simonpasquier (commented):
It might be worth looking if other fields in the daemonset's status wouldn't bring us the same information? In particular `numberAvailable` and `numberReady`...

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#daemonsetstatus-v1-apps

simonpasquier:
> It might be worth looking if other fields in the daemonset's status wouldn't bring us the same information? In particular numberAvailable and numberReady...

Scratch that, it doesn't seem to provide the information we need.

pkg/client/client.go
```diff
@@ -974,7 +974,21 @@ func (c *Client) WaitForDaemonSetRollout(ds *appsv1.DaemonSet) error {
...
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == v1.NodeReady {
+					if condition.Status != v1.ConditionTrue {
+						nodeNotReadyCount++
+					}
+				}
+			}
+		}
+
+		if d.Status.NumberUnavailable != 0 && d.Status.NumberUnavailable != nodeNotReadyCount {
```
simonpasquier, line 991: another way to look at it would be to compare the number of ready nodes against `NumberAvailable`. If there's a difference, CMO could log something like `expected %d ready pods for %q daemonset, got only %d`

pkg/client/client.go
```diff
@@ -974,7 +974,21 @@ func (c *Client) WaitForDaemonSetRollout(ds *appsv1.DaemonSet) error {
...
 			return false, nil
 		}
-		if d.Status.NumberUnavailable != 0 {
+
+		var nodeNotReadyCount int32
+		nodeList, err := c.kclient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+		nodes := nodeList.Items
+		for _, node := range nodes {
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == v1.NodeReady {
```
PhilipGough, line 983: Won't this only catch nodes that are in ready state? what about nodes that are under other conditions (eg disk pressure) are we happy to ignore them?

pkg/client/client.go

pkg/client/client.go

pkg/client/client.go

pkg/client/client.go
```diff
@@ -976,7 +976,18 @@ func (c *Client) WaitForDaemonSetRollout(ds *appsv1.DaemonSet) error {
...
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == v1.NodeReady && condition.Status == v1.ConditionTrue {
+					nodeReadyCount++
+				}
+			}
+		}
+
+		if d.Status.NumberAvailable != nodeReadyCount {
 			lastErr = errors.Errorf("got %d unavailable nodes",
 				d.Status.NumberUnavailable)
```
simonpasquier, line 992:
```suggestion
			lastErr = errors.Errorf("expected %d ready pods for %q daemonset, got only %d",
				nodeReadyCount, ds.GetName(), d.Status.NumberAvailable)
```

pkg/client/client.go
```diff
@@ -974,9 +974,20 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
 				d.Status.DesiredNumberScheduled, d.Status.UpdatedNumberScheduled)
 			return false, nil
 		}
-		if d.Status.NumberUnavailable != 0 {
-			lastErr = errors.Errorf("got %d unavailable nodes",
-				d.Status.NumberUnavailable)
+
+		var nodeReadyCount int32
+		nodeList, err := c.kclient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
```
dgrisonnet, line 979: let's pass the context `ctx` here instead of `TODO()`

pkg/client/client.go
```diff
@@ -974,9 +974,20 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
...
+		for _, node := range nodeList.Items {
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == v1.NodeReady && condition.Status == v1.ConditionTrue {
+					nodeReadyCount++
+				}
+			}
+		}
+
+		if d.Status.NumberAvailable != nodeReadyCount {
+			lastErr = errors.Errorf("expected %d ready pods for %q daemon set, got %d ",
```
dgrisonnet, line 989:
```suggestion
			lastErr = errors.Errorf("expected %d ready pods for %q daemonset, got %d ",
```

pkg/client/client.go

pkg/client/client.go

simonpasquier (commented):
/lgtm

PR #1281, opened 2021-07-14
Send etcd telemetry rules

lilic:
This is the first batch of etcd metrics we are interested in, these will help in adjusting our alerts and also seeing what these values are per provider as they differ.

@smarterclayton as mentioned before, we want to use this data to adjust our alerts and define our SLO in the future. Total cardinality is 9 + 6 series = 15 series.

cc @hexfusion @openshift/openshift-team-monitoring 

* [x ] No user facing changes, so no entry in CHANGELOG was needed.


lilic:
/retest

🤷‍♀️

jsonnet/rules.libsonnet
```diff
@@ -548,6 +548,31 @@ function(params) {
         },
       ],
     },
+    {
+      name: 'openshift-etcd-telemetry.rules',
+      rules: [
+        {
+          expr: 'histogram_quantile(0.90, sum(rate(etcd_disk_wal_fsync_duration_seconds_bucket{job="etcd"}[5m])) by (instance, le))',
```
hexfusion, line 555: do we want to use irate vs rate?

Documentation/data-collection.md
```diff
@@ -333,6 +333,21 @@ data:
     # (@red-hat-data-services/odh, Open Data Hub team) The total number of users of RHODS using each component.
     - '{__name__="rhods_total_users"}'
     #
+    # (etcd team) 99th procentile of etcd wal fsync duration.
+    - '{__name__="histogram_quantile_99:etcd_disk_wal_fsync_duration_seconds_bucket:sum"}'
+    #
+    # (etcd team) 90th procentile of etcd wal fsync duration.
+    - '{__name__="histogram_quantile_90:etcd_disk_wal_fsync_duration_seconds_bucket:sum"}'
```
hexfusion, line 340: what is the cost of adding this 90th percentile? I am not sure its worth it.

hexfusion: could we instead include p99 for `etcd_disk_backend_commit_duration_seconds_bucket` ?

hexfusion (commented):
Thanks for doing this few notes I was hoping we could also report the following on disk state metrics below. Curious the cost. The use case is having an understanding of how performance is impacted by the size of the state file and the ability for the operator to improve this with the defrag controller. 
```
etcd_mvcc_db_total_size_in_bytes

etcd_mvcc_db_total_size_in_use_in_bytes
```

Documentation/data-collection.md
lilic: I thought the 90th showed a good picture when I was running these queries, but happy to replace it with `etcd_disk_backend_commit_duration_seconds_bucket `.

lilic:
My thought I would get the feel of things in terms of cardinality and we can add `etcd_mvcc_db_total_size_in_bytes
` and rest batch in a follow-up as they are not part of our SLO or alerting improvements. But if @smarterclayton is happy to approve those as well, can add them here :D

jsonnet/rules.libsonnet
lilic: As we are doing a sum here, we want to use rate, irate would not give us the right picture. But agreed overall for spikes irate is good when trying to represent it on a dashboard, but here we want a more accurate picture when sending via telemetry.

hexfusion: thanks for explanation

smarterclayton:
Approved for 15 cardinality (etcd is fundamental to success of a cluster)

lilic:
@openshift/openshift-team-monitoring please take a look, thanks!

assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -409,6 +409,21 @@ spec:
       labels:
         workload_type: deploymentconfig
       record: namespace_workload_pod:kube_pod_owner:relabel
+  - name: openshift-etcd-telemetry.rules
+    rules:
+    - expr: sum by (instance) (etcd_mvcc_db_total_size_in_bytes{job="etcd"})
+      record: cluster:etcd_mvcc_db_total_size_in_bytes:sum
+    - expr: histogram_quantile(0.99, sum(rate(etcd_disk_wal_fsync_duration_seconds_bucket{job="etcd"}[5m]))
+        by (instance, le))
+      record: histogram_quantile_99:etcd_disk_wal_fsync_duration_seconds_bucket:sum
```
dgrisonnet, line 418: For recording rules based on histogram quantiles, we usually add generic names to the recording rules along with quantile label that will distinguish the different percentile that can be computed via this recording rule. This makes it easier to extend the recording rule if we need to compute more percentiles at some point.
```suggestion
      labels:
        quantile: "0.99"
      record: etcd_quantile:etcd_disk_wal_fsync_duration_seconds:histogram_quantile
```

lilic: Yes I know, but wasn't sure if that is possible to be sent via telemetry or in which format?

Documentation/data-collection.md
```diff
@@ -333,6 +333,21 @@ data:
     # (@red-hat-data-services/odh, Open Data Hub team) The total number of users of RHODS using each component.
     - '{__name__="rhods_total_users"}'
     #
+    # (etcd team) 99th procentile of etcd wal fsync duration.
```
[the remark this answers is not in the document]
lilic: Wed don't have a team yet where you can ping us sadly, I am working on creating one hence the temp thing here. Can I add it later maybe?

hexfusion (approved):
LGTM

assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -409,6 +409,21 @@ spec:
       labels:
         workload_type: deploymentconfig
       record: namespace_workload_pod:kube_pod_owner:relabel
+  - name: openshift-etcd-telemetry.rules
+    rules:
+    - expr: sum by (instance) (etcd_mvcc_db_total_size_in_bytes{job="etcd"})
+      record: cluster:etcd_mvcc_db_total_size_in_bytes:sum
+    - expr: histogram_quantile(0.99, sum(rate(etcd_disk_wal_fsync_duration_seconds_bucket{job="etcd"}[5m]))
+        by (instance, le))
```
simonpasquier, line 417: (suggestion) I find it more readable when the `by()` clause is right after `sum`.

```suggestion
    - expr: histogram_quantile(0.99, sum by (instance, le) (rate(etcd_disk_wal_fsync_duration_seconds_bucket{job="etcd"}[5m]))
       )
```

assets/cluster-monitoring-operator/prometheus-rule.yaml
simonpasquier: I'm not aware of any limitations on the telemetry side that would prevent metrics from being sent.

simonpasquier (commented):
a few minor comments otherwise LGTM

assets/cluster-monitoring-operator/prometheus-rule.yaml
lilic: So we just allowlist this rule and any label matches would be sent.

Documentation/data-collection.md
```diff
@@ -333,6 +333,23 @@ data:
     # (@red-hat-data-services/odh, Open Data Hub team) The total number of users of RHODS using each component.
     - '{__name__="rhods_total_users"}'
     #
+    # owners: (@etcd team)
```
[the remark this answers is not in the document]
lilic: We are working on adding a new team, but might take some time, that's not the etcd team. Plan was to adjust this later on.

assets/cluster-monitoring-operator/prometheus-rule.yaml
dgrisonnet: Hmm good point, I think that if we were to add the label to the telemeter query today, it would apply to all the metrics as `alertstate="firing"` currently. As far as I understand it, labels are currently global to the telemeter query, so if we were to add recording rules with different quantiles later, we wouldn't be able to select only one to send via telemetry.

lilic: Yes, not sure how much of a win defining them this way would be as I remember we had some issues on the label side in the past. But up to you two really, happy to go any route there.

lilic:
@simonpasquier @dgrisonnet addressed to adapt to new formatting of the comments, the recording rule is still not resolved, let me know how you want to solve that part, thanks!

assets/cluster-monitoring-operator/prometheus-rule.yaml
simonpasquier: > So we just allowlist this rule and any label matches would be sent. 

We could narrow-down the selected series in `manifests/0000_50_cluster-monitoring-operator_04-config.yaml` to sent only the 99th percentile value (e.g. `{__name__="etcd_quantile:etcd_disk_wal_fsync_duration_seconds:histogram_quantile",quantile="0.99"}`) but I'm not sure it's worth the effort.

lilic: Yes exactly my thinking. I also feel like querying might be more awkward as majority rules are sent directly.

lilic:
/retest

Strange failure 🤔 `assert_tenancy_model_is_enforced_for_rules `

lilic:
@simonpasquier @dgrisonnet friendly ping, please take a look again, thanks! :)

assets/cluster-monitoring-operator/prometheus-rule.yaml
dgrisonnet: I would personally prefer having a generic recording rule narrowed down in telemetry. If we ever need to extend the recording rule to get another percentile, we would be able to keep it under the same recording rule which would be easier to manipulate and query.

lilic: Can we add that when we extend the recording rules, otherwise I fear premature optimization here? (We are already at our budget right now, so we most likely won't add it, is why I am asking :) )

dgrisonnet: It doesn't have to be a recording rule sent via telemetry necessarily.

I essentially just want to make sure that we stick to the original format that we used for the other quantile recording rules. If y'all think we should make an exception here, I am fine with it.

Documentation/sample-metrics.md
```diff
@@ -13,7 +13,7 @@ return the full set of metrics that the Telemeter client captures:
...
...ests:sum:max_over_time:2m|cluster:telemetry_selected_series:count|openshift:prometheus_tsdb_head_series:sum|openshift:prometheus_tsdb_head_samples_appended_total:sum|monitoring:container_memory_working_set_bytes:sum|monitoring:haproxy_server_http_responses_total:sum|rhmi_status|cluster_legacy_scheduler_policy|cluster_master_schedulable|che_workspace_status|che_workspace_started_total|che_workspace_failure_total|che_workspace_start_time_seconds_sum|che_workspace_start_time_seconds_count|cco_credentials_mode|cluster:kube_persistentvolume_plugin_type_counts:sum|visual_web_terminal_sessions_total|acm_managed_cluster_info|cluster:vsphere_vcenter_info:sum|cluster:vsphere_esxi_version_total:sum|cluster:vsphere_node_hw_version_total:sum|openshift:build_by_strategy:sum|rhods_aggregate_availability|rhods_total_users|etcd_quantile:etcd_disk_wal_fsync_duration_seconds:histogram_quantile|cluster:etcd_mvcc_db_total_size_in_bytes:sum|etcd_quantile:etcd_network_peer_round_trip_time_seconds_bucket:histogram_quantile|cluster:etcd_mvcc_db_total_size_in_use_in_bytes:sum|etcd_quantile:etcd_disk_backend_commit_duration_seconds_bucket:histogram_quantile",alertstate=~"firing|",quantile=~"0.99|0.99|0.99|"}
```
lilic, line 16: So it seems like we have 3 x 0.99 :) I don't think its an issue in Prometheus just looks odd and won't allow for sending one telemetry with just 0.90 quantile for example.

lilic: @simonpasquier @dgrisonnet done, please take a look. I am fine with this, if you are :)

simonpasquier: :smile: it should be fine and it's only documented for users that want to see which metrics are forwarded (so no impact on telemeter-client).

jsonnet/rules.libsonnet
```diff
@@ -548,6 +548,40 @@ function(params) {
         },
       ],
     },
+    {
+      name: 'openshift-etcd-telemetry.rules',
+      rules: [
+        {
+          expr: 'sum by (instance) (etcd_mvcc_db_total_size_in_bytes{job="etcd"})',
+          record: 'cluster:etcd_mvcc_db_total_size_in_bytes:sum',
```
simonpasquier, line 556:
```suggestion
          record: 'instance:etcd_mvcc_db_total_size_in_bytes:sum',
```

jsonnet/rules.libsonnet
```diff
@@ -548,6 +548,40 @@ function(params) {
...
+        },
+        {
+          expr: 'histogram_quantile(0.99, sum by (instance, le) (rate(etcd_disk_wal_fsync_duration_seconds_bucket{job="etcd"}[5m])))',
+          labels: {
+            quantile: '0.99',
+          },
+          record: 'etcd_quantile:etcd_disk_wal_fsync_duration_seconds:histogram_quantile',
```
simonpasquier, line 563:
```suggestion
          record: 'instance:etcd_disk_wal_fsync_duration_seconds:histogram_quantile',
```

jsonnet/rules.libsonnet
```diff
@@ -548,6 +548,40 @@ function(params) {
...
+        },
+        {
+          expr: 'histogram_quantile(0.99, sum by (instance, le) (rate(etcd_network_peer_round_trip_time_seconds_bucket{job="etcd"}[5m])))',
+          labels: {
+            quantile: '0.99',
+          },
+          record: 'etcd_quantile:etcd_network_peer_round_trip_time_seconds_bucket:histogram_quantile',
```
simonpasquier, line 570:
```suggestion
          record: 'instance:etcd_network_peer_round_trip_time_seconds:histogram_quantile',
```

jsonnet/rules.libsonnet
```diff
@@ -548,6 +548,40 @@ function(params) {
...
+        },
+        {
+          expr: 'sum by (instance) (etcd_mvcc_db_total_size_in_use_in_bytes{job="etcd"})',
+          record: 'cluster:etcd_mvcc_db_total_size_in_use_in_bytes:sum',
```
simonpasquier, line 574:
```suggestion
          record: 'instance:etcd_mvcc_db_total_size_in_use_in_bytes:sum',
```

jsonnet/rules.libsonnet
```diff
@@ -548,6 +548,40 @@ function(params) {
...
+        },
+        {
+          expr: 'histogram_quantile(0.99, sum by (instance, le) (rate(etcd_disk_backend_commit_duration_seconds_bucket{job="etcd"}[5m])))',
+          labels: {
+            quantile: '0.99',
+          },
+          record: 'etcd_quantile:etcd_disk_backend_commit_duration_seconds_bucket:histogram_quantile',
```
simonpasquier, line 581:
```suggestion
          record: 'instance:etcd_disk_backend_commit_duration_seconds:histogram_quantile',
```

simonpasquier (commented):
small nits to better comply with recording rules [best practices](https://prometheus.io/docs/practices/rules/).

simonpasquier (commented):
/lgtm

lilic:
> Forbidden - CSRF token invalid

🤷‍♀️

lilic:
/cherry-pick 4.8

This won't work probably...

PR #1282, opened 2021-07-14
generate client key and certificates

stlaz:
This is a PR to wire the creation of client certificates for metrics scraping. It uses library-go idiomatic controller logic which is not used yet in CMO.

Kindly asking for feedback. If the general direction is accepted, I will continue with wiring those certificates to Prometheus.


stlaz:
This PR includes some fixes to the original of https://github.com/openshift/cluster-monitoring-operator/pull/1252, namely:
1. the client CA is now passed in as a config map, there's no need to have it as a secret
2. the added RBAC now includes permissions to handle events as that's a new thing for the operator
3. the CSRs now use the `generateName` instead of a single name to prevent the name of the CSR being taken and thus preventing us from creating a CSR and progressing further. This fix also adds the label required by the [CPO to approve the CSR](https://github.com/openshift/cluster-policy-controller/blob/369412114eca966117d977a9d301a95d70c92993/pkg/cmd/controller/csr.go)

stlaz:
looks like the failures in the kube-rbac-proxy are persisting, I'll try to check what's going on

stlaz:
The problem was in the client-ca being missing for the kube-rbac-proxy-thanos container in the prometheus-k8s workload. Nevertheless, the kube-rbac-proxy should not panic in such cases but that's something that can be investigated outside of this PR.

stlaz:
Opened up https://github.com/brancz/kube-rbac-proxy/issues/131 to track the kube-rbac-proxy issue

stlaz:
added fixes for further failing kube-rbac-proxies that did not have the client-ca configured

stlaz:
mhm, the tech debt is showing as the operator refuses to progress to the client-ca creation while the prometheus-operator fails to deploy because it's missing the client-ca :| In an ideal world, this would be handled by two separate controllers.

 I'll see if rearranging the order is possible.

pkg/tasks/prometheus.go
```diff
@@ -65,6 +65,21 @@ func (t *PrometheusTask) Run() error {
...
+	cm, err := t.factory.MetricsClientCACM(apiAuthConfigmap)
+	if err != nil {
+		return errors.Wrap(err, "initializing Metrics Client CA failed")
+	}
+
+	err = t.client.CreateOrUpdateConfigMap(cm)
+	if err != nil {
+		return errors.Wrap(err, "reconciling Metrics Client CA ConfigMap failed")
+	}
+
```
simonpasquier, line 82: this needs to be moved outside of the Prometheus task (probably a separate task group).

stlaz: I was wondering about that, will do!

stlaz: I added it as a new task in front of the prometheus-operator one, let's see what happens.

stlaz:
the related test errors from the "common" test runs seem to have passed and the cluster installation is now successful.

I can see that the operator test still fails. Very likely because of the kube-rbac-proxy bug that's prevents our client-cert-authenticated metrics scraping requests from succeeding. That's a shame. I'll try to see if I can fix that.

simonpasquier:
Yes it looks like the error is now with the kube-rbac-proxy of the prometheus-operator in the openshift-user-workload-monitoring namespace [1].

Maybe you could only enable client TLS authentication for a few targeted service monitors (e.g. node_exporter, kubelet and kube-state-metrics)?

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/1282/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-operator/1417043735262269440/artifacts/e2e-agnostic-operator/gather-extra/artifacts/pods/openshift-user-workload-monitoring_prometheus-operator-6cc4bfb495-w4w9f_kube-rbac-proxy.log

stlaz:
I believe I've found the issue, we should be green once https://github.com/brancz/kube-rbac-proxy/pull/132 makes it to the release payload

go.mod
```diff
@@ -18,13 +18,13 @@ require (
 	github.com/prometheus/prometheus v1.8.2-0.20210518124745-6eeded0fdf76 // v1.8.2 is misleading as Prometheus does not have v2 module. This is pointing to v2.26.1, the same as in prometheus-operator v0.47.1
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.21.0
-	k8s.io/apiextensions-apiserver v0.21.0
-	k8s.io/apimachinery v0.21.0
-	k8s.io/apiserver v0.21.0
+	k8s.io/api v0.21.1
```
sthaha, line 21: shouldn't the `replace( ... )` change as well?

sthaha:
Would the PR also include tests?

stlaz:
I'm using the current e2e tests of this repository (quite successfully, actually, issues were found and are being fixed).

For unit tests, please refer to https://github.com/openshift/library-go/blob/8cc52ce9b72b591c2c1946c81c21f823bbc43a1d/pkg/operator/csr/controller_test.go#L33

go.mod
stlaz: thanks, fixed

stlaz:
fixed the go.mod in the latest revision

simonpasquier:
@sthaha I agree with @stlaz here about tests. The purpose of the change is to switch some service monitors to use TLS client auth (instead of bearer tokens) and the CMO e2e tests already [ensure](https://github.com/openshift/cluster-monitoring-operator/blob/2f35619d5b51058f918acbdb794109d5a6c8f163/test/e2e/main_test.go#L115-L135) that all monitoring components are scraped successfully so we should be good if all tests pass.

slaskawi:
@simonpasquier Just FYI - @stlaz started his PTO today and I'll be looking at this PR when he's off. Please let me know if we need any additional changes.

sthaha:
> @sthaha I agree with @stlaz here about tests. The purpose of the change is to switch some service monitors to use TLS client auth (instead of bearer tokens) and the CMO e2e tests already [ensure](https://github.com/openshift/cluster-monitoring-operator/blob/2f35619d5b51058f918acbdb794109d5a6c8f163/test/e2e/main_test.go#L115-L135) that all monitoring components are scraped successfully so we should be good if all tests pass.

Thank you both 🙇, I was wondering how this change is validated :)

sthaha:
> mhm, the tech debt of running things sequentially is showing as the operator refuses to progress to the client-ca creation while the prometheus-operator fails to deploy because it's missing the client-ca 😐 In an ideal world, this would be handled by two separate controllers.

@simonpasquier Should this go into our backlog?

stlaz:
> simonpasquier Should this go into our backlog?
you may want to consider library-go for your operator, I trust that the folks of #forum-apiserver would be willing to help, me including :)

simonpasquier (commented):
/lgtm

The CI is happy with the change so am I :)
Thanks @s-urbaniak @stlaz @slaskawi for the great work!

slaskawi:
:tada:

PR #1286, opened 2021-07-15
Remove use of deprecated APIs being removed in Kubernetes v1.22

simonpasquier:


go.mod
```diff
@@ -18,18 +18,18 @@ require (
...
 	k8s.io/klog/v2 v2.8.0
 	k8s.io/kube-aggregator v0.20.0
 	k8s.io/metrics v0.19.4
 )
 
 replace (
-	k8s.io/api => k8s.io/api v0.21.0
-	k8s.io/apimachinery => k8s.io/apimachinery v0.21.0
-	k8s.io/client-go => k8s.io/client-go v0.21.0
+	k8s.io/api => k8s.io/api v0.21.2
```
prashbnair, line 32: I think we need the replace only for the client-go library.

simonpasquier: correct! updated.

PR #1288, opened 2021-07-19
Bug 1982778: Favour HTTP probes over exec for Thanos query

PhilipGough:
This is cherry pick of #1277 which backports to 4.8.z

### DIFF
--- a/jsonnet/node-exporter.libsonnet
+++ b/jsonnet/node-exporter.libsonnet
@@ -138,7 +138,19 @@ function(params)
                       // gather that data (especially for bare metal clusters), and
                       // add flags to collect the node_cpu_info metric + metrics
                       // from the text file.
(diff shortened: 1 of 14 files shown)


jsonnet/node-exporter.libsonnet
```diff
@@ -138,7 +138,19 @@ function(params)
...
+                            [
+                              '--collector.cpu.info',
+                              '--collector.textfile.directory=' + textfileDir,
+
+                              // The cpufreq collector seems to be causing high
+                              // load on some nodes with lots of cores.  Disable
+                              // it temporarily as a workaround.
+                              //
+                              // https://bugzilla.redhat.com/show_bug.cgi?id=1972076
+                              // https://github.com/prometheus/node_exporter/issues/1880
```
prashbnair, line 151: Why do we want to do this? It does not seem related to the bug reported

PhilipGough: correct, that shouldnt be there :)

PR #1290, opened 2021-07-19
Remove context field from structs

slashpai:
Following the discussion in slack we need to remove context from struct field. Removing the context from struct field and passing it as first argument to functions that require it

### DIFF
--- a/pkg/client/status_reporter.go
+++ b/pkg/client/status_reporter.go
@@ -36,15 +36,13 @@ const (
 type StatusReporter struct {
 	client                clientv1.ClusterOperatorInterface
 	clusterOperatorName   string
-	ctx                   context.Context
 	namespace             string
 	userWorkloadNamespace string
 	version               string
 }
 
-func NewStatusReporter(ctx context.Context, client clientv1.ClusterOperatorInterface, name string, namespace, userWorkloadNamespace, version string) *StatusReporter {
+func NewStatusReporter(client clientv1.ClusterOperatorInterface, name string, namespace, userWorkloadNamespace, version string) *StatusReporter {
 	return &StatusReporter{
-		ctx:                   ctx,
 		client:                client,
 		clusterOperatorName:   name,
 		namespace:             namespace,
@@ -69,11 +67,11 @@ func (r *StatusReporter) relatedObjects() []v1.ObjectReference {
 	}
 }
 
-func (r *StatusReporter) SetDone(degradedConditionMessage string, degradedConditionReason string) error {
-	co, err := r.client.Get(r.ctx, r.clusterOperatorName, metav1.GetOptions{})
+func (r *StatusReporter) SetDone(ctx context.Context, degradedConditionMessage string, degradedConditionReason string) error {
+	co, err := r.client.Get(ctx, r.clusterOperatorName, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		co = r.newClusterOperator()
-		co, err = r.client.Create(r.ctx, co, metav1.CreateOptions{})
+		co, err = r.client.Create(ctx, co, metav1.CreateOptions{})
 	}
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
@@ -102,7 +100,7 @@ func (r *StatusReporter) SetDone(degradedConditionMessage string, degradedCondit
 		co.Status.Versions = nil
 	}
 
-	_, err = r.client.UpdateStatus(r.ctx, co, metav1.UpdateOptions{})
+	_, err = r.client.UpdateStatus(ctx, co, metav1.UpdateOptions{})
 	return err
 }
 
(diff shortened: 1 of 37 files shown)


slashpai:
@fpetkovski Could you please review?

cmd/operator/main.go
```diff
@@ -203,7 +203,7 @@ func Main() int {
 
 	wg, ctx := errgroup.WithContext(ctx)
 
-	wg.Go(func() error { return o.Run(ctx.Done()) })
+	wg.Go(func() error { return o.Run(ctx, ctx.Done()) })
```
simonpasquier, line 206: I would rather change the signature of Run() since it could use `ctx.Done()`

```suggestion
	wg.Go(func() error { return o.Run(ctx) })
```

test/e2e/framework/framework.go
```diff
@@ -143,7 +142,7 @@ func New(kubeConfigPath string) (*Framework, cleanUpFunc, error) {
 	// Prometheus client depends on setup above.
 	f.ThanosQuerierClient, err = NewPrometheusClientFromRoute(
 		openshiftRouteClient,
-		f.Ctx,
+		ctx,
```
simonpasquier, line 145: (nit) to follow the conventions, `ctx` should be the first argument.

pkg/client/status_reporter.go
```diff
@@ -34,15 +34,13 @@ const (
...
 	client                clientv1.ClusterOperatorInterface
 	clusterOperatorName   string
-	ctx                   context.Context
 	namespace             string
 	userWorkloadNamespace string
 	version               string
 }
 
-func NewStatusReporter(ctx context.Context, client clientv1.ClusterOperatorInterface, name string, namespace, userWorkloadNamespace, version string) *StatusReporter {
+func NewStatusReporter(client clientv1.ClusterOperatorInterface, name string, namespace, userWorkloadNamespace, version string) *StatusReporter {
```
sthaha, line 42: nit: `name string,` -> `name`

cmd/operator/main.go
sthaha: Is `operator.Run` trying to satisfy an interface that has `Run(ctx, stopCh <-struct{})`?

pkg/operator/operator.go
```diff
@@ -381,20 +378,20 @@ func (o *Operator) handleEvent(obj interface{}) {
 	o.enqueue(cmoConfigMap)
 }
 
-func (o *Operator) worker() {
-	for o.processNextWorkItem() {
+func (o *Operator) worker(ctx context.Context) {
+	for o.processNextWorkItem(ctx) {
```
sthaha, line 382: @simonpasquier I am guessing that if the context gets cancelled, the next work item be processed. Could you please confirm my hunch?

sthaha (commented):
The changes looks good to me but I think we need a test that validates the expected behaviour if a context is `cancelled` 
WDYT  @slashpai @simonpasquier ?

pkg/client/status_reporter.go
```diff
@@ -27,24 +27,22 @@ import (
...
-	asExpectedReason   string = "AsExpected"
-	StorageNotConfiguredMessage = "Prometheus is running without persistent storage which can lead to data loss during upgrades and cluster disruptions. Please refer to the official documentation to see how to configure storage for Prometheus: https://docs.openshift.com/container-platform/4.8/monitoring/configuring-the-monitoring-stack.html"
-	StorageNotConfiguredReason  = "PrometheusDataPersistenceNotConfigured"
+	unavailableMessage          string = "Rollout of the monitoring stack failed and is degraded. Please investigate the degraded status error."
+	asExpectedReason            string = "AsExpected"
+	StorageNotConfiguredMessage        = "Prometheus is running without persistent storage which can lead to data loss during upgrades and cluster disruptions. Please refer to the official documentation to see how to configure storage for Prometheus: https://docs.openshift.com/container-platform/4.8/monitoring/configuring-the-monitoring-stack.html"
+	StorageNotConfiguredReason         = "PrometheusDataPersistenceNotConfigured"
```
slashpai, line 33: This was changed when ran go fmt

pkg/operator/operator.go
simonpasquier: good catch! yes I agree that we should verify whether the context has been cancelled before processing the next item.

simonpasquier: at the same time, `Run()` currently doesn't wait for any of the goroutines to stop before exiting so it would need to be addressed not only for `worker()` IMHO. I'm ok with having a follow-up PR to "fix" this.

simonpasquier:
@slashpai could you rebase?

slashpai:
> @slashpai could you rebase?

Sure I will get that done when login tomorrow. Are we good with the changes or anymore changes are needed?

slashpai:
@simonpasquier rebased

PR #1293, opened 2021-07-21
MON-1749: Allow users to disable the local Alertmanager

fpetkovski:
The ACM would like to funnel all alerts from a fleet of OpenShift
clusters through a single external Alertmanager. As a result, they
would like to trim down the monitoring stack and disable the local
alertmanager.

This commit exposes a binary field in the cluster-monitoring-configmap
which allows users to fully disable the built-in alertmanager.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1054,8 +1054,13 @@ func (f *Factory) ThanosRulerAlertmanagerConfigSecret() (*v1.Secret, error) {
 	}
 	s.Namespace = f.namespaceUserWorkload
 
+	alertmanagerEnabled := f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.IsEnabled()
 	amConfigs := f.config.GetThanosRulerAlertmanagerConfigs()
-	if amConfigs == nil {
+	if amConfigs == nil && alertmanagerEnabled {
+		return s, nil
+	}
+	if amConfigs == nil && !alertmanagerEnabled {
+		s.StringData["alertmanagers.yaml"] = `alertmanagers: []`
 		return s, nil
 	}
 
@@ -1065,7 +1070,11 @@ func (f *Factory) ThanosRulerAlertmanagerConfigSecret() (*v1.Secret, error) {
 		return nil, err
 	}
 
-	s.StringData["alertmanagers.yaml"] += "\n" + string(additionalConfig)
+	if alertmanagerEnabled {
+		s.StringData["alertmanagers.yaml"] += "\n" + string(additionalConfig)
+	} else {
+		s.StringData["alertmanagers.yaml"] = `alertmanagers:` + "\n" + string(additionalConfig)
+	}
 
 	return s, nil
 }
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1056,11 +1056,10 @@ func (f *Factory) ThanosRulerAlertmanagerConfigSecret() (*v1.Secret, error) {
 
 	alertmanagerEnabled := f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.IsEnabled()
 	amConfigs := f.config.GetThanosRulerAlertmanagerConfigs()
-	if amConfigs == nil && alertmanagerEnabled {
-		return s, nil
-	}
-	if amConfigs == nil && !alertmanagerEnabled {
-		s.StringData["alertmanagers.yaml"] = `alertmanagers: []`
+	if amConfigs == nil {
+		if !alertmanagerEnabled {
+			s.StringData["alertmanagers.yaml"] = `alertmanagers: []`
+		}
 		return s, nil
 	}
 
--- a/test/e2e/alertmanager_test.go
+++ b/test/e2e/alertmanager_test.go
@@ -17,6 +17,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	statusv1 "github.com/openshift/api/config/v1"
 	"io/ioutil"
 	"net/http"
 	"testing"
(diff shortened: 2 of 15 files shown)


prashbnair (approved):

prashbnair:
lgtm

PhilipGough (approved):
/lgtm

test/e2e/framework/framework.go
```diff
@@ -80,6 +83,14 @@ func New(kubeConfigPath string) (*Framework, cleanUpFunc, error) {
 		return nil, nil, errors.Wrap(err, "creating kubeClient failed")
 	}
 
+	jsonClientConfig := rest.CopyConfig(config)
```
simonpasquier, line 86: we already create `openshiftRouteClient` just below, why not use it?

simonpasquier (commented):
/hold
just one small comment, I think we want to see the impact on the console too before merging.

test/e2e/framework/framework.go
fpetkovski: Good point, addressed.

simonpasquier:
/lgtm cancel

testing locally CMO goes degraded when Alertmanager is disabled

```
Failed to rollout the stack. Error: running task Updating configuration sharing failed: failed to retrieve Alertmanager host: getting Route object failed: routes.route.openshift.io "alertmanager-main" not found
```

fpetkovski:
Ack, do we still need that public URL though? Can we remove it from the shared config?

simonpasquier:
@PhilipGough has https://github.com/openshift/cluster-monitoring-operator/pull/1223 in progress to remove the shared configmap but it's still waiting on code changes in the console repository.

fpetkovski:
@simonpasquier we had a chat with @kyoto and his assessment is that it's fine to merge this PR before the console changes are in place. The reason for this once the AM is disabled, the console shows a message that it cannot connect to it and there isn't a major degradation to the UI. 

Also, @kyoto will speak to the PM during planning to see how much bandwidth he has for the next sprint. If he has no time to handle the console changes, I will take over the work and he will review the PR.

simonpasquier:
Thanks for the heads-up, I agree with @kyoto that the current UX in the console is good enough (disabling Alertmanager is a deliberate choice so you shouldn't be too surprised that the console fails to find the local Alertmanager).

simonpasquier:
I don't see any test in `ConfigSharingTask` that verifies whether the local Alertmanager is disabled (just like we do for Grafana)? Am I missing something?

https://github.com/openshift/cluster-monitoring-operator/blob/e3023648dee4860e07b04e329f2a64c4f50061ef/pkg/tasks/configsharing.go#L63-L73

fpetkovski:
Based on one of your previous comments, I thought we would wait for @PhilipGough's task to remove the shared config to be merged. Let me know if I should handle the case in this PR instead.

simonpasquier:
Sorry for the confusion :) Yes let's handle it in this PR because we're not sure when the other one is going to be merged.

fpetkovski:
@simonpasquier to make the review easier, I added a new commit which handles this edge case and removes the public url from the shared configmap when AM is disabled.

simonpasquier (commented):
/lgtm

pkg/manifests/manifests.go
```diff
@@ -1054,8 +1054,13 @@ func (f *Factory) ThanosRulerAlertmanagerConfigSecret() (*v1.Secret, error) {
...
+	alertmanagerEnabled := f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.IsEnabled()
 	amConfigs := f.config.GetThanosRulerAlertmanagerConfigs()
-	if amConfigs == nil {
+	if amConfigs == nil && alertmanagerEnabled {
+		return s, nil
+	}
+	if amConfigs == nil && !alertmanagerEnabled {
+		s.StringData["alertmanagers.yaml"] = `alertmanagers: []`
 		return s, nil
 	}
```
simonpasquier, line 1065:
```suggestion
	if amConfigs == nil {
		if !alertmanagerEnabled {
			s.StringData["alertmanagers.yaml"] = `alertmanagers: []`
		}
		return s, nil
	}
```

test/e2e/alertmanager_test.go
```diff
@@ -377,6 +377,10 @@ func TestAlertmanagerDisabling(t *testing.T) {
 		{name: "assert service alertmanager-main does not exist", assertion: f.AssertServiceDoesNotExist("alertmanager-main", f.Ns)},
 		{name: "assert service alertmanager-operated does not exist", assertion: f.AssertServiceDoesNotExist("alertmanager-operated", f.Ns)},
 		{name: "assert serviceaccount alertmanager-main does not exist", assertion: f.AssertServiceAccountDoesNotExist("alertmanager-main", f.Ns)},
+		{name: "assert role monitoring-alertmanager-edit exists does not exist", assertion: f.AssertRoleDoesNotExist("monitoring-alertmanager-edit", "openshift-monitoring")},
```
simonpasquier, line 380:
```suggestion
		{name: "assert role monitoring-alertmanager-edit does not exist", assertion: f.AssertRoleDoesNotExist("monitoring-alertmanager-edit", "openshift-monitoring")},
```

dgrisonnet:
/retitle MON-1749: Allow users to disable the local Alertmanager

fpetkovski:
@juzhao if this is good to merge, could you add the qe approval label?

@jan--f could you ping the px and docs representatives for our team?

sferich888:
Would users have to turn off the local 'alertmanager' for the information to be aggregated to ACM? What if I wanted to get alerts on cluster; and within ACM? Would that be possible? 


cc: @jcarter12

fpetkovski:
@sferich888 users can turn the local alertmanager and configure external alertmanagers as independent features. So you they can do both depending on how they configure CMO.

pkg/manifests/manifests.go
```diff
@@ -1065,7 +1069,11 @@ func (f *Factory) ThanosRulerAlertmanagerConfigSecret() (*v1.Secret, error) {
 		return nil, err
 	}
 
-	s.StringData["alertmanagers.yaml"] += "\n" + string(additionalConfig)
+	if alertmanagerEnabled {
+		s.StringData["alertmanagers.yaml"] += "\n" + string(additionalConfig)
+	} else {
+		s.StringData["alertmanagers.yaml"] = `alertmanagers:` + "\n" + string(additionalConfig)
```
raptorsun, line 1075: just curious, why we care additional config for alter manager while it is disabled? 🤔

PR #1296, opened 2021-07-23
Bug 1978091: fix node_exporter recording rules for cluster network dashboards

simonpasquier:
https://github.com/openshift/cluster-monitoring-operator/pull/1214 introduced a change in the node_exporter recording rules that breaks the console. And since node_exporter metrics are scraped every 15 seconds, a time range of 1m is a correct choice.


PhilipGough (approved):
/lgtm

PR #1305, opened 2021-07-29
WIP: changes for static auth

prashbnair:

### DIFF
--- a/assets/alertmanager/kube-rbac-proxy-secret.yaml
+++ b/assets/alertmanager/kube-rbac-proxy-secret.yaml
@@ -15,4 +15,9 @@ stringData:
       "rewrites":
         "byQueryParameter":
           "name": "namespace"
+      "static":
+        - "apiGroup": "monitoring.coreos.com"
+          "namespace": "{{ .Value }}"
+          "resource": "prometheusrules"
+          "resourceRequest": true
 type: Opaque
--- a/assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
+++ b/assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
@@ -9,11 +9,17 @@ metadata:
 stringData:
   config.yaml: |-
     "authorization":
+      "rewrites":
+        "byQueryParameter":
+          "name": "namespace"
+      "static":
+        - "apiGroup": "metrics.k8s.io/v1beta1"
+          "namespace": "{{ .Value }}"
+          "resource": "pods"
+          "resourceRequest": true
       "resourceAttributes":
         "apiVersion": "metrics.k8s.io/v1beta1"
         "namespace": "{{ .Value }}"
         "resource": "pods"
-      "rewrites":
-        "byQueryParameter":
-          "name": "namespace"
+
 type: Opaque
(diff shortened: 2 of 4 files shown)


assets/alertmanager/kube-rbac-proxy-secret.yaml
```diff
@@ -15,4 +15,9 @@ stringData:
       "rewrites":
         "byQueryParameter":
           "name": "namespace"
+      "static":
+        - "apiGroup": "monitoring.coreos.com"
+          "namespace": "{{ .Value }}"
+          "resource": "prometheusrules"
+          "resourceRequest": true
```
s-urbaniak, line 22: we don't need static authorization for the /rules endpoint. The purpose is to be able to scrape metrics in the face of a network partition or a failed apiserver.

assets/thanos-querier/kube-rbac-proxy-rules-secret.yaml
```diff
@@ -16,6 +16,11 @@ stringData:
         "apiGroup": "monitoring.coreos.com"
         "namespace": "{{ .Value }}"
         "resource": "prometheusrules"
+      "static":
+        - "apiGroup": "monitoring.coreos.com"
+          "namespace": "{{ .Value }}"
+          "resource": "prometheusrules"
+          "resourceRequest": true
```
s-urbaniak, line 23: see above

assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
```diff
@@ -9,11 +9,17 @@ metadata:
...
   config.yaml: |-
     "authorization":
+      "rewrites":
+        "byQueryParameter":
+          "name": "namespace"
+      "static":
+        - "apiGroup": "metrics.k8s.io/v1beta1"
+          "namespace": "{{ .Value }}"
+          "resource": "pods"
+          "resourceRequest": true
```
s-urbaniak, line 19: this would actually not verify against a concrete user. the purpose of static authorization is to bind it against a concrete user, see https://github.com/openshift/kube-rbac-proxy/pull/43/files#diff-c6b1500c9d0e568b9d4e4712df03c09f384ce19722c3befbd59847dc4fbfd431R43-R50.

In this case we also need to check against the `system:serviceaccount:openshift-monitoring:prometheus-k8s` user, so the section would need to add:
```
        - user:
            name: system:serviceaccount:openshift-monitoring:prometheus-k8s
```

s-urbaniak:
generally: you need to modify the underlying jsonnet code, all files in `assets` are generated.

assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
prashbnair: Could you please explain why it needs to be hard coded to this SA. If a user is logged in, isnt that user token used?

s-urbaniak: you don't assert if a user is logged in here. the current declaration is actually a no-op. the purpose of this is use case is to authorize prometheus service accounts only to be able to scrape its own metrics without performing a subject access review against API server. See a more elaborate rationale in https://github.com/deads2k/openshift-enhancements/blob/master/enhancements/monitoring/client-cert-scraping.md

prashbnair:
> generally: you need to modify the underlying jsonnet code, all files in assets are generated.
Yes, I am aware, I was trying to quickly test out something.

prashbnair:
Closing this out, will raise a new one with changes.

PR #1307, opened 2021-07-30
Bug 1988291: pkg/client/client.go: Add retry logic for daemonset create

slashpai:

Allow retry for daemon set creation as commented in https://github.com/openshift/cluster-monitoring-operator/pull/1257/#pullrequestreview-700719930

cc: @dgrisonnet @simonpasquier 


sdodson:
/override ci/prow/e2e-agnostic-upgrade
This job is not passing regularly and the retry is meant to work around noted api downtime or unrespnsiveness.

PR #1308, opened 2021-07-30
[WIP] Expose remote write

jan--f:
_This is a WIP PR as https://github.com/prometheus-operator/prometheus-operator/pull/4180 is still being reviewed_

This exposes `remote_write`to users. CMO gets its own `remote_write` struct in order to hide unsafe fields (like `BearerToken`) and to potentially add fields later so simpler specification for this feature.


go.mod
```diff
@@ -29,3 +29,5 @@ require (
 )
 
 replace k8s.io/client-go => k8s.io/client-go v0.21.2
+
+replace github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.48.1 => github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.49.1-0.20210803083158-78e0ea966b65
```
simonpasquier, line 33: I would stick with v0.49.0 and leave aside Authorization and OAuth2 support for now (it can be added in 4.10 once we upgrade to a version of the Prometheus operator supporting them).

jan--f: Oh yeah, good point ty.

jan--f:
cc @juzhao @sferich888 tests for this are added in #1328 
@rkratky doc input was provided in https://issues.redhat.com/browse/MON-1687

simonpasquier (commented):
/lgtm

PR #1312, opened 2021-08-04
jsonnet: Support exluding namespaces from user-workload monitoring

bison:
This adds support for setting an `openshift.io/user-monitoring` label
on namespaces to exclude them from user-workload monitoring.  This is
in addition to the exclusion of namespaces that have set a `true`
value for the `openshift.io/cluster-monitoring` label.

### DIFF
--- a/test/e2e/user_workload_monitoring_test.go
+++ b/test/e2e/user_workload_monitoring_test.go
@@ -46,7 +46,6 @@ const (
 	userWorkloadTestNs = "user-workload-test"
 )
 
-
 type scenario struct {
 	name      string
 	assertion func(*testing.T)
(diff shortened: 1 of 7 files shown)


test/e2e/user_workload_monitoring_test.go
```diff
@@ -1181,6 +1181,61 @@ func assertEnforcedTargetLimit(limit uint64) func(*testing.T) {
...
+	})
+
+	// Add opt-out label to namespace.
+	ns, err := f.KubeClient.CoreV1().Namespaces().Get(ctx, userWorkloadTestNs, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("Failed to fetch user-workload namespace: %v", err)
+	}
+
+	labels := ns.GetLabels()
+	labels["openshift.io/user-monitoring"] = "false"
```
sferich888, line 1201: Should we use a more 'explicit' label? `openshift.io/user-moitoring-excluded`?

sferich888: I understand that `true` means that you want to 'monitor' the namespace; and `false` indicates that you don't want to monitor; however not having the label does the same thing *(I think). Thus making sure this is clearly documented will be EXTREMELY important.

bison: I think the label was chosen to be consistent with the existing `openshift.io/cluster-monitoring` label. I don't have much of an opinion on the naming. I agree it needs to be documented. I will prep a docs PR.

bison:
I've added a docs PR for this: https://github.com/openshift/openshift-docs/pull/35289

Hi @rkratky, could you have a look at that, and if everything looks good and we think we can get the docs in for 4.9, then can we get no-feature-freeze approval for this?

PR #1318, opened 2021-08-11
MON-1679: adding static auth to KSM

prashbnair:

### DIFF
--- a/jsonnet/components/kube-state-metrics.libsonnet
+++ b/jsonnet/components/kube-state-metrics.libsonnet
@@ -2,6 +2,7 @@ local tmpVolumeName = 'volume-directive-shadow';
 local tlsVolumeName = 'kube-state-metrics-tls';
 
 local kubeStateMetrics = import 'github.com/prometheus-operator/kube-prometheus/jsonnet/kube-prometheus/components/kube-state-metrics.libsonnet';
+local generateSecret = import '../utils/generate-secret.libsonnet';
 
 function(params)
   local cfg = params;
@@ -72,6 +73,8 @@ function(params)
       },
     },
 
+    kubeRbacProxySecret: generateSecret.staticAuthSecret(cfg.namespace, cfg.commonLabels, 'kube-state-metrics-kube-rbac-proxy-config'),
+
     // This removes the upstream addon-resizer and all resource requests and
     // limits. Additionally configures the kube-rbac-proxies to use the serving
     // cert configured on the `Service` above.
@@ -94,6 +97,7 @@ function(params)
                         '--tls-cert-file=/etc/tls/private/tls.crt',
                         '--tls-private-key-file=/etc/tls/private/tls.key',
                         '--client-ca-file=/etc/tls/client/client-ca.crt',
+                        '--config-file=/etc/kube-rbac-policy/config.yaml',
                       ],
                       terminationMessagePolicy: 'FallbackToLogsOnError',
                       volumeMounts: [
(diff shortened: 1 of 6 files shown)


prashbnair:
@jan--f @s-urbaniak  I have implemented the changes for KSM and if this looks good, I can implement it for other components.

simonpasquier:
lgtm
Have you tried it with https://github.com/openshift/kube-rbac-proxy/pull/51 to make sure that it works without the hardcoded authorizer?

simonpasquier:
Looking at https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/1318/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-operator/1426158852574810112 it seems that the e2e operator tests are failing because we hit the 30 minutes timeout...

prashbnair:
> Have you tried it with openshift/kube-rbac-proxy#51 to make sure that it works without the hardcoded authorizer?

Yes

prashbnair:
> Looking at https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/1318/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-operator/1426158852574810112 it seems that the e2e operator tests are failing because we hit the 30 minutes timeout...

This does not seem related to my change, the test that is failing is for thanos querier and I have not made any changes in that area.

jan--f:
The approach looks good I think. Maybe the secret could have a more descriptive name, i.e. mention static auth for the metric path?

simonpasquier:
> This does not seem related to my change, the test that is failing is for thanos querier and I have not made any changes in that area.

Correct I wasn't suggesting that it was introduced by your PR.

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -72,6 +72,34 @@ function(params)
       },
     },
 
+    kubeRbacProxySecret: {
+      apiVersion: 'v1',
+      kind: 'Secret',
+      metadata: {
+        name: 'kube-state-metrics-kube-rbac-proxy',
```
simonpasquier, line 79: As @jan--f suggested, the secret could be more descriptive 

```suggestion
        name: 'kube-state-metrics-kube-rbac-proxy-config',
```

prashbnair:
/retitle MON-1679 adding static auth to KSM

jsonnet/components/kube-state-metrics.libsonnet
prashbnair: Done

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -107,6 +136,11 @@ function(params)
                           name: 'metrics-client-ca',
                           readOnly: false,
                         },
+                        {
+                          mountPath: '/etc/kube-rbac-policy',
+                          name: 'kube-state-metrics-kube-rbac-proxy-config',
+                          readOnly: false,
```
simonpasquier, line 142: (nit)

```suggestion
                          readOnly: true,
```

simonpasquier (commented):
/hold waiting for 4.10
LGTM, I think you can go ahead with the other components. Maybe there' some way to have jsonnet functions to avoid code duplication?

prashbnair:
Added jsonnet function to generate the secret.

simonpasquier (commented):
Can you do the same change for the other components?
* node-exporter
* openshift-state-metrics
* prometheus-operator
* prometheus-operator-user-workload
* prometheus-user-workload
* telemeter-client

prashbnair:
Can we do the changes in stages, once this PR is merged I will start work on the other components and raise PR's for each of them.

simonpasquier:
sure, lets do it in stages.

/lgtm

/label px-approved
/label docs-approved
/label qe-approved

The change has no user-facing impact and the CI should fail if the metrics collection is broken.

PR #1319, opened 2021-08-11
jsonnet: Add missing namespace labels to alerting rules

bison:
All OpenShift alerts should include a namespace label for routing.
We've queried Telemeter data and identified several shipped by CMO
that don't include a namespace.  This adds the labels to these by
either adjusting the expressions in the rules, or simply adding static
labels were appropriate.

See: https://issues.redhat.com/browse/MON-939

### DIFF
--- a/jsonnet/rules.libsonnet
+++ b/jsonnet/rules.libsonnet
@@ -381,6 +381,9 @@ function(params) {
           },
           labels: {
             severity: 'warning',
+            // All OpenShift alerts should have a namespace label.
+            // See: https://issues.redhat.com/browse/MON-939
+            namespace: 'openshift-monitoring',
           },
         },
         {
@@ -414,6 +417,9 @@ function(params) {
           },
           labels: {
             severity: 'info',
+            // All OpenShift alerts should have a namespace label.
+            // See: https://issues.redhat.com/browse/MON-939
+            namespace: 'kube-system',
           },
         },
         {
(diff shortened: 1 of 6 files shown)


jsonnet/rules.libsonnet
```diff
@@ -381,6 +381,9 @@ function(params) {
           },
           labels: {
             severity: 'warning',
+            // All OpenShift alerts should have a namespace label.
+            // See: https://issues.redhat.com/browse/MON-939
+            namespace: 'openshift-monitoring',
```
raptorsun, line 386: Shall we add a namespace to other alerts in this file, too?  for example the previous alert "ClusterMonitoringOperatorReconciliationErrors".

bison: This PR adds the namespace label to alerts that we know are missing them. In most cases that means adding a static label like this, because the query used in the rule is cross namespace, or can't be grouped by namespace for some other reason. The other alerts will already include a namespace label because the queries produce results with one. So, no need to change anything in most cases.

bison:
I pinged everyone in JIRA also, so please forgive the spam, but we'd like to get this in via the no-feature-freeze process for the 4.9 release. Can @sferich888 @juzhao and @rkratky please have a look here?

This is just ensuring that all the alerts we ship have a namespace label in the final output. Part of [MON-939](https://issues.redhat.com/browse/MON-939). No docs changes needed.

juzhao:
test with the PR, there is namespace label for the alerting rules
/label qe-approved

PR #1322, opened 2021-08-16
Bug 1989438: changing error message to indicate deployment status

prashbnair:

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -795,7 +795,8 @@ func (c *Client) WaitForDeploymentRollout(ctx context.Context, dep *appsv1.Deplo
 			return false, nil
 		}
 		if d.Status.UpdatedReplicas != d.Status.Replicas {
-			lastErr = errors.Errorf("expected %d replicas, got %d updated replicas",
+			lastErr = errors.Errorf("the number of pods targeted by the deployment (%d pods) is different "+
+				"from the number of pods targeted by the deployment that have the desired template spec (%d pods)",
 				d.Status.Replicas, d.Status.UpdatedReplicas)
 			return false, nil
 		}


prashbnair:
When we use a node affinity rule that cannot be complied with, the deployment stalls and does not make any progress. In this state, the replica numbers reported by the status is incorrect. This is because new replicas have been brought up to be assigned as per new affinity rules. This PR checks if the deployment is progressing and reports an error if it is not.

pkg/client/client.go
```diff
@@ -794,6 +794,15 @@ func (c *Client) WaitForDeploymentRollout(ctx context.Context, dep *appsv1.Deplo
 				d.Generation, d.Status.ObservedGeneration)
 			return false, nil
 		}
+
+		deploymentConditions := d.Status.Conditions
+
+		for _, deploymentCondition := range deploymentConditions {
+			if deploymentCondition.Type == appsv1.DeploymentProgressing && deploymentCondition.Status == v1.ConditionFalse {
```
simonpasquier, line 801:
```suggestion
			if deploymentCondition.Type == appsv1.DeploymentProgressing && deploymentCondition.Status != v1.ConditionTrue {
```

simonpasquier:
I'm not sure that it would "fix" the issue: I see that `DeploymentProgressing` would be updated to `v1.ConditionFalse` when it hits the deployment's progress deadline which is 600s by default while CMO waits for 300s for the deployment to be complete.
TBH I'm not sure that we need to change the logic and maybe updating the log message would make it less confusing?

```
lastErr = errors.Errorf("the number of pods targeted by the deployment (%d pods) is different from the number of pods targeted by the deployment that have the desired template spec (%d pods)", d.Status.Replicas, d.Status.UpdatedReplicas)
```

prashbnair:
I dont think that message will fix the issue. The number of replicas is actually incorrect because it brings up a replica but it cannot be scheduled because there is no node. So the count is actually always one more than the required number of replicas.

dgrisonnet:
The error log is actually valid in my opinion. I tried to reproduce the bug and the deployment is always targeting N+1 pods because, during the rolling update of the deployment, the new pod can't be scheduled to a node and is thus left in a pending state. So we end having a deployment targeting the N existing pods and the new pending pod. That said I don't think we should change any logic, but maybe we could improve the error message.

prashbnair:
/retitle Bug 1989438: changing error message to indicate deployment status

juzhao:
tested with the PR, the error now looks normal, see https://bugzilla.redhat.com/show_bug.cgi?id=1989438#c4
/label qe-approved

PR #1328, opened 2021-08-19
test/e2e: add remote_write http e2e test

jan--f:
This only adds a test for an unencrypted enpoint. An mTLS test will follow in a separate PR, but use the same test function.

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -15,13 +15,21 @@
 package e2e
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"strings"
 	"testing"
 	"time"
 
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+
+	osConfigv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
+	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
+	"github.com/pkg/errors"
 )
 
 func TestPrometheusMetrics(t *testing.T) {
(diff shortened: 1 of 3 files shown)


test/e2e/prometheus_test.go
```diff
@@ -92,3 +97,180 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+			t.Fatal(promLogErr)
+		}
+		if i, _ := promLogs.WriteString(promLogs1); i == 0 {
+			t.Fatal(promLogErr)
+		}
+
+		rwEndpointOpts := metav1.ListOptions{
+			LabelSelector: fields.SelectorFromSet(fields.Set(map[string]string{
+				"group": name,
+			})).String(),
```
dgrisonnet, line 257: This can be simplified by using [labels.FormatLabels](https://pkg.go.dev/k8s.io/apimachinery/pkg/labels#FormatLabels).

test/e2e/prometheus_test.go
```diff
@@ -92,3 +97,180 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		promLogs1, err := f.GetLogs(f.Ns, "prometheus-k8s-1", "prometheus")
+		if err != nil {
+			t.Fatal(err)
+		}
+		var promLogs strings.Builder
+		promLogErr := "prometheus logs are empty"
```
dgrisonnet, line 246: I am not too familiar with RW so this error message is quite confusing to me. I think we should add what was expected to make it clearer?

test/e2e/prometheus_test.go
```diff
@@ -92,3 +97,180 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+			t.Fatal(promLogErr)
+		}
+
+		rwEndpointOpts := metav1.ListOptions{
+			LabelSelector: fields.SelectorFromSet(fields.Set(map[string]string{
+				"group": name,
+			})).String(),
+		}
+
+		rwEndpointPodList, err := f.KubeClient.CoreV1().Pods(f.Ns).List(context.Background(), rwEndpointOpts)
```
dgrisonnet, line 260:
```suggestion
		rwEndpointPodList, err := f.KubeClient.CoreV1().Pods(f.Ns).List(ctx, rwEndpointOpts)
```

test/e2e/prometheus_test.go
```diff
@@ -92,3 +97,180 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+	for _, scenario := range []struct {
+		name  string
+		proto string
+	}{
+		// check remote write logs
+		{
+			name:  "assert remote write to http works",
+			proto: "http",
+		},
+	} {
```
dgrisonnet, line 220: Since you wrote this test with the intent of having multiple test cases in the future, it would be great to split them into sub-tests via `t.Run()` and differentiate them via their names.

test/e2e/prometheus_test.go
```diff
@@ -92,3 +97,180 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+	  name: certs
+        - mountPath: /etc/certs/tls
+          name: rw-target-tls
+        - mountPath: /etc/certs/ca
+          name: serving-certs-ca-bundle
+`
+	deployment := &appsv1.Deployment{}
+	if err := yaml.NewYAMLOrJSONDecoder(bytes.NewReader([]byte(targetDeployment)), 100).Decode(&deployment); err != nil {
+		t.Fatal(err)
+	}
```
dgrisonnet, line 166: nit: we could reuse https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/manifests/manifests.go#L3680-L3688

jan--f:
> suggestion: instead of grepping the logs, you could check remote write metrics such as `prometheus_remote_write_samples_total` and `prometheus_remote_write_samples_failed_total`?

I spend some time looking into this but its not straight forward. 
Iiuc:
`prometheus_remote_write_samples_total` tracks the number of samples that went into remote write POST requests.
`prometheus_remote_write_samples_failed_total` only increases if prometheus considers a request failed due to a non-recoverable error. A 404 on the POST request doesn't seem to do the trick.
`prometheus_remote_storage_samples_retried_total` is close, when remote write works well this should be zero. However this does seem brittle as well, as retrying is generally not a bad thing.


We could watch `prometheus_remote_storage_samples_pending` and if that is equal to `prometheus_remote_storage_shards*prometheus_remote_storage_shard_capacity+prometheus_remote_storage_max_samples_per_send` that would indicates nothing is being send. Wdyt?

jan--f:
@fpetkovski @simonpasquier ptal, I addressed review comments and added a test for an mTLS secured endpoint.

test/e2e/prometheus_test.go
```diff
@@ -92,3 +97,223 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		t.Run(scenario.name, checkRemoteWrite(name, ctx))
+	}
+}
+
+func checkRemoteWrite(rwEndpointName string, ctx context.Context) func(*testing.T) {
+	return func(t *testing.T) {
+		remoteWriteCheckLogs(rwEndpointName, ctx, t)
```
fpetkovski, line 256: Do we still want to make assertions on the logs? I thought the idea was to use metrics instead since they are more stable than logs.

fpetkovski (commented):
Overall this looks good, I just left comment about the general intent of the test.

test/e2e/utils.go
```diff
@@ -73,3 +79,65 @@ func getThanosRules(body []byte, expGroupName, expRuleName string) error {
 	}
 	return fmt.Errorf("'%s' alert not found in '%s' group", expRuleName, expGroupName)
 }
+
+func createSelfSignedMTLSArtifacts(s *v1.Secret) error {
```
fpetkovski, line 83: Just a small suggestion: if you don't need to share the root CA between the writer and receiver, you can also use [GenerateSelfSignedCertKey](https://pkg.go.dev/k8s.io/client-go/util/cert#GenerateSelfSignedCertKey) to simplify this function.

test/e2e/prometheus_test.go
jan--f: I would argue yes. Errors in the logs will give a clearer indication to issues in the mTLS setup for example. Checking the metric give a clearer signal that the requests are actually happening.

Where do you expect instability regarding the logs or any downside with checking the logs?

test/e2e/utils.go
jan--f: If we only needed a serving certificate we would use the [Service CA operator](https://github.com/openshift/service-ca-operator), but for mTLS we need a server certificate and a client certificate signed be the same CA (or at least a mutually trusted CAs).

test/e2e/prometheus_test.go
fpetkovski: I don't expect instability per se, but I also do not think there is any guarantee that log messages would remain unchanged in future Prometheus versions. For example, someone might decide to reword a message for clarity, change the verbosity level of a message, change the format, add punctuation etc. Also, logs are usually meant to be read by humans. On the other hand, metrics are considered an API and are also intended for machines. Removing them is usually a breaking change which is documented in some release notes.

That being said, I think that in this particular test case it's fine to check the logs if using only metrics is not sufficient to verify correct behavior.

test/e2e/utils.go
jan--f: > You can usually specify the CA used for validating the certificate from the other end.
Not sure what you mean by "the other end" here?

We need a client cert that the server can verify and a server cert that the client can verify. We solve this here by creating on CA (that signed both the server and client cert) and passing that CA to both parties.

test/e2e/utils.go
```diff
@@ -73,3 +79,65 @@ func getThanosRules(body []byte, expGroupName, expRuleName string) error {
...
+
+		crt, key, err := cfg.GetPEMBytes()
+		if err != nil {
+			return errors.Wrap(err, "error getting PEM bytes for server certificate")
+		}
+		s.Data["server.crt"] = crt
+		s.Data["server.key"] = key
+		s.Data["server-ca.pem"] = append(crt, newCABytes...)
+	}
+	// create clien keys the same way
```
simonpasquier, line 122:
```suggestion
	// create client keys the same way
```

test/e2e/utils.go
```diff
@@ -73,3 +79,65 @@ func getThanosRules(body []byte, expGroupName, expRuleName string) error {
...
+	}
+
+	newCABytes, newCAKeyBytes, err := newCA.Config.GetPEMBytes()
+	if err != nil {
+		return errors.Wrap(err, "error getting PEM bytes from CA")
+	}
+
+	s.Data["ca.crt"] = newCABytes
+	s.Data["ca.key"] = newCAKeyBytes
+	// create serving certs maybe according to secret keys?
```
simonpasquier, line 104: not sure about this comment

test/e2e/utils.go
```diff
@@ -73,3 +79,65 @@ func getThanosRules(body []byte, expGroupName, expRuleName string) error {
...
+		)
+		if err != nil {
+			return errors.Wrap(err, "error making server certificate")
+		}
+
+		crt, key, err := cfg.GetPEMBytes()
+		if err != nil {
+			return errors.Wrap(err, "error getting PEM bytes for server certificate")
+		}
+		s.Data["server.crt"] = crt
```
simonpasquier, line 118: do we need this entry? IIUC only `server.key` and `server-ca.pem` are used.

test/e2e/prometheus_test.go
```diff
@@ -92,3 +97,226 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+
+		remoteWriteCheckMetrics(ctx, t)
+	}
+}
+
+func remoteWriteCheckLogs(rwEndpointName string, ctx context.Context, t *testing.T) {
```
simonpasquier, line 265:
```suggestion
func remoteWriteCheckLogs(ctx context.Context,  rwEndpointName string, t *testing.T) {
```

test/e2e/prometheus_test.go
```diff
@@ -92,3 +97,226 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+	time.Sleep(1 * time.Minute)
+	for _, pod := range []string{
+		"prometheus-k8s-0",
+		"prometheus-k8s-1",
+	} {
+		f.ThanosQuerierClient.WaitForQueryReturn(
+			t, 1*time.Minute, fmt.Sprintf(`ceil(delta(prometheus_remote_storage_samples_pending{pod="%s"}[1m]))`, pod),
+			func(v int) error {
+				if v == 0 {
+					return fmt.Errorf("prometheus_remote_storage_samples_pending indcates no remote write progress")
```
simonpasquier, line 315:
```suggestion
					return fmt.Errorf("prometheus_remote_storage_samples_pending indicates no remote write progress")
```

test/e2e/utils.go
fpetkovski: Sorry about the confusion, what I was trying to suggest is shortening the function like this:
```
func createSelfSignedMTLSArtifacts(s *v1.Secret) error {
	certBytesServer, keyBytesServer, _ := cert.GenerateSelfSignedCertKey("openshift-cluster-monitoring-test", nil, nil)
	certBytesClient, keyBytesClient, _ := cert.GenerateSelfSignedCertKey("openshift-cluster-monitoring-test", nil, nil)

	s.Data["server.crt"] = certBytesServer
	s.Data["server.key"] = keyBytesServer
	s.Data["client.crt"] = certBytesClient
	s.Data["client.key"] = keyBytesClient
	s.Data["ca.crt"] = append(certBytesClient, certBytesServer...)

	return nil
}
```
However, I now see that this will fail because the instrumented app uses the same file for the tls certificate and CA certificate :(.

So let's not waste too much time with this since it might end up being a rabbit hole. Maybe once we replace the instrumented app with Prometheus as a remote write target we can reconsider this again.

test/e2e/prometheus_test.go
```diff
@@ -92,3 +97,226 @@ func TestPrometheusAlertmanagerAntiAffinity(t *testing.T) {
...
+`,
+		},
+	} {
+		rw := fmt.Sprintf(scenario.rwSpec, deployedService.Spec.ClusterIP+":"+scenario.port)
+
+		cmoConfigMap := fmt.Sprintf(`prometheusK8s:
+  logLevel: debug
+  remoteWrite: %s
+`, rw)
+		if err := f.OperatorClient.CreateOrUpdateConfigMap(ctx, configMapWithData(t, cmoConfigMap)); err != nil {
```
simonpasquier, line 249: we have to wait until the configuration has been rolled out successfully to avoid races?

test/e2e/utils.go
jan--f: Yes `server-ca.pem` is mostly a convenience field for the sample app, as that expects the certificate and CA on one pem chunk. Since this lives in `utils` I left `server.key` in there for future use cases, I figured it wouldn't harm anything.

test/e2e/prometheus_test.go
jan--f: Yes that is probably a good idea, added.

test/e2e/utils.go
jan--f: yeah same here :thinking:

test/e2e/utils.go
jan--f: Another hitch here is that `GenerateSelfSignedCertKey` creates a server cert specifically. I.e. this certificate will be missing `x509.ExtKeyUsageClientAuth` in the `KeyUsage` list.

:+1: for revisiting this at a later stage.

test/e2e/utils.go
simonpasquier: :+1:

PR #1331, opened 2021-08-20
Rule patching can use index to distinguish alert / record rules under the same name

raptorsun:

The function patchOrExcludeRule uses an additional field "index" to distinguish different rules under the same name (alert/record field is considered as its name). Rules under the same group are ordered by alphabetic order in output.

For example, the following patch will add the field "patch:'A1' to the second rule of "alert:'A'". 

```jsonnet
local patchedRules = [
  {
    name: 'alertmanager.rules',
    rules: [
      {
        alert: 'A',
        index: 1,
        patch: 'A1',
        labels: {
          severity: 'warning',
        },
      },
      {
        alert: 'B',
        patch: 'B0',
        labels: {
          severity: 'warning',
        },
      },
      {
        alert: 'C',
        patch: 'C0',
        labels: {
          severity: 'warning',
        },
      },
      {
        alert: 'D',
        patch: 'D0',
        labels: {
          severity: 'warning',
        },
      },
    ],
  },
];

local rulesToPatch = [
  {
    name: 'alertmanager.rules',
    rules: [
      {
        alert: 'A',
        field: 'A0 rule',
        labels: {
          severity: 'warning',
        },
      },
      {
        alert: 'A',
        field: 'A1 rule',
        labels: {
          severity: 'warning',
        },
      },
      {
        alert: 'B',
        field: 'B0 rule',
        labels: {
          severity: 'warning',
        },
      },
    ],
  },
];
```


raptorsun:
[Another PR](https://github.com/prometheus-operator/kube-prometheus/pull/1351) with this rule patching function has been created in [prometheus-operator/kube-prometheus ](https://github.com/prometheus-operator) repository, too.

dgrisonnet:
Since we don't need this now in CMO, it would be better to add this functionality in kube-prometheus and bring it to CMO by reusing the upstream jsonnet script, wdyt?

raptorsun:
> Since we don't need this now in CMO, it would be better to add this functionality in kube-prometheus and bring it to CMO by reusing the upstream jsonnet script, wdyt?

Totally agree, I will close this PR, and use [this PR in kube-prometheus](https://github.com/prometheus-operator/kube-prometheus/pull/1351) to add this functionality instead.

PR #1334, opened 2021-08-22
add support for RemoteRead

eranra:
Align with Prometheus operator:: https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#remotereadspec

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -95,6 +95,7 @@ type PrometheusK8sConfig struct {
 	ExternalLabels      map[string]string                    `json:"externalLabels"`
 	VolumeClaimTemplate *monv1.EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate"`
 	RemoteWrite         []monv1.RemoteWriteSpec              `json:"remoteWrite"`
+	RemoteRead          []monv1.RemoteReadSpec               `json:"remoteRead"`
 	TelemetryMatches    []string                             `json:"-"`
 }
 


PhilipGough:
Hi @eranra - what is the goal of this PR? Adding the remote read to the config effectively won't achieve anything unless it is handled by the logic in the operator.

Aside from that, is there something driving this change? JIRA, specific use case etc?

PhilipGough (commented):
Hi @eranra - what is the goal of this PR? Adding the remote read to the config effectively won't achieve anything unless it is handled by the logic in the operator.

Aside from that, is there something driving this change? JIRA, specific use case etc?

eranra:
Hi, @PhilipGough thanks for the quick review ...  if you can look on https://github.com/openshift/cluster-monitoring-operator/pull/1334#issuecomment-904433015 this new "Recording Rules" feature from Loki allows using Prometheus Remote-Write and send log-based metrics into Prometheus. This is supported with the generic Prometheus Operator but not supported with CMO.

Can you advise on the areas in the code that need to be handled to add the support ... I will do my best to learn those in CMO and add the needed code. Make sense ?

FY: @periklis @Red-GV

alanconway:
@eranra I think it makes sense for what we want to do but it sounds like we need to make it fit in the bigger picture. Carry on!

adrianchifor:
Any updates on this? We use remote write to Cortex for long term metrics and would like to keep querying thanos-querier/prometheus: serve last 30d from disk, the rest remote read from cortex.

PhilipGough:
@adrianchifor as I mentioned in the thread this would need to come through official channels. If there is a use case for this, it is probably best to create an RFE via RH support ticket. Thanks

PR #1336, opened 2021-08-23
Bug 1991504: changes for moving from sha1 to bcrypt

prashbnair:


simonpasquier:
Checking the [oauth-proxy](https://github.com/openshift/oauth-proxy/blob/c4b9b826513af32e4c0854e5f8b088bb3fee1c20/htpasswd.go#L61) implementation, we should add a [version prefix](https://en.wikipedia.org/wiki/Bcrypt#Versioning_history) to the bcrypt'ed value for oauth-proxy to accept the password.

/hold

feel free to remove the hold once you've tested on a local cluster that Grafana works as expected.

prashbnair:
> Checking the oauth-proxy implementation, we should add a version prefix to the bcrypt'ed value for oauth-proxy to accept the password.

I have tested the changes with Grafana and it works.  The
bcrypt.GenerateFromPassword([]byte(password), bcrypt.MinCost) generates the required prefix.

simonpasquier:
/lgtm
/retest
/bugzilla refresh

Thanks Prashant for the update!

PR #1337, opened 2021-08-23
Bug 1996718: Fix ksm metric label allowlist flag

jan--f:


jan--f:
This should fix the failure seen in https://github.com/openshift/kube-state-metrics/pull/56

raptorsun (approved):

PR #1338, opened 2021-08-24
MON-975: Descope possible RBAC rules to monitoring namespaces

arajkumar:
This PR descopes CMO RBAC rules to openshift-monitoring and openshift-user-workload-monitoring namespaces instead of cluster-wide access.

Prior to the change, the following auth test would give `yes` for all namespaces.

```
➜  oc auth can-i get deployment --as=system:serviceaccount:openshift-monitoring:cluster-monitoring-operator -n openshift-insights
no
➜  oc auth can-i get deployment --as=system:serviceaccount:openshift-monitoring:cluster-monitoring-operator -n kube-system
no
```


simonpasquier:
/lgtm

Thanks! I think it' a move in the right direction. We could also tidy up the permissions for the prometheus-operator and prometheus service accounts but it's another effort (and one that should be conducted upstream first).
@jan--f @dgrisonnet @fpetkovski for another look.

arajkumar:
Thanks @simonpasquier 

/hold to get opinion from @jan--f @dgrisonnet @fpetkovski.

fpetkovski:
Does it make sense to add some tests in the CI for RBAC? The `kubectl auth can-i` command[1] can be useful for this. We could add positive but also and negative cases to make sure CMO cannot, e.g., create secrets in other namespaces etc.

[1] https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-can-i-em-

arajkumar:
@fpetkovski Good idea, anyways I had used `kubectl auth can-i` to test the things manually. The same can be plugged with prow CI with a proper script.

I would prefer to do it as a separate PR, WDYT?

Another idea is to use [audit2rbac](https://github.com/liggitt/audit2rbac) tool upon completion of e2e to test against a baseline RBAC rules. This would also help us to catch unintended RBAC elevations from CMO dependencies.

fpetkovski:
> I would prefer to do it as a separate PR, WDYT?

I think we have to do it as a separate PR because it has to end up in openshift/release. If you are confident with the changes you've made so far, we can merge this PR first and add the tests in a follow up PR. 

> Another idea is to use audit2rbac tool upon completion of e2e to test against a baseline RBAC rules. This would also help us to catch unintended RBAC elevations from CMO dependencies.

I am not familiar with audit2rbac so I cannot say which approach is better, feel free to choose the one you think is better suited for this :)

.gitignore
```diff
@@ -10,3 +10,6 @@ tmp/
 
 # This file gets generated as part of `make run-local`
 operator.log
+
+# GoLand project files
+.idea
```
jan--f, line 15: Let's keep IDE/editor specific settings out of this file. Better add this to your personal config, either [globally](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) or [per repo](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file).

jan--f (changes requested):
One nit but otherwise this looks good :+1:

.gitignore
arajkumar: Fixed it. Thank you.

arajkumar:
/label docs-approved
/label qe-approved
/label px-approved

This change is not affecting any of the user facing features, instead it just descopes RBAC rules.

PR #1339, opened 2021-08-24
Bug 1996491: adding label check for node when creating daemon set

prashbnair:


prashbnair:
Adding check that the node is a linux node for the count of nodes on which the node exporter can be installed.

pkg/client/client.go
```diff
@@ -992,7 +992,12 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
 		for _, node := range nodeList.Items {
 			for _, condition := range node.Status.Conditions {
 				if condition.Type == v1.NodeReady && condition.Status == v1.ConditionTrue {
-					nodeReadyCount++
+					labelMap := node.Labels
+					for name, value := range labelMap {
```
arajkumar, line 996: You can make use of LabelSelector in `metav1.ListOptions`. It is as equvalent as `kubectl get nodes -l "kubernetes.io/os=linux"`

arajkumar (changes requested):

pkg/client/client.go

juzhao: I am kindly think `kubernetes.io/os=linux` maybe not proper, especially  for windows nodes, we should be able to schedule node-exporter to each node, including windows node which don't have `kubernetes.io/os=linux`  label

pkg/client/client.go
```diff
@@ -988,7 +988,7 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
 		}
 
 		var nodeReadyCount int32
-		nodeList, err := c.kclient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+		nodeList, err := c.kclient.CoreV1().Nodes().List(ctx, metav1.ListOptions{LabelSelector: "kubernetes.io/os=linux"})
```
simonpasquier, line 991: it would be better to use `ds.Spec.Template.Spec.NodeSelector`

pkg/client/client.go
arajkumar: @juzhao node_exporter is meant only for Linux.

pkg/client/client.go
prashbnair: done

arajkumar (approved):
/lgtm
/hold please wait for @simonpasquier's approval.

simonpasquier (commented):
/lgtm
/hold cancel

PR #1326, opened 2021-08-18
WIP:Sync kube prom

PhilipGough:
* [ ] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #1340, opened 2021-08-24
pin dependencies for upcoming relase

jan--f:


.ci-operator.yaml
```diff
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: golang-1.14
+  tag: golang-1.16
```
simonpasquier, line 4: :+1:

jsonnet/jsonnetfile.json
```diff
@@ -35,7 +35,7 @@
           "subdir": "jsonnet"
         }
       },
-      "version": "master",
+      "version": "release-4,9",
```
simonpasquier, line 38:
```suggestion
      "version": "release-4.9",
```

jan--f: oh boy, good catch.

simonpasquier:
/hold
while checking the changes I've noticed that the `etcd_object_counts` metric from the API server would be dropped while the metric is being used by 2 Telemetry rules. I've submitted https://github.com/openshift/cluster-monitoring-operator/pull/1345 to fix the rules and use the new metric (`apiserver_storage_objects`) instead.

jan--f:
we'll also wait for #1326

PR #1341, opened 2021-08-24
Bug 1933847: enable hard affinity + PodDisruptionBudget for Prometheus and Thanos Ruler pods

simonpasquier:
~~This requires #1330 (on 4.9 and 4.8.z) as well as https://github.com/openshift/release/pull/21564 to be merged first.~~

We need https://github.com/openshift/cluster-monitoring-operator/pull/1472 to be merged (backport of #1330 to release-4.9) to ensure that before the pods spread on different nodes before the upgrade to 4.10 starts (otherwise the new hard anti-affinity rules might block the process).


jan--f:
This looks good, I have one question though: in #1330 we actively balance prometheus and alertmanager. Here we set hard affinity settings for prometheus and thanos ruler. Should alertmanager also get hard affinity settings?

simonpasquier:
@jan--f yes we have https://bugzilla.redhat.com/show_bug.cgi?id=1955489 to track this but it implies reducing the number of replicas to 2 and using the `minReadySeconds` field in the Alertmanager CRD which is why I plan to tackle it in another PR.

simonpasquier:
While #1472 has merged in 4.9 a couple of hours, we'll keep on holding this PR a bit to make sure that we haven't introduced any regression.

PR #1354, opened 2021-08-31
[MON-1851] upgrade unused rule checking script

raptorsun:

This PR improves unused rule checking by including openshift console code and rule expression references in CMO component.

[JIRA ticket MON-1851](https://issues.redhat.com/browse/MON-1851)

### DIFF
--- a/hack/check-rec-rule-usage.sh
+++ b/hack/check-rec-rule-usage.sh
@@ -4,19 +4,45 @@ set -e
 TMP=$(mktemp -d)
 echo "Created temporary directory at $TMP"
 
+# This directory should be a local copy of github.com/openshift/console
+CONSOLE_DIR="${1:-../console}"
+
+if [ ! -d "$CONSOLE_DIR" ]; then
+    echo "Couldn't find a local copy of github.com/openshift/console: directory $CONSOLE_DIR does not exist"
+    exit 1
+fi
+
 findrules="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/find-rules.sh"
 
 "$findrules" >"$TMP/rule-groups"
-# extract recording rules
+# Extract recording rules
 tmp/bin/gojsontoyaml -yamltojson <"$TMP/rule-groups" | jq -s '.. | objects | select(.record) | .record' | tr -d '"' | sort -u >"$TMP/rec-rules"
-# and alerting rules
-tmp/bin/gojsontoyaml -yamltojson <"$TMP/rule-groups" | jq -s '.. | objects | select(.alert) | .expr' | grep "\w\+{" -o | tr -d "{" | sort -u >"$TMP/used-rules"
-# and telemetry client's metrics
-tmp/bin/gojsontoyaml -yamltojson <manifests/*-config.yaml | jq '.data."metrics.yaml"' | sed "s;\\\\n;\n;g" | grep "^-\s'{__name__=\\\\\"" | sed -e "s/^- '{__name__=\\\\\"//" | sed -e "s/\\\\\"[^}]*}'//" | sort | uniq >>"$TMP/used-rules"
 
-# the first grep outputs all recording rules that are used neither in alerts nor telemetry metrics
+# Extract alerting rules
+tmp/bin/gojsontoyaml -yamltojson <"$TMP/rule-groups" | jq -s '.. | objects | select(.alert) | .expr' | grep "\w\+{" -o | tr -d "{" | sort -u >"$TMP/used-rules"
+
+# Extract recording rules which are used in other recording or alerting rules.
+go run -mod=vendor hack/promql_rule/main.go "$TMP/rule-groups" >"$TMP/rule-expr-names"
+grep -Fxf "$TMP/rec-rules" "$TMP/rule-expr-names" >>"$TMP/used-rules"
+
+# Extract Telemetry Client's metrics
+# Put plain text matching rule names into used-rules
+tmp/bin/gojsontoyaml -yamltojson <manifests/*-config.yaml | jq '.data."metrics.yaml"' | sed "s;\\\\n;\n;g" | grep "^-\s'{__name__=\\\\\"" | sed -e "s/^- '{__name__=\\\\\"//" | sed -e "s/\\\\\"[^}]*}'//" | sort | uniq >>"$TMP/used-rules"
+# Put regex pattern matching rule names into used-rules-regex
+tmp/bin/gojsontoyaml -yamltojson <manifests/*-config.yaml | jq '.data."metrics.yaml"' | sed "s;\\\\n;\n;g" | grep "^-\s'{__name__=~\\\\\"" | sed -e "s/^- '{__name__=~\\\\\"//" | sed -e "s/\\\\\"[^}]*}'//" | sort | uniq >"$TMP/used-rules-regex"
+# The first grep outputs all recording rules that are used neither in alerts nor telemetry metrics
 # and then greps for the remaining rules in the dashboard defs and outputs the ones that are not found
-grep -Fxvf "$TMP/used-rules" "$TMP/rec-rules" | while read -r r; do grep "$r" assets/grafana/dashboard-definitions.yaml -q || echo "$r"; done >"$TMP/unused-rules"
+grep -Fxvf "$TMP/used-rules" "$TMP/rec-rules" | while read -r r; do grep "$r" assets/grafana/dashboard-definitions.yaml -q || echo "$r"; done >"$TMP/unused-rules-fixstr"
+grep -Exvf "$TMP/used-rules-regex" "$TMP/unused-rules-fixstr" >"$TMP/unused-rules-cmo"
+
+# Find out the rules used in console. The console TypeScript codes contain reference to rule names by exact text matching.
+find "${CONSOLE_DIR}" -type f -regex ".*\.ts[x]*" | while read -r filename; do
+    grep -Ff "$TMP/unused-rules-cmo" "$filename" | { grep -Eo "[a-zA-Z_][a-zA-Z0-9_:]*" || true; } >>"$TMP/used-rules-console-tmp"
+done
+sort "$TMP/used-rules-console-tmp" | uniq | { grep -Fxf "$TMP/unused-rules-cmo" || true; } >"$TMP/used-rules-console"
+# Eliminate rules used by console from unused rules in CMO.
+# Now we get rules neither used in CMO nor in console.
+grep -Fxvf "$TMP/used-rules-console" "$TMP/unused-rules-cmo" >"$TMP/unused-rules"
 
 echo "Found $(wc -l <"$TMP/unused-rules") unused rules"
 echo "Find the unused rules in $TMP/unused-rules"
--- a/hack/find-rules.sh
+++ b/hack/find-rules.sh
@@ -3,8 +3,11 @@ set -e
 
 # find all rule groups and aggregate them
 # returns a json list of groups
-
-find assets/ -type f -name "*prometheus-rule.yaml" | while read -r f
-do
-    tmp/bin/gojsontoyaml -yamltojson < "$f" | jq .spec
+find assets/ -type f -name "*.yaml" | while read -r f; do
+    JSONIFIED=$(tmp/bin/gojsontoyaml -yamltojson <"$f")
+    KIND=$(jq .kind <<<"$JSONIFIED")
+    if [ "$KIND" != "\"PrometheusRule\"" ]; then
+        continue
+    fi
+    jq .spec <<<"$JSONIFIED"
 done | jq -s '[.[]] | { groups: map(.groups[]) }'
--- a/hack/promql_rule/main.go
+++ b/hack/promql_rule/main.go
@@ -1,0 +1,94 @@
+// Copyright 2021 The Cluster Monitoring Operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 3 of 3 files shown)


hack/check-rec-rule-usage.sh
```diff
@@ -1,22 +1,41 @@
...
+
+# and referenced rules in other rules' expressions
+tmp/bin/gojsontoyaml -yamltojson <"$TMP/rule-groups" | jq -s '.. | objects | select(.expr) | .expr' | sed 's/\\n/\n/g' | sed 's/[^a-zA-Z0-9_]/\n/g' | sort -u | uniq >"$TMP/rule-expr-tokens"
+
+grep -Fxf "$TMP/rec-rules" "$TMP/rule-expr-tokens" || true >>"$TMP/used-rules"
+
 # and telemetry client's metrics
 tmp/bin/gojsontoyaml -yamltojson <manifests/*-config.yaml | jq '.data."metrics.yaml"' | sed "s;\\\\n;\n;g" | grep "^-\s'{__name__=\\\\\"" | sed -e "s/^- '{__name__=\\\\\"//" | sed -e "s/\\\\\"[^}]*}'//" | sort | uniq >>"$TMP/used-rules"
 
+tmp/bin/gojsontoyaml -yamltojson <manifests/*-config.yaml | jq '.data."metrics.yaml"' | sed "s;\\\\n;\n;g" | grep "^-\s'{__name__=~\\\\\"" | sed -e "s/^- '{__name__=~\\\\\"//" | sed -e "s/\\\\\"[^}]*}'//" | sort | uniq >"$TMP/used-rules-regex"
```
dgrisonnet, line 30: can you add a comment to this line so that it is easier to understand what it does?

hack/check-rec-rule-usage.sh
```diff
@@ -1,22 +1,41 @@
...
 # the first grep outputs all recording rules that are used neither in alerts nor telemetry metrics
 # and then greps for the remaining rules in the dashboard defs and outputs the ones that are not found
-grep -Fxvf "$TMP/used-rules" "$TMP/rec-rules" | while read -r r; do grep "$r" assets/grafana/dashboard-definitions.yaml -q || echo "$r"; done >"$TMP/unused-rules"
+grep -Fxvf "$TMP/used-rules" "$TMP/rec-rules" | while read -r r; do grep "$r" assets/grafana/dashboard-definitions.yaml -q || echo "$r"; done >"$TMP/unused-rules-fixstr"
+grep -Exvf "$TMP/used-rules-regex" "$TMP/unused-rules-fixstr" >"$TMP/unused-rules-cmo"
+
+find ${CONOSLE_DIR} -type f -regex ".*\.ts[x]*" | while read -r filename; do grep -Ff "$TMP/unused-rules-cmo" $filename || true; done | sed 's/[^a-zA-Z0-9_]/\n/g' | sort | uniq | grep -Fxf "$TMP/unused-rules-cmo" >"$TMP/used-rules-console"
+
+grep -Fxvf $TMP/used-rules-console $TMP/unused-rules-cmo >$TMP/unused-rules
```
dgrisonnet, line 38: Same here, would you mind adding some comments?

dgrisonnet (commented):
A few nits but it looks good to me overall

hack/check-rec-rule-usage.sh
```diff
@@ -1,22 +1,41 @@
 #!/usr/bin/env bash
 set -e
-
 TMP=$(mktemp -d)
 echo "Created temporary directory at $TMP"
 
+CONOSLE_DIR=$1
```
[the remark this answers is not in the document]
simonpasquier: maybe have a default value? 

```suggestion
CONSOLE_DIR="{$1:../console}"
```

the other option would be to check out the console codebase in a temporary directory if no argument is given.

hack/check-rec-rule-usage.sh
```diff
@@ -1,22 +1,41 @@
...
+
+if [ ! -d $CONOSLE_DIR ]; then
+    echo "Directory $CONOSLE_DIR does not exist"
```
simonpasquier, line 9:
```suggestion
    echo "Couldn't find a local copy of github.com/openshift/console: directory $CONSOLE_DIR does not exist"
```

hack/check-rec-rule-usage.sh
```diff
@@ -1,22 +1,41 @@
...
+    exit 1
+fi
+
 findrules="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/find-rules.sh"
```
simonpasquier, line 13: I notice that `find-rules.sh` processes files that match `*prometheus-rule.yaml` but it misses at least `assets/prometheus-k8s/prometheus-rule-thanos-sidecar.yaml`. It would be better to parse all assets and match on `kind: PrometheusRule`.

raptorsun: `find-rules.sh` modified, now using `kind: PrometheusRule` to find out rules.

hack/check-rec-rule-usage.sh
```diff
@@ -1,22 +1,46 @@
...
-# and alerting rules
+
+# Extract alerting rules
 tmp/bin/gojsontoyaml -yamltojson <"$TMP/rule-groups" | jq -s '.. | objects | select(.alert) | .expr' | grep "\w\+{" -o | tr -d "{" | sort -u >"$TMP/used-rules"
-# and telemetry client's metrics
-tmp/bin/gojsontoyaml -yamltojson <manifests/*-config.yaml | jq '.data."metrics.yaml"' | sed "s;\\\\n;\n;g" | grep "^-\s'{__name__=\\\\\"" | sed -e "s/^- '{__name__=\\\\\"//" | sed -e "s/\\\\\"[^}]*}'//" | sort | uniq >>"$TMP/used-rules"
 
-# the first grep outputs all recording rules that are used neither in alerts nor telemetry metrics
+# Extract referenced rules in other rules' expressions
+tmp/bin/gojsontoyaml -yamltojson <"$TMP/rule-groups" | jq -s '.. | objects | select(.expr) | .expr' | sed 's/\\n/\n/g' | sed 's/[^a-zA-Z0-9_]/\n/g' | sort -u | uniq >"$TMP/rule-expr-tokens"
```
simonpasquier, line 24: this part seems brittle to me: looking at the `rule-expr-tokens` file output, it contains way more items than just rules. IMHO the only way we can safely identify whether a recording rule is used by another rule is to parse the PromQL expressions for all rules. This would have to be done in Go using the promql parser (check `pkg/promqlgen/promqlgen.go` for an example).

raptorsun: agree, promql parser shall be used to extract referenced rules. I will hold the PR until this modification is done. 
/hold

raptorsun: Promql parser is used now to extract referenced rule names.

hack/check-rec-rule-usage.sh
```diff
@@ -1,22 +1,47 @@
 #!/usr/bin/env bash
-set -e
```
dgrisonnet, line 2: any reasons for unsetting this option? Usually when a command has an error, we always want the script to exit

hack/find-rules.sh
```diff
@@ -3,8 +3,10 @@ set -e
 
 # find all rule groups and aggregate them
 # returns a json list of groups
-
-find assets/ -type f -name "*prometheus-rule.yaml" | while read -r f
-do
-    tmp/bin/gojsontoyaml -yamltojson < "$f" | jq .spec
+find assets/ -type f -name "*.yaml" | while read -r f; do
+    KIND=$(tmp/bin/gojsontoyaml -yamltojson <"$f" | jq .kind)
```
dgrisonnet, line 7: instead of converting the file twice from yaml to json, let's store the output in a variable and then extract the kind / spec

hack/promqlrule.go
```diff
@@ -0,0 +1,75 @@
+package main
```
dgrisonnet, line 1: we should add a license header to this file

hack/promqlrule.go
```diff
@@ -0,0 +1,75 @@
...
+		selectors := promql.ExtractSelectors(expr)
+		for _, selector := range selectors {
+			for _, label := range selector {
+				if label.Name == "__name__" {
+					metricNamesMap[label.Value] = true
+				} else {
+					metricNamesMap[label.Name] = true
+				}
+			}
+		}
```
dgrisonnet, line 44: I don't that this will yield exactly what you are expecting. Instead of just recording metrics, you'll also get some extract labels.
It would be more appropriate to have the following piece if you only want to record metric names:
```go
promql.Inspect(expr, func(node Node, _ []Node) error {
  vs, ok := node.(*promql.VectorSelector)
  if ok {
    metricNameMap[vs.Name] = struct{}{}
  }
  return nil
})
```

hack/promqlrule.go
```diff
@@ -0,0 +1,75 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+
+	promql "github.com/prometheus/prometheus/promql/parser"
+)
+
+func extractMetricNames(file *os.File) (map[string]bool, error) {
+	metricNamesMap := map[string]bool{}
```
dgrisonnet, line 14: A set is usually defined as: `map[string]struct{}`

hack/promqlrule.go
raptorsun: Updated now, ready to review again :)

hack/check-rec-rule-usage.sh
```diff
@@ -4,19 +4,46 @@ set -e
...
+# The first grep outputs all recording rules that are used neither in alerts nor telemetry metrics
 # and then greps for the remaining rules in the dashboard defs and outputs the ones that are not found
-grep -Fxvf "$TMP/used-rules" "$TMP/rec-rules" | while read -r r; do grep "$r" assets/grafana/dashboard-definitions.yaml -q || echo "$r"; done >"$TMP/unused-rules"
+grep -Fxvf "$TMP/used-rules" "$TMP/rec-rules" | while read -r r; do grep "$r" assets/grafana/dashboard-definitions.yaml -q || echo "$r"; done >"$TMP/unused-rules-fixstr"
+grep -Exvf "$TMP/used-rules-regex" "$TMP/unused-rules-fixstr" >"$TMP/unused-rules-cmo"
+
+# Find out the rules used in console. The console TypeScript codes contain reference to rule names by exact text matching.
+find "${CONSOLE_DIR}" -type f -regex ".*\.ts[x]*" | while read -r -d '' filename; do
+    grep -Ff "$TMP/unused-rules-cmo" "$filename" || true
+done | sed 's/[^a-zA-Z0-9_]/\n/g' | sort | uniq | grep -Fxf "$TMP/unused-rules-cmo" || true >"$TMP/used-rules-console"
```
simonpasquier, line 43: that part doesn't seem to work because pod:container_cpu_usage:sum is listed in the unused rules while it's used by the console code.

hack/check-rec-rule-usage.sh
```diff
@@ -4,19 +4,46 @@ set -e
...
 tmp/bin/gojsontoyaml -yamltojson <"$TMP/rule-groups" | jq -s '.. | objects | select(.record) | .record' | tr -d '"' | sort -u >"$TMP/rec-rules"
-# and alerting rules
+
+# Extract alerting rules
 tmp/bin/gojsontoyaml -yamltojson <"$TMP/rule-groups" | jq -s '.. | objects | select(.alert) | .expr' | grep "\w\+{" -o | tr -d "{" | sort -u >"$TMP/used-rules"
-# and telemetry client's metrics
-tmp/bin/gojsontoyaml -yamltojson <manifests/*-config.yaml | jq '.data."metrics.yaml"' | sed "s;\\\\n;\n;g" | grep "^-\s'{__name__=\\\\\"" | sed -e "s/^- '{__name__=\\\\\"//" | sed -e "s/\\\\\"[^}]*}'//" | sort | uniq >>"$TMP/used-rules"
 
-# the first grep outputs all recording rules that are used neither in alerts nor telemetry metrics
+# Extract referenced rules in other rules' expressions
```
simonpasquier, line 24:
```suggestion
# Extract recording rules which are used in other recording or alerting rules.
```

raptorsun:
Script has been updated to fix the bug of not finding rules from console.
Now the script considers the following rules as unused:

- cluster:capacity_cpu_cores_hyperthread_enabled:sum
- cluster:capacity_cpu_sockets_hyperthread_enabled:sum
- cluster:container_cpu_usage:ratio
- cluster:container_spec_cpu_shares:ratio
- cluster:hyperthread_enabled_nodes
- cluster:memory_usage:ratio
- cluster:node_cpu:ratio
- cluster:node_cpu:sum_rate5m
- cluster_quantile:scheduler_binding_duration_seconds:histogram_quantile
- cluster_quantile:scheduler_e2e_scheduling_duration_seconds:histogram_quantile
- cluster_quantile:scheduler_scheduling_algorithm_duration_seconds:histogram_quantile
- code:cluster:ingress_http_request_count:rate5m:sum
- namespace:container_memory_usage_bytes:sum

simonpasquier (commented):
/lgtm

simonpasquier:
/hold cancel
/label qe-approved
/label px-approved
/label docs-approved

this script is an internal tool for the montioring developers that want to check if we ship rules that aren't used so no user impact.

raptorsun:
The test on go-fmt is still failing on an unmodified file `hack/telemeter_query.go`. 🤔

raptorsun:
A PR is created to fix the falling test `ci/prow/go-fmt`
https://github.com/openshift/cluster-monitoring-operator/pull/1497

simonpasquier (commented):
could you move `hack/promqlrule.go` to its own sub-directory, (e.g. `hack/rulestometrics_parser/main.go` but exact naming to be defined)?

raptorsun:
The tool program now parses directly rule-groups and gives back metric names in alphabetic order.

PR #1365, opened 2021-09-07
Bug 2001566: Enabling alerts for prometheus operator in UWM

prashbnair:

### DIFF
--- a/jsonnet/components/prometheus-operator-user-workload.libsonnet
+++ b/jsonnet/components/prometheus-operator-user-workload.libsonnet
@@ -6,8 +6,6 @@ function(params)
   local cfg = params;
   operator(cfg) + {
 
-    mixin:: null,
-    prometheusRule:: null,
     '0alertmanagerCustomResourceDefinition':: {},
     '0alertmanagerConfigCustomResourceDefinition':: {},
     '0prometheusCustomResourceDefinition':: {},
@@ -111,6 +109,22 @@ function(params)
         annotations+: {
           'service.beta.openshift.io/serving-cert-secret-name': 'prometheus-operator-user-workload-tls',
         },
+      },
+    },
+    prometheusRule+: {
+      metadata+: {
+        name: 'prometheus-operator-user-workload-rules',
+      },
+      spec+: {
+        groups:
+          std.mapWithIndex(
+            function(i, v) if i == 0 then
+              v {
+                name: 'prometheus-operator-user-workload',
+                rules: v.rules,
+              },
+            super.groups
+          ),
       },
     },
     serviceMonitor+: {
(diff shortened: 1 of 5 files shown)


pkg/client/client.go
```diff
@@ -794,6 +794,15 @@ func (c *Client) WaitForDeploymentRollout(ctx context.Context, dep *appsv1.Deplo
 				d.Generation, d.Status.ObservedGeneration)
 			return false, nil
 		}
+
+		deploymentConditions := d.Status.Conditions
+
```
fpetkovski, line 799: tiny nit
```suggestion
```

pkg/tasks/prometheusoperator_user_workload.go
```diff
@@ -117,6 +117,15 @@ func (t *PrometheusOperatorUserWorkloadTask) create(ctx context.Context) error {
 		return errors.Wrap(err, "waiting for Prometheus Operator CRs to become available failed")
 	}
 
+	pr, err := t.factory.PrometheusOperatorUserWorkloadPrometheusRule()
```
fpetkovski, line 120: Should we change the destroy function to delete the rules when UWM is disabled?

prashbnair: Done

jsonnet/components/prometheus-operator-user-workload.libsonnet
```diff
@@ -113,6 +111,22 @@ function(params)
         },
       },
     },
+    prometheusRule+: {
```
arajkumar, line 114: Should we consider reusing the same alerts defined in PM namespace for prometheus(by modifying [prometheusOperatorSelector](https://github.com/prometheus-operator/kube-prometheus/blob/main/jsonnet/kube-prometheus/components/prometheus-operator.libsonnet#L33) config) instead of redefining the alerts again in UWM namespace?

i.e.
```
 prometheusOperatorSelector: 'job="prometheus-operator",namespace~="openshift-monitoring|openshift-user-workload-monitoring"'
```

arajkumar: It shall be done just by adding `prometheusOperatorSelector` config value in https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/main.jsonnet#L273.

```
prometheusOperatorSelector: 'job="prometheus-operator", namespace=~"%(namespace)s|%(namespaceUserWorkload)s"' % $.values.common,
```

simonpasquier: yes we should rather tweak `prometheusOperatorSelector` 
/hold

prashbnair: Closing out this PR and will raise a fresh one with the new approach.

PR #1366, opened 2021-09-07
Bug 2000490: jsonnet: Drop unnecessary kube-state-metrics alerts

bison:
We don't configure sharding for kube-state-metrics, so there's no
reason to ship the critical alerts for them, or maintain runbooks for
them, etc. This drops the two critical alerts related to sharding.

### DIFF
--- a/jsonnet/utils/sanitize-rules.libsonnet
+++ b/jsonnet/utils/sanitize-rules.libsonnet
@@ -27,6 +27,14 @@ local excludedRules = [
     ],
   },
   {
+    name: 'kube-state-metrics',
+    rules: [
+      // We do not configure sharding for kube-state-metrics.
+      { alert: 'KubeStateMetricsShardingMismatch' },
+      { alert: 'KubeStateMetricsShardsMissing' },
+    ],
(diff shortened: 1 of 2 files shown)


jsonnet/utils/sanitize-rules.libsonnet
```diff
@@ -26,6 +26,14 @@ local excludedRules = [
       { alert: 'TargetDown' },
     ],
   },
+  {
+    name: 'kube-state-metrics',
+    rules: [
+      // We do not conifugre sharding for kube-state-metrics.
```
arajkumar, line 32: nit:(typo)

```suggestion
      // We do not configure sharding for kube-state-metrics.
```

arajkumar (approved):
lgtm with a nit 👍🏼

jsonnet/utils/sanitize-rules.libsonnet
bison: Oops. Fixed. Thanks!

simonpasquier (commented):
/lgtm

PR #1367, opened 2021-09-07
Bug 1997028: drop [$(POD_IP)] from --grpc-address argument for thanos sidecar

jan--f:


slashpai (approved):

simonpasquier (commented):
/lgtm

PR #1363, opened 2021-09-06
Bug 1999521: Use secureJsonData.basicAuthPassword field for Grafana authentication

prashbnair:
* [ ] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #1368, opened 2021-09-07
Format the codebase and fix shellcheck

fpetkovski:
This PR formats the codebase and fixes shellcheck in the CI.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -2166,21 +2166,21 @@ type GrafanaDatasources struct {
 }
 
 type SecureJSONData struct {
-	BasicAuthPassword string           `json:"basicAuthPassword"`
+	BasicAuthPassword string `json:"basicAuthPassword"`
 }
 
 type GrafanaDatasource struct {
-	Access            string           `json:"access"`
-	BasicAuth         bool             `json:"basicAuth"`
-	SecureJSONData    SecureJSONData   `json:"secureJsonData"`
-	BasicAuthUser     string           `json:"basicAuthUser"`
-	Editable          bool             `json:"editable"`
-	JsonData          *GrafanaJsonData `json:"jsonData"`
-	Name              string           `json:"name"`
-	OrgId             int              `json:"orgId"`
-	Type              string           `json:"type"`
-	Url               string           `json:"url"`
-	Version           int              `json:"version"`
+	Access         string           `json:"access"`
+	BasicAuth      bool             `json:"basicAuth"`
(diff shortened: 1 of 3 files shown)


jan--f (changes requested):
This seems to break `hack/generate-versions.sh`

Also if we want the scripts in hack shellcheck'ed, should we add that to `hack/shellcheck.sh`?

fpetkovski:
Good catch. `generate-versions` does not pass shellcheck and I tried to fix it, but it turns out I committed the wrong "fix". 

The scripts in `hack` already seem to be shellcheck-ed, this is how I found the issue in `generate-versions`. Did you have something else in mind?

pkg/manifests/manifests.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/openshift/cluster-monitoring-operator/pkg/promqlgen"
 	"github.com/pkg/errors"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"golang.org/x/crypto/bcrypt"
```
slashpai, line 34: This change is covered in https://github.com/openshift/cluster-monitoring-operator/pull/1363/files#diff-1f9d2d18d4cabd8262e585b932330183f7cb9b544ac9f27c8bcb3ef9c8d14376R34   :)

jan--f:
> The scripts in hack already seem to be shellcheck-ed, this is how I found the issue in generate-versions. Did you have something else in mind?

hmm no, just wondering how that got through our CI then, as that runs a required shellcheck job :thinking:

fpetkovski:
We never verified it in the CI :) But we should now: https://github.com/openshift/release/pull/21638

pkg/manifests/manifests.go
fpetkovski: Good catch. 
I don't think there is a problem with having this change in two PRs, it should not cause any conflicts. 
Also, I would like to enable go linting checks in the CI and for that PR to be merged, the codebase has to be properly formatted.

jan--f (changes requested):
Imho it would be better to run `install` in a sub shell, instead of explicitly cd'ing back. `cd -` is quite dependent on context whereas  `(install)` will leave the context of the main script untouched.

simonpasquier (commented):
/lgtm

fpetkovski:
@jan--f good idea, done

fpetkovski:
/label px-approved
no px needed

/label docs-approved
no user-facing docs needed

/label qe-approved
no qe needed

PR #1369, opened 2021-09-08
docs: new development doc for new comers

sthaha:
Explains:
*  how to run CMO locally
*  Update jsonnet bundles

### DIFF
--- a/Documentation/development.md
+++ b/Documentation/development.md
@@ -1,0 +1,111 @@
+# CMO Development
+
+
+## Run CMO Locally
+### Assumptions
+* Should have a cluster running
+* Has setup Go development env
+
+### Build and Run CMO
+
+The `make` target `run-local` builds the `operator` and runs CMO locally
+
+```shell
+export KUBECONFIG=/path/to/kubeconfig/file
+make run-local
+```
+
+### Modify jsonnet
+
+**TL;DR:**
+1. build operator
+1. use `hack/local-cmo.sh` to run the operator
+1. modify assets/yamls
+1. backport the changes to jsonnet
+
+
+Sometimes we may not want to rebuild the operator. E.g. you may need
+to figure out how the stack behaves after applying a change to generated
+kubernetes manifest.  In other words, to modify the YAML in `assets/`
+directory directly without needing to change the `jsonnet/` or
+the `vendored` jsonnet files.
+
+In these cases it is easier to change those yaml files manually, run
+the operator (with those changed files) and later tweak `jsonnet`, than
+to figuring out which parts of jsonnet code you may need to change to
+fix an issue.
+
+It is easier to follow the following workflow
+#### Setup
+1. Start openshift cluster ( e.g. using `cluster-bot` )
+1. `export KUBECONFIG=/path/to/kubeconfig`
+1. Build operator - `make operator`
+
+#### Amend Assets
+1. Amend yaml files in `assets/` directory
+1. Run `hack/local-cmo.sh` (**does not** build the operator)
+1. ☝☝☝ _rinse and repeat_
+
+#### Finally:
+1. Port the changes back to `jsonnet` or vendored `jsonnet`
+
+
+
+## Updating individual vendored jsonnet code
+
+NOTE: `jb update <repo-url>/<jsonnet-subdir>` doesn't seem to work since it
+does not update the dependencies (at the time of writing)
+
+Steps to update a bundle e.g. `kube-prometheus`:
+
+```shell
+cd jsonnet
+
+```
+
+Edit ``jsonnetfile.json ``and remove everything except the bundle you want to update
+E.g. remove everything but *kube-prometheus* from the jsonnetfile.json file
+
+```
+jb update
+
+```
+
+`git add -p ` only the changes to the `version` and `sum`. E.g. the diff would
+show as follows, add only those ignore the bundles that got deleted
+
+```diff
+diff --git a/jsonnet/jsonnetfile.lock.json b/jsonnet/jsonnetfile.lock.json
+index 6d5adc26..eead7874 100644
+--- a/jsonnet/jsonnetfile.lock.json
++++ b/jsonnet/jsonnetfile.lock.json
+@@ -28,8 +28,8 @@
+           "subdir": "grafonnet"
+         }
+       },
+-      "version": "3082bfca110166cd69533fa3c0875fdb1b68c329",
+-      "sum": "4/sUV0Kk+o8I+wlYxL9R6EPhL/NiLfYHk+NXlU64RUk="
++      "version": "05fb200ee1a1816fc1b4c522071d5606d8dd71c1",
++      "sum": "mEoObbqbyVaXrHFEJSM2Nad31tOvadzIevWuyNHHBgI="
+     },
+```
+
+Revert back the changes to jsonnetfile.json and jsonnetfile.lock.json
+```
+git restore --  jsonnetfile.json jsonnnet.lock.json
+
+```
+
+Now you should have a `jsonnetfile.lock.json` that has only the updates to
+the bundle `kube-prometheus`. Remove the  `vendor` and reinstall all bundles.
+
+```
+# reinstall all jsonnet based on the new lockfile
+rm -rf vendor ; jb install
+
+# Now you should have only the changes that is part of latest kube-prometheus
+cd ..
+make generate
+git add everything and commit
+
+```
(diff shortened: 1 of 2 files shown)


Documentation/development.md
```diff
@@ -0,0 +1,83 @@
...
+Steps to update a bundle e.g. `kube-prometheus`:
+
+```shell
+cd jsonnset
+
+
+# edit jsonnset.json and remove all but the bundle you want to update
+# remove everything but *kube-prometheus* from the jsonnet.json file
+jb update
+cp lockfile.json backup
```
arajkumar, line 69:
```suggestion
cp jsonnetfile.lock.json backup
```

Documentation/development.md
```diff
@@ -0,0 +1,83 @@
...
+# restore the original jsonnet.json file
+git checkout -- jsonnet.json # (restore to original)
+
+#apply all changes from backup to lockfile
+vimdiff backup lockfile.json
```
arajkumar, line 74: Actually you can use `git add -p` to ease the partial change staging.

Documentation/development.md
```diff
@@ -0,0 +1,83 @@
...
+### Build and Run CMO
+
+The `make` target `run-local` builds the `operator` and runs CMO locally
+
+```shell
+export KUBECONFIG=/path/to/kubeconfig/file
+make run-local
+```
+
+### Fix jsonnet
```
fpetkovski, line 18: I think we need a more specific heading here. Maybe something like "working with jsonnet" and a subsection for different operations. I find Fix jsonnet is a bit blunt and unspecific.

Documentation/development.md
```diff
@@ -0,0 +1,83 @@
...
+1. Amend yaml files
+1. ☝☝☝ _rinse and repeat_
+
+Finally:
+1. Port the changes back to `jsonnet` or vendored `jsonnet`
+
+
+
+
+## Updating vendored jsonnsets
```
fpetkovski, line 55:
```suggestion
## Updating vendored jsonnet code
```

Documentation/development.md
```diff
@@ -0,0 +1,83 @@
...
+
+NOTE: `jb update <repo-url>/<jsonnet-subdir>` doesn't seem to work since it
```
fpetkovski, line 57: Is this a bug or is it expected? Is it related to https://github.com/jsonnet-bundler/jsonnet-bundler/issues/142?

Documentation/development.md
```diff
@@ -0,0 +1,83 @@
...
+does not update the dependencies (at the time of writing)
+
+Steps to update a bundle e.g. `kube-prometheus`:
+
+```shell
+cd jsonnset
```
fpetkovski, line 63:
```suggestion
cd jsonnet
```

Documentation/development.md
```diff
@@ -0,0 +1,83 @@
...
+### Fix jsonnet
+
+**TL;DR:**
+1. build operator
+1. use `hack/local-cmo.sh`
+1. modify assets/yamls
+1. backport
+
+
+Sometimes we may not want to rebuild the operator. E.g. you may need
```
fpetkovski, line 27: Do we have examples of such use cases?

sthaha: Yes, we do, I got this information from Pawel and it seems like a common use-case

Say some changes need to go into jsonnet (including vendored bundle pulled from different git repo)
The simpler worflow seems to modify assets directly, make it work and then backport these changes to jssonet (including the vendored jsonnet in a different git repos)

sthaha: In fact it was the reason why we removed a change that I had put into local-run.sh to build `operator` every single time.

Documentation/development.md
sthaha: Have you tried it and seen if it is any easier than using a merge tool? I found the `git add -p` harder to get right.

Documentation/development.md
sthaha: Good catch :)

Documentation/development.md
sthaha: That could be the issue ... I haven't looked further into it. 

When I tried to update just `kube-prometheus` what I found was that it only updated the `kube-prometheus` and not any transitive dependencies. E.g. it didn't update the `kubernetes-mixin`. 

This was the workaround that worked.

Documentation/development.md
```diff
@@ -0,0 +1,84 @@
...
+# E.g. remove everything but *kube-prometheus* from the jsonnetfile.json file
+
+jb update
+cp jsonnetfile.lock.json backup.json
+
+# restore the original jsonnetfile.json file
+git checkout -- jsonnetfile.json # (restore to original)
+
+# apply all changes from backup to lockfile
+vimdiff backup.json  jsonnnetfile.lock.json
```
prashbnair, line 75: In line 69 you are copying the lock file to backup , so there will not be any changes that need to be moved to the lock file.

sthaha: There should be changes since `jb update` should have changed the lockfile 
Note  that you will need to 

> # edit jsonnetfile.json and remove all but the bundle you want to update
> # E.g. remove everything but *kube-prometheus* from the jsonnetfile.json file

before running the update.

Documentation/development.md
fpetkovski: It seems that the same problem is being reported in that issue
```
What actually seems to be happening for transitive dependencies is that jb is retrieving the versions specified in the jsonnetfile.json in the dependency, rather than our local lock file.
```
I suggest we set up a ticket and fix this problem. We can still document the workaround temporarily and remove it once the problem with jsonnet-bundler is resolved.

Documentation/development.md
arajkumar: I found it to be bit easier.

Documentation/development.md
```diff
@@ -0,0 +1,84 @@
...
+1. Amend yaml files in `assets/` directory
+1. Run `hack/local-cmo.sh` (**does not** build the operator)
+1. ☝☝☝ _rinse and repeat_
+
+#### Finally:
+1. Port the changes back to `jsonnet` or vendored `jsonnet`
+
+
+
+## Updating vendored jsonnet code
```
dgrisonnet, line 54: The usual way I do it which might be easier is:
```sh
cd jsonnet
jb update
# add only the dependencies that you want to update
git add -p -- jsonnetfile.lock.json
git checkout .
git reset HEAD -- jsonnetfile.lock.json
cd ..
make clean; make generate
git add -u
```

dgrisonnet (commented):
Great initiative @sthaha!

One thing that could be documented also is when should running kube-prometheus for tests purposes is considered good enough. Running CMO locally takes a lot of time compared to just spinning up a kube-prometheus stack and apply manifests changes, so I think a lot of newcomers could benefit from that and save a lot of time and effort. wdyt?

Documentation/development.md
sthaha: The problem I think I find with `git add -p` is that  finding transitive dependencies of the **only** package that you want to update could be harder compared to the approach you suggested. 

I will add both and point this as the first approach to take.

sthaha:
@dgrisonnet 

> One thing that could be documented also is when should running kube-prometheus for tests purposes is considered good enough. 

I am not entirely sure when this should be done. For almost all work I have done until now (except for prometheus), I have been using CMO explicitly. Could you shed some light into this? It would also be great if we could capture as a `todo` in jira issue or GH issue so that we can close this PR.

arajkumar (approved):
lgtm

Documentation/development.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+Revert back the changes to jsonnetfile.json and jsonnetfile.lock.json
+```
+git restore --  jsonnetfile.json jsonnnet.lock.json
+
+```
+
+Now you should have a `jsonnetfile.lock.json` that has only the updates to
+the bundle `kube-prometheus`. Remove the  ``vendor`` and reinstall all bundles.
```
simonpasquier, line 100:
```suggestion
the bundle `kube-prometheus`. Remove the  `vendor` and reinstall all bundles.
```

simonpasquier (commented):
Thanks for taking the time of writing this! One small typo spotted otherwise I think it's good enough to be merged and we can still improve further.

simonpasquier:
/lgtm
/hold
@dgrisonnet feel free to remove the hold if you're happy with the current change

simonpasquier:
/label px-approved
/label qe-approved
/label docs-approved

this change has no user impact since this documentation targets only developers contributing to this repository.

dgrisonnet:
All good from my side :rocket: 

I'll have a look at adding some information related to kube-prometheus later on, but I wouldn't block this PR based on that.

/unhold

PR #1370, opened 2021-09-08
Bug 2001566: enabling alerts for prometheus operator in uwm

prashbnair:

### DIFF
--- a/jsonnet/main.jsonnet
+++ b/jsonnet/main.jsonnet
@@ -271,7 +271,7 @@ local inCluster =
         mixin+: {
           ruleLabels: $.values.common.ruleLabels,
           _config+: {
-            prometheusSelector: 'job=~"prometheus-k8s|prometheus-user-workload"',
+            prometheusOperatorSelector: 'job="prometheus-operator", namespace=~"%(namespace)s|%(namespaceUserWorkload)s"' % ($.values.common),
           },
         },
         tlsCipherSuites: $.values.common.tlsCipherSuites,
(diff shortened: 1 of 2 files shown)


jsonnet/main.jsonnet
```diff
@@ -271,7 +271,7 @@ local inCluster =
         mixin+: {
           ruleLabels: $.values.common.ruleLabels,
           _config+: {
-            prometheusSelector: 'job=~"prometheus-k8s|prometheus-user-workload"',
+            prometheusOperatorSelector: 'job="prometheus-operator"',
```
arajkumar, line 274: We should restrict it to only PM and UWM namespaces.

arajkumar (changes requested):

jsonnet/main.jsonnet
```diff
@@ -271,7 +271,7 @@ local inCluster =
         mixin+: {
           ruleLabels: $.values.common.ruleLabels,
           _config+: {
-            prometheusSelector: 'job=~"prometheus-k8s|prometheus-user-workload"',
+            prometheusOperatorSelector: 'job="prometheus-operator", namespace=~"openshift(-|-user-workload-)monitoring"',
```
arajkumar, line 274: We should take namespace info from config.

```suggestion
            prometheusOperatorSelector: 'job="prometheus-operator", namespace=~"%(namespace)s|%(namespaceUserWorkload)s"' % $.values.common,
```

arajkumar (changes requested):

PR #1364, opened 2021-09-07
Bug 1987263: fsSpaceFillingUpWarningThreshold not aligned to Kubernetes Garbage Collection Threshold

arajkumar:
The goal of this PR is to pick changes related to https://github.com/prometheus-operator/kube-prometheus/pull/1357.  It also updates jsonnet dep version to master as we are on 4.10 track.

* [ ] I added CHANGELOG entry for this change.
* [X] No user facing changes, so no entry in CHANGELOG was needed.

PR #1371, opened 2021-09-08
Unpin jsonnet dependencies

dgrisonnet:
Since the release-4.10 branch is now open, we can unpin the jsonnet dependencies that were pinned for the 4.9 release.


arajkumar:
It has been already done as part of https://github.com/openshift/cluster-monitoring-operator/pull/1364

dgrisonnet:
Ah my bad I missed it, thank you for the heads-up @arajkumar

PR #1372, opened 2021-09-09
Bug 2000536:  jsonnet: Remove redundant KSM alerts

PhilipGough:
Removes kube-state-metrics alerts that will not fire on OCP

### DIFF
--- a/jsonnet/utils/sanitize-rules.libsonnet
+++ b/jsonnet/utils/sanitize-rules.libsonnet
@@ -103,6 +103,14 @@ local excludedRules = [
       { alert: 'ThanosQueryRangeLatencyHigh' },
     ],
   },
+  {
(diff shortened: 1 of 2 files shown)


jsonnet/utils/sanitize-rules.libsonnet
```diff
@@ -103,6 +103,13 @@ local excludedRules = [
       { alert: 'ThanosQueryRangeLatencyHigh' },
     ],
   },
+  {
+    name: 'kube-state-metrics',
```
simonpasquier, line 107: can you add a comment explaining why these are removed? thanks :)

slashpai:
This looks like duplicate of https://github.com/openshift/cluster-monitoring-operator/pull/1366/files?

PhilipGough:
Closed as this duplicates the work done in #1366

PR #1373, opened 2021-09-10
Support enabling the `query_log_file` config for Prometheus

PhilipGough:
Enables setting https://prometheus.io/docs/guides/query-log/#enable-the-query-log on/off via the ConfigMap for both platform and user workload monitoring.

### DIFF
--- a/examples/user-workload/README.md
+++ b/examples/user-workload/README.md
@@ -18,25 +18,26 @@ data:
 Configuration example is located in this directory, with the following supported configuration fields:
 ```
 prometheusOperator:
-logLevel     string
-nodeSelector map[string]string
-tolerations  []v1.Toleration
+  logLevel     string
+  nodeSelector map[string]string
+  tolerations  []v1.Toleration
 
 thanosRuler:
-logLevel     string
-nodeSelector map[string]string
-tolerations  []v1.Toleration
-resources           *v1.ResourceRequirements
-volumeClaimTemplate *v1.PersistentVolumeClaim
+  logLevel            string
+  nodeSelector        map[string]string
+  tolerations         []v1.Toleration
+  resources           *v1.ResourceRequirements
+  volumeClaimTemplate *v1.PersistentVolumeClaim
 
 prometheus:
-logLevel     string
-nodeSelector map[string]string
-tolerations  []v1.Toleration
-retention string
-resources           *v1.ResourceRequirements
-externalLabels      map[string]string
-volumeClaimTemplate *v1.PersistentVolumeClaim
-hostport            string
-remoteWrite         []monv1.RemoteWriteSpec
+  logLevel            string
+  nodeSelector        map[string]string
+  tolerations         []v1.Toleration
+  retention           string
+  resources           *v1.ResourceRequirements
+  externalLabels      map[string]string
+  volumeClaimTemplate *v1.PersistentVolumeClaim
+  hostport            string
+  remoteWrite         []monv1.RemoteWriteSpec
+  queryLogFile        string
 ```
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1376,6 +1376,10 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
 		}
 	}
 
+	if f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.QueryLogFile != "" {
+		p.Spec.QueryLogFile = f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.QueryLogFile
+	}
+
 	telemetryEnabled := f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.IsEnabled()
 	if telemetryEnabled && f.config.RemoteWrite {
 
@@ -1668,6 +1672,10 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret) (*monv1.Prometheus,
 
 	if f.config.Images.Thanos != "" {
 		p.Spec.Thanos.Image = &f.config.Images.Thanos
+	}
+
+	if f.config.UserWorkloadConfiguration.Prometheus.QueryLogFile != "" {
+		p.Spec.QueryLogFile = f.config.UserWorkloadConfiguration.Prometheus.QueryLogFile
 	}
 
 	for i, container := range p.Spec.Containers {
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -1035,6 +1035,7 @@ func TestPrometheusK8sConfiguration(t *testing.T) {
     datacenter: eu-west
   remoteWrite:
   - url: "https://test.remotewrite.com/api/write"
+  queryLogFile: /tmp/test
 ingress:
   baseAddress: monitoring-demo.staging.core-os.net
 `)
@@ -1167,6 +1168,10 @@ ingress:
 
 	if p.Spec.RemoteWrite[0].URL != "https://test.remotewrite.com/api/write" {
 		t.Fatal("Prometheus remote-write is not configured correctly")
+	}
+
+	if p.Spec.QueryLogFile != "/tmp/test" {
+		t.Fatal("Prometheus query log is not configured correctly")
 	}
 }
 
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -151,6 +151,7 @@ func TestClusterMonitorPrometheusK8Config(t *testing.T) {
 	data := fmt.Sprintf(`prometheusK8s:
   logLevel: debug
   retention: 10h
+  queryLogFile: /tmp/test.log
   tolerations:
     - operator: "Exists"
   externalLabels:
@@ -198,6 +199,10 @@ func TestClusterMonitorPrometheusK8Config(t *testing.T) {
 		{
 			name:      "assert remote write url value in set in CR",
 			assertion: assertRemoteWriteWasSet(f.Ns, crName, "https://test.remotewrite.com/api/write"),
+		},
+		{
+			name:      "assert query log file value is set and correct",
+			assertion: assertQueryLogValueEquals(f.Ns, crName, "/tmp/test.log"),
 		},
 		{
 			name:      "assert rule for Thanos sidecar exists",
@@ -523,6 +528,7 @@ func TestUserWorkloadMonitorPrometheusK8Config(t *testing.T) {
   enforcedTargetLimit: 10
   logLevel: debug
   retention: 10h
+  queryLogFile: /tmp/test.log
   tolerations:
     - operator: "Exists"
   externalLabels:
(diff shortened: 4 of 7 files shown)


examples/user-workload/README.md
```diff
@@ -39,4 +39,5 @@ externalLabels      map[string]string
 volumeClaimTemplate *v1.PersistentVolumeClaim
 hostport            string
 remoteWrite         []monv1.RemoteWriteSpec
+queryLogFile string
```
sthaha, line 42: nit: `go fmt` ?

pkg/manifests/manifests.go
```diff
@@ -1281,6 +1281,10 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
 		}
 	}
 
+	if f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.QueryLogFile != "" {
```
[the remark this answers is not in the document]
sthaha: you might find this `go idiom` useful
```go
if logFile := f.config....QueryLogFile; logFile != "" {
  p.spec.QueryLogFile =  logFile
}
```

pkg/manifests/manifests_test.go
```diff
@@ -1053,6 +1054,10 @@ ingress:
 	if p.Spec.RemoteWrite[0].URL != "https://test.remotewrite.com/api/write" {
 		t.Fatal("Prometheus remote-write is not configured correctly")
 	}
+
+	if p.Spec.QueryLogFile != "/tmp/test" {
+		t.Fatal("Prometheus query logis not configured correctly")
```
sthaha, line 1059: nit: `log is`

test/e2e/config_test.go
```diff
@@ -1110,3 +1120,22 @@ func assertRemoteWriteWasSet(namespace, crName, urlValue string) func(t *testing
...
+func assertQueryLogWasSet(namespace, crName, value string) func(t *testing.T) {
+	return func(t *testing.T) {
+		err := framework.Poll(time.Second, time.Minute*5, func() error {
+			prom, err := f.MonitoringClient.Prometheuses(namespace).Get(context.Background(), crName, metav1.GetOptions{})
+			if err != nil {
+				t.Fatal("failed to get required prometheus cr", err)
+			}
+
+			if prom.Spec.QueryLogFile != value {
+				return fmt.Errorf("expected query log file value not found")
```
sthaha, line 1133: might want to consider using `%w` to wrap `value` in Error returned.

test/e2e/config_test.go
```diff
@@ -794,6 +800,10 @@ func TestUserWorkloadMonitorPrometheusK8Config(t *testing.T) {
 			name: "assert remote write url value in set in CR",
 			f:    assertRemoteWriteWasSet(f.UserWorkloadMonitoringNs, crName, "https://test.remotewrite.com/api/write"),
 		},
+		{
+			name: "assert query log file value is set and correct",
```
sthaha, line 804: Should we have a test that validates that after query_log_file is set, and then reset, the queryLogFile becomes `""` ?

sthaha (commented):
/lgtm 
some minor comments that you may want to address

pkg/manifests/manifests.go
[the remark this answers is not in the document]
PhilipGough: we could indeed copy correctly. If you look further up this function though you will see similar pattern https://github.com/openshift/cluster-monitoring-operator/pull/1373/files/4a3d1247b17a8fc816096224041213bf351d8adf#diff-1f9d2d18d4cabd8262e585b932330183f7cb9b544ac9f27c8bcb3ef9c8d14376R1251-R1257

Often I find for readability it is best to match the existing style, wdyt?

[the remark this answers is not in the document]
sthaha: I agree, it is better to follow the style in general but it in particular instance, I am not quite convinced that it was intended as a convention, rather I feel it is something  the original author may have missed ...  but again, this is only a `nit`, so please feel free to ignore :-)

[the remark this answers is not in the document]
PhilipGough: no its a valid point, and I agree that we should make the code idiomatic. I'm just not sure this PR is the right place to make those changes. Perhaps we should curate a list of things we need to "tidy" in the codebase? I've done a similar cleanup on the e2e tests already.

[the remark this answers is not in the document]
sthaha: Yes, makes sense to move the cleanup to a different PR 🤗

test/e2e/config_test.go
PhilipGough: personally I don' think so, because we don't do it for any other config option and not entirely sure we should treat this any differently.

e2e tests are expensive time wise, and I feel we already rely on them more heavily than we should. Testing the happy path in e2e tests is sufficient in my mind unless its critical we cover other cases.

Having said that, lets leave it open and see what others think.

sthaha: Do we require a e2e test for this ? Can't this be done in the `manifest_test.go` itself?

PhilipGough: It's already a path that is tested. The pattern as of now in regards to config is that we test the happy path as is the case with this test.

I did already add a unit test in `manifest_test.go` for same.

examples/user-workload/README.md
PhilipGough: done

pkg/manifests/manifests.go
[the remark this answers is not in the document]
simonpasquier: Late to the party but technically we don't need the `if` check since an empty value disables query logging.

https://github.com/prometheus-operator/prometheus-operator/blob/2a802a189b1230ef04e82c3e5ecad9de80668db3/pkg/prometheus/promcfg.go#L264-L268

https://github.com/prometheus/prometheus/blob/1270b87970baeb926fcce64552db5c744ffaf83f/cmd/prometheus/main.go#L558-L561

simonpasquier (commented):
/lgtm

PhilipGough:
@bburt-rh are we okay to add a docs label here? The docs PR is https://github.com/openshift/openshift-docs/pull/36495

bburt-rh:
@PhilipGough Is this for the 4.9 or 4.10 release?

bburt-rh:
@PhilipGough Starting with 4.10 PRs, engineering can start adding the docs-approved label themselves. See Simon's email from earlier today, and lmk if you have questions.

PhilipGough:
@bburt-rh I did see the note from Simon today, however I had asked the question yesterday :)

Thanks

/label docs-approved

bburt-rh:
@PhilipGough perfectly timed, I think!  ;-)

PhilipGough:
@sferich888  can this be considered for `px-approval` ?

simonpasquier (commented):
/lgtm

PhilipGough:
@sferich888 another nudge on this one for px-approval if we could?

juzhao:
@PhilipGough 
I see the queryLogFile size would be increased as time goes by, do we have limit for it?
sh-4.4$ du -h /tmp/test-cluster.log
6.6M /tmp/test-cluster.log


sh-4.4$ du -h /tmp/test-cluster.log
11M	/tmp/test-cluster.log
```

**Comment by juzhao** [comment]
tested with the PR, enabled UWM and set
```
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    enableUserWorkload: true
    prometheusK8s:
      retention: 3h
      logLevel: debug
      queryLogFile: /tmp/test-cluster.log
      volumeClaimTemplate:
        spec:
          volumeMode: Filesystem
          resources:
            requests:
              storage: 10Gi
```
and
```
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    prometheus:
      logLevel: warn
      queryLogFile: /tmp/test-uwm.log
      volumeClaimTemplate:
        spec:
          volumeMode: Filesystem
          resources:
           requests:
              storage: 10Gi
```
create prometheusrules under user namespace
```
kind: PrometheusRule
metadata:
  name: example-alert
  namespace: ns1
spec:
  groups:
  - name: example
    rules:
    - alert: HighErrors
      expr: (sum(rate(http_requests_total{code!~"2.."}[5m])) / sum(rate(http_requests_total[5m])))
        * 100 > 10
    - alert: TestAlert
      annotations:
        message: This is an alert meant to ensure that the entire alerting pipeline
          is functional.
      expr: vector(1)
      labels:
        severity: none
```
we can see the query logs for openeshift-monitoring prometheus, but can not find from UWM prometheus
```
# oc -n openshift-user-workload-monitoring rsh -c prometheus prometheus-user-workload-0
sh-4.4$ cat /tmp/test-uwm.log
empty result
```

**Comment by PhilipGough** [comment]
@juzhao no, we have no limit for it. We will explicitly call this out in the docs and this should be enabled only as a temporary troubleshooting solution -> https://github.com/openshift/openshift-docs/pull/36495/files#diff-c934df74a3d8f522bb7fae910d0e0b4cb8f5a92facbe61d093a16297e8c9803dR12

**Comment by PhilipGough** [comment]
@juzhao - following your exact config I am not seeing the same behaviour when this PR is deployed:

```shell
oc -n openshift-user-workload-monitoring exec -c prometheus prometheus-user-workload-0 -- cat /etc/prometheus/config_out/prometheus.env.yaml
```
produces (shortened) output showing the configmap has set the correct config as per https://prometheus.io/docs/guides/query-log/#logging-all-the-queries-to-a-file

```yaml
global:
  evaluation_interval: 30s
  scrape_interval: 30s
  external_labels:
    prometheus: openshift-user-workload-monitoring/user-workload
    prometheus_replica: prometheus-user-workload-0
  query_log_file: /tmp/test-uwm.log
rule_files:
- /etc/prometheus/rules/prometheus-user-workload-rulefiles-0/*.yaml
scrape_configs: []
```

Run the following query:
```shell
 oc -n openshift-user-workload-monitoring exec -c prometheus prometheus-user-workload-0 -- curl 'http://localhost:9090/api/v1/query?query=up'

```

Followed by: 

 ```shell
oc -n openshift-user-workload-monitoring exec -c prometheus prometheus-user-workload-0 -- cat /tmp/test-uwm.log | jq
```

Results in:

```json
{
  "httpRequest": {
    "clientIP": "127.0.0.1",
    "method": "GET",
    "path": "/api/v1/query"
  },
  "params": {
    "end": "2021-11-24T11:32:15.595Z",
    "query": "up",
    "start": "2021-11-24T11:32:15.595Z",
    "step": 0
  },
  "stats": {
    "timings": {
      "evalTotalTime": 3.5229e-05,
      "resultSortTime": 0,
      "queryPreparationTime": 2.2002e-05,
      "innerEvalTime": 7.712e-06,
      "execQueueTime": 7.8172e-05,
      "execTotalTime": 0.000137558
    }
  },
  "ts": "2021-11-24T11:32:15.595Z"

```

PhilipGough:
Screenshot 2021-11-24 at 11 25 27

PR #1377, opened 2021-09-14
[MON-1786] Enable audit logs by default for Prometheus Adapter

sthaha:
Enabling audit logs helps troubleshoot issues in Prometheus Adapter. The
details of the audit level can be tweaked by selecting an audit profile
just like in OpenShift API server.

The following profiles are provided which corresponds to [Audit Log Levels](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#audit-policy)
 - Request
 - RequestResponse
 - None
 - *Metadata* (the default audit profile)
 
User can pick any of the profile by modifying the CMO configmap as follows

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    k8sPrometheusAdapter:
      audit:
        profile: Request
```

CMO picks up the change and (re)deploys `prometheus-adapter` with appropriate `--audit-` flags. 


assets/prometheus-adapter/config-map.yaml
```diff
@@ -1,5 +1,21 @@
 apiVersion: v1
 data:
+  audit-policy-metadata.yaml: |-
```
sthaha, line 3: @dgrisonnet if we are happy with the approach, I will submit the jsonnet changes to `prometheus-adapter.libsonnet`

pkg/manifests/config.go
```diff
@@ -268,6 +268,7 @@ type OpenShiftStateMetricsConfig struct {
 type K8sPrometheusAdapter struct {
 	NodeSelector map[string]string `json:"nodeSelector"`
 	Tolerations  []v1.Toleration   `json:"tolerations"`
+	AuditLevel   string            `json:"auditLevel"`
```
sthaha, line 271: @dgrisonnet https://docs.openshift.com/container-platform/4.8/security/audit-log-policy-config.html#configuring-audit-policy_audit-log-policy-config  uses `profiles`. 

 We could do the same where each profile has a different meaning `Default, WriteRequestBodies, or AllRequestBodies`  but in our case, the default is to not enable any audit logs. I  prefer this implementation where the control is the `Level` of `Audit Logs` to theirs where each profile has a different meaning.  WDYT?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 4.9
 
+- [#1377](https://github.com/openshift/cluster-monitoring-operator/pull/1377) Allow OpenShift users to configure audit logs for prometheus-adapter
```
slashpai, line 5: This should be in 4.10

pkg/manifests/manifests.go
```diff
@@ -1719,6 +1719,18 @@ func (f *Factory) PrometheusAdapterConfigMapPrometheus() (*v1.ConfigMap, error)
...
+func auditLogsEnabled(level string) bool {
+	switch level {
+	case "None",
+		"Metadata",
+		"Request",
+		"RequestResponse":
+		return level != "None" // no audit logs for None
+	default:
+		return false
+	}
```
PhilipGough, line 1731: this function seems a bit confusing to me, since it checks a list of magic strings without any context that are essentially ignored, perhaps it could just be 

```	
if level == "" || level == "None" {
		return false
	}
```

PhilipGough (changes requested):
Overall looks good. Would be good to have some link in the commit message, PR or changelog etc as some context as to where the various options are coming from.

Probably needs an e2e test in https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/config_test.go?

pkg/manifests/manifests.go
sthaha: `None, Metadata, Request, RequestResponse` are the only valid `Log Levels` for which an audit-policy is defined. 
 

```
if level == "" || level == "None" { return false }
```
the above would return `true` if user sets the loglevel to "foobar", wouldn't it?  or am I missing something?

pkg/manifests/manifests.go
```diff
@@ -1769,6 +1783,14 @@ func (f *Factory) PrometheusAdapterDeployment(apiAuthSecretName string, requesth
 		},
 	)
 
+	if auditLogsEnabled(config.AuditLevel) {
+		lvl := strings.ToLower(config.AuditLevel)
```
sthaha, line 1787: @PhilipGough continuing from the previous comment ... 

providing a `valid` log level other than `none` adds `--audit-` flags to the `prometheus adapter`

pkg/manifests/manifests.go
PhilipGough: yes you are correct, makes sense now, I think removing `None` is cleaner for sure.

PhilipGough: A link to https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#audit-policy might be useful as a comment on this function. 

Alternatively, and probably preferably the func could take a concrete type from `"k8s.io/apiserver/pkg/apis/audit"`

PhilipGough: Lastly would we want to have a warning log in the case where an invalid option has been provided perhaps?

pkg/manifests/manifests.go
```diff
@@ -1769,6 +1783,14 @@ func (f *Factory) PrometheusAdapterDeployment(apiAuthSecretName string, requesth
...
+		spec.Containers[0].Args = append(spec.Containers[0].Args,
+			fmt.Sprintf("--audit-policy-file=/etc/adapter/audit-policy-%s.yaml", lvl),
+			"--audit-log-path=/tmp/audit.log",
```
prashbnair, line 1790: Do you want to set a maxage or maxsize for this to ensure that we do not run out of space?

pkg/manifests/manifests.go
sthaha: I agree with a warning log albeit k8s way would be to update the status of "some" resource (surely an overkill 😝) 
I did think of a warning log but the rest of the code seems to not warn anything 🤔 ... I will add a log message.

pkg/manifests/manifests.go
sthaha: I agree ... 2 things 
1. not too sure what we want to set as an upper limit (any thoughts? ) 
1. Please note /tmp is a `tmpfs` mount (in our deployment)

pkg/manifests/config.go
simonpasquier: Apparently there's also a `None` profile:

https://github.com/openshift/api/blob/cc0db1116639638254e87a564902833f1ee006d5/config/v1/types_apiserver.go#L68-L82

The advantage of using the same approach is that it reduces the cognitive load for users (e.g. consistent with the other OpenShift components) and we can easily link to the existing OpenShift docs to explain what each profile does. The flip side is that it may require more work for CMO though we can probably reuse [existing code](https://github.com/openshift/library-go/blob/fd8365501a47f82b85f5433f821ab047ba19011a/pkg/operator/apiserver/audit/audit_policies.go#L84-L114) to generate the policy configmap. Another disadvantage is that it would be harder/impossible to support the same in kube-prometheus.

dgrisonnet: I also think that reusing the same format as the other OpenShift components would make it easier for the end-users, so I would side with using the same levels as the apiserver.

Instead of using a string to store the audit level we could use the same format as the apiserver and have an [Audit](https://github.com/openshift/library-go/blob/fd8365501a47f82b85f5433f821ab047ba19011a/vendor/github.com/openshift/api/config/v1/types_apiserver.go#L84-L116) object from which we could generate a profile based on that with [GetAuditPolicy](https://github.com/openshift/library-go/blob/fd8365501a47f82b85f5433f821ab047ba19011a/pkg/operator/apiserver/audit/audit_policies.go#L88-L114).

> Another disadvantage is that it would be harder/impossible to support the same in kube-prometheus.

Since these profiles are OpenShift specific, we won't be able to have the same in both upstream and downstream if we were to follow the route of consistency. That said, with the current implementation of kube-prometheus, I would not expect audit logs to be configured by default and only configurable in an addon. So I think it is fine even if we were to diverge here.

>  the default is to not enable any audit logs

Thinking about it again, I don't see any disadvantages to having audit logs by default. I think we should be consistent with the apiserver here and set it to `Default` profile by default. This would improve the observability of the apiserver as a whole and if audit logs were deemed valuable enough to be enabled by default in the apiserver, that should also be the case for aggregated APIs to cover all API resources.

pkg/manifests/manifests.go
```diff
@@ -1769,6 +1790,19 @@ func (f *Factory) PrometheusAdapterDeployment(apiAuthSecretName string, requesth
 		},
 	)
 
+	enableAudit, err := auditLogsEnabled(config.AuditLevel)
```
dgrisonnet, line 1793: Even if the audit log level is set to `None`, we could have everything set up in prometheus-adapter to produce audit logs. I think that we should not introduce a logic saying that this particular level will produce or not logs and stick to set up everything by default and leave it up to the apiserver to handle the logs based on the policy level.

pkg/manifests/manifests.go
dgrisonnet: I think we should follow what the apiserver is doing and have an initContainer setup the directory on the node that we would then use as a volume to store the logs: https://github.com/openshift/cluster-openshift-apiserver-operator/blob/ff289ebf4cd2df621801dd6b76a932ac39f1833b/bindata/v3.11.0/openshift-apiserver/deploy.yaml#L46-L61

sthaha: @simonpasquier I was basing this on https://github.com/sthaha/cluster-monitoring-operator/blob/b728aa6c986e51c4fea9f1111038fb2b3af7a723/assets/prometheus-adapter/deployment.yaml#L60

pkg/manifests/manifests.go
sthaha: @PhilipGough shall I mark this as resolved?

assets/prometheus-adapter/config-map.yaml
sthaha: I am bit lost here, 
* Where would the `policy object` be stored?  IIUC we need an actual audit-policy `file` that can be read by the api server. 
* Are you suggesting that CMO generate the policy file at runtime, create a config-map  and then inject that into the promethes-adapter (update the deployment), if so, I am not sure what advantage it gives us 🤔

pkg/manifests/manifests.go
simonpasquier: Using `hostPath` is probably problematic security-wise since IIUC it imposes to run as root:
https://github.com/openshift/cluster-openshift-apiserver-operator/commit/95e60ab3a8a9f20a53d523e43d71820c91f55f99

pkg/manifests/config.go
simonpasquier: If we manage to find a solution for limiting the size of the logs, it's fine by me.

dgrisonnet: We can set a sane default for the `--audit-log-maxsize` flag of prometheus-adapter. If the size is exceeded, the log file will be rotated.

assets/prometheus-adapter/config-map.yaml
dgrisonnet: As per our offline discussion, we should stick to having a configmap containing all policy manifests, but it should be separated from the prometheus-adapter config.

pkg/manifests/manifests.go
dgrisonnet: good point, do you know of any alternative we could have? In the end, all the logs are stored on the node's disk under /var/log, and I believe that is what the must-gather also expects.

sthaha: In must-gather scripts the `default` SA has all permission so we should be able to `oc exec|cp` from the adapter pod.

I will create a new mount for `/var/log/adapter/[audit.log]` instead of using `/tmp` 

will set the max log size to 100MB, or  do you want that to be configurable  as well @simonpasquier , @dgrisonnet ?

Makefile
```diff
@@ -14,7 +14,8 @@ export GOPROXY
 
 PKGS=$(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
 GOLANG_FILES:=$(shell find . -name \*.go -print)
-ASSETS=$(shell grep -oh '[^"]*/.*\.yaml' pkg/manifests/manifests.go | sed 's/^/assets\//')
+ASSETS=$(shell grep -oh '[^"]*/.*\.yaml' pkg/manifests/manifests.go \
+          | grep -v '%.*yaml' | sed 's/^/assets\//')
```
sthaha, line 18: `-v %.yaml` is need because `%s-policy.yaml` is used in manifest.go and that isn't a valid asset.

jsonnet/components/prometheus-adapter-audit.libsonnet
```diff
@@ -0,0 +1,117 @@
+// NOTE: audit rules are copied from openshift/library-go/pkg/operator/apiserver/audit/manifests/*.yaml
+// to maintain the same UX as OpenShift apiserver
+
+// createProfile prepends base audit rules to the rules passed
+local createProfile(rules) =
+  // see openshift/library-go/pkg/operator/apiserver/audit/manifests/base-policy.yaml
+  local baseAuditPolicy = |||
+    apiVersion: audit.k8s.io/v1
```
arajkumar, line 8: Can't this be defined as jsonnet and later convert to string(for configmap) using std.toString()? It bit odd to see k8s objects being defined as string inside jsonnet.

pkg/manifests/manifests_test.go
```diff
@@ -1573,6 +1574,136 @@ alertmanagerMain:
...
+		}
+		assertDeepEqual(t, test.expectedArgs, auditArgs,
+			"k8s-prometheus-adapter audit is not configured correctly")
+
+	}
+
+}
+
+func assertDeepEqual(t *testing.T, expected, got interface{}, msg string) {
+	if !reflect.DeepEqual(expected, got) {
```
arajkumar, line 1696: nit:

Is it possible to use `https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff` for better test failure actual vs expected comparison?

arajkumar (commented):
Overall looks good to me with few nits.

sthaha:
@simonpasquier @dgrisonnet , I have removed the `WIP`.  Does this PR require changes to any doc other than the openshift/docs (repo) ?

jsonnet/components/prometheus-adapter-audit.libsonnet
```diff
@@ -0,0 +1,117 @@
+// NOTE: audit rules are copied from openshift/library-go/pkg/operator/apiserver/audit/manifests/*.yaml
+// to maintain the same UX as OpenShift apiserver
```
dgrisonnet, line 2: We shouldn't copy the manifests directly since we are exposing audit logs on different resources than them. In our case, we will only expose audit logs for NodeMetrics and PodMetrics, so it doesn't really make sense to have a policy about routes and oauth since the adapter will never be responsible for these resources.

jsonnet/components/prometheus-adapter-audit.libsonnet
```diff
@@ -0,0 +1,117 @@
...
+
+// createProfile prepends base audit rules to the rules passed
+local createProfile(rules) =
+  // see openshift/library-go/pkg/operator/apiserver/audit/manifests/base-policy.yaml
+  local baseAuditPolicy = |||
```
dgrisonnet, line 7: I would expect a catch-all rule with `level: None` here

jsonnet/components/prometheus-adapter-audit.libsonnet
```diff
@@ -0,0 +1,117 @@
...
+      - "/healthz"
+      - "/readyz"
+  |||;
+
+  |||
+    %s
+    %s
+  ||| % [baseAuditPolicy, rules];
+
+local defaultRules = |||
```
dgrisonnet, line 36: I would expect a catch-all rule with `level: Metadata` here

jsonnet/components/prometheus-adapter-audit.libsonnet
```diff
@@ -0,0 +1,117 @@
...
+      resources: ["oauthaccesstokens", "oauthauthorizetokens"]
+  # A catch-all rule to log all other requests at the Metadata level.
+  - level: Metadata
+    # Long-running requests like watches that fall under this rule will not
+    # generate an audit event in RequestReceived.
+    omitStages:
+    - "RequestReceived"
+|||;
+
+local allRequestBodiesRules = |||
```
dgrisonnet, line 54: I would expect a catch-all rule with `level: RequestResponse` here

jsonnet/components/prometheus-adapter-audit.libsonnet
```diff
@@ -0,0 +1,117 @@
...
+  - level: Metadata
+    resources:
+    - group: "oauth.openshift.io"
+      resources: ["oauthclients"]
+  # catch-all rule to log all other requests with request and response payloads
+  - level: RequestResponse
+|||;
+
+
+local writeRequestBodiesRules = |||
```
dgrisonnet, line 70: @sthaha thinking about it again, this one doesn't make sense for prometheus-adapter since we won't receive any write requests on the metrics endpoint, only read ones. With that in mind, I am thinking that it might be better to stick to the upstream levels: https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#audit-policy in the end. wdyt?

pkg/manifests/manifests_test.go
```diff
@@ -1573,6 +1574,136 @@ alertmanagerMain:
...
+	tt := []struct {
+		config       string
+		expectedArgs []string
+		err          error
+	}{{
+		config: ``,
+		expectedArgs: []string{
+			"--audit-policy-file=/etc/audit/default-profile.yaml",
+			"--audit-log-path=/var/log/adapter/audit.log",
+			"--audit-log-maxsize=100",
```
juzhao, line 1588: I see there is `--audit-log-maxage` paramater for prometheus adapter, is there a default value for it?
```--audit-log-maxage int                                    The maximum number of days to retain old audit log files based on the timestamp encoded in their filename.```

File: pkg/manifests/manifests_test.go (hunk 10 above)
Comment (sthaha) [reply pkg/manifests/manifests_test.go:1696]: it is but and I think it is better to use https://pkg.go.dev/gotest.tools but that is a bigger change and this only refactors existing code albeit in new test.

File: jsonnet/components/prometheus-adapter-audit.libsonnet (hunk 9 above)
Comment (sthaha) [reply jsonnet/components/prometheus-adapter-audit.libsonnet:8]: `data` part of the configmap which is a string is represented like this everywhere in jsonnet .e.g. PTAL at prometheus-config.yaml. This configmap's data is storing an object/configuration that isn't an actual k8s resource but is an object that encoded/decoded using k8s schema and encoders.

File: jsonnet/components/prometheus-adapter-audit.libsonnet [hunk 17]
```diff
@@ -0,0 +1,30 @@
...
+  rules: [{ level: level }],
+};
+
+
+{
+  volume: {
+    name: 'audit-log',
+    path: '/var/log/adapter',
+  },
+  cm: {
```
arajkumar, line 19: nit: can we expand it? cm => configMap?
```suggestion
  configMap: {
```

pkg/manifests/manifests_test.go
arajkumar: Ack

pkg/manifests/manifests_test.go
```diff
@@ -1599,6 +1601,131 @@ alertmanagerMain:
...
+				t.Logf("%s\n\n", test.config)
+				t.Fatalf("Expected error %q but got %q", test.err, err)
+			}
+			break
+		}
+
+		adapterArgs := d.Spec.Template.Spec.Containers[0].Args
+		auditArgs := []string{}
+		for _, arg := range adapterArgs {
+			if strings.HasPrefix(arg, "--audit-") {
```
dgrisonnet, line 1706: Could we check only `--audit-policy-file` here? That would avoid checking the values of `--audit-log-path` and `--audit-log-maxsize` everytime even though they are not configurable by the users.

pkg/manifests/manifests_test.go
```diff
@@ -1599,6 +1601,131 @@ alertmanagerMain:
...
+			t.Fatal(err)
+		}
+		f := NewFactory("openshift-monitoring", "openshift-user-workload-monitoring", c,
+			defaultInfrastructureReader(), &fakeProxyReader{}, NewAssets(assetsPath))
+
+		d, err := f.PrometheusAdapterDeployment("foo", map[string]string{
+			"requestheader-allowed-names":        "",
+			"requestheader-extra-headers-prefix": "",
+			"requestheader-group-headers":        "",
+			"requestheader-username-headers":     "",
```
dgrisonnet, line 1691: I don't think we will need these headers for this particular test.

pkg/manifests/manifests_test.go
```diff
@@ -1599,6 +1601,131 @@ alertmanagerMain:
...
+    audit:
+      profile: Foobar  # should generate an error
+`,
+		err: ErrConfigValidation,
+	}}
+
+	for _, test := range tt {
+		c, err := NewConfigFromString(test.config)
+		if err != nil {
+			t.Logf("%s\n\n", test.config)
```
dgrisonnet, line 1681: Instead of logging the whole config for each failure, could you give a name to each test case and run them in subtests with `t.Run()`? This would already give us enough insights when debugging test failures.

pkg/manifests/manifests_test.go
sthaha: yes, I think `--audit-log-maxsize` (which is also optional) should be sufficient.

jsonnet/components/prometheus-adapter-audit.libsonnet
```diff
@@ -0,0 +1,30 @@
+local profile(level) = {
```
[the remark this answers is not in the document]
sthaha: Thanks for the tip, I will submit it 🙇

pkg/manifests/config.go
```diff
@@ -268,6 +269,11 @@ type OpenShiftStateMetricsConfig struct {
 type K8sPrometheusAdapter struct {
 	NodeSelector map[string]string `json:"nodeSelector"`
 	Tolerations  []v1.Toleration   `json:"tolerations"`
+	Audit        *Audit            `json:"audit"`
+}
+
+type Audit struct {
+	Profile auditv1.Level `json:"profile"`
```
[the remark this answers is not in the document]
sthaha: I thought about it but felt it is better to keep it `generic` so that more profiles can be added if needed while keeping the `UX` (profile) the same

pkg/manifests/manifests_test.go
sthaha: we could but I will keep at least one check since something needs to validate the args are set properly

pkg/manifests/manifests_test.go
sthaha: Initial I did try this but I got an error
```go
  d, err := f.PrometheusAdapterDeployment("foo", map[string]string{})  // or nil
```

```
   manifests_test.go:1682: Expected error "invalid value for config" but got "value not found in extension api server authentication configmap: key requestheader-allowed-names is missing"
-
```

pkg/manifests/manifests_test.go
sthaha: I am not sure I agree ... e.g  below provides a more readable test output 
```
=== RUN   TestK8sPrometheusAdapterAuditLog/None
    manifests_test.go:1686:
          k8sPrometheusAdapter:
            audit:
             profile: Noxne


    manifests_test.go:1687: Expected error %!q(<nil>) but got "invalid value for config - adapter audit profile: Noxne"
```
Than
```
=== RUN   TestK8sPrometheusAdapterAuditLog/None
    manifests_test.go:1706: k8s-prometheus-adapter audit is not configured correctly
        got:
                []string{"--audit-policy-file=/etc/audit/metadata-profile.yaml", "--audit-log-path=/var/log/adapter/audit.log", "--audit-log-maxsize=100"}

        expected:
                []string{"--audit-policy-file=/etc/audit/none-profile.yaml", "--audit-log-path=/var/log/adapter/audit.log", "--audit-log-maxsize=100"}
```
However, I have changed it

jsonnet/components/prometheus-adapter-audit.libsonnet
sthaha: I purposely kept it to `cm` to avoid the confusion with a `configMap` that is used elsewhere in the project.

jsonnet/components/prometheus-adapter.libsonnet
```diff
@@ -220,6 +241,16 @@ function(params)
...
+
+    configmapAuditProfiles: {
+      apiVersion: 'v1',
+      kind: 'ConfigMap',
+      metadata: {
+        name: audit.cm.name,
+        namespace: cfg.namespace,
+      },
+      data: audit.profiles,
+    },
```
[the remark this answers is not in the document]
sthaha: Can this be done when the file is moved and accepted upstream since we have to modify this anyway at that time?

sthaha:
@sferich888  Could you please (px) approve this PR?

sthaha:
@dgrisonnet here is the upstream PR - https://github.com/prometheus-operator/kube-prometheus/pull/1438 which improves on the current implementation (hopefully :-) )

simonpasquier:
/lgtm cancel
we need to make sure that the logs aren't kept forever which might be the case in the current version

pkg/manifests/manifests.go
```diff
@@ -1821,6 +1857,19 @@ func (f *Factory) PrometheusAdapterDeployment(apiAuthSecretName string, requesth
...
 
+	if err := validateAuditProfile(config.Audit.Profile); err != nil {
+		return nil, err
+	}
+
+	profile := strings.ToLower(string(config.Audit.Profile))
+	spec.Containers[0].Args = append(spec.Containers[0].Args,
+		fmt.Sprintf("--audit-policy-file=/etc/audit/%s-profile.yaml", profile),
+		"--audit-log-path=/var/log/adapter/audit.log",
+		"--audit-log-maxsize=100", // 100 MB
```
sthaha, line 1868: @simonpasquier @dgrisonnet I have tested that this works fine and that a max of  5 backups are kept. The `compression` does seems to reduce the size quite a bit (by a factor of 20) . 

```
-rw-------. 1 1000420000 1000420000  46828 Oct 19 04:58 audit-2021-10-19T04-58-50.055.log.gz
-rw-------. 1 1000420000 1000420000  46640 Oct 19 04:59 audit-2021-10-19T04-59-36.002.log.gz
-rw-------. 1 1000420000 1000420000  47372 Oct 19 05:00 audit-2021-10-19T05-00-21.835.log.gz
-rw-------. 1 1000420000 1000420000  46677 Oct 19 05:01 audit-2021-10-19T05-01-07.529.log.gz
-rw-------. 1 1000420000 1000420000  47168 Oct 19 05:01 audit-2021-10-19T05-01-56.287.log.gz
-rw-------. 1 1000420000 1000420000 899215 Oct 19 05:02 audit.log
```
Do you we want to compress the logs?

sferich888:
Who set qe-approved on this? I ask because its not QE acked in https://issues.redhat.com/browse/MON-1988 (as part of the planning processes). 

/label px-approved

sthaha:
> Who set qe-approved on this? I ask because its not QE acked in https://issues.redhat.com/browse/MON-1988 (as part of the planning processes).

If you are asking why QE approval was set rather than who ... I believe it  was given based [on the previous epic](https://issues.redhat.com/browse/MON-1853).  [MON-1988](https://issues.redhat.com/browse/MON-1988) was created recently to narrow down the scope of otherwise a large/never-ending epic

simonpasquier:
/label docs-approved

as per agreement with the docs team

PR #1382, opened 2021-09-15
Update KSM version to match what we deploy

fpetkovski:
The versions check was broken in the CI until recently and operand
versions in jsonnet/versions.yaml were not validated properly. Now that
the CI job is fixed, all version checks are starting to fail for newly
created PRs.
    
This commit regenerates the versions to make sure they are correct.


fpetkovski:
Closing this since the `versions` job is allowed to fail and we can update all dependencies in a batch

/close

PR #1384, opened 2021-09-16
Add cluster:vmi_request_cpu_cores:sum into Telemetry

erkanerol:
OpenShift Virtualization team wants to know how much CPUs are requested by VMIs to get more insights about VMIs sizes.

Origin of the recording rule: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1530

Signed-off-by: Erkan Erol <eerol@redhat.com>


simonpasquier:
/lgtm

did you get the acknowledgement from @smarterclayton to add this metric?

simonpasquier:
/hold
until explicit approval

dgrisonnet:
We also need to hold until https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1530 is merged.

erkanerol:
@sthaha  @simonpasquier  @dgrisonnet I had a comment in my PR and I need to update the name of the metric. The question: Can we use different prefix other than "cnv"?  "cnv"   refers downstream while we are adding this metric into our upstream repo. Therefore, we want to use something different like "vmis"? Can we use different prefixes according to domain even all of them belong to same organization ("kubevirt") ?

dgrisonnet:
Usually, the prefix that we use downstream is the scope of the recording rule, so it can be something `cluster` or `openshift`. But since you already added a telemetry rule with `cnv` as a prefix, I think it would be better to stick with that downstream. That said, you can always have different naming downstream and upstream, you just have to map the upstream recording rule to a downstream one in: https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/cluster-monitoring-operator/prometheus-rule.yaml.

You could add a new recording rule similar to:
```yaml
    - expr: your_upstream_recording_rule
      record: cnv:vmi_request_cpu_cores:sum
```

Also, if it doesn't really make sense to have the recording rule upstream, you can have it strictly downstream. This is also an option.

erkanerol:
@dgrisonnet Thanks for the explanation. I changed the prefix as "cluster" and the PR in hyperconverged-cluster-operator is merged now.  Could you please review this PR again?

dgrisonnet:
/lgtm

Still waiting for @smarterclayton approval before unholding.

erkanerol:
@smarterclayton May I get a review?

smarterclayton:
Approved for this metric with cardinality one.

erkanerol:
@dgrisonnet is "versions" test broken?

dgrisonnet:
No, it is just an informative job that isn't required.

/label px-approved
/label qe-approved
/label docs-approved

No user-facing changes, this PR adds metrics to telemetry that are only available to Red Hat employees.

erkanerol:
@dgrisonnet  The bot seems trying to fix that job.

PR #1385, opened 2021-09-16
Bug 2004051: changing the condition for error in daemon set creation

prashbnair:

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -998,8 +998,10 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
 				}
 			}
 		}
-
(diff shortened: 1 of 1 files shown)


pkg/client/client.go
```diff
@@ -997,8 +997,9 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
 				}
 			}
 		}
-
-		if d.Status.NumberAvailable != nodeReadyCount {
+		// In BZ-2004051 it has been noticed that the number of available pods can be greater than
+		// the node ready count.
```
arajkumar, line 1001: nit:
Can we add the reason here in the comment instead of pointing to a BZ?

simonpasquier (commented):
/lgtm

PR #1391, opened 2021-09-20
test:e2e: Fixes assertion logic for missing secret

PhilipGough:
EDIT - we were expecting this test to fail since the code prior to the change looked like a no-op but we should merge this fix regardless.


simonpasquier:
/lgtm
/label docs-approved
/label px-approved
/label qe-approved

no need for docs/QE/PX approval since it's only changing the e2e tests and they pass.

PhilipGough:
closing this - this helped us surface a bug but in the end we don't need this assertion and in fact the secret won't be cleaned up - this overall fix is taken care of in https://github.com/openshift/cluster-monitoring-operator/pull/1378 but master isn't technically incorrect

PR #1393, opened 2021-09-21
static auth changes for node-exporter

prashbnair:


pkg/manifests/manifests.go
```diff
@@ -774,6 +775,17 @@ func (f *Factory) NodeExporterPrometheusRule() (*monv1.PrometheusRule, error) {
 	return f.NewPrometheusRule(f.assets.MustNewAssetReader(NodeExporterPrometheusRule))
 }
 
+func (f *Factory) NodeExporterRBACProxySecret() (*v1.Secret, error) {
```
arajkumar, line 778: Should we add this to TestUnconfiguredManifests test?

/hold

prashbnair: Done

arajkumar (approved):
/lgtm

PR #1395, opened 2021-09-24
Bug 2007677: Reinstate dropped metrics

PhilipGough:
Pulls in changes from https://github.com/prometheus-operator/kube-prometheus/pull/1396


simonpasquier (commented):
/lgtm

PR #1396, opened 2021-09-27
chore: update versions according to downstream latest

arajkumar:
Following component versions are updated.,

kubeStateMetrics: 2.1.1
nodeExporter: 1.2.2
prometheus: 2.30.0


simonpasquier (commented):
/label px-approved
/label docs-approved

No change from docs and PX point of view (it's only aligning the resource labels with the versions shipped in OCP).

cc @juzhao @lihongyan1 for QE verification.

sthaha (commented):
/lgtm

PR #1400, opened 2021-09-27
Bug 1808240: prom-label-proxy: set --error-on-replace

jan--f:
This commit enables the new prom-label-proxy flag in order to return an
error in case a label would have been silently replaced. Before the user
would simply see a result that would not respond to the query. Returning
an error follows the principle of least surprise.

### DIFF
--- a/test/e2e/user_workload_monitoring_test.go
+++ b/test/e2e/user_workload_monitoring_test.go
@@ -588,66 +588,94 @@ func assertTenancyForMetrics(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	for _, q := range []string{"up", `up{namespace="should-be-overwritten"}`, fmt.Sprintf(`up{namespace="%s"}`, userWorkloadTestNs)} {
-		t.Logf("Running query %q", q)
+	for _, tc := range []struct {
+		name      string
+		query     string
+		expStatus int
+	}{
+		{
+			name:      "expect to add namespace label",
+			query:     "up",
+			expStatus: http.StatusOK,
+		},
+		{
+			name:      "expect no change and status 200",
+			query:     fmt.Sprintf(`up{namespace="%s"}`, userWorkloadTestNs),
+			expStatus: http.StatusOK,
+		},
+		{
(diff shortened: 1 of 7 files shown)


jan--f:
I'm not sure if the console needs a corresponding change. cc @kyoto for visibility.

simonpasquier:
I think we need a new upstream release of https://github.com/prometheus-community/prom-label-proxy and sync our downstream before we can make the change?

jan--f:
/hold
waiting on https://github.com/prometheus-community/prom-label-proxy/pull/87

test/e2e/user_workload_monitoring_test.go
```diff
@@ -588,8 +588,29 @@ func assertTenancyForMetrics(t *testing.T) {
...
 	}
 
-	for _, q := range []string{"up", `up{namespace="should-be-overwritten"}`, fmt.Sprintf(`up{namespace="%s"}`, userWorkloadTestNs)} {
-		t.Logf("Running query %q", q)
+	for _, tc := range []struct {
+		query     string
+		expStatus int
+	}{
+		{
+			// expect to add namespace label
```
dgrisonnet, line 596: Instead of adding comments to describe each test, it would be better to add a name to the structure and expose it in the form of subtests via [t.Run()](https://pkg.go.dev/testing#T.Run).

test/e2e/user_workload_monitoring_test.go
```diff
@@ -609,11 +630,17 @@ func assertTenancyForMetrics(t *testing.T) {
 				},
 			)
 
-			b, err := client.PrometheusQuery(q)
+			b, err := client.PrometheusQueryWithStatus(tc.query, tc.expStatus)
 			if err != nil {
 				return err
 			}
 
+			if tc.expStatus/100 != 2 {
```
dgrisonnet, line 638:
```suggestion
			if tc.expStatus != http.StatusOK {
```

test/e2e/user_workload_monitoring_test.go
jan--f: Thanks, good point.

PR #1402, opened 2021-09-28
Bug 2007677: Adjust dropped cAdvisor metrics

PhilipGough:
This change drops pod-centric metrics that have a non empty `container` label.  For context it is best to look at the detailed discussion on https://github.com/prometheus-operator/kube-prometheus/pull/1402 and the issues it links to


slashpai:
@PhilipGough Can you add the description in PR

dgrisonnet (commented):
/lgtm

PR #1392, opened 2021-09-21
[WIP] [MON-1850] Replace bearer token by client TLS certificate for metrics scraping

raptorsun:
Do Not Merge. Still in progress.
Remove bearer token from service monitors.

* [ ] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #1405, opened 2021-09-29
static auth changes for telemeter client

prashbnair:


raptorsun:
shall we set `keyFile` and `certFile` in assets/telemeter-client/service-monitor.yaml, too?

raptorsun:
I have the impression we are on a very similar task 😅. 
I'm modifying the telemeter client to use TLS clients certificates, too. 
https://github.com/openshift/cluster-monitoring-operator/pull/1392

prashbnair:
> shall we set keyFile and certFile in assets/telemeter-client/service-monitor.yaml, too?

IMO, they are different tasks. I feel that I should not be adding as it does not have anything to do with static auth.
Please refer to #1393

raptorsun:
> > shall we set keyFile and certFile in assets/telemeter-client/service-monitor.yaml, too?
> 
> IMO, they are different tasks. I feel that I should not be adding as it does not have anything to do with static auth. Please refer to #1393

OK, I will add these in #1392 .

PR #1406, opened 2021-09-29
Remove prom-label-proxy from Prometheus

fpetkovski:


fpetkovski:
/label px-approved
/label docs-approved

Not needed since this is internal refactoring

manifests/0000_50_cluster-monitoring-operator_02-role.yaml
```diff
@@ -374,18 +374,6 @@ rules:
   - /metrics
   verbs:
   - get
-- apiGroups:
```
fpetkovski, line 377: @arajkumar do you have any idea why this is showing up in the diff now, and how it wasn't caught before?

arajkumar: Looks like a rebase problem on your side. Can you try once again?

fpetkovski: This PR is already rebased against latest master. It seems like there is something strange with either the jsonnet code or with how we generate the yaml files. Touching anything in the cluster role in prometheus.jsonnet will also affect the CMO cluster role.

fpetkovski: So it looks like this change is driven by the deletion of these lines: https://github.com/openshift/cluster-monitoring-operator/pull/1406/files#diff-cf7ce97df96f13b990e1fd5d01c1cb685e705bb4b0fcaf039187a92c029b30acL138-L147

fpetkovski: Ok, looks like this works as expected. The CMO role is an aggregation of all operand roles: https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/main.jsonnet#L336

arajkumar: Thats correct, I was actually looking that specific asset without the actual jsonnet changes :)

assets/prometheus-k8s/cluster-role.yaml
```diff
@@ -18,18 +18,6 @@ rules:
   - /metrics
   verbs:
   - get
-- apiGroups:
```
raptorsun, line 21: just curious, why they are no longer useful?

arajkumar: Because this PR also changes kube-rbac-proxy to use static authorizer. So It doesn't need to do SAR for each request to api-server to validate the Bearer token.

CC: @prashbnair

pkg/tasks/prometheus.go
```diff
@@ -125,9 +125,9 @@ func (t *PrometheusTask) Run(ctx context.Context) error {
 		return errors.Wrap(err, "initializing Prometheus RBAC proxy Secret failed")
 	}
 
-	err = t.client.CreateIfNotExistSecret(ctx, rs)
+	err = t.client.CreateOrUpdateSecret(ctx, rs)
```
raptorsun, line 128: why do we need to update the secret here?

fpetkovski: Because the existing secret already contains configuration for authorization through token access reviews. Now we need to update the config to change the authz method.

assets/prometheus-k8s/cluster-role.yaml
fpetkovski: Looks like we need these permissions after role because oauth-proxy also works with SARs.

pkg/tasks/prometheus.go
raptorsun: thanks,  i may need make a similar change to my PR, too 🤔
https://github.com/openshift/cluster-monitoring-operator/pull/1392

assets/prometheus-k8s/service.yaml
```diff
@@ -16,9 +16,9 @@ spec:
   - name: web
```
prashbnair, line 16: Should we not remote this port as the prom-label-proxy is no longer available.

prashbnair: My mistake, closing this out.

simonpasquier (commented):
/lgtm

PR #1409, opened 2021-09-30
Set arguments in a deterministic order

fpetkovski:
When setting arguments for the prometheus-operator deployment, we
flatten them out into a map for easier processing before merging them
into a list again. This causes the arguments to always end up in a
different order and re-triggers a new rollout of the prometheus-operator
deployment.

This PR changes the implementation to preserve the order of the
prometheus-operator arguments so that new pods are rolled out only when
the arguments change.


fpetkovski:
/assign 

@dgrisonnet

jan--f:
This looks good, though lets reference the commit that introduced this for future us. Something like `Partially reverts 857f03cd03af807912d837c0cde3ff18213efafe`?

fpetkovski:
Looks like they decided to merge the revert after all, so now this PR contains the revert of the revert and a new commit on top that fixes the issue. I've also referenced the original commit that introduced the problem.

dgrisonnet:
Great catch @fpetkovski 
/lgtm

Removing the reference to https://bugzilla.redhat.com/show_bug.cgi?id=2008911, since it was fixed by https://github.com/openshift/cluster-monitoring-operator/pull/1407.
/retitle Set arguments in a deterministic order

Also, let's make sure we actively monitor the CI health in the next couple of days.

fpetkovski:
We did not add docs initially for this feature
/label docs-approved

PX did not change from the original implementation
/label px-approved

dgrisonnet:
Leaving QE approval to @juzhao considering that this change was previously the cause of an urgent BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2008911

/assign @juzhao

simonpasquier:
/retitle Revert of revert "Configure prometheus operator TLS based on the cluster APIServer config" with argument order fix

simonpasquier:
/retitle Revert of revert "Configure prometheus operator TLS ..." with argument order fix

dgrisonnet:
Friendly ping to @juzhao @lihongan since this PR requires QE approval.

PR #1410, opened 2021-09-30
Adjust prometheus-adapter queries intervals

dgrisonnet:
Readjust the intervals used in prometheus-adapter queries to take into
account the scrape interval of kubelet, node-exporter and
windows-exporter.

### DIFF
--- a/assets/prometheus-adapter/config-map.yaml
+++ b/assets/prometheus-adapter/config-map.yaml
@@ -7,13 +7,13 @@ data:
         "containerQuery": |
           sum by (<<.GroupBy>>) (
             irate (
-                container_cpu_usage_seconds_total{<<.LabelMatchers>>,container!="",pod!=""}[4m]
+                container_cpu_usage_seconds_total{<<.LabelMatchers>>,container!="",pod!=""}[120s]
             )
           )
         "nodeQuery": |
           sum by (<<.GroupBy>>) (
             1 - irate(
-              node_cpu_seconds_total{mode="idle"}[4m]
+              node_cpu_seconds_total{mode="idle"}[60s]
             )
             * on(namespace, pod) group_left(node) (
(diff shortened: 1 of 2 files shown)


dgrisonnet:
Holding until @simonpasquier approval since it will most likely impact the potential increase of kubelet scrape interval. 
/hold

This change is mostly meant to expose the jsonnet parameters so that we don't forget about them. I am fine even if we have to set a fixed interval to the kubelet parameter for the time being.

simonpasquier:
I see your point but as you mentioned, we want to keep some flexibility around scrape intervals. Lets use the `rangeIntervals` config but account for larger scrape intervals (e.g. keep the `4m`). In practice `irate(...[2m])` and `irate(...[4m])` yield the same result since irate() looks at the last 2 points in the range interval.

dgrisonnet:
@simonpasquier the current approach is accounting for 4 times the scrape interval, so we are already resilient to 50% of scrapes failures. Do you think that this isn't resilient enough based on the current state of the monitoring stack? I am fine to increase that to 75%, but it feels a bit unnecessary considering that the stack should now be HA.

simonpasquier (commented):
I'm all for setting explicit values. Given that irate() only works with the last 2 datapoints from the range interval, a value greater than the usual "4 x scrape interval" rule doesn't really matter.
Thus I would prefer to keep a safety margin and continue to use larger intervals, the main reason being that it might be possible to set larger scrape intervals in the future.

dgrisonnet:
> Thus I would prefer to keep a safety margin and continue to use larger intervals, the main reason being that it might be possible to set larger scrape intervals in the future.

The modification of the scrape interval is exactly the reason why I wanted to have the window depend on the scrape intervals so that we wouldn't have to worry about having queries and scrape intervals out-of-sync.

That said, I agree that changing the value to a greater one doesn't really matter with irate, so if you prefer dealing with explicit values, it is fine by me.

PR #1411, opened 2021-09-30
Change kube-rbac-proxy authentication for UWM Prometheus

fpetkovski:
The UWM Prometheus kube-rbac-proxy currently works with subject access
reviews to verify whether the caller is allowed to access Prometheus endpoints.
    
This commit changes the authentication method to a static one based on
the CN in the certificate of the caller.

### DIFF
--- a/assets/prometheus-user-workload/cluster-role.yaml
+++ b/assets/prometheus-user-workload/cluster-role.yaml
@@ -19,18 +19,6 @@ rules:
   verbs:
   - get
 - apiGroups:
-  - authentication.k8s.io
-  resources:
-  - tokenreviews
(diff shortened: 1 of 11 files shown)


simonpasquier:
This PR is about switching kube-rbac-proxy to use TLS client certificates for authentication rather than removing prom-label-proxy?

fpetkovski:
Ah correct, I mechanically added the same description as for https://github.com/openshift/cluster-monitoring-operator/pull/1406.

/retitle Change kube-rbac-proxy authentication for UWM Prometheus

simonpasquier (commented):
/lgtm

PR #1412, opened 2021-09-30
Add aliases for doc, qe, px approvers

arajkumar:
Instead of referring the respective folks by their GH usernames, add aliases to ease the ping. 


PhilipGough (approved):
/lgtm

OWNERS_ALIASES
```diff
@@ -0,0 +1,9 @@
+# See the OWNERS docs at https://go.k8s.io/owners#owners_aliases
+
+aliases:
+  qe-approvers:
+  - juzhao
```
dgrisonnet, line 5: I think we should also add @lihongyan1

OWNERS_ALIASES
```diff
@@ -0,0 +1,10 @@
...
+  - lihongyan1
+  doc-approvers:
+  - rkratky
```
simonpasquier, line 8: we need @bburt-rh instead

arajkumar: @simonpasquier , looks like @bburt-rh is not yet part of openshift GH org. Should I revert back to @ rkratky?

OWNERS_ALIASES
```diff
@@ -0,0 +1,10 @@
+# See the OWNERS docs at https://go.k8s.io/owners#owners_aliases
+
+aliases:
+  qe-approvers:
+  - juzhao
+  - lihongyan1
+  doc-approvers:
+  - bburt-rh
```
simonpasquier, line 8: after discussing with the docs team, we agreed that the engineering team could apply the docs-approved label themselves as long as the epic has already been ack-ed by the docs team.

PR #1416, opened 2021-10-06
MON-1666: CMO deployment: pass enabled-remote-write

jan--f:
in order to switch telemeter over to Prometheus remote write.

### DIFF
--- a/manifests/0000_50_cluster-monitoring-operator_05-deployment-ibm-cloud-managed.yaml
+++ b/manifests/0000_50_cluster-monitoring-operator_05-deployment-ibm-cloud-managed.yaml
@@ -63,6 +63,7 @@ spec:
         - -images=prom-label-proxy=quay.io/openshift/origin-prom-label-proxy:latest
         - -images=k8s-prometheus-adapter=quay.io/openshift/origin-k8s-prometheus-adapter:latest
         - -images=thanos=quay.io/openshift/origin-thanos:latest
+        - -enabled-remote-write
         env:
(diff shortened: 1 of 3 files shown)


simonpasquier:
I've noticed from the logs that Prometheus enables sending of metadata by default but I presume that we don't want this for telemeter so it should be turned off explicitly in the [RemoteWrite](https://github.com/openshift/cluster-monitoring-operator/blob/53f1eb70e19e86f8a75e977627f5e2c8aef91d17/pkg/manifests/manifests.go#L1334) spec.

jan--f:
> I've noticed from the logs that Prometheus sends metadata by default but I presume that we don't want this for telemeter. I believe that it should be turned off explicitly in the [RemoteWrite](https://github.com/openshift/cluster-monitoring-operator/blob/53f1eb70e19e86f8a75e977627f5e2c8aef91d17/pkg/manifests/manifests.go#L1334) spec.

Yeak makes sense. Tbh I'm not 100% sure yet what the impact is and whether telemeter can actually make use of this metadata. I'll investigate more, but until then lets turn it off.

jan--f:
2021-10-14-152738_1620x898
Going by the `prometheus_remote_storage_bytes_total` metric of prometheus, it seems like the requests should be around 70kB large (we have roughly 500kB over 7 requests and ~413kB over 6 requests).

Should we start with a 128kB request limit on the telemeter side?

jan--f:
/close
Needs further research

PR #1419, opened 2021-10-07
jsonnet/components/control-plane.libsonnet: decrease etcd interval

lilic:

etcd scrape interval is 30 seconds, which sometimes doesn't allow us to
catch every metric we need, especially during bootstrap process around
learners.


dgrisonnet:
I think the CPU usage increase is pretty much negligible in most cases. Even for SNO we would be looking at around 5-10 millicores of CPU usage increase which is a lot for this use case, but considering the amount of additional information that we would get from etcd, I think it is OK.
/lgtm

Leaving approval to @simonpasquier.
/hold

PR #1421, opened 2021-10-07
Bug 2011698: Allow namespace label in metric allow list

slashpai:
Customer need this label for custom alerting which was available in previous OCP versions which went away with KSM v2 update. Adding it back by allowing in metric label allow list


PhilipGough (approved):
/lgtm

PR #1426, opened 2021-10-12
KSM: add kube_*_annotations to --metric-denylist arg

jan--f:
These metrics were added in
https://github.com/kubernetes/kube-state-metrics/pull/1468/ but for us
don't add any valuable information.


jan--f:
This filters metrics in an upcoming KSM release, that won't add any information, no user-visible changes.
/label docs-approved
/label px-approved
/label qe-approved

PR #1427, opened 2021-10-12
KSM: use new --use-apiserver-cache argument

jan--f:
This cli arg was added in
https://github.com/kubernetes/kube-state-metrics/pull/1548 and is
available since KSM v2.2.0.


jan--f:
/close
Closing this for now, needs more discussion and potential exposing to admins.

PR #1430, opened 2021-10-13
add E2E test checking image registry is well replaced for containers.

raptorsun:
This PR adds a test to verify image registry of each container is the same as cluster-monitoring-operator has. 

### DIFF
--- a/test/e2e/framework/helpers.go
+++ b/test/e2e/framework/helpers.go
@@ -73,6 +73,25 @@ func (f *Framework) MustGetStatefulSet(t *testing.T, name, namespace string) *ap
 	return statefulSet
 }
 
+// MustGetPods return all pods from `namespace` within 5 minutes or fail
+func (f *Framework) MustGetPods(t *testing.T, namespace string) *v1.PodList {
+	t.Helper()
+	var pods *v1.PodList
+	err := wait.Poll(time.Second, 5*time.Minute, func() (bool, error) {
+		pl, err := f.KubeClient.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return false, nil
+		}
+
+		pods = pl
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("failed to get pods in namespace %s - %s", namespace, err.Error())
+	}
+	return pods
+}
+
 func ensureCreatedByTestLabel(obj metav1.Object) {
 	// only add the label if it doesn't exist yet, leave existing values
 	// untouched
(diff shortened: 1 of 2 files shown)


test/e2e/framework/helpers.go
```diff
@@ -73,6 +73,25 @@ func (f *Framework) MustGetStatefulSet(t *testing.T, name, namespace string) *ap
 	return statefulSet
 }
 
+// MustGetDeployment `name` from `namespace` within 5 minutes or fail
+func (f *Framework) MustGetDeployment(t *testing.T, name, namespace string) *appsv1.Deployment {
+	t.Helper()
+	var deployment *appsv1.Deployment
+	err := wait.Poll(time.Second, 5*time.Minute, func() (bool, error) {
+		ss, err := f.KubeClient.AppsV1().Deployments(namespace).Get(ctx, name, metav1.GetOptions{})
```
dgrisonnet, line 81: nit: ss means statefulset here, so we might want to find another name

simonpasquier:
This is missing the node_exporter daemonset :)
Instead of checking high-level resources (e.g. deployment, statefulset, ...), I would check that all pods from openshift-monitoring and openshift-user-workload-monitoring use correct images (meaning that UWM needs to be enabled).

> If that assumption is correct, would it make sense to also verify that the images passed down to CMO via the --images option are also used

The trigger for creating this e2e test was when reviewing #1392 (see https://github.com/openshift/cluster-monitoring-operator/pull/1392#pullrequestreview-776018095). I think that the main risk is to forget overriding a particular image. Having a typo in the  image name should already fail the e2e tests since starting the container with wrong CLI arguments would break.  

@raptorsun can you add a commit on top that removes an image override (for instance [here](https://github.com/openshift/cluster-monitoring-operator/blob/ce2e4a2acc738dbb5db4aeb9707610dda4d69dad/pkg/manifests/manifests.go#L477)), just to ensure that the test would fail in the future? Once it's verified, you can revert it.

raptorsun:
Test is working :)  
Test fails when image registry of grafana container differs from CMO container:
1. [commit ](https://github.com/openshift/cluster-monitoring-operator/pull/1430/commits/b1249d7eab51b84e1b4c7a107bc963af910bac4d)switch grafana to default container
2. [failing test](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/1430/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-operator/1449013664475516928) detecting the change above

simonpasquier:
/lgtm

/label docs-approved
/label px-approved
/label qe-approved

This change adds a new end-to-end test and doesn't modify the behavior of the operator itself.

PR #1436, opened 2021-10-18
Bug 2012039: Allow namespace label in metric allow list

slashpai:
Customer need this label for custom alerting which was available in previous OCP versions which went away with KSM v2 update. Adding it back by allowing in metric label allow list


juzhao:
tested with the PR, issue is fixed, see from https://bugzilla.redhat.com/show_bug.cgi?id=2012039#c2
/lgtm

PR #1439, opened 2021-10-19
Bug 2015386: jsonnet: Add PodDisruptionBudget to KSM metric allow list

PhilipGough:


PhilipGough:
We can merge this, but it will have no effect until https://github.com/kubernetes/kube-state-metrics/pull/1623 is pulled in by CMO.

PR #1442, opened 2021-10-19
WIP: Add required labels on components

slashpai:


simonpasquier:
I think that we should update the top-level `commonLabels` to include `'app.kubernetes.io/managed-by': 'cluster-monitoring-operator'`, then make sure that we have common labels defined for each component and apply them to every linked resource (like we do with thanos-querier for instance).

https://github.com/openshift/cluster-monitoring-operator/blob/03557bb5dc9e50c52e2fead842e6c1b6b9aeea65/jsonnet/main.jsonnet#L71-L74

https://github.com/openshift/cluster-monitoring-operator/blob/03557bb5dc9e50c52e2fead842e6c1b6b9aeea65/jsonnet/components/thanos-querier.libsonnet#L17-L26

@arajkumar @dgrisonnet does it make sense from your POV?

jsonnet/components/alertmanager.libsonnet
```diff
@@ -13,6 +13,10 @@ function(params)
         name: 'alertmanager-trusted-ca-bundle',
         namespace: cfg.namespace,
         labels: {
+          'app.kubernetes.io/name': 'alertmanager-trusted-ca-bundle',
+          'app.kubernetes.io/managed-by': 'cluster-monitoring-operator',
+          'app.kubernetes.io/part-of': 'openshift-monitoring',
+          'app.kubernetes.io/component': 'configmap',
```
dgrisonnet, line 19: Instead of redefining values for `component` and `name`, we should reuse the ones from upstream: `https://github.com/prometheus-operator/kube-prometheus/blob/main/jsonnet/kube-prometheus/components/alertmanager.libsonnet#L10-L15`. Generally the name shouldn't be the one of the resource but the name of the component so `alertmanager` in this case and the component should be the `role` of the component in the cluster so `alert-router` here.

What could be done to make this change more generic is to make the `app.kubernetes.io/part-of` and `app.kubernetes.io/part-of` labels configurable at the top level of kube-prometheus and then just apply the upstream commonLabels to the new resource that we introduce downstream.

PR #1443, opened 2021-10-20
pkg/client/client.go: Remove dead code

slashpai:
Removing the code that is not been used

cc: @simonpasquier @dgrisonnet 

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -52,7 +52,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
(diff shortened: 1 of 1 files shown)


pkg/client/client.go
```diff
@@ -776,12 +775,9 @@ func (c *Client) CreateOrUpdateDeployment(ctx context.Context, dep *appsv1.Deplo
 			if err != nil {
 				return errors.Wrap(err, "deleting Deployment object failed")
 			}
-			retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-				return c.CreateDeployment(ctx, required)
-			})
-
-			if retryErr != nil {
-				return errors.Wrap(retryErr, "creating Deployment object failed after update failed")
+			err = c.CreateDeployment(ctx, required)
```
arajkumar, line 778: @slashpai May I know why `RetryOnConflict` is no longer useful?

dgrisonnet: The exact background can be found here: https://github.com/openshift/cluster-monitoring-operator/pull/1333#issuecomment-902802506

simonpasquier:
/label qe-approved
/label px-approved
/label docs-approved

The change is only removing code that would never be used: a create operation should never fail with `StatusReasonConflict`.

PR #1444, opened 2021-10-20
Bug 2013920: Bump jsonnet dependencies to latest

arajkumar:
This PR bumps jsonnet dependencies to latest which also fixes Bug 2013920.

### DIFF
--- a/assets/control-plane/prometheus-rule.yaml
+++ b/assets/control-plane/prometheus-rule.yaml
@@ -531,6 +531,17 @@ spec:
       labels:
         namespace: kube-system
         severity: critical
+  - name: kubernetes-system-kube-proxy
+    rules:
+    - alert: KubeProxyDown
+      annotations:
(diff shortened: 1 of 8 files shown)


assets/control-plane/prometheus-rule.yaml
```diff
@@ -531,6 +531,17 @@ spec:
...
+    rules:
+    - alert: KubeProxyDown
+      annotations:
+        description: KubeProxy has disappeared from Prometheus target discovery.
+        summary: Target disappeared from Prometheus target discovery.
+      expr: |
+        absent(up{job="kube-proxy"} == 1)
+      for: 15m
+      labels:
+        severity: critical
```
simonpasquier, line 544: we need approval from @openshift/sre-alert-sme whether they feel that it's a valid critical alert. If so we need a runbook in openshift/runbook.
Also the alert needs a static `namespace: kube-system` label.

PR #1445, opened 2021-10-20
Switch hosted control plane check to controlPlaneTopology

csrwng:
Before 4.9, the only platform that had an external/hosted control plane
was IBM cloud. However, starting in 4.10, IBM cloud will also be
available with a self-hosted control plane, and other platforms will
also support an external control plane via Hypershift. This changes the
hosted control plane check to use the
Infrastructure.Status.ControlPlaneTopology field instead of the
platform to match the latest state.


rtheis (approved):
/lgtm

PR #1447, opened 2021-10-21
static auth changes for prometheus operator

prashbnair:


simonpasquier (commented):
/lgtm

PR #1449, opened 2021-10-21
Bug 2016352: Fix cert injection

jan--f:


jan--f:
We probably don't even have to create an empty `data` property, will test this.

jan--f:
/hold
There is something weird going on with the CI jobs.

jan--f:
CI is borked, need to open a fresh pr :/

PR #1461, opened 2021-10-29
Bug 2018455: Keep container_fs_usage_bytes metric

PhilipGough:
Cherry pick of #1460 


juzhao:
checked with the PR, could see the project and pod filesystem usage on UI, see https://bugzilla.redhat.com/show_bug.cgi?id=2018455#c1
/lgtm

PR #1466, opened 2021-11-03
pkg: add tlsSecurityProfile to kube-rbac-proxy in prometheus-operator

slashpai:
This PR covers tls security profile for kube-rbac-proxy in prometheus-operator in platform and uwm

[MON-1651](https://issues.redhat.com/browse/MON-1651)

### DIFF
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -759,6 +759,7 @@ func TestPrometheusOperatorConfiguration(t *testing.T) {
 		"prometheus-operator":        "docker.io/openshift/origin-prometheus-operator:latest",
 		"prometheus-config-reloader": "docker.io/openshift/origin-prometheus-config-reloader:latest",
 		"configmap-reloader":         "docker.io/openshift/origin-configmap-reloader:latest",
+		"kube-rbac-proxy":            "docker.io/openshift/origin-kube-rbac-proxy:latest",
 	})
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


slashpai:
cc: @juzhao

pkg/manifests/manifests_test.go
```diff
@@ -779,26 +780,45 @@ func TestPrometheusOperatorConfiguration(t *testing.T) {
...
 		}
 
-		if strings.HasPrefix(d.Spec.Template.Spec.Containers[0].Args[i], PrometheusOperatorWebTLSMinTLSVersionFlag) {
-			prometheusWebTLSVersionArg = d.Spec.Template.Spec.Containers[0].Args[i]
+		if container.Name == "kube-rbac-proxy" {
+			if d.Spec.Template.Spec.Containers[i].Image != "docker.io/openshift/origin-kube-rbac-proxy:latest" {
+				t.Fatalf("%s image incorrectly configured", container.Name)
+			}
+
+			for j := range d.Spec.Template.Spec.Containers[i].Args {
```
fpetkovski, line 813: Should we extract a function `containsArg` or something similar for improved readability?

PhilipGough: +1

slashpai: Not sure if we need function. Looks like this could help with readability. WDYT?

```go
containsArgs :=  d.Spec.Template.Spec.Containers[i].Args
for j := range containsArgs {
```

fpetkovski: This still leaves the deeply nested for + if statements. Even a double nested for loop is a code-smell, and here we have 4-5 levels of nesting. Extracting the for loop into a single function will reduce the nesting depth. 

In addition, we are repeating this code chunk too often which is error prone, especially because it is part of the tests.

PhilipGough: @slashpai in case it is useful, we already have a similar extraction in the e2e tests - https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/config_test.go#L708

slashpai: Ya that makes sense. I will update

slashpai: addressed

slashpai:
@fpetkovski @PhilipGough can you review again? :)

pkg/manifests/manifests_test.go
```diff
@@ -779,26 +780,29 @@ func TestPrometheusOperatorConfiguration(t *testing.T) {
...
+			if container.Image != "docker.io/openshift/origin-prometheus-operator:latest" {
+				t.Fatalf("%s image incorrectly configured", container.Name)
+			}
 
-		if strings.HasPrefix(d.Spec.Template.Spec.Containers[0].Args[i], PrometheusOperatorWebTLSMinTLSVersionFlag) {
-			prometheusWebTLSVersionArg = d.Spec.Template.Spec.Containers[0].Args[i]
+			if expectContainerArg(d.Spec.Template.Spec.Containers, PrometheusConfigReloaderFlag+"docker.io/openshift/origin-prometheus-config-reloader:latest") != "" {
+				prometheusReloaderFound = true
+			}
+			prometheusWebTLSCipherSuitesArg = expectContainerArg(d.Spec.Template.Spec.Containers, PrometheusOperatorWebTLSCipherSuitesFlag)
```
fpetkovski, line 798: Thanks for the change @slashpai, this looks much better now. I am just wondering if we should call the function something more descriptive. Is the function supposed to make an assertion, or return the value for the field? If it's the former then it should be `assertContainerArgEquals`, otherwise it should be something like `getContainerArgValue`

PhilipGough: Same comment. I think with the current naming I would just expect the test to fail if the assertion was not met. Renaming to `getContainerArgValue` seems more appropriate.

Also, looking at the code, we seem to already range over the containers so perhaps `getContainerArgValue` does not need to take a slice

slashpai: I updated the func name to `getContainerArgValue()`
`d.Spec.Template.Spec.Containers` is of type of `[]v1.Container` so that should be the type right?

slashpai: @fpetkovski @PhilipGough Please review :)

arajkumar:
/retitle pkg: add tlsSecurityProfile to kube-rbac-proxy in prometheus-operator

pkg/manifests/manifests_test.go
```diff
@@ -779,26 +780,29 @@ func TestPrometheusOperatorConfiguration(t *testing.T) {
...
+			if getContainerArgValue(d.Spec.Template.Spec.Containers, PrometheusConfigReloaderFlag+"docker.io/openshift/origin-prometheus-config-reloader:latest") != "" {
+				prometheusReloaderFound = true
+			}
+			prometheusWebTLSCipherSuitesArg = getContainerArgValue(d.Spec.Template.Spec.Containers, PrometheusOperatorWebTLSCipherSuitesFlag)
+			prometheusWebTLSVersionArg = getContainerArgValue(d.Spec.Template.Spec.Containers, PrometheusOperatorWebTLSMinTLSVersionFlag)
+		case "kube-rbac-proxy":
+			if container.Image != "docker.io/openshift/origin-kube-rbac-proxy:latest" {
+				t.Fatalf("%s image incorrectly configured", container.Name)
+			}
+			kubeRbacProxyTLSCipherSuitesArg = getContainerArgValue(d.Spec.Template.Spec.Containers, KubeRbacProxyTLSCipherSuitesFlag)
```
fpetkovski, line 804: Isn't this going to get the value for any container, not the just for kube-rbac-proxy?

slashpai: correct

fpetkovski: But the variable says `kubeRbacProxyTLSCipherSuitesArg`. It also means that the tests even will pass if we set the argument in the wrong container.

slashpai: since we are trying get only argument value for `KubeRbacProxyTLSCipherSuitesFlag variable is named as `kubeRbacProxyTLSCipherSuitesArg`. Test will not pass since we compare expected value and this value. I had issue initially since it was not extracting argument correctly but once I fixed the function it was matching with expected

slashpai: This line https://github.com/openshift/cluster-monitoring-operator/pull/1466/files#diff-b388a528a3665a46f4dfa8eb4a8c02de116acb8f85e4611f12f11dc337d7f7b5R826

fpetkovski: The value of `KubeRbacProxyTLSCipherSuitesFlag` is `--tls-cipher-suites=`. The prometheus adapter also uses the same arg name[1]. Furthermore, the newly defined helper function is called `getContainerArgValue` which indicates that it will get the argument value for a specific container. 

I would suggest that we add the container name as an argument to the function and resolve any ambiguity and potential misuses. It will also make the function much more reusable because it will be safe to use it to get the arg value for individual containers.

[1] https://github.com/openshift/cluster-monitoring-operator/blob/361da15b129da47a19ef4f2ae7c5b28eebabdd8c/pkg/manifests/manifests.go#L261

slashpai: Ok that sounds good. I will update

slashpai: updated. Please review :)

fpetkovski:
nice work :+1: 

/lgtm

PhilipGough (approved):
/lgtm

PR #1350, opened 2021-08-30
MON-1708: Enforce label scrape limits in UWM

slashpai:
[MON-1708](https://issues.redhat.com/browse/MON-1708) -> Enforce label scrape limits. This feature is available in prometheus-operator 0.50 only

* [x] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #1467, opened 2021-11-05
Add bodysize limit for metric scraping.

raptorsun:
* [ ] No user-facing changes, so no entry in CHANGELOG was needed.

This PR adds an option to CMO config map to activate bodysize limit on metrics scraping, which can prevent potential OOM problems when scraping metric endpoints responding with an oversized HTTP body.

The dependency upgrade PR is PR #1468 . This functionality requires Prometheus-Operator 0.51+. So I upgrade it to 0.52.

Here is the [JIRA ticket](https://issues.redhat.com/browse/MON-1838).

FIeld prometheusK8s.enforcedBodySizeLimit to CMO ConfigMap, accepting the size format from Prometheus:
 - Empty value and 0 mean no limit 
 - string "automatic" for automatically set the limit according the cluster pod capacity.
 -  [0-9]+[A-Z][a-z]*B for a customized size.

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -16,19 +16,33 @@ package manifests
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 
 	configv1 "github.com/openshift/api/config/v1"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	poperator "github.com/prometheus-operator/prometheus-operator/pkg/operator"
 	v1 "k8s.io/api/core/v1"
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
+	"k8s.io/klog/v2"
 )
 
 const (
 	DefaultRetentionValue = "15d"
+
+	// Limit the body size from scrape queries
+	// Assumptions: one node has in average 110 pods, each pod exposes 400 metrics, each metric is expressed by on average 250 bytes.
+	// 1.5x the size for a safe margin,
+	// minimal HA requires 3 nodes. it rounds to 47.2 MB (49,500,000 Bytes).
+	minimalSizeLimit = 3 * 1.5 * 110 * 400 * 250
+
+	// A value of Prometheusk8s.enforceBodySizeLimit,
+	// meaning the limit will be automatically calculated based on cluster capacity.
+	automaticBodySizeLimit = "automatic"
 )
 
 type Config struct {
@@ -186,6 +200,12 @@ type PrometheusK8sConfig struct {
 	TelemetryMatches    []string                             `json:"-"`
 	AlertmanagerConfigs []AdditionalAlertmanagerConfig       `json:"additionalAlertmanagerConfigs"`
 	QueryLogFile        string                               `json:"queryLogFile"`
+	/* EnforcedBodySizeLimit accept 3 kind of values:
+	 * 1. empty value: no limit
+	 * 2. a value in Prometheus size format, e.g. "64MB"
+	 * 3. string "automatic", which means the limit will be automatically calculated based on cluster capacity.
+	 */
+	EnforcedBodySizeLimit string `json:"enforcedBodySizeLimit,omitempty"`
 }
 
 type AdditionalAlertmanagerConfig struct {
@@ -329,7 +349,6 @@ func NewConfig(content io.Reader) (*Config, error) {
 	res := &c
 	res.applyDefaults()
 	c.UserWorkloadConfiguration = NewDefaultUserWorkloadMonitoringConfig()
-
 	return res, nil
 }
 
@@ -475,6 +494,42 @@ func (c *Config) HTTPSProxy() string {
 // NoProxy implements the ProxyReader interface.
 func (c *Config) NoProxy() string {
 	return c.ClusterMonitoringConfiguration.HTTPConfig.NoProxy
+}
+
+// PodCapacityReader returns the maximum number of pods that can be scheduled in a cluster.
+type PodCapacityReader interface {
+	PodCapacity(context.Context) (int, error)
+}
+
+func (c *Config) LoadEnforcedBodySizeLimit(pcr PodCapacityReader, ctx context.Context) error {
+	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit == "" {
+		return nil
+	}
+
+	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit == automaticBodySizeLimit {
+		podCapacity, err := pcr.PodCapacity(ctx)
+		if err != nil {
+			return fmt.Errorf("error fetching pod capacity: %v", err)
+		}
+		c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit = calculateBodySizeLimit(podCapacity)
+		return nil
+	}
+
+	return poperator.ValidateSizeField(c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit)
+
+}
+
+func calculateBodySizeLimit(podCapacity int) string {
+	const samplesPerPod = 400 // 400 samples per pod
+	const sizePerSample = 200 // 200 Bytes
+
+	bodySize := podCapacity * samplesPerPod * sizePerSample
+	if bodySize < minimalSizeLimit {
+		klog.Infof("Calculated scrape body size limit %v is too small, using default value %v instead", bodySize, minimalSizeLimit)
+		bodySize = minimalSizeLimit
+	}
+
+	return fmt.Sprintf("%dMB", int(math.Ceil(float64(bodySize)/(1024*1024))))
 }
 
 func NewConfigFromString(content string) (*Config, error) {
--- a/pkg/manifests/config_test.go
+++ b/pkg/manifests/config_test.go
@@ -16,6 +16,8 @@ package manifests
 
 import (
 	"bytes"
+	"context"
+	"errors"
 	"io/ioutil"
 	"os"
 	"testing"
@@ -192,7 +194,7 @@ func TestHttpProxyConfig(t *testing.T) {
 	conf := `http:
   httpProxy: http://test.com
   httpsProxy: https://test.com
-  noProxy: https://example.com	
+  noProxy: https://example.com
 `
 
 	c, err := NewConfig(bytes.NewBufferString(conf))
@@ -232,5 +234,102 @@ func TestHttpProxyConfig(t *testing.T) {
 		if test.got() != test.expect {
 			t.Errorf("testcase %d: expected enabled %s, got %s", i, test.expect, test.got())
 		}
+	}
+}
+
+type fakePodCapacity struct {
+	capacity int
+	err      error
+}
+
+func (fpc *fakePodCapacity) PodCapacity(context.Context) (int, error) {
+	return fpc.capacity, fpc.err
+}
+
+func TestLoadEnforcedBodySizeLimit(t *testing.T) {
+
+	mc_10 := fakePodCapacity{capacity: 10, err: nil}
+	mc_1000 := fakePodCapacity{capacity: 1000, err: nil}
+	mc_err := fakePodCapacity{capacity: 1000, err: errors.New("error")}
+	for _, tt := range []struct {
+		name                string
+		config              string
+		expectBodySizeLimit string
+		expectError         bool
+		pcr                 PodCapacityReader
+	}{
+		{
+			name:                "empty config",
+			config:              "",
+			expectBodySizeLimit: "",
+			expectError:         false,
+			pcr:                 &mc_10,
+		},
+		{
+			name:                "disable body size limit",
+			config:              `{"prometheusK8s": {"enforcedBodySizeLimit": "0"}}`,
+			expectBodySizeLimit: "0",
+			expectError:         false,
+			pcr:                 &mc_10,
+		},
+		{
+			name:                "normal size format",
+			config:              `{"prometheusK8s": {"enforcedBodySizeLimit": "10KB"}}`,
+			expectBodySizeLimit: "10KB",
+			expectError:         false,
+			pcr:                 &mc_10,
+		},
+		{
+			name:                "invalid size format",
+			config:              `{"prometheusK8s": {"enforcedBodySizeLimit": "10EUR"}}`,
+			expectBodySizeLimit: "",
+			expectError:         true,
+			pcr:                 &mc_10,
+		},
+		{
+			name:                "automatic deduced limit: error when getting pods capacity",
+			config:              `{"prometheusK8s": {"enforcedBodySizeLimit": "automatic"}}`,
+			expectBodySizeLimit: "",
+			expectError:         true,
+			pcr:                 &mc_err,
+		},
+		{
+			name:                "automatically deduced limit: minimal 48MB",
+			config:              `{"prometheusK8s": {"enforcedBodySizeLimit": "automatic"}}`,
+			expectBodySizeLimit: "48MB",
+			expectError:         false,
+			pcr:                 &mc_10,
+		},
+		{
+			name:                "automatically deduced limit: larger than minimal 16MB",
+			config:              `{"prometheusK8s": {"enforcedBodySizeLimit": "automatic"}}`,
+			expectBodySizeLimit: "77MB",
+			expectError:         false,
+			pcr:                 &mc_1000,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			c, err := NewConfigFromString(tt.config)
+			if err != nil {
+				t.Fatalf("config parsing error")
+			}
+
+			err = c.LoadEnforcedBodySizeLimit(tt.pcr, context.Background())
+			if tt.expectError {
+				if err == nil {
+					t.Fatalf("expected error, got nil")
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("expected no error, got error %v", err)
+			}
+
+			if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit != tt.expectBodySizeLimit {
+				t.Fatalf("incorrect EnforcedBodySizeLimit is set: got %s, expected %s",
+					c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit,
+					tt.expectBodySizeLimit)
+			}
+		})
 	}
 }
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1674,6 +1674,10 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 		p.Spec.Secrets = append(p.Spec.Secrets, getAdditionalAlertmanagerSecrets(f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.AlertmanagerConfigs)...)
 	}
 
+	if f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit != "" {
+		p.Spec.EnforcedBodySizeLimit = f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit
+	}
+
 	return p, nil
 }
 
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -15,6 +15,7 @@
 package manifests
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/url"
@@ -1533,8 +1534,13 @@ ingress:
 		t.Fatal("Prometheus image is not configured correctly")
 	}
 
+	if p.Spec.EnforcedBodySizeLimit != "" {
+		t.Fatal("EnforcedBodySizeLimit is not set to empty by default")
+	}
+
 	kubeRbacProxyTLSCipherSuitesArg := ""
 	kubeRbacProxyMinTLSVersionArg := ""
+
 	for _, container := range p.Spec.Containers {
 		switch container.Name {
 		case "prometheus-proxy":
@@ -1809,6 +1815,43 @@ func TestPrometheusRetentionConfigs(t *testing.T) {
 	}
 }
 
+func TestPrometheusK8sConfigurationBodySizeLimit(t *testing.T) {
+	pcr := &fakePodCapacity{
+		capacity: 1000,
+		err:      nil,
+	}
+	ctx := context.Background()
+
+	c, err := NewConfigFromString(`
+prometheusK8s:
+    enforcedBodySizeLimit: "10MB"
+  `)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = c.LoadEnforcedBodySizeLimit(pcr, ctx)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	f := NewFactory("openshift-monitoring", "openshift-user-workload-monitoring", c, defaultInfrastructureReader(), &fakeProxyReader{}, NewAssets(assetsPath), &APIServerConfig{}, nil)
+	p, err := f.PrometheusK8s(
+		&v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+		&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// the body size limit value is not set at configuration parsing time.
+	if p.Spec.EnforcedBodySizeLimit != "10MB" {
+		t.Fatalf("EnforcedBodySizeLimit is not configured correctly, expected 10MB but got %v", p.Spec.EnforcedBodySizeLimit)
+	}
+
+}
 func TestPrometheusK8sAdditionalAlertManagerConfigsSecret(t *testing.T) {
 	testCases := []struct {
 		name           string
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -829,6 +829,12 @@ func (o *Operator) Config(ctx context.Context, key string) (*manifests.Config, e
 	}
 	o.userWorkloadEnabled = *c.ClusterMonitoringConfiguration.UserWorkloadEnabled
 
+	err = c.LoadEnforcedBodySizeLimit(o.client, ctx)
+	if err != nil {
+		c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit = ""
+		klog.Warningf("Error loading enforced body size limit, no body size limit will be enforced: %v", err)
+	}
+
 	// Only fetch the token and cluster ID if they have not been specified in the config.
 	if c.ClusterMonitoringConfiguration.TelemeterClientConfig.ClusterID == "" || c.ClusterMonitoringConfiguration.TelemeterClientConfig.Token == "" {
 		err := c.LoadClusterID(func() (*configv1.ClusterVersion, error) {
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -295,3 +295,50 @@ func remoteWriteCheckMetrics(ctx context.Context, t *testing.T, promClient *fram
 		)
 	}
 }
+
+func TestBodySizeLimit(t *testing.T) {
+	const (
+		bodySizeLimitSmall = "1MB"
+	)
+
+	cm := f.MustGetConfigMap(t, clusterMonitorConfigMapName, f.Ns)
+	cmBackup := cm.DeepCopy()
+	cmBackup.ObjectMeta.ResourceVersion = ""
+	cmBackup.ObjectMeta.UID = ""
+	cmBackup.ObjectMeta.CreationTimestamp = metav1.Time{}
+
+	restoreConfig := func() {
+		f.MustCreateOrUpdateConfigMap(t, cmBackup)
+	}
+
+	defer restoreConfig()
+
+	f.PrometheusK8sClient.WaitForQueryReturn(
+		t, 5*time.Minute, `ceil(sum(increase(prometheus_target_scrapes_exceeded_body_size_limit_total{job="prometheus-k8s"}[5m])))`,
+		func(v int) error {
+			if v > 0 {
+				return fmt.Errorf("expected prometheus_target_scrapes_exceeded_body_size_limit_total does not increase up but got %v increase in last 5 minutes", v)
+			}
+
+			return nil
+		},
+	)
+
+	data := fmt.Sprintf(`prometheusK8s:
+  logLevel: debug
+  enforcedBodySizeLimit: %s
+`, bodySizeLimitSmall)
+	f.MustCreateOrUpdateConfigMap(t, configMapWithData(t, data))
+
+	f.PrometheusK8sClient.WaitForQueryReturn(
+		t, 5*time.Minute, `ceil(sum(increase(prometheus_target_scrapes_exceeded_body_size_limit_total{job="prometheus-k8s"}[5m])))`,
+		func(v int) error {
+			if v == 0 {
+				return fmt.Errorf("expected prometheus_target_scrapes_exceeded_body_size_limit_total to increase but no increase is observed in last 5 minutes")
+			}
+
+			return nil
+		},
+	)
+
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - [#1630](https://github.com/openshift/cluster-monitoring-operator/pull/1630) Expose retention size settings for UWM Prometheus
 - [#1640](https://github.com/openshift/cluster-monitoring-operator/pull/1640) Deploy standalone admission webhook for HA.
 - [#1651](https://github.com/openshift/cluster-monitoring-operator/pull/1651) Allow retention to be configurable for Thanos-Ruler in UWM
+- [#1467](https://github.com/openshift/cluster-monitoring-operator/pull/1467) Add bodysize limit for metric scraping
 
 ## 4.10
 
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1524,6 +1524,24 @@ func (c *Client) DeleteRole(ctx context.Context, role *rbacv1.Role) error {
 	return err
 }
 
+func (c *Client) PodCapacity(ctx context.Context) (int, error) {
+	nodes, err := c.kclient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return 0, err
+	}
+	var podCapacityTotal int64
+	for _, node := range nodes.Items {
+		podsCount, succeeded := node.Status.Capacity.Pods().AsInt64()
+		if !succeeded {
+			klog.Warningf("Cannot get pod capacity from node: %s. Error: %v", node.Name, err)
+			continue
+		}
+		podCapacityTotal += podsCount
+	}
+
+	return int(podCapacityTotal), nil
+}
+
 // mergeMetadata merges labels and annotations from `existing` map into `required` one where `required` has precedence
 // over `existing` keys and values. Additionally function performs filtering of labels and annotations from `exiting` map
 // where keys starting from string defined in `metadataPrefix` are deleted. This prevents issues with preserving stale
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -25,6 +25,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
@@ -1835,4 +1836,51 @@ func TestCreateOrUpdateValidatingWebhookConfiguration(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestPodCapacity(t *testing.T) {
+	ctx := context.Background()
+	node1 := v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "node1",
+		},
+		Status: v1.NodeStatus{
+			Capacity: v1.ResourceList{
+				v1.ResourcePods: resource.MustParse("100"),
+			},
+		},
+	}
+	node2 := v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "node2",
+		},
+		Status: v1.NodeStatus{
+			Capacity: v1.ResourceList{
+				v1.ResourcePods: resource.MustParse("50"),
+			},
+		},
+	}
+	nodeList := v1.NodeList{
+		Items: []v1.Node{
+			node1,
+			node2,
+		},
+	}
+	t.Run("sum 2 nodes pod capacity", func(st *testing.T) {
+
+		c := Client{
+			kclient: fake.NewSimpleClientset(nodeList.DeepCopy()),
+		}
+
+		podCapacity, err := c.PodCapacity(ctx)
+
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if podCapacity != 150 {
+			t.Fatalf("expected pods capacity 150, got %d", podCapacity)
+		}
+	})
+
 }


raptorsun:
Tested with clusterbot it is well set to 2MB body size limit when adding the option limitScrapeBodySize: true in CMO config map.

```
$k exec -it prometheus-k8s-0 -n openshift-monitoring -- cat /etc/prometheus/config_out/prometheus.env.yaml  | grep body_size
  body_size_limit: 2MB
  body_size_limit: 2MB
  body_size_limit: 2MB
```

With the default value 2MB limit all targets are correctly scraped.

When setting the body size limit to 500KB part of the targets shows an error of exceeding the limit as shown in the screenshot below. (CMO running locally)
image

jan--f:
Nice, this looks good to me. Is the scenario of hitting a scrape limit already covered by an existing alert?

I think it should definitely be alerted upon, as otherwise user might not get relevant metrics without being aware.

raptorsun:
> Nice, this looks good to me. Is the scenario of hitting a scrape limit already covered by an existing alert?
> 
> I think it should definitely be alerted upon, as otherwise user might not get relevant metrics without being aware.

Thanks for pointing out the alert.  I almost forget it 😀
There is a metric `prometheus_target_scrapes_exceeded_body_size_limit_total` in prometheus to trigger alert when scrape body exceeds the size limit. I complete the PR with a commit adding this alert.

pkg/manifests/config.go
```diff
@@ -29,6 +29,11 @@ import (
 
 const (
 	DefaultRetentionValue = "15d"
+
+	// Limit the body size from scrape queries
+	// Assumptions: one node has maximum 110 pods, each pod exposes 100 metrics, each metric is expressed by at most 100 bytes.
+	// 1.5x the size for a safe margin, it rounds up to 2MB.
+	DefaultScrapeBodySizeLimit = "2MB"
```
fpetkovski, line 36: Do we want to allow users to set this value themselves instead of us setting an arbitrary value?
Looking at the metrics coming from kube-state-metrics in the daily cluster, the 2MB are can very easily be exceeded

image

jsonnet/utils/sanitize-rules.libsonnet
```diff
@@ -326,6 +326,26 @@ local patchedRules = [
   },
 ];
 
+local addedRules = [
```
arajkumar, line 329: Can this be defined in upstream prometheus mixin?

pkg/manifests/config.go
arajkumar: I agree to a certain extent, we can't limit the body size for UWM, however we can judge the same for PM.

fpetkovski: I'm not sure if we can even make an estimate for PM because it depends on the size of the cluster. We have basic clusters with just a few pods, and also build clusters with ~10K pods. Exporters like KSM and NE will produce data with different size depending on the size of the cluster, so I'm not sure how we could cap this properly.

jsonnet/utils/sanitize-rules.libsonnet
raptorsun: Yes, I will submit a change to upstream prometheus mixins later. Once it is fit into upstream, we can clean it from CMO.

pkg/manifests/config.go
raptorsun: We can let the user to choose a size they see fit. How about `limitScrapeBodySize `be a string option instead of boolean? and the user can set it to the following values:
1. a size in Prometheus config format such as "10MB"
2. a string "default" to use the default value we provided "2MB"
3. empty. no limit will be set.

raptorsun:
A new version is pushed :) 
`limitScrapeBodySize` is a string now accepting the following 3 values:
1. a size in Prometheus config format such as "10MB"
2. a string "default" to use the default value we provided "2MB"
3. empty. no limit will be set.

pkg/manifests/config.go
fpetkovski: I personally would have only 1. and 3., but I also don't know what the actual requirements are. Maybe @simonpasquier could chime in.

juzhao:
from the struct file, I think we only need to set enforcedBodySizeLimit in configmap for prometheusK8s(openshift-monitoring) and prometheus(openshift-user-workload-monitoring), LimitScrapeBodySize in ClusterMonitoringConfiguration struct is useless, should be removed
```
type PrometheusRestrictedConfig struct {
...
	EnforcedBodySizeLimit string	`json:"enforcedBodySizeLimit,omitempty"`
}

type PrometheusK8sConfig struct {
...
	EnforcedBodySizeLimit string	`json:"enforcedBodySizeLimit,omitempty"`
}
```

raptorsun:
> from the struct file, I think we only need to set enforcedBodySizeLimit in configmap for prometheusK8s(openshift-monitoring) and prometheus(openshift-user-workload-monitoring), limitScrapeBodySize in ClusterMonitoringConfiguration struct is useless, should be removed
> 
> ```
> type PrometheusRestrictedConfig struct {
> ...
> 	EnforcedBodySizeLimit string	`json:"enforcedBodySizeLimit,omitempty"`
> }
> 
> type PrometheusK8sConfig struct {
> ...
> 	EnforcedBodySizeLimit string	`json:"enforcedBodySizeLimit,omitempty"`
> }
> 
> type ClusterMonitoringConfiguration struct {
> ...
> 	LimitScrapeBodySize 	string	`json:"limitScrapeBodySize,omitempty"`
> }
> ```

Good point. The only use for LimitScrapeBodySize option is to give a default value to the users less savvy on Prometheus or its metric endpoints. 
Shall we keep the option LimitScrapeBodySize at CMO level config? @simonpasquier @fpetkovski @arajkumar

jsonnet/utils/sanitize-rules.libsonnet
raptorsun: PR created in Prometheus to add this alert to mixins: https://github.com/prometheus/prometheus/pull/9873
Hope it can be merged someday :)

simonpasquier:
>  Shall we keep the option LimitScrapeBodySize at CMO level config? @simonpasquier @fpetkovski @arajkumar

For UWM, admins already have some guard rails to prevent abuses with `enforcedSampleLimit` and `enforcedTargetLimit`. Exposing the `body_size_limit` option on top of them might provide additional safety but it's definitely hard for an admin to know in advance what is an adequate value.

The idea of using `body_size_limit` stemmed from an issue that happened during the k/k 1.22 rebase in openshisft/origin. There was a new admission webhoook metric that exposed so many timeseries (> 1M) with so many labels that the prometheus-k8s pods were immediately OOM-killed when they tried to parse the response body. IIRC the response body size was 300M and the initial reaction was that Prometheus should have rejected the response upfront because given the cluster size, no target should be exposing so much data. I think that based on the maximum number of pods that the cluster can schedule, we can come up with a reasonable value for `body_size_limit` (with a comfortable margin to avoid failing the targets). It should be possible for cluster admins to turn off the limit but I'm less convinced that allowing an admin to define the exact value is useful.

fpetkovski:
I see, in that case we should make the limit very liberal to make sure we don't suddenly dropping scrapes for valid exporters.

simonpasquier:
We've discussed this offline with @raptorsun.

* We may want to expose the `body_size_limit` configuration for UWM eventually but lets tackle platform monitoring first (side-note: we had https://github.com/openshift/cluster-monitoring-operator/pull/1350 to enable more controls on UWM targets but it got stale).
* CMO should come up with a default value that makes sense given the "size" of the cluster. The main factor is the maximum number of pods that can be running on the: the more pods, the more samples exposed by kube-state-metrics. We'll make use of our own CI clusters to establish the correlation factor between the number of pods and the number of samples.
* The "out-of-the"box" limit should be clamped to a lower value that works for the minimum supported configuration. We will do some testing on a "cluster-bot" cluster with tools like https://github.com/prometheus-community/avalanche to better understand the actual limits (e.g. how much data Prometheus can handle from a single target before facing constant OOMs). We want the limit to avoid Prometheus crashing but we don't want targets to be dismissed aggresively.
* As a safety mechanism, cluster admins should be able to override the "computed" limit in the CMO configmap, either by defining their own absolute value or disabling the limit altogether.

raptorsun:
Default body size limit is calculated from cluster configuration using the following formula:
body_size_limit = sum(node.status.capacity.pods) * sample_per_pod * size_per_sample
where sample_per_pod is 50 samples and size_per_sample is 200 bytes.

Tests for cluster configuration dependant default limit setting:
1. start cluster, we have a default content in CMO config map `cluster-monitoring-config`
2. set limitScrapeBodySize to 10MB, Prometheus configuration is set to have the expected `body_size_limit: 10MB` for each target.
3. set limitScrapeBodySize to "default",  the calculated limit is `body_size_limit: 14648KB`, set on each target.
4. remove limitScrapeBodySize, the configuration is back to its initial state.

Config maps and correspondent Prometheus configuration files are saved in this link below:
[CMO-Prometheus-configs.zip](https://github.com/openshift/cluster-monitoring-operator/files/7648967/CMO-Prometheus-configs.zip)

raptorsun:
Added a log when the calculated body size limit is smaller than 2MB.

raptorsun:
The new version gives CMO config option limitScrapeBodySize an overriding power to prometheusK8s.enforcedBodySizeLimit. If limitScrapeBodySize  is not set, the option prometheusK8s.enforcedBodySizeLimit will be overriden to empty. 
I think this is not a problem because limitScrapeBodySize can set any value for enforcedBodySizeLimit. 
If I missed something about this option, please share your idea here :)

juzhao:
> The new version gives CMO config option limitScrapeBodySize an overriding power to prometheusK8s.enforcedBodySizeLimit. If limitScrapeBodySize is not set, the option prometheusK8s.enforcedBodySizeLimit will be overriden to empty. I think this is not a problem because limitScrapeBodySize can set any value for enforcedBodySizeLimit. If I missed something about this option, please share your idea here :)

1. for openshift-monitoring
scenario 1: set prometheusK8s.enforcedBodySizeLimit = 10MB only, no limitScrapeBodySize setting，body_size_limit is empty for each target
```
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |-
    enableUserWorkload: true
    prometheusK8s:
      enforcedBodySizeLimit: 10MB
```
no result from config file
```
# oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- cat /etc/prometheus/config_out/prometheus.env.yaml | grep body_size
no result
```
scenario 2: set prometheusK8s.enforcedBodySizeLimit = 10MB and limitScrapeBodySize， body_size_limit is decided by limitScrapeBodySize setting, not by prometheusK8s.enforcedBodySizeLimit
```
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |-
    limitScrapeBodySize: default
    enableUserWorkload: true
    prometheusK8s:
      enforcedBodySizeLimit: 10MB
```
no result from config file
```
# oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- cat /etc/prometheus/config_out/prometheus.env.yaml | grep body_size
  body_size_limit: 14648KB
  body_size_limit: 14648KB
```
2. for user workload monitoring
scenario 1: set limitScrapeBodySize only for UWM, 
```
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    limitScrapeBodySize: default
```
deploy application in user project, body_size_limit is empty for each target. Is it expected? the code does not consider this scenario
```
# oc -n openshift-user-workload-monitoring exec -c prometheus prometheus-user-workload-0 -- cat /etc/prometheus/config_out/prometheus.env.yaml | grep body_size
no result
```
scenario 2: set limitScrapeBodySize and prometheus.enforcedBodySizeLimit = 10MB  for UWM, 
```
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    limitScrapeBodySize: default
    prometheus:
      enforcedBodySizeLimit: 10MB
```
deploy application in user project, body_size_limit is decided by prometheus.enforcedBodySizeLimit
```
# oc -n openshift-user-workload-monitoring exec -c prometheus prometheus-user-workload-0 -- cat /etc/prometheus/config_out/prometheus.env.yaml | grep body_size
  body_size_limit: 10MB
```

juzhao:
@raptorsun @simonpasquier @jan--f any idea for the above comments?

raptorsun:
just pushed a modification that PrometheusK8s config enforcedBodySizeLimit is not overriden to 0 when CMO config limitScrapeBodySize is empty.

raptorsun:
Now empty value for `limitScrapeBodySize `does not override `prometheusK8s.enforcedBodySizeLimit`.

```
apiVersion: v1
data:
  config.yaml: |-
    prometheusK8s:
      enforcedBodySizeLimit: 10MB
      volumeClaimTemplate:
        metadata:
          name: prometheus-data
          annotations:
            openshift.io/cluster-monitoring-drop-pvc: "yes"
        spec:
          resources:
            requests:
              storage: 20Gi
kind: ConfigMap
metadata:
```
```
➜  clusterbot  oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- cat /etc/prometheus/config_out/prometheus.env.yaml | grep body_size_limit | wc -l
70
```

jan--f:
/lgtm
/label docs-approved
This looks good, adding `docs-approved` label as per agreement.
Ping @juzhao for qe approval.

juzhao:
bugs:
https://bugzilla.redhat.com/show_bug.cgi?id=2065037
https://bugzilla.redhat.com/show_bug.cgi?id=2065035
if 2065037 is by design, QE could add approve label, since 2065035 is a low severity bug

raptorsun:
> bugs: https://bugzilla.redhat.com/show_bug.cgi?id=2065037 https://bugzilla.redhat.com/show_bug.cgi?id=2065035 if 2065037 is by design, QE could add approve label, since 2065035 is a low severity bug

Thank you for raising these concerns :) 
The bug [2065037](https://bugzilla.redhat.com/show_bug.cgi?id=2065037) is indeed part of our design that limitScrapeBodySize setting takes precedence over prometheusK8s.enforcedBodySizeLimit.
The bug [2065035](https://bugzilla.redhat.com/show_bug.cgi?id=2065035) is something I overlooked. I'm going to make another PR to add a verification mechanism to the value of limitScrapeBodySize.

juzhao:
based on the previous comment, and bug 2065035 is a low severity bug, we could merge the code now
/label qe-approved

jsonnet/utils/sanitize-rules.libsonnet
simonpasquier: Can you add a TODO so we don't forget?

simonpasquier (commented):
/hold
i want to give a last review :)

pkg/manifests/config.go
```diff
@@ -400,6 +417,21 @@ func (c *Config) applyDefaults() {
 	}
 }
 
+func (c *Config) ActivatePromEnforceBodyLimit(defaultSize string) error {
+	if c.ClusterMonitoringConfiguration.LimitScrapeBodySize == "default" {
```
simonpasquier, line 421: What about `automatic` rather than `default`? I read "default" as "a fixed value" while it's dynamic in fact.

simonpasquier (commented):
Sorry for the very late review! My main cnncert is whether we need to have 2 parameters in the CMO config that interact with the body size limit as I find it a bit confusing.

pkg/manifests/config.go
```diff
@@ -400,6 +417,21 @@ func (c *Config) applyDefaults() {
...
+	} else if c.ClusterMonitoringConfiguration.LimitScrapeBodySize == "" {
+		return nil
+	}
+	c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit = c.ClusterMonitoringConfiguration.LimitScrapeBodySize
+	return nil
+}
+
+func (c *Config) DeactivatePromEnforceBodyLimit() {
+	c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit = ""
+}
```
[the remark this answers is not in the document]
raptorsun: yup, it will be removed.

pkg/manifests/config.go
raptorsun: good idea, default gives an impression of the value being static.

pkg/manifests/config.go
```diff
@@ -105,6 +112,7 @@ type ClusterMonitoringConfiguration struct {
 	K8sPrometheusAdapter     *K8sPrometheusAdapter        `json:"k8sPrometheusAdapter"`
 	ThanosQuerierConfig      *ThanosQuerierConfig         `json:"thanosQuerier"`
 	UserWorkloadEnabled      *bool                        `json:"enableUserWorkload"`
+	LimitScrapeBodySize      string                       `json:"limitScrapeBodySize,omitempty"`
```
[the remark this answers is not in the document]
raptorsun: mainly 2 considerations:
1. In CMO level config EnforcedBodySizeLimit hides details in PrometheusK8sConfig from the user.
2. It makes it easier to manage the cause default/automatic.

pkg/operator/operator.go
```diff
@@ -822,6 +822,12 @@ func (o *Operator) Config(ctx context.Context, key string) (*manifests.Config, e
 	}
 	o.userWorkloadEnabled = *c.ClusterMonitoringConfiguration.UserWorkloadEnabled
 
+	if c.ClusterMonitoringConfiguration.LimitScrapeBodySize == "default" {
+		c.ActivatePromEnforceBodyLimit(o.calculateBodySizeLimit(ctx))
+	} else if manifests.ValidateSizeFormat(c.ClusterMonitoringConfiguration.LimitScrapeBodySize) {
+		c.ActivatePromEnforceBodyLimit(c.ClusterMonitoringConfiguration.LimitScrapeBodySize)
+	}
```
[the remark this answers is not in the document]
raptorsun: This is much better decoupling than the current version. I'm going to put LoadEnforcedBodySizeLimit into config.go where keeps all the class methods of Config while leave the interface in manifest.go where all interfaces are.

raptorsun:
New version is ready to review :)

raptorsun:
ready to review again :)

raptorsun:
This new version adds the fIeld prometheusK8s.enforcedBodySizeLimit to CMO ConfigMap, accepting the size format from Prometheus:
- Empty value means automatic bodysize limit.
- 0 for no limit.
- [0-9]+[A-Z][a-z]*B

CMO config map has no longer limitScrapeBodySize at its root level, avoiding confusion.

pkg/manifests/config.go
```diff
@@ -467,6 +476,35 @@ func (c *Config) NoProxy() string {
...
+	if err := pm.ValidateSizeField(enforcedBodySizeLimit); err != nil {
+		return err
+	}
+	c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit = enforcedBodySizeLimit
+	return nil
+}
+
+func calculateBodySizeLimit(podCapacity int) string {
+	const samplePerPod = 50
+	const sizePerSample = 200 // 200 Bytes
```
simonpasquier, line 497: These consts don't match with the comment from `minimalSizeLimit`. We need to use the same inputs and probably document in more details how we got these numbers. 

```
// Limit the body size from scrape queries
// Assumptions: one node has maximum 110 pods, each pod exposes 100 metrics, each metric is expressed by at most 100 bytes.
// 1.5x the size for a safe margin, it rounds up to 2MB.
```

pkg/manifests/config.go
```diff
@@ -16,19 +16,27 @@ package manifests
...
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 
 	configv1 "github.com/openshift/api/config/v1"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	pm "github.com/prometheus-operator/prometheus-operator/pkg/operator"
```
simonpasquier, line 26: (nit)

```suggestion
	poperator "github.com/prometheus-operator/prometheus-operator/pkg/operator"
```

pkg/manifests/config.go
```diff
@@ -321,7 +330,7 @@ func NewConfig(content io.Reader) (*Config, error) {
 	res.applyDefaults()
 	c.UserWorkloadConfiguration = NewDefaultUserWorkloadMonitoringConfig()
 
-	return res, nil
+	return res, err
```
simonpasquier, line 333:
```suggestion
```

pkg/manifests/config.go
```diff
@@ -467,6 +476,35 @@ func (c *Config) NoProxy() string {
 	return c.ClusterMonitoringConfiguration.HTTPConfig.NoProxy
 }
 
+func (c *Config) LoadEnforcedBodySizeLimit(pcr PodsCapacityReader, ctx context.Context) error {
+	enforcedBodySizeLimit := c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit
+	if enforcedBodySizeLimit == "" {
```
simonpasquier, line 481: From my POV, it is more readable to return early in case the value isn't empty. Then the function can focus on computing the correct value.

```suggestion
	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit != "" {
		return pm.ValidateSizeField(c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit)
	}
```

pkg/manifests/config.go
```diff
@@ -467,6 +476,35 @@ func (c *Config) NoProxy() string {
...
+	const samplePerPod = 50
+	const sizePerSample = 200 // 200 Bytes
+
+	bodySize := podCapacity * samplePerPod * sizePerSample
+	if bodySize < minimalSizeLimit {
+		bodySize = minimalSizeLimit
+		klog.Infof("Calculated scrape body size limit is too small, using default value %v instead", minimalSizeLimit)
+	}
+
+	return fmt.Sprintf("%dKB", int(bodySize/1024))
```
simonpasquier, line 505: Lets take another safety margin and round up the value to the next MB.

```suggestion
	return fmt.Sprintf("%dMB", math.Ceil(float64(bodySize)/(1024*1024))))
```

pkg/manifests/config_test.go
```diff
@@ -234,3 +235,79 @@ func TestHttpProxyConfig(t *testing.T) {
...
+func TestEnforceBodySizeLimit(t *testing.T) {
+
+	mc_10 := fakePodsCapacity(10)
+	mc_1000 := fakePodsCapacity(1000)
+	for _, tt := range []struct {
+		name                string
+		config              string
+		expectBodySizeLimit string
+		expectError         bool
+		mockClient          *fakePodsCapacity
```
simonpasquier, line 253: we can use the interface type directly.

```suggestion
		pcr          PodsCapacityReader
```

pkg/manifests/config_test.go
```diff
@@ -234,3 +235,79 @@ func TestHttpProxyConfig(t *testing.T) {
...
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			c, err := NewConfigFromString(tt.config)
+			if tt.expectError && err == nil {
+				t.Fatalf("expected error, got nil")
+			} else if tt.expectError && err != nil {
+				return // expected error, got error, skip size verification
+			} else if !tt.expectError && err != nil {
+				t.Fatalf("expected no error, got error %v", err)
+			}
```
simonpasquier, line 299: IIUC NewConfigFromString() should never fail.

```suggestion
			if  err != nil {
				t.Fatalf("expected no error, got error %v", err)
			}
```

pkg/manifests/config_test.go
```diff
@@ -234,3 +235,79 @@ func TestHttpProxyConfig(t *testing.T) {
 		}
 	}
 }
+
+type fakePodsCapacity int
+
+func (fpc *fakePodsCapacity) PodsCapacity(context.Context) (int, error) {
+	return int(*fpc), nil
+}
+func TestEnforceBodySizeLimit(t *testing.T) {
```
simonpasquier, line 244:
```suggestion
func TestLoadEnforcedBodySizeLimit(t *testing.T) {
```

pkg/manifests/config_test.go
```diff
@@ -234,3 +235,79 @@ func TestHttpProxyConfig(t *testing.T) {
...
+			expectError:         false,
+			mockClient:          &mc_10,
+		},
+		{
+			name:                "automatically deduced limit: larger than minimal 2MB",
+			config:              `{"prometheusK8s": {"enforcedBodySizeLimit": ""}}`,
+			expectBodySizeLimit: "9765KB",
+			expectError:         false,
+			mockClient:          &mc_1000,
+		},
```
simonpasquier, line 289: we can ave 2 more test cases:
* one with a fake PodsCapacityReader that always returns an error.
* one with an invalid user input.

pkg/manifests/manifests.go
```diff
@@ -318,6 +319,11 @@ type ProxyReader interface {
 	NoProxy() string
 }
 
+// PodsCapacityReader returns the maximum number of pods that can be scheduled in a cluster.
+type PodsCapacityReader interface {
+	PodsCapacity(context.Context) (int, error)
+}
+
```
simonpasquier, line 326: (nit) it can be moved next to LoadEnforcedBodySizeLimit()

pkg/manifests/manifests_test.go
```diff
@@ -1469,6 +1474,31 @@ func TestPrometheusQueryLogFileConfig(t *testing.T) {
 	}
 }
 
+func TestPrometheusK8sConfigurationBodySizeLimitDefault(t *testing.T) {
+	c, err := NewConfigFromString(`
+prometheusK8s:
+    enforceBodySizeLimit: "10MB"
```
simonpasquier, line 1480:
```suggestion
    enforcedBodySizeLimit: "10MB"
```

pkg/manifests/manifests_test.go
```diff
@@ -1469,6 +1474,31 @@ func TestPrometheusQueryLogFileConfig(t *testing.T) {
...
+		&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// the body size limit value is not set at configuration parsing time.
+	if p.Spec.EnforcedBodySizeLimit != "" {
+		t.Fatalf("EnforcedBodySizeLimit is not configured correctly, expected empty string, but got %v", p.Spec.EnforcedBodySizeLimit)
+	}
```
simonpasquier, line 1499: it should be the opposite

pkg/operator/operator.go
```diff
@@ -829,6 +829,11 @@ func (o *Operator) Config(ctx context.Context, key string) (*manifests.Config, e
 	}
 	o.userWorkloadEnabled = *c.ClusterMonitoringConfiguration.UserWorkloadEnabled
 
+	err = c.LoadEnforcedBodySizeLimit(o.client, ctx)
+	if err != nil {
+		klog.Warningf("Error loading enforced body size limit. Proceeding without it. Error: %v", err)
```
simonpasquier, line 834: Nothing in LoadEnforcedBodySizeLimit() takes care of resetting the value if something bad happens (e.g. if the config contains an invalid size value). This should be added.

test/e2e/framework/client.go
```diff
@@ -313,6 +313,42 @@ func GetFirstValueFromPromQuery(body []byte) (int, error) {
 	return v, nil
 }
 
+// GetFirstValueFromPromQuery takes a query api response body and returns the
+// value of the first timeseries. If body contains multiple timeseries
+// GetFirstValueFromPromQuery errors.
```
simonpasquier, line 318:
```suggestion
// GetFirstFloatValueFromPromQuery takes a query api response body and returns the
// value of the first timeseries. If body contains multiple timeseries
// GetFirstFloatValueFromPromQuery errors.
```

test/e2e/framework/client.go
```diff
@@ -313,6 +313,42 @@ func GetFirstValueFromPromQuery(body []byte) (int, error) {
...
+func GetFirstFloatValueFromPromQuery(body []byte) (float64, error) {
```
simonpasquier, line 319: there's room to factorize code between GetFirstFloatValueFromPromQuery() and GetFirstValueFromPromQuery().

test/e2e/prometheus_test.go
```diff
@@ -318,3 +318,61 @@ func remoteWriteCheckMetrics(ctx context.Context, t *testing.T) {
...
+  enforcedBodySizeLimit: %s
+  volumeClaimTemplate:
+    metadata:
+      name: %s
+    annotations:
+      openshift.io/cluster-monitoring-drop-pvc: "yes"
+    spec:
+      resources:
+        requests:
+          storage: %s
```
simonpasquier, line 363: I don't think we need persistent storage for the test?

test/e2e/prometheus_test.go
```diff
@@ -318,3 +318,61 @@ func remoteWriteCheckMetrics(ctx context.Context, t *testing.T) {
...
+`, bodySizeLimitSmall, pvcClaimName, storage)
+	f.MustCreateOrUpdateConfigMap(t, configMapWithData(t, data))
+
+	f.PrometheusK8sClient.WaitForQueryReturnFloat(
+		t, 5*time.Minute, `sum(increase(prometheus_target_scrapes_exceeded_body_size_limit_total{job="prometheus-k8s"}[5m]))`,
```
simonpasquier, line 368: If you don't want to add the new WaitForQueryReturnFloat() function, you can also tweak the query so it returns always an integer.

```suggestion
		t, 5*time.Minute, `count(increase(prometheus_target_scrapes_exceeded_body_size_limit_total{job="prometheus-k8s"}[5m] > 0))`,
```

we can also check that the returned value is 2 (prometheus-k8s-0 and prometheus-k8s-1).

pkg/manifests/config.go
raptorsun: yup, it is more readable in this way :)

pkg/manifests/config_test.go
raptorsun: agree, these checks should be done for LoadEnforcedBodySizeLimit()

pkg/manifests/config.go
raptorsun: Going to check again on average size of metrics and count of metrics per pod. Then update the comments.

pkg/manifests/manifests_test.go
raptorsun: will add LoadEnforcedBodySizeLimit after config parsing and test the automatic deduced bodysize limit.

pkg/operator/operator.go
raptorsun: will use minimal body size limit in case of invalid size value.

test/e2e/framework/client.go
raptorsun: Their common logics are going to extract into private function `getFirstRawValueFromPromQuery(body []byte) (*gabs.Container, error)`

test/e2e/prometheus_test.go
raptorsun: I'm going to remove the persistent storage for a test. if it goes well, I will remove that definitely.

test/e2e/prometheus_test.go
raptorsun: good idea for rounding the number :D we dont need to create a float version for these functions. will modify the code accordingly.

pkg/manifests/config.go
```diff
@@ -16,19 +16,28 @@ package manifests
...
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
+	"k8s.io/klog/v2"
 )
 
 const (
 	DefaultRetentionValue = "15d"
+
+	// Limit the body size from scrape queries
+	// Assumptions: one node has maximum 110 pods, each pod exposes 100 metrics, each metric is expressed by at most 100 bytes.
+	// 1.5x the size for a safe margin, it rounds up to 2MB.
```
raptorsun, line 39: todo: refine this minimal size limit.

pkg/operator/operator.go
```diff
@@ -829,6 +829,13 @@ func (o *Operator) Config(ctx context.Context, key string) (*manifests.Config, e
 	}
 	o.userWorkloadEnabled = *c.ClusterMonitoringConfiguration.UserWorkloadEnabled
 
+	err = c.LoadEnforcedBodySizeLimit(o.client, ctx)
+	if err != nil {
+		c.UseMinimalEnforcedBodySizeLimit()
+		klog.Warningf("Error loading enforced body size limit. Using minimal body size limit %s. Error: %v",
+			c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit, err)
+	}
```
simonpasquier, line 837: I would disable the limit if an error is returned.

```suggestion
	if err := c.LoadEnforcedBodySizeLimit(o.client, ctx);  err != nil {
		c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit = ""
		klog.Warningf("Error loading enforced body size limit, no body size limit will be enforced. Error: %v",
			err)
	}
```

pkg/manifests/config_test.go
```diff
@@ -234,3 +236,98 @@ func TestHttpProxyConfig(t *testing.T) {
...
+			}
+
+			err = c.LoadEnforcedBodySizeLimit(tt.pcr, context.Background())
+			if tt.expectError && err == nil {
+				t.Fatalf("expected error, got nil")
+			} else if tt.expectError && err != nil {
+				return // expected error, got error, skip size verification
+			} else if !tt.expectError && err != nil {
+				t.Fatalf("expected no error, got error %v", err)
+			}
```
simonpasquier, line 324:
```suggestion
			if tt.expectError {
				if err == nil {
					t.Fatalf("expected error, got nil")
				}
				return
			} 
			
			if err != nil {
				t.Fatalf("expected no error, got error %v", err)
			}
```

raptorsun:
New minimal body size limit is revised to 4MB.
One node should have maximum 110 pods according [to Kubernetes document](https://kubernetes.io/docs/setup/best-practices/cluster-large/). 
The following numbers are observed from [a cluster running e2e tests with extra metric feature enabled in Prometheus](https://github.com/openshift/cluster-monitoring-operator/pull/1627):
- Each pod exposes ~100 metrics `max(scrape_samples_scraped/kubelet_running_pods)`
- Each metric is expressed by ~250 bytes `avg(scrape_body_size_bytes / scrape_samples_scraped) by(job)`

1.5 times the size for a safe margin, it rounds to 4MB (4125000 Bytes).

The chart below shows how many samples are collected from a pod
image


The chart below shows large metrics average around 250bytes per sample. 
image

raptorsun:
Here is a [PR](https://github.com/openshift/runbooks/pull/47) to add alert PrometheusScrapeBodySizeLimitHit  into runbook.

raptorsun:
Formula for body size limit has been adjusted with new assumptions on scaping:
- 400 samples per pod
- 200 bytes per sample
- each node uses no more than 60% of its declared pod capacity (this is 250 in test cluster, kubernetes document recommends 110 max).

jan--f:
This looks good with some of Simons suggestions.
One naming nit from my side: This PR currently uses `podsCapacity` in most places, but sometimes `podCapacity` (mostly as variable names I think). I would like to see consistency there and my vote would be to use `podCapacity`.

pkg/manifests/config_test.go
raptorsun: 2 cases were added.

pkg/manifests/config.go
```diff
@@ -315,19 +325,18 @@ func (cfg *TelemeterClientConfig) IsEnabled() bool {
 	return true
 }
 
-func NewConfig(content io.Reader) (*Config, error) {
+func NewConfig(content io.Reader) (res *Config, err error) {
```
simonpasquier, line 328: I'm not a big fan of named return values as I find that they make the code less readable. Maybe this is only me but I don't feel either that this change is required here.

pkg/manifests/config.go
```diff
@@ -473,6 +482,46 @@ func (c *Config) NoProxy() string {
 	return c.ClusterMonitoringConfiguration.HTTPConfig.NoProxy
 }
 
+// PodCapacityReader returns the maximum number of pods that can be scheduled in a cluster.
+type PodCapacityReader interface {
+	PodCapacity(context.Context) (int, error)
+}
+
+func (c *Config) LoadEnforcedBodySizeLimit(pcr PodCapacityReader, ctx context.Context) error {
+	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit == "automatic" {
```
simonpasquier, line 491: Maybe define a `const` for "automatic".

pkg/manifests/config.go
```diff
@@ -173,17 +182,18 @@ type RemoteWriteSpec struct {
...
+	NodeSelector          map[string]string                    `json:"nodeSelector"`
+	Tolerations           []v1.Toleration                      `json:"tolerations"`
+	Resources             *v1.ResourceRequirements             `json:"resources"`
+	ExternalLabels        map[string]string                    `json:"externalLabels"`
+	VolumeClaimTemplate   *monv1.EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate"`
+	RemoteWrite           []RemoteWriteSpec                    `json:"remoteWrite"`
+	TelemetryMatches      []string                             `json:"-"`
+	AlertmanagerConfigs   []AdditionalAlertmanagerConfig       `json:"additionalAlertmanagerConfigs"`
+	QueryLogFile          string                               `json:"queryLogFile"`
+	EnforcedBodySizeLimit string                               `json:"enforcedBodySizeLimit,omitempty"`
```
simonpasquier, line 196: Can you add a comment about how using the parameter?
* an empty value means no enforcement
* "automatic" means that CMO picks up a value based on the cluster capacity
* A fixed size can be defined too.

pkg/manifests/config.go
simonpasquier, line 491: you can handle the "return-early" case first.

```suggestion
	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit == "" {
		return nil
	}

	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit == "automatic" {
```

pkg/manifests/config.go
```diff
@@ -473,6 +482,46 @@ func (c *Config) NoProxy() string {
...
+	}
+
+	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit != "" {
+		return poperator.ValidateSizeField(c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit)
+	}
+
+	return nil
+}
+
+func (c *Config) UseMinimalEnforcedBodySizeLimit() {
```
simonpasquier, line 507: this doesn't seem to be used but should be incorporated in `calculateBodySizeLimit()` I believe.

pkg/manifests/config.go
```diff
@@ -473,6 +482,46 @@ func (c *Config) NoProxy() string {
...
+	c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit = fmt.Sprintf("%dMB", int(math.Ceil(float64(minimalSizeLimit)/(1024*1024))))
+}
+
+func calculateBodySizeLimit(podCapacity int) string {
+	const samplesPerPod = 400       // 400 samples per pod
+	const sizePerSample = 200       // 200 Bytes
+	const loadFactorPercentage = 60 // assume 80% of the maximum pods capacity per node is used
```
simonpasquier, line 514: I would assume that the full capacity can be used.

pkg/manifests/config.go
```diff
@@ -473,6 +482,46 @@ func (c *Config) NoProxy() string {
...
+
+	bodySize := loadFactorPercentage * podCapacity / 100 * samplesPerPod * sizePerSample
+	if bodySize < minimalSizeLimit {
+		bodySize = minimalSizeLimit
+		klog.Infof("Calculated scrape body size limit is too small, using default value %v instead", minimalSizeLimit)
```
simonpasquier, line 519: we could log both values (e.g. "calculated body size limit = ... is too small, using ... instead")

pkg/operator/operator.go
```diff
@@ -829,6 +829,12 @@ func (o *Operator) Config(ctx context.Context, key string) (*manifests.Config, e
 	}
 	o.userWorkloadEnabled = *c.ClusterMonitoringConfiguration.UserWorkloadEnabled
 
+	err = c.LoadEnforcedBodySizeLimit(o.client, ctx)
+	if err != nil {
+		c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit = ""
+		klog.Warningf("Error loading enforced body size limit, no body size limit will be enforced. Error: %v", err)
```
simonpasquier, line 835:
```suggestion
		klog.Warningf("Error loading enforced body size limit, no body size limit will be enforced: %v", err)
```

test/e2e/prometheus_test.go
```diff
@@ -295,3 +295,50 @@ func remoteWriteCheckMetrics(ctx context.Context, t *testing.T, promClient *fram
...
+	)
+
+	data := fmt.Sprintf(`prometheusK8s:
+  logLevel: debug
+  enforcedBodySizeLimit: %s
+`, bodySizeLimitSmall)
+	f.MustCreateOrUpdateConfigMap(t, configMapWithData(t, data))
+
+	f.PrometheusK8sClient.WaitForQueryReturn(
+		t, 5*time.Minute, `ceil(sum(increase(prometheus_target_scrapes_exceeded_body_size_limit_total{job="prometheus-k8s"}[5m])))`,
```
simonpasquier, line 334:
```suggestion
		t, 5*time.Minute, `sum(increase(prometheus_target_scrapes_exceeded_body_size_limit_total{job="prometheus-k8s"}[5m]))`,
```

raptorsun: need to keep the `ceil` function to round the result to an integer as `WaitForQueryReturn` requires.

pkg/manifests/config.go
```diff
@@ -474,6 +492,43 @@ func (c *Config) NoProxy() string {
...
+	return poperator.ValidateSizeField(c.ClusterMonitoringConfiguration.PrometheusK8sConfig.EnforcedBodySizeLimit)
+
+}
+
+func calculateBodySizeLimit(podCapacity int) string {
+	const samplesPerPod = 400        // 400 samples per pod
+	const sizePerSample = 200        // 200 Bytes
+	const loadFactorPercentage = 100 // assume 100% of the maximum pods capacity per node is used
+
+	bodySize := loadFactorPercentage * podCapacity / 100 * samplesPerPod * sizePerSample
```
simonpasquier, line 523: I think we can get rid of `loadFactorPercentage`.

```suggestion
	bodySize :=  podCapacity * samplesPerPod * sizePerSample
```

pkg/manifests/config.go
```diff
@@ -16,19 +16,32 @@ package manifests
...
+	"k8s.io/klog/v2"
 )
 
 const (
 	DefaultRetentionValue = "15d"
+
+	// Limit the body size from scrape queries
+	// Assumptions: one node has in average 110 pods, each pod exposes 400 metrics, each metric is expressed by on average 250 bytes.
+	// 1.5x the size for a safe margin, it rounds to 16MB (16,500,000 Bytes).
+	minimalSizeLimit = 1.5 * 110 * 400 * 250
```
simonpasquier, line 40: We probably need a "safe" lower-bound value for the "automatically computed"  value but I'm not sure that this value is accurate. I would take a typical CI cluster, load it with as many pods/secret/configmaps as possible and measure the body size returned by kube-state-metrics `/metrics`.

raptorsun: Having tested with 100k secrets and 100k config maps with kube-burner, KSM scrape target gives us ~32MB body size. 
So this lower bound is tripled to 48MB in new commit. It should be large enough I suppose :) 

Screenshot from 2022-05-10 12-09-02

jan--f (approved):
This lgtm

simonpasquier (approved):
/lgtm

PR #1470, opened 2021-11-08
pkg: add tlsSecurityProfile to kube-rbac-proxy in node-exporter

slashpai:
This PR covers tls security profile for kube-rbac-proxy in node-exporter

[MON-2022](https://issues.redhat.com/browse/MON-2022)


slashpai:
cc: @juzhao

slashpai:
/assign @fpetkovski 
Since you already reviewed other TLS Profile PR's :)

slashpai:
There is issue in e2e test since its a daemonset while others are deployments. Updating the e2e test

slashpai:
@simonpasquier Are we good to add other components in same PR since we are done with most of helper functions needed and other logic?. Its same code we need other components as well, we can create separate commit per component

fpetkovski:
/lgtm

I would say let's get this PR merged and we can add all other components in one last PR.

PR #1473, opened 2021-11-09
WIP: Add support for user-defined alert overrides

bison:
This adds a new controller that watches for user-defined overrides of
platform alerting rules, generates new alerting rules with the
overrides applied, then generates an alert relabel config to drop the
original alerts.

The easiest way to test this is probably to spin up a cluster from this PR with cluster-bot. Then you can create overrides by creating an `AlertOverrides` resource. There's an example in `examples/alert-overrides.yaml`:

```yaml
---
apiVersion: monitoring.openshift.io/v1alpha1
kind: AlertOverrides
metadata:
  name: alert-overrides
spec:
  overrides:
  - selector:
      alert: "Watchdog"
    action: "patch"
    labels:
      patched: "true"

  rules:
  - alert: "UserDefinedAlert"
    expr: "vector(1)"
    for: "1m"
    labels:
      custom: "true"
```


bison:
We decided to go a different route. See: #1596 & #1590 -- Closing this.

PR #1474, opened 2021-11-10
pkg: Implements tlsSecurityProfile to remaining components using kube-rbac-proxy

slashpai:
This PR implements tls security profile for remaining components (except cmo) using kube-rbac-proxy per suggestion in last PR #1470. Each component is implemented in separate commit.

cc: @fpetkovski @simonpasquier @juzhao 

[MON-1651](https://issues.redhat.com/browse/MON-1651)

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -2449,6 +2449,7 @@ func (f *Factory) GrafanaDeployment(proxyCABundleCM *v1.ConfigMap) (*appsv1.Depl
 			}
 		case "kube-rbac-proxy-metrics":
 			d.Spec.Template.Spec.Containers[i].Image = f.config.Images.KubeRbacProxy
+			d.Spec.Template.Spec.Containers[i].Args = f.setTLSSecurityConfiguration(container.Args, KubeRbacProxyTLSCipherSuitesFlag, KubeRbacProxyMinTLSVersionFlag)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


slashpai:
aah I missed CMO in this PR, I will add it quick

simonpasquier (commented):
/lgtm

Not directly related to this PR I think so it can be addressed afterwards but `correctCipherSuiteArg` and `correctMinTLSVersion` return as soon as they find one container which matching arguments which doesn't work properly for pods with multiple kube-rbac-proxy containers (e.g. it wouldn't check the subsequent kube-rbac-proxy containers).
Also I would rather see the test checking at the pod level than deployment, statefulset and daemonset.

https://github.com/openshift/cluster-monitoring-operator/blob/4ecf1715bf3f572edb1c424cf3185f37719521fe/test/e2e/tls_security_profile_test.go#L150-L157

juzhao:
remaining components (except cmo) using kube-rbac-proxy now honors the global tlsSecurityProfile setting
/lgtm

slashpai:
> /lgtm
> 
> Not directly related to this PR I think so it can be addressed afterwards but `correctCipherSuiteArg` and `correctMinTLSVersion` return as soon as they find one container which matching arguments which doesn't work properly for pods with multiple kube-rbac-proxy containers (e.g. it wouldn't check the subsequent kube-rbac-proxy containers). Also I would rather see the test checking at the pod level than deployment, statefulset and daemonset.
> 
> https://github.com/openshift/cluster-monitoring-operator/blob/4ecf1715bf3f572edb1c424cf3185f37719521fe/test/e2e/tls_security_profile_test.go#L150-L157

Will takecare of it in separate PR

PR #1478, opened 2021-11-15
MON-1872: Use kube-thanos jsonnet libs for thanos ruler

PhilipGough:

### DIFF
--- a/jsonnet/components/thanos-ruler.libsonnet
+++ b/jsonnet/components/thanos-ruler.libsonnet
@@ -1,495 +1,485 @@
-// TODO(paulfantom): This should use upstream kube-thanos project.
 local generateCertInjection = import '../utils/generate-certificate-injection.libsonnet';
+local ruler = import 'github.com/thanos-io/kube-thanos/jsonnet/kube-thanos/kube-thanos-rule.libsonnet';
 
-function(params) {
-  local cfg = params,
+local defaults = {
+  volumeClaimTemplate: {},
+  serviceMonitor: true,
+};
 
-  mixin:: (import 'github.com/thanos-io/thanos/mixin/alerts/rule.libsonnet') {
-    targetGroups: {},
-    rule+:: {
-      selector: 'job="thanos-ruler"',
-    },
-  },
+function(params)
+  local cfg = defaults + params;
+  local tr = ruler(cfg);
 
-  thanosRulerPrometheusRule: {
-    apiVersion: 'monitoring.coreos.com/v1',
-    kind: 'PrometheusRule',
-    metadata: {
-      name: 'thanos-ruler',
-      namespace: cfg.namespace,
-    },
-    spec: $.mixin.prometheusAlerts,
-  },
-
-  trustedCaBundle: generateCertInjection.trustedCNOCaBundleCM(cfg.namespace, 'thanos-ruler-trusted-ca-bundle'),
-
-  route: {
-    apiVersion: 'v1',
-    kind: 'Route',
-    metadata: {
-      name: 'thanos-ruler',
-      namespace: cfg.namespace,
-    },
-    spec: {
-      to: {
-        kind: 'Service',
-        name: 'thanos-ruler',
-      },
-      port: {
-        targetPort: 'web',
-      },
-      tls: {
-        termination: 'Reencrypt',
-        insecureEdgeTerminationPolicy: 'Redirect',
+  tr {
+    mixin:: (import 'github.com/thanos-io/thanos/mixin/alerts/rule.libsonnet') {
+      targetGroups: {},
+      rule+:: {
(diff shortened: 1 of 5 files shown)


jan--f:
/retitle MON-1872: Use kube-thanos jsonnet libs for thanos ruler

jsonnet/main.jsonnet
```diff
@@ -284,17 +284,14 @@ local inCluster =
...
         namespace: $.values.common.namespaceUserWorkload,
         replicas: 2,
-        labels: {
-          'app.kubernetes.io/name': 'user-workload',
-        },
-        selectorLabels: {
-          'app.kubernetes.io/name': 'thanos-ruler',
-          'thanos-ruler': 'user-workload',
-        },
+        serviceMonitor: true,
```
simonpasquier, line 291: Should we move it to `jsonnet/components/thanos-ruler.libsonnet` directly? I feel that it would be clearer since it's not something that we need to customize.
I've got the same question for `volumeClaimTemplate` below.

assets/thanos-ruler/thanos-ruler.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
     thanosRulerName: user-workload
   name: user-workload
   namespace: openshift-user-workload-monitoring
+  selectorLabels:
```
simonpasquier, line 8: hmm not sure why this gets added :)

PhilipGough: I had to add that because of the comment https://github.com/openshift/cluster-monitoring-operator/pull/1478#discussion_r751215716

If we don't want to update the selector on the service, then we need to set this in order to get the labels on the service to match the pods in the stateful set.

Basically I don't think we can pick up the defaults here and retain our existing service selector without failing our e2e tests. [Sample](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/1478/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-operator/1460246190468632576)

juzhao:
why 
`app.kubernetes.io/name: thanos-rule` instead of `app.kubernetes.io/name: thanos-ruler`

PhilipGough:
@juzhao that is being set via `https://github.com/thanos-io/kube-thanos/blob/main/jsonnet/kube-thanos/kube-thanos-rule.libsonnet#L6`

PhilipGough:
/label docs-approved

no docs changes required

jsonnet/components/thanos-ruler.libsonnet
```diff
@@ -1,495 +1,485 @@
...
+
+
+    thanosRuler: {
+      apiVersion: 'monitoring.coreos.com/v1',
+      kind: 'ThanosRuler',
+      metadata: {
+        name: cfg.crName,
+        namespace: tr.config.namespace,
+        labels: {
+          thanosRulerName: cfg.crName,
```
simonpasquier, line 306: this is definitely not introduced by your PR but I wonder why we have this label on the ThanosRuler resource? :thinking:

jsonnet/components/thanos-ruler.libsonnet
```diff
@@ -1,495 +1,485 @@
...
 
-  podDisruptionBudget: {
-    apiVersion: 'policy/v1',
-    kind: 'PodDisruptionBudget',
-    metadata: {
-      name: 'thanos-ruler-' + cfg.name,
-      namespace: cfg.namespace,
-      labels: {
-        thanosRulerName: cfg.name,
+    podDisruptionBudget: {
```
simonpasquier, line 465: there's no way to manage PDB in kube-thanos? If not, it would seem like a good addition

jsonnet/components/thanos-ruler.libsonnet
PhilipGough: I'm not sure. I can remove it if we like. I was just trying to keep the diff in this PR to a minimum.

jsonnet/components/thanos-ruler.libsonnet
PhilipGough: not in ruler libsonnet. We could propose it and I'm happy to contribute that, but I'm not sure it should block this PR? Or do we prefer to tackle it as one?

simonpasquier: agreed that it doesn't block this PR as long as we have a proper TODO comment.

jsonnet/components/thanos-ruler.libsonnet
simonpasquier: We should try to remove it and see what the CI says but not in this PR.

simonpasquier:
/lgtm
/label px-approved

no functional change.

PR #1479, opened 2021-11-15
Fix TestUserWorkloadMonitoringAlerting for Prometheus Operator v0.52

raptorsun:
Prometheus Operator version 0.52 has a new feature to append a postfix to the rule file name to avoid name collision. This breaks the E2E test [TestUserWorkloadMonitoringAlerting ](https://github.com/openshift/cluster-monitoring-operator/blob/4ecf1715bf3f572edb1c424cf3185f37719521fe/test/e2e/user_workload_monitoring_test.go#L733)which assumes rule file without this postfix.

This PR update that test so it can pass after upgrading Prometheus Operator to version 0.52.

rule file name before 0.52: ` /etc/prometheus/rules/prometheus-user-workload-rulefiles-0/user-workload-test-prometheus-example-rule-leaf.yaml`
rule file name after 0.52: `/etc/prometheus/rules/prometheus-user-workload-rulefiles-0/user-workload-test-prometheus-example-rule-leaf-f4c6359f-9510-4c1d-9b44-5aa014b3aacd.yaml`

### DIFF
--- a/test/e2e/user_workload_monitoring_test.go
+++ b/test/e2e/user_workload_monitoring_test.go
@@ -804,7 +804,6 @@ func assertTenancyForRules(t *testing.T) {
 		}
 
 		type testData struct {
-			file      string
 			ruleType  string
 			name      string
 			namespace string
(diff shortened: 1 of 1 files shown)


fpetkovski:
That assertion is testing an implementation detail of Prometheus Operator, there's no guarantee that the file name won't change in the future. Do we still need to have this test?

raptorsun:
> That assertion is testing an implementation detail of Prometheus Operator, there's no guarantee that the file name won't change in the future. Do we still need to have this test?

I think this test is for verifying the rule files are created correctly. 
Maybe we should ignore file names and check only name + namespace + rule type?

simonpasquier:
I agree with @fpetkovski: the test shouldn't check the filename.

raptorsun:
Test updated, no checks on file name now :)

simonpasquier (commented):
/lgtm

simonpasquier:
/label qe-approved
/label docs-approved
/label px-approved

this is only updating the e2e tests, no functional change.

PR #1480, opened 2021-11-16
test/e2e: fix assertTenancyForLabels test

simonpasquier:
The test didn't check the last returned error.

### DIFF
--- a/test/e2e/user_workload_monitoring_test.go
+++ b/test/e2e/user_workload_monitoring_test.go
@@ -575,13 +575,19 @@ func assertUserWorkloadRules(t *testing.T) {
 func assertTenancyForMetrics(t *testing.T) {
 	const testAccount = "test-metrics"
 
-	_, err := f.CreateServiceAccount(userWorkloadTestNs, testAccount)
+	err := framework.Poll(2*time.Second, 10*time.Second, func() error {
+		_, err := f.CreateServiceAccount(userWorkloadTestNs, testAccount)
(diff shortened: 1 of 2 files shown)


test/e2e/user_workload_monitoring_test.go
```diff
@@ -893,50 +899,51 @@ func assertTenancyForRules(t *testing.T) {
...
+		_, err = f.CreateRoleBindingFromClusterRole(userWorkloadTestNs, testAccount, "admin")
+		return err
+	})
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	var token string
-	err = framework.Poll(5*time.Second, 5*time.Minute, func() error {
+	err = framework.Poll(5*time.Second, time.Minute, func() error {
```
raptorsun, line 920: just curious, how can we know the proper timeout value for an API call in Kubernetes?

raptorsun (approved):
/lgtm

test/e2e/user_workload_monitoring_test.go
simonpasquier: I'm reducing the number of retries because the test doesn't wait for a change to be propagated. Here the retries are needed to deal with transient API errors so a time out of 1 minute is more than enough.

simonpasquier:
/label qe-approved
/label docs-approved
/label px-approved

this is only updating the e2e tests, no functional change.

PR #1483, opened 2021-11-17
MON-1659: adding relatedObjects to cluster operator manifest

prashbnair:

### DIFF
--- a/manifests/0000_50_cluster-monitoring-operator_06-clusteroperator.yaml
+++ b/manifests/0000_50_cluster-monitoring-operator_06-clusteroperator.yaml
@@ -11,3 +11,31 @@ status:
   versions:
   - name: operator
     version: "0.0.1-snapshot"
+  relatedObjects:
+    - group: ''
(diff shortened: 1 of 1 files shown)


prashbnair:
/retitle MON-1659: adding relatedObjects to cluster operator manifest

simonpasquier:
Do we still need the cluster monitoring operator to [update](https://github.com/openshift/cluster-monitoring-operator/blob/a6bc9824035ceb8dbfe7c53cf0c138bfb2ec5643/pkg/client/status_reporter.go#L49-L63) the relatedObjects field?

prashbnair:
> Do we still need the cluster monitoring operator to update the relatedObjects field?

I dont think it is required but I see that it is being set in 2 places
1. When the operator is being setup
2. When the operator conditions are set

Do we want to remove both of them?

simonpasquier:
After looking at what other operators are doing, it is fine to have CMO updating the `relatedObjects` field.

/lgtm

PR #1484, opened 2021-11-17
Add metrics for Node Tuning Operator to telemetry

dagrayvid:
This PR will add 2 metrics from the [Node Tuning Operator](https://github.com/openshift/cluster-node-tuning-operator) to telemetry. 

1. `nto_degraded_info`, tells whether NTO is in the degraded state. Will be used to gain insight into NTO downtime
2. `nto_custom_profiles:count` counts the number of nodes which are using custom TuneD profiles. Custom means that the profiles are not shipped by NTO, so any profile other than `openshift`, `openshift-node`, `openshift-control-plane`. This will give us an idea of how often users are applying custom tunings to nodes.

Signed-off-by: David Gray <dagray@redhat.com>


simonpasquier (commented):
Hey! Thanks for the PR.

I'm wondering whether the information about NTO being degraded doesn't already exist in Telemetry via other means like `cluster_operator_conditions{name="node-tuning",condition="Degraded"} == 1` or `alerts{alertname="NTODegraded"}`?

dagrayvid:
Thanks @simonpasquier for the review.

> I'm wondering whether the information about NTO being degraded doesn't already exist in Telemetry via other means like cluster_operator_conditions{name="node-tuning",condition="Degraded"} == 1 or alerts{alertname="NTODegraded"}?

 I went back and forth on this, mainly because the guide for adding metrics to telemetry suggested adding a metric for gauging the health of your component first. But I agree it can be removed because nto_degraded_info is redundant to information from cluster_operator_conditions. It should now be removed, PTAL.

simonpasquier (commented):
/lgtm
/hold

waiting for an official approval from @smarterclayton but the addition shouldn't be controversial since it's only add one series.

simonpasquier:
/hold cancel

Clayton already approved :)

PR #1485, opened 2021-11-18
fix generate-versions.sh to fetch version from packages.json

raptorsun:
A bugfix to fetch version from packages.json, useful for grafana package for now.


simonpasquier (commented):
it may be more readable to have a special case for Grafana (e.g `if $repo = "grafana" then fetch from package.json else fetch from VERSION`)?

raptorsun:
> it may be more readable to have a special case for Grafana (e.g `if $repo = "grafana" then fetch from package.json else fetch from VERSION`)?

that's better. I'm going to update the PR soon.

simonpasquier (commented):
/lgtm

simonpasquier:
/label px-approved
/label qe-approved
/label docs-approved

internal change, no user facing impact.

PR #1489, opened 2021-11-22
Bug 1955489: enable hard-anti affinity and PDB for Alertmanager

simonpasquier:
This change introduces hard pod anti-affinity rules and pod disruption
budgets for Alertmanager to ensure the maximum availability for the
Alertmanager cluster in the event of nodes going down (either due to
upgrades or unexpected outages). The cluster monitoring operator updates
the `Upgradeable` condition to false when it detects that the pods
aren't correctly spread to ensure that an upgrade only happens in safe
configurations.

The change also decreases the number of Alertmanager replicas from 3 to
2 to be consistent with the other monitoring components as well as the
  HA conventions stating that in general OpenShift component should run
with a replica count of 2 [1]. In addition with 3 replicas, it is
impossible to enable hard anti-affinity on nodes since 2 worker nodes is
a supported deployment for OCP.

The initial idea of running 3 replicas was to guarantee the replication
of data (silences + notifications) during pod roll-outs even if the user
didn't configure persistent storage. However given that no pod
disruption budget were defined, there was no guarantee that Kubernetes
would always keep one Alertmanager pod running.  With hard anti-affinity
and PDB, we are now sure that at least one Alertmanager pod is kept
running. And we are also setting up a startup probe that waits for at
least 20 seconds meaning that Kubernetes should wait for about 20
seconds after a new Alertmanager pod is running before considering
rolling out the next one. This interval of time should be more than
enough for the new Alertmanager to synchronize its data from the older
peer.

### DIFF
--- a/assets/alertmanager/alertmanager.yaml
+++ b/assets/alertmanager/alertmanager.yaml
@@ -23,17 +23,6 @@ spec:
         - openshift-monitoring
         topologyKey: kubernetes.io/hostname
   containers:
-  - name: alertmanager
-    startupProbe:
-      exec:
-        command:
-        - sh
-        - -c
-        - exec curl http://localhost:9093/-/ready
-      initialDelaySeconds: 20
-      periodSeconds: 1
-      successThreshold: 1
-      timeoutSeconds: 3
   - args:
     - -provider=openshift
     - -https-address=:9095
--- a/assets/alertmanager/alertmanager.yaml
+++ b/assets/alertmanager/alertmanager.yaml
@@ -12,19 +12,28 @@ metadata:
 spec:
   affinity:
     podAntiAffinity:
-      preferredDuringSchedulingIgnoredDuringExecution:
-      - podAffinityTerm:
-          labelSelector:
-            matchLabels:
-              app.kubernetes.io/component: alert-router
-              app.kubernetes.io/instance: main
-              app.kubernetes.io/name: alertmanager
-              app.kubernetes.io/part-of: openshift-monitoring
-          namespaces:
-          - openshift-monitoring
-          topologyKey: kubernetes.io/hostname
-        weight: 100
+      requiredDuringSchedulingIgnoredDuringExecution:
+      - labelSelector:
+          matchLabels:
+            app.kubernetes.io/component: alert-router
+            app.kubernetes.io/instance: main
+            app.kubernetes.io/name: alertmanager
+            app.kubernetes.io/part-of: openshift-monitoring
+        namespaces:
+        - openshift-monitoring
+        topologyKey: kubernetes.io/hostname
   containers:
+  - name: alertmanager
+    startupProbe:
+      exec:
+        command:
+        - sh
+        - -c
+        - exec curl http://localhost:9093/-/ready
+      initialDelaySeconds: 20
+      periodSeconds: 1
+      successThreshold: 1
+      timeoutSeconds: 3
   - args:
     - -provider=openshift
     - -https-address=:9095
@@ -150,7 +159,7 @@ spec:
       app.kubernetes.io/part-of: openshift-monitoring
       app.kubernetes.io/version: 0.22.2
   priorityClassName: system-cluster-critical
-  replicas: 3
+  replicas: 2
   resources:
     requests:
       cpu: 4m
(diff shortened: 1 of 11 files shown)


juzhao:
see from https://bugzilla.redhat.com/show_bug.cgi?id=1955489#c11, tested with the PR, Alertmanager Statefulsets now have 2 replicas and hard affinity set but pods can not be started

simonpasquier:
@juzhao yes the PR is still WIP and the CI fails for the same reason as you've noticed.

assets/alertmanager/alertmanager.yaml
```diff
@@ -149,7 +158,7 @@ spec:
       app.kubernetes.io/part-of: openshift-monitoring
       app.kubernetes.io/version: 0.22.2
   priorityClassName: system-cluster-critical
-  replicas: 3
+  replicas: 2
```
fpetkovski, line 161: Is this intended?

simonpasquier: yes I'll add more explanation to the commit message

simonpasquier: I've copied https://github.com/openshift/cluster-monitoring-operator/pull/1489/commits/0a05fb0daffd93ba743d15eca90be8262de694bf commit message into the PR description. Let me know if this is clear enough :)

simonpasquier:
/skip
/retitle Bug 1955489: enable hard-anti affinity and PDB for Alertmanager

pkg/manifests/manifests.go
```diff
@@ -463,6 +464,34 @@ func (f *Factory) AlertmanagerMain(host string, trustedCABundleCM *v1.ConfigMap)
...
 		}
+	} else if f.infrastructure.HighlyAvailableInfrastructure() {
+		// When no persistent storage is configured, add a startup probe to
+		// ensure that the Alertmanager container has time to replicate data
+		// from other peers before declaring itself as ready. This allows
+		// silences and notifications to be preserved on roll-outs. We assume
+		// that 20 seconds is enough for a full synchronization (this is twice
+		// the time Alertmanager waits before declaring that it can start
+		// sending notfications).
+		a.Spec.Containers = append(a.Spec.Containers,
```
fpetkovski, line 475: Should this change be made in the prometheus operator?

assets/alertmanager/pod-disruption-budget.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels:
+    app.kubernetes.io/component: alert-router
+    app.kubernetes.io/instance: main
+    app.kubernetes.io/name: alertmanager
+    app.kubernetes.io/part-of: openshift-monitoring
+    app.kubernetes.io/version: 0.22.2
+  name: alertmanager-main
+  namespace: openshift-monitoring
+spec:
+  maxUnavailable: 1
```
juzhao, line 13: `maxUnavailable` is fine, maybe `minAvailable` is better
```
# oc -n openshift-monitoring get pdb 
NAME                 MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
alertmanager-main    N/A             1                 1                     39m
prometheus-adapter   1               N/A               1                     50m
prometheus-k8s       1               N/A               1                     39m
thanos-querier-pdb   1               N/A               1                     38m
```

simonpasquier: `maxUnavailable` comes from upstream (https://github.com/prometheus-operator/kube-prometheus/blob/6d013d4e4f980ba99cfdafa9432819d484e2f829/jsonnet/kube-prometheus/components/alertmanager.libsonnet#L154) and my understanding is that because kube-prometheus deploys 3 replicas of Alertmanager, the choice was either `maxUnavailable: 1` or `minAvailable: 2`. Ideally we should settle on one field and automatically calculate the budget.

From https://kubernetes.io/docs/tasks/run-application/configure-pdb/

> The use of maxUnavailable is recommended as it automatically responds to changes in the number of replicas of the corresponding controller.

I'll check with the workloads team if there's any strong recommendation.

pkg/manifests/manifests.go
simonpasquier: Yes it makes sense to follow up upstream. BTW I notice that we have no readiness probe for Alertmanager...

assets/alertmanager/pod-disruption-budget.yaml
jan--f: :+1: for using `maxUnavailable` here. The note about the number of replicas is relevant here imo.

jan--f:
/lgtm
This looks great!

PR #1494, opened 2021-11-26
pkg/operator: move config sharing task at the end

simonpasquier:
The config sharing task depends on resources that are created by other
tasks (like Prometheus, Alertmanager, ...). It makes sense to move it as
the last task being executed by the operator.


simonpasquier:
/label px-approved
/label qe-approved
/label docs-approved
Overriding the labels because there are no visible change and the CI is green.

PR #1495, opened 2021-11-29
Bug 1953264: client: don't update resource fields that are managed externally

jan--f:
Some resources (ConfigMaps, Secrets, ValidatingWebhookConfigurations)
have fields, that are managed by the service-ca-operator. If we find
these fields populated we want to retain those values.
Otherwise CMO and service-ca-operator will compete on every sync loop
that CMO runs. While we don't watch those resources (that would lead to
an infinte competition) we do update them on every sync loop, i.e. CMO
will eventually stop competing with service-ca-operator, but until then
this causes unnecessary churn and can cause TLS errors during times when
CMO removes the TLS data that service-ca-operator populates.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -295,6 +295,14 @@ func (c *Client) CreateOrUpdateValidatingWebhookConfiguration(ctx context.Contex
 
 	required := w.DeepCopy()
 	required.ResourceVersion = existing.ResourceVersion
+	// retain the CABundle that service-ca-operator created if the proper annotation is found
+	if val, ok := required.Annotations["service.beta.openshift.io/inject-cabundle"]; ok && val == "true" {
+		for i := range required.Webhooks {
+			if len(existing.Webhooks[i].ClientConfig.CABundle) > 0 {
+				required.Webhooks[i].ClientConfig.CABundle = existing.Webhooks[i].ClientConfig.CABundle
+			}
+		}
+	}
 	_, err = admclient.Update(ctx, required, metav1.UpdateOptions{})
 	return errors.Wrap(err, "updating ValidatingWebhookConfiguration object failed")
 }
@@ -1113,9 +1121,42 @@ func (c *Client) CreateOrUpdateSecret(ctx context.Context, s *v1.Secret) error {
 
 	required := s.DeepCopy()
 	mergeMetadata(&required.ObjectMeta, existing.ObjectMeta)
-
+	// Check if the Secret has an owner reference to a Service, that carries
+	// the annotation with key
+	// service.beta.openshift.io/serving-cert-secret-name and the Secrets
+	// name as the value.
+	// This means that service-ca-operator controls and populates the two
+	// data fields tls.crt and tls.key. We want to retain those on updates
+	// if they existing and are not empty.
+	if c.maybeHasServiceCAData(ctx, required) {
+		if v, ok := existing.Data["tls.crt"]; ok && len(v) > 0 {
+			required.Data["tls.crt"] = v
+		}
+		if v, ok := existing.Data["tls.key"]; ok && len(v) > 0 {
+			required.Data["tls.key"] = v
+		}
+	}
 	_, err = sClient.Update(ctx, required, metav1.UpdateOptions{})
 	return errors.Wrap(err, "updating Secret object failed")
+}
+
+// maybeHasServiceCAData checks if the passed Secret s has at least one owner reference that
+// points to a Service with the annotation service.beta.openshift.io/serving-cert-secret-name: s.name
+func (c *Client) maybeHasServiceCAData(ctx context.Context, s *v1.Secret) bool {
+	for _, owner := range s.OwnerReferences {
+		if owner.Kind != "Service" {
+			continue
+		}
+		sclient := c.kclient.CoreV1().Services(s.GetNamespace())
+		svc, err := sclient.Get(ctx, owner.Name, metav1.GetOptions{})
+		if err != nil {
+			return false
+		}
+		if secName, ok := svc.Annotations["service.beta.openshift.io/serving-cert-secret-name"]; ok && secName == s.Name {
+			return true
+		}
+	}
+	return false
 }
 
 func (c *Client) CreateIfNotExistSecret(ctx context.Context, s *v1.Secret) error {
@@ -1162,6 +1203,12 @@ func (c *Client) CreateOrUpdateConfigMap(ctx context.Context, cm *v1.ConfigMap) 
 
 	required := cm.DeepCopy()
 	mergeMetadata(&required.ObjectMeta, existing.ObjectMeta)
+	if val, ok := required.Annotations["service.beta.openshift.io/inject-cabundle"]; ok && val == "true" {
+		// retain any service-ca data that service-ca-operator has created
+		if v, ok := existing.Data["service-ca.crt"]; ok && len(v) > 0 {
+			required.Data["service-ca.crt"] = v
+		}
+	}
 
 	_, err = cmClient.Update(ctx, required, metav1.UpdateOptions{})
 	return errors.Wrap(err, "updating ConfigMap object failed")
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -1113,6 +1119,12 @@ func (c *Client) CreateOrUpdateSecret(ctx context.Context, s *v1.Secret) error {
 
 	required := s.DeepCopy()
 	mergeMetadata(&required.ObjectMeta, existing.ObjectMeta)
+	if v, ok := existing.Data["tls.crt"]; ok && len(v) > 0 {
```
simonpasquier, line 1122: shouldn't we check if the service CA annotation is present and if yes, assume that the `Data` field is fully managed by the service CA operator (e.g. CMO only manages the metadata)?

jan--f: Checking the annotation is a good point yes.

While it's true for CMO, that only the service-ca-operator will touch the data field, simply by convention, I wouldn't put this in the code. The service-ca-operator will only update the keys that are checked here. Other potentially present keys would not be altered.
The current code will already remove data keys added by users. Should we ever add any fields to any Secret manifest, we might produce a subtle bug by copying the whole data field here.

jan--f: oh and there won't be an annotation on the Secret that we can check. But an annotation check on ConfigMaps and Vaildatingwebhooks makes sense.

jan--f: On a Secret several fields will be added by servce-ca-operator, though not sure if checking any of those makes sense:
- service-ca-operator will add a `ManagedField` like so:
```
managedFields:
    - manager: service-ca-operator
      ...
  ```
  - A type field is added `type: kubernetes.io/tls`
  - An annotation with the expiry date is added `annotations: service.beta.openshift.io/expiry: '2023-11-29T06:22:59Z'`
  
  All these would work, not sure if we should check on of those though.

Comment (simonpasquier) [reply pkg/client/client.go:1124]: Right I confused the annotation that CMO puts on the service with the resulting secret... 
From what I can tell:
* For any configmap with the `config.openshift.io/inject-trusted-cabundle=true` label ([custom CA bundle](https://docs.openshift.com/container-platform/4.9/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki)) or `service.beta.openshift.io/inject-cabundle=true` annotation ([service CA bundle](https://docs.openshift.com/container-platform/4.9/security/certificates/service-serving-certificate.html#add-service-certificate-configmap_service-serving-certificate)), CMO should keep the Data field untouched.
* The same goes for validating webhooks and API services with the `service.beta.openshift.io/inject-cabundle=true` annotation.
* For secrets, the situation is a bit tricky... Maybe CMO can assume that a secret of type `kubernetes.io/tls` with at least one annotation starting with the `service.beta.openshift.io/` prefix is managed by the service CA operator and the data shouldn't be modified?

Comment (fpetkovski) [reply pkg/client/client.go:1124]: I wonder if this is the right place to make the change because now `CreateOrUpdateSecret` needs domain specific knowledge about managing secret contents. Instead, I would expect the caller to manage the contents of the secret, and `CreateOrUpdateSecret` should just persist whatever is provided by the caller. 

Also, when it comes to TLS secrets, shouldn't CMO just create these secrets and never ever update them?

File: pkg/client/client_test.go [hunk 2]
```diff
@@ -459,6 +494,48 @@ func TestCreateOrUpdateConfigMap(t *testing.T) {
...
+			initialData: map[string]string{
+				"service-ca.crt": "",
+			},
+			updatedAnnotations: map[string]string{
+				"service.beta.openshift.io/inject-cabundle": "true",
+			},
+			expectedAnnotations: map[string]string{
+				"service.beta.openshift.io/inject-cabundle": "true",
+			},
+		},
```
raptorsun, line 538: will it be useful for a case where both annotation and certificate are absent?

pkg/client/client.go
jan--f: The various `CreateOrUpdate` functions in CMO already do this today. Since this is not a generic interface I don't see this as an issue.

Regarding the TLS secrets, CMO today follows the paradigm that all resources are immutable. As this PR shows allowances need to be made, but I don't see a reason to not keep reverting unwanted changes (say labels a user might attach to a TLS secret). I'm also unclear how we would reconcile changes that we want CMO to deploy on a version update.

simonpasquier: > Also, when it comes to TLS secrets, shouldn't CMO just create these secrets and never ever update them?

I think that CMO still needs to update the secrets because CMO may add/remove metadata (labels, annotations, owner references) between versions ([example](https://github.com/openshift/cluster-monitoring-operator/commit/0bc17b108cac36b09c10c355e0533101f9efdd3f#diff-0855bf7fb0e348899b2128ea27ba8725f294c9d7a28609925de4ccd1f94667b0)).

pkg/client/client.go
```diff
@@ -1113,11 +1121,44 @@ func (c *Client) CreateOrUpdateSecret(ctx context.Context, s *v1.Secret) error {
...
 	required := s.DeepCopy()
 	mergeMetadata(&required.ObjectMeta, existing.ObjectMeta)
-
+	// Check if the Secret has an owner reference to a Service, that carries
+	// the annotation with key
+	// service.beta.openshift.io/serving-cert-secret-name and the Secrets
+	// name as the value.
+	// This means that service-ca-operator controls and populates the two
+	// data fields tls.crt and tls.key. We want to retain those on updates
+	// if they existing and are not empty.
```
simonpasquier, line 1130: `if they are existing` or `if they exist`?

pkg/client/client.go
```diff
@@ -1113,11 +1121,44 @@ func (c *Client) CreateOrUpdateSecret(ctx context.Context, s *v1.Secret) error {
...
+// points to a Service with the annotation service.beta.openshift.io/serving-cert-secret-name: s.name
+func (c *Client) maybeHasServiceCAData(ctx context.Context, s *v1.Secret) bool {
+	for _, owner := range s.OwnerReferences {
+		if owner.Kind != "Service" {
+			continue
+		}
+		sclient := c.kclient.CoreV1().Services(s.GetNamespace())
+		svc, err := sclient.Get(ctx, owner.Name, metav1.GetOptions{})
+		if err != nil {
+			return false
```
simonpasquier, line 1153: why not return the error?

pkg/client/client.go
```diff
@@ -1113,11 +1121,44 @@ func (c *Client) CreateOrUpdateSecret(ctx context.Context, s *v1.Secret) error {
...
-
+	// Check if the Secret has an owner reference to a Service, that carries
+	// the annotation with key
+	// service.beta.openshift.io/serving-cert-secret-name and the Secrets
+	// name as the value.
+	// This means that service-ca-operator controls and populates the two
+	// data fields tls.crt and tls.key. We want to retain those on updates
+	// if they existing and are not empty.
+	if c.maybeHasServiceCAData(ctx, required) {
+		if v, ok := existing.Data["tls.crt"]; ok && len(v) > 0 {
```
simonpasquier, line 1132: (nit) I would have overwrite the `Data` field with the content from the existing secret.

simonpasquier (commented):
just a few nits but I'm fine with the merged code :)

pkg/client/client.go
jan--f: I went with the more specific approach here, since `service-ca-operator` only manages those fields too. If we retain the whole data field, we would potentially keep user content in the data field.

pkg/client/client.go
jan--f: I wasn't sure how to handle an error here. This current implementation tries to err on the side of caution and remove the tls data if we can't find an owning service for any reason. I suppose it would make sense to check if the error means the service doesn't exist.
I still don't have a clear intuition though how to handle another error. I think the sync loop will probably not even get here if there is another error (lets say we can't communication with the apiserver). And if it does, assuming this data is nothing we care about is still safe. We could also retain the data in that case, though it likely won't make a huge difference.
Actually, now that I think about it, this shouldn't `return` but `continue`. There could be another `OwnerReference` that we can retrieve.

PR #1498, opened 2021-11-30
hack/local-cmo.sh: fix CVO override

simonpasquier:
Since [1] got merged, CVO checks that the manifest's group matches with
the override definition hence the `/v1` needs to be removed.

[1] https://github.com/openshift/cluster-version-operator/pull/689


simonpasquier:
/hold

Just need to verify locally that it's indeed fixing the issue.

simonpasquier:
/label qe-approved
/label px-approved
/label docs-approved

No user facing change, this script is used by monitoring developers to test changes locally.

simonpasquier:
tested successfully

PR #1492, opened 2021-11-25
MON-1656: adding new target to makefile for verification

prashbnair:
* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

PR #1499, opened 2021-12-01
MON-1656: removing all the check tasks from other targets

prashbnair:
[PR](https://github.com/openshift/cluster-monitoring-operator/pull/1492)  created a new target named verify in the MakeFile. which runs all the check tasks. It is now running as a part of every PR build. 

This PR removes the check tasks that were spread across multiple targets.


arajkumar (approved):
/lgtm

PR #1505, opened 2021-12-03
jsonnet: Add external labels to platform Prometheus

PhilipGough:
See https://issues.redhat.com/browse/MON-2089

In order to effectively support multi-tenant, self-service alerting,
it is useful to allow identifying the source of the alert,
(i.e., platform vs UWM).

This change adds the external label 'openshift_io_alert_source': 'platform'
to platform monitoring Prometheus.


PhilipGough:
/label docs-approved

there is no customer facing docs impact here

PhilipGough:
/label px-approved

no product impact on this change

PhilipGough:
checking config after this change:

```yaml
global:
  scrape_interval: 30s
  scrape_timeout: 10s
  evaluation_interval: 30s
  external_labels:
    openshift_io_alert_source: platform
    prometheus: openshift-monitoring/k8s
    prometheus_replica: prometheus-k8s-1
```

simonpasquier (commented):
I realize that using external labels isn't the right approach since it would add the label to the time series too. I guess that we need to use `AdditionalAlertRelabelConfigs` instead.

PhilipGough:
/hold

After chatting with @simonpasquier we decided we may not want to progress this right now with this solution.

PhilipGough:
Closing in favour of #1508

PR #1506, opened 2021-12-06
Bug 2027272: Humanize bytes value for KubeMemoryOvercommit

PhilipGough:


jan--f (commented):
/lgtm

juzhao:
tested with PR, the value is humanized, example
alert description now is
`Cluster has overcommitted memory resource requests for Pods by 76.45G bytes and cannot tolerate node failure.`
/lgtm

PR #1507, opened 2021-12-07
test/e2e/framework: retry configmap actions

simonpasquier:

### DIFF
--- a/test/e2e/framework/helpers.go
+++ b/test/e2e/framework/helpers.go
@@ -22,7 +22,10 @@ const (
 func (f *Framework) MustCreateOrUpdateConfigMap(t *testing.T, cm *v1.ConfigMap) {
 	t.Helper()
 	ensureCreatedByTestLabel(cm)
-	if err := f.OperatorClient.CreateOrUpdateConfigMap(ctx, cm); err != nil {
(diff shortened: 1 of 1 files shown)


test/e2e/framework/helpers.go
```diff
@@ -22,16 +22,22 @@ const (
...
 
 // MustDeleteConfigMap or fail the test
 func (f *Framework) MustDeleteConfigMap(t *testing.T, cm *v1.ConfigMap) {
 	t.Helper()
-	if err := f.OperatorClient.DeleteConfigMap(ctx, cm); err != nil {
-		t.Fatalf("failed to configmap configmap - %s", err.Error())
+	err := Poll(time.Second, 10*time.Second, func() error {
+		return f.OperatorClient.DeleteConfigMap(ctx, cm)
+	})
+	if err != nil {
```
PhilipGough, line 39: ~should we check for error not found here in case we hit a race? In the end, we wont care what deleted the CM, just that it is gone IMO.~

Never-mind, I see the calling function implements this already. Maybe we should rename that to `DeleteConfigMapIfExists` at some point for clarity.

PhilipGough (approved):
/lgtm

PR #1508, opened 2021-12-07
MON-2089: Ensure we can identify the source of alerts from Platform Prometheus

PhilipGough:
See https://issues.redhat.com/browse/MON-2089

In order to effectively support multi-tenant, self-service alerting,
it is useful to allow identifying the source of the alert,
(i.e., platform vs UWM).

This change adds the external label 'openshift_io_alert_source': 'platform'
to platform monitoring Prometheus.

### DIFF
--- a/test/e2e/alertmanager_test.go
+++ b/test/e2e/alertmanager_test.go
@@ -16,6 +16,7 @@ package e2e
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -508,4 +509,49 @@ func TestAlertmanagerDisabling(t *testing.T) {
 			t.Run(assertion.name, assertion.assertion)
 		}
 	})
+}
+
+func TestAlertManagerHasAdditionalAlertRelabelConfigs(t *testing.T) {
+	const (
+		expectPlatformLabel      = "openshift_io_alert_source"
+		expectPlatformLabelValue = "platform"
+	)
+
+	type Alerts []struct {
+		Labels map[string]string `json:"labels"`
+	}
+
+	var alerts Alerts
+
+	err := framework.Poll(5*time.Second, time.Minute, func() error {
+		resp, err := f.AlertmanagerClient.Do("GET", "/api/v2/alerts", nil)
+		if err != nil {
+			return err
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK {
+			return fmt.Errorf("expecting 200 status code, got %d (%q)", resp.StatusCode, resp.Body)
+		}
+
+		if err := json.NewDecoder(resp.Body).Decode(&alerts); err != nil {
+			return fmt.Errorf("error decoding alert response")
+		}
+
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, alert := range alerts {
+		v, found := alert.Labels[expectPlatformLabel]
+		if !found {
+			t.Fatal("expected correct label to be present")
+		}
+
+		if v != expectPlatformLabelValue {
+			t.Fatalf("expected correct value for %s but got %s", expectPlatformLabel, v)
+		}
+	}
 }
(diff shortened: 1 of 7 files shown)


PhilipGough:
Checked the Prometheus config to ensure it contains the `alert_relabel_configs`:

```yaml
global:
  scrape_interval: 30s
  scrape_timeout: 10s
  evaluation_interval: 30s
  external_labels:
    prometheus: openshift-monitoring/k8s
    prometheus_replica: prometheus-k8s-0
alerting:
  alert_relabel_configs:
  - separator: ;
    regex: prometheus_replica
    replacement: $1
    action: labeldrop
  - separator: ;
    regex: (.*)
    target_label: openshift_io_alert_source
    replacement: platform
    action: replace
```

simonpasquier:
the change looks good to me but can you add an e2e test checking that the label is present in Alertmanager?

test/e2e/alertmanager_test.go
```diff
@@ -509,3 +510,48 @@ func TestAlertmanagerDisabling(t *testing.T) {
...
+			t.Fatal("expected correct label to be present")
+		}
+
+		if v != expectPlatformLabelValue {
+			t.Fatalf("expected correct value for %s but got %s", expectPlatformLabel, v)
+		}
+	}
+}
+
+type alerts []struct {
```
raptorsun, line 555: It may be better to move this type to line 520 as an inline declaration if we are not using it for other tests?

test/e2e/alertmanager_test.go
```diff
@@ -509,3 +510,48 @@ func TestAlertmanagerDisabling(t *testing.T) {
...
+	)
+
+	type Alerts []struct {
+		Labels map[string]string `json:"labels"`
+	}
+
+	var alerts Alerts
+
+	err := framework.Poll(5*time.Second, time.Minute, func() error {
+		resp, err := f.AlertmanagerClient.Do("GET", "/api/v2/alerts", nil)
```
simonpasquier, line 527: we could also filter on alerts with `openshift_io_alert_source="platform"`. Similar to:

https://github.com/openshift/cluster-monitoring-operator/blob/1824f9c9a39f54734298dd10e5d20d42c8247995/test/e2e/alertmanager_test.go#L399-L402

test/e2e/alertmanager_test.go
```diff
@@ -509,3 +510,48 @@ func TestAlertmanagerDisabling(t *testing.T) {
...
+	}
+
+	for _, alert := range alerts {
+		v, found := alert.Labels[expectPlatformLabel]
+		if !found {
+			t.Fatal("expected correct label to be present")
+		}
+
+		if v != expectPlatformLabelValue {
+			t.Fatalf("expected correct value for %s but got %s", expectPlatformLabel, v)
```
simonpasquier, line 554: `v` would be an empty string

raptorsun: why should we worry about an empty string v? It will fail the condition here expectPlatformLabelValue = "platform".

simonpasquier: oops sorry my bad!

test/e2e/alertmanager_test.go
PhilipGough: would we want to filter though? my understanding was the test should check that all alerts have the label applied?

simonpasquier: as long as no UWM alerts is firing, it should be fine indeed.

PhilipGough: well that is a good point actually, although we did put a fair amount of work in to ensure the tests are isolated and each run is cleaned up, so if it fails in its current format, then it probably highlights a different problem  :)

PhilipGough:
@sferich888 - can we consider this for px approval?

juzhao:
tested with the PR, 
1. for the alerts from Platform Prometheus, only see the "openshift_io_alert_source": "platform" label in alertmanager API and UI,  no such label from thanos-querier/prometheus API, is it expected?
resullt see from: http://pastebin.test.redhat.com/1016231
2. for the alerts from user workload, do we have the same task to identify the alerts are from user project?

PhilipGough:
1. It is expected since the additional label is used to differentiate alerts when routing them in Alertmanager

2. For now, user defined alerts can be identified with openshift_io_alert_source=""

PR #1509, opened 2021-12-08
NE-704: add NLB usage metrics for network edge

tjungblu:

exposing the metric from https://github.com/openshift/cluster-monitoring-operator/pull/1509


tjungblu:
/hold 
until we merge the other PR :)

simonpasquier (commented):
Looking at https://github.com/openshift/cluster-ingress-operator/pull/675 the metric has a `name` label. What could be the cardinality of that label?

tjungblu:
@simonpasquier yeah good catch, I don't think that would even be necessary. Let me patch that out of the metric.

jsonnet/rules.libsonnet
```diff
@@ -513,6 +513,11 @@ function(params) {
           record: 'cluster:usage:openshift:ingress_request_total:irate5m',
           // The instantaneous rate of openshift requests per second arriving at the ingress controllers
         },
+        {
+          expr: 'sum(ingress_controller_aws_nlb_active) or absent(__does_not_exist__)*0',
```
simonpasquier, line 517:
```suggestion
          expr: 'sum(ingress_controller_aws_nlb_active) or vector(0)',
```

simonpasquier:
/lgtm

/label px-approved
/label qe-approved
/label docs-approved

PX, docs and QE have alreay approved in https://issues.redhat.com/browse/NE-704. Given that it's only adding 1 series per cluster, I expect that @smarterclayton is ok with it so I'm not holding the PR.

PR #1510, opened 2021-12-09
Bug 2012426: Add namespace label for all thanos alerts

arajkumar:
This PR adds namespace label through thanos `targetGroup` config
variable[1].

[1] https://github.com/thanos-io/thanos/blob/d1acaea2a11a3e4db6bb435c98dea63c517e3530/mixin/config.libsonnet#L7-L26


arajkumar:
Upstream PR to add namespace for thanos sidecar alerts: https://github.com/prometheus-operator/kube-prometheus/pull/1543

PR #1512, opened 2021-12-13
[MON-1950] Remove UI access from the Thanos routes

raptorsun:

### DIFF
--- a/jsonnet/components/thanos-querier.libsonnet
+++ b/jsonnet/components/thanos-querier.libsonnet
@@ -37,6 +37,7 @@ function(params)
         labels: tq.config.commonLabels,
       },
       spec: {
+        path: '/api',
         to: {
           kind: 'Service',
           name: 'thanos-querier',
--- a/assets/thanos-querier/route.yaml
+++ b/assets/thanos-querier/route.yaml
@@ -10,6 +10,7 @@ metadata:
   name: thanos-querier
   namespace: openshift-monitoring
 spec:
+  path: /api
   port:
     targetPort: web
   tls:
--- a/assets/thanos-ruler/route.yaml
+++ b/assets/thanos-ruler/route.yaml
@@ -4,6 +4,7 @@ metadata:
   name: thanos-ruler
   namespace: openshift-user-workload-monitoring
 spec:
+  path: /api
   port:
     targetPort: web
   tls:
--- a/jsonnet/components/thanos-ruler.libsonnet
+++ b/jsonnet/components/thanos-ruler.libsonnet
@@ -40,6 +40,9 @@ function(params)
         namespace: tr.config.namespace,
       },
       spec: {
+        // restrict to Thanos Rule API endpoint only
+        // ref: https://github.com/thanos-io/thanos/blob/v0.24.0/cmd/thanos/rule.go#L657
+        path: '/api',
         to: {
           kind: 'Service',
           name: tr.config.name,
--- a/test/e2e/thanos_ruler_test.go
+++ b/test/e2e/thanos_ruler_test.go
@@ -49,7 +49,6 @@ func TestUserWorkloadThanosRulerWithAdditionalAlertmanagers(t *testing.T) {
 				{"assert thanos ruler rollout", assertThanosRulerDeployment},
 				{"create additional alertmanager", createAlertmanager},
 				{"create alerting rule that always fires", createPrometheusRule},
-				{"start alertmanager port forward", startAlertmanagerPortForward},
 				{"verify alertmanager received the alert", verifyAlertmanagerAlertReceived},
 			},
 		},
@@ -117,20 +116,16 @@ func createPrometheusRule(t *testing.T) {
 	}
 }
 
-func startAlertmanagerPortForward(t *testing.T) {
-	if err := f.StartPortForward(
-		"https",
-		"alertmanager-alertmanager-e2e-test-0",
-		f.UserWorkloadMonitoringNs,
-		"9093",
-	); err != nil {
+func verifyAlertmanagerAlertReceived(t *testing.T) {
+
+	host, cleanUp, err := f.ForwardPort(t, "alertmanager-operated", 9093)
+	if err != nil {
 		t.Fatal(err)
 	}
-}
+	t.Cleanup(cleanUp)
 
-func verifyAlertmanagerAlertReceived(t *testing.T) {
-	err := framework.Poll(time.Second, 5*time.Minute, func() error {
-		resp, err := http.Get("http://localhost:9093/api/v2/alerts")
+	err = framework.Poll(time.Second, 5*time.Minute, func() error {
+		resp, err := http.Get(fmt.Sprintf("http://%s/api/v2/alerts", host))
 		if err != nil {
 			return err
 		}


raptorsun:
About test "[sig-instrumentation] Prometheus when installed on the cluster should start and expose a secured proxy and unsecured metrics":
This test of calling Thanos querier root URL without authentication should return [401 unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) instead of [403 forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403).
[The test code ](https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus.go#L584) sends a request using the curl command `curl -k -s -o /dev/null -w 403 https://thanos-querier.openshift-monitoring.svc:9091` without providing any information for authentication. A response with status code 401 should be expected.
I'm going to submit a PR to openshift/origin to correct it.

assets/thanos-querier/deployment.yaml
```diff
@@ -80,59 +80,45 @@ spec:
...
-            port: 9091
+            port: web
             scheme: HTTPS
           initialDelaySeconds: 5
           periodSeconds: 5
         resources:
           requests:
             cpu: 1m
-            memory: 20Mi
+            memory: 15Mi
```
simonpasquier, line 115: i would leave the requests as-is

raptorsun: OK, I will change it to 20MiB. 
From which criteria do we know this kube-rbac-proxy should require 20MiB memory while its neighbor container kube-rbac-proxy for tenancy requires 15MiB? 🤔

simonpasquier: https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#resources-and-limits describes how to determine resource requests for OpenShift components. The other kube-rbac-proxy is only for serving /metrics while this container would serve the /api/v1/* requests so I would expect that the latter uses more resources. I think it is wiser to leave the requests untouched for now and check later if we can decrease the values.

pkg/manifests/manifests.go
```diff
@@ -222,8 +222,7 @@ var (
 	ThanosQuerierServiceMonitor         = "thanos-querier/service-monitor.yaml"
 	ThanosQuerierPrometheusRule         = "thanos-querier/prometheus-rule.yaml"
 	ThanosQuerierRoute                  = "thanos-querier/route.yaml"
-	ThanosQuerierOauthCookieSecret      = "thanos-querier/oauth-cookie-secret.yaml"
```
simonpasquier, line 225: CMO needs to remove the associated objects so that we don't leave anything behind after upgrading from 4.9 to 4.10.

raptorsun:
> About test "[sig-instrumentation] Prometheus when installed on the cluster should start and expose a secured proxy and unsecured metrics": This test of calling Thanos querier root URL without authentication should return [401 unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) instead of [403 forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403). [The test code ](https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus.go#L584) sends a request using the curl command `curl -k -s -o /dev/null -w 403 https://thanos-querier.openshift-monitoring.svc:9091` without providing any information for authentication. A response with status code 401 should be expected. Same goes for the test "[sig-instrumentation][sig-builds][Feature:Builds] Prometheus when installed on the cluster should start and expose a secured proxy and verify build metrics",too. I'm going to submit a PR to openshift/origin to correct it.

here is the PR on openshift/origin: https://github.com/openshift/origin/pull/26695/files

jsonnet/components/thanos-querier.libsonnet
```diff
@@ -408,27 +389,25 @@ function(params)
                 ],
               },
               {
-                name: 'oauth-proxy',
-                image: 'quay.io/openshift/oauth-proxy:latest',  //FIXME(paulfantom)
+                // This proxy takes over the old oauth proxy and is used to proxy prometheus metrics as well as queries from Admin console.
```
simonpasquier, line 392: I wouldn't mention the previous existence of OAuth proxy.
 
```suggestion
                // This proxy is used to proxy requests from the Admin console to the Thanos API.
```

jsonnet/components/thanos-querier.libsonnet
```diff
@@ -438,50 +417,37 @@ function(params)
...
                   {
-                    mountPath: '/etc/proxy/secrets',
-                    name: 'secret-thanos-querier-oauth-cookie',
+                    mountPath: '/etc/kube-rbac-proxy',
+                    name: 'secret-' + $.kubeRbacProxyWebSecret.metadata.name,
                   },
                 ],
               },
               {
+                // This proxy is used by frontend Developer console, with enforced namespace label.
```
simonpasquier, line 450:
```suggestion
                // This proxy is used to proxy requests from the Developer console to the Thanos API.
                // It enforces the presence of the namespace query parameter.
```

pkg/tasks/thanos_querier.go
```diff
@@ -63,55 +62,24 @@ func (t *ThanosQuerierTask) Run(ctx context.Context) error {
...
 
-	s, err := t.factory.ThanosQuerierOauthCookieSecret()
+	rs, err := t.factory.ThanosQuerierRBACProxySecret()
 	if err != nil {
-		return errors.Wrap(err, "initializing Thanos Querier OAuth Cookie Secret failed")
+		return errors.Wrap(err, "initializing Thanos Querier RBAC proxy Secret failed")
 	}
 
-	err = t.client.CreateIfNotExistSecret(ctx, s)
+	err = t.client.CreateIfNotExistSecret(ctx, rs)
```
simonpasquier, line 70:
```suggestion
	err = t.client.CreateOrUpdateSecret(ctx, rs)
```

pkg/tasks/thanos_querier.go
```diff
@@ -63,55 +62,24 @@ func (t *ThanosQuerierTask) Run(ctx context.Context) error {
...
 	}
 
-	rs, err := t.factory.ThanosQuerierRBACProxySecret()
+	rs, err = t.factory.ThanosQuerierRBACProxyWebSecret()
 	if err != nil {
-		return errors.Wrap(err, "initializing Thanos Querier RBAC proxy Secret failed")
+		return errors.Wrap(err, "initializing Thanos Querier RBAC proxy Web Secret failed")
 	}
 
 	err = t.client.CreateIfNotExistSecret(ctx, rs)
```
simonpasquier, line 80: if should be `CreateOrUpdateSecret` here too.

pkg/tasks/thanos_querier.go
```diff
@@ -268,6 +236,18 @@ func (t *ThanosQuerierTask) Run(ctx context.Context) error {
 	if err != nil {
 		return errors.Wrap(err, "reconciling Thanos Querier PrometheusRule failed")
 	}
+	// TODO: remove removeRemnantsOauthProxy when we upgrade from version 4.11
```
simonpasquier, line 239:
```suggestion
	// TODO: remove removeRemnantsOauthProxy in 4.11.
```

raptorsun:
TestUserWorkloadMonitoringMetrics

simonpasquier:
I think that the `ci/prow/e2e-agnostic-operator` failure needs an investigation.

simonpasquier (commented):
we should also try to remove the `-openshift-sar={"resource": "namespaces", "verb": "get"}` argument from oauth-proxy:

https://github.com/openshift/cluster-monitoring-operator/blob/781a7ca5b6c4242cad7d47eeef91853e53e58cc1/jsonnet/components/thanos-querier.libsonnet#L456

~~otherwise a user accessing the route with a browser could see the "login with openshift" page but it would fail if they click the button~~

simonpasquier:
/hold
we should tackle the Thanos ruler route in openshift-user-workload-monitoring too.

raptorsun:
> /hold we should tackle the Thanos ruler route in openshift-user-workload-monitoring too.

Route is restricted to /api for Thanos Ruler now :)

raptorsun:
Hello @kyoto , Is there some reference to Thanos Ruler API in openshift/console that needs to be tested?

simonpasquier:
/skip
/hold cancel

@raptorsun the console backend only uses the Thanos query API.

simonpasquier:
/label docs-approved
/label px-approved

The associated epic (https://issues.redhat.com/browse/MON-1962) has the acks from docs and PX.

raptorsun:
> /skip /hold cancel
> 
> @raptorsun the console backend only uses the Thanos query API.

Thank you :) No additional test will be required for Thanos Ruler then.

raptorsun:
Hello @juzhao , this PR is ready for QE testing.

juzhao:
> Hello @juzhao , this PR is ready for QE testing.

@lihongyan1 will test it

raptorsun:
> > About test "[sig-instrumentation] Prometheus when installed on the cluster should start and expose a secured proxy and unsecured metrics": This test of calling Thanos querier root URL without authentication should return [401 unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) instead of [403 forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403). [The test code ](https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus.go#L584) sends a request using the curl command `curl -k -s -o /dev/null -w 403 https://thanos-querier.openshift-monitoring.svc:9091` without providing any information for authentication. A response with status code 401 should be expected. Same goes for the test "[sig-instrumentation][sig-builds][Feature:Builds] Prometheus when installed on the cluster should start and expose a secured proxy and verify build metrics",too. I'm going to submit a PR to openshift/origin to correct it.
> 
> here is the PR on openshift/origin: https://github.com/openshift/origin/pull/26695/files

The PR on openshift/origin is no longer required for this PR. We can safely ignore that now.

kyoto:
@raptorsun Yes, all the Observe > Alerting pages will need to be tested for both the Administrator and Developer Consoles as well as the alert notifications (at the top right of the Console UI).

CC @lihongyan1

lihongyan1:
Tested the pr following steps in case
https://polarion.engineering.redhat.com/polarion/redirect/project/OSE/workitem?id=OCP-47420
https://polarion.engineering.redhat.com/polarion/redirect/project/OSE/workitem?id=OCP-47520

Everything looks well, except a minor issue
https://bugzilla.redhat.com/show_bug.cgi?id=2045975

PR #1519, opened 2021-12-17
Enable the /api/v1/series endpoint on the Thanos tenancy port

simonpasquier:
Follow-up of #1299


simonpasquier:
/hold 
the operator e2e failure looks legit

jan--f:
/lgtm
/label px-approved
/label docs-approved
/label qe-approved
Adding labels since this doesn't change any core functionality but rather exposes an optional feature.

PR #1523, opened 2021-12-21
Bug 2033751: Bump go dependencies to latest version

arajkumar:
This PR does the following

1. Bumps all go deps to latest
2. Fixes build errors
3. Revendors
4. Adds a [make target](https://github.com/prometheus-operator/prometheus-operator/blob/dec443d201f008a5d30c62b153b2558638cdc34e/Makefile#L165-L172) to ease the go deps bump


pkg/operator/operator.go
```diff
@@ -330,6 +330,9 @@ func New(
 		"OpenShiftMonitoringClientCertRequester",
 	)
 
+	if err != nil {
+		return nil, err
```
simonpasquier, line 334:
```suggestion
		return nil, errors.Wrap(err, "failed to create client certificate controller")
```

Makefile
```diff
@@ -88,6 +88,13 @@ vendor:
 	go mod vendor
 	go mod verify
 
+.PHONY: update-go-deps
+update-go-deps:
+	for m in $$(go list -mod=readonly -m -f '{{ if and (not .Indirect) (not .Main)}}{{.Path}}{{end}}' all); do \
+		go get $$m; \
+	done
+	@echo "Don't forget to run 'make tidy'"
```
simonpasquier, line 96:
```suggestion
	@echo "Don't forget to run 'make vendor'"
```

go.mod
```diff
@@ -6,28 +6,31 @@ require (
...
-	github.com/prometheus-operator/prometheus-operator v0.52.1
-	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.52.1
-	github.com/prometheus-operator/prometheus-operator/pkg/client v0.52.1
+	github.com/prometheus-operator/prometheus-operator v0.53.1
+	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.53.1
+	github.com/prometheus-operator/prometheus-operator/pkg/client v0.53.1
 	github.com/prometheus/client_golang v1.11.0
-	github.com/prometheus/prometheus v1.8.2-0.20211005150130-f29caccc4255 // v1.8.2 is misleading as Prometheus does not have v2 module. This is pointing to v2.26.1, the same as in prometheus-operator v0.52.1
-	golang.org/x/crypto v0.0.0-20210616213533-5ff15b29337e
+	github.com/prometheus/prometheus v2.5.0+incompatible
```
simonpasquier, line 18: why not use the pseudo-version directly here? In general, I prefer to use `replace` only for dependencies that can't be resolved properly by `go mod` (like k8s.io/client-go in this file).

```suggestion
	github.com/prometheus/prometheus v1.8.2-0.20211214150951-52c693a63be1
```

damemi:
/retitle Bug 2033751: Bump go dependencies to latest version
Feel free to change to another BZ if you want to track it that way. We are tracking all the other bumps with this one

PR #1527, opened 2022-01-04
Reconcile AlertManagerConfig in UWM from platform Prometheus

PhilipGough:
Related to https://github.com/openshift/enhancements/pull/937 - this changes allows alerts created in user space when UWM is enabled to be routed via AlertManagerConfig to platform alertmanager

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - [#1491](https://github.com/openshift/cluster-monitoring-operator/pull/1491) Rename alerts `AggregatedAPIErrors to KubeAggregatedAPIErrors` and `AggregatedAPIDown to KubeAggregatedAPIDown`.
 - [#1488](https://github.com/openshift/cluster-monitoring-operator/pull/1488) Removing the alert HighlyAvailableWorkloadIncorrectlySpread.
 - [#1858](https://github.com/openshift/cluster-monitoring-operator/pull/1858) Allow suppression of storage alerts via PersistentVolumeClaim label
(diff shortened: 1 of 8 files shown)


pkg/client/client.go
```diff
@@ -439,6 +440,26 @@ func (c *Client) CreateOrUpdateAlertmanager(ctx context.Context, a *monv1.Alertm
 	return errors.Wrap(err, "updating Alertmanager object failed")
 }
 
+func (c *Client) CreateOrUpdateAlertmanagerConfig(ctx context.Context, a *monv1alpha.AlertmanagerConfig) error {
```
simonpasquier, line 443: should this be moved to the `test/` directory since CMO itself doesn't manage AlertmanagerConfig resources?

pkg/client/client.go
```diff
@@ -698,6 +719,18 @@ func (c *Client) DeletePrometheusRuleByNamespaceAndName(ctx context.Context, nam
 	return nil
 }
 
+func (c *Client) DeleteAlertManagerConfigByNamespaceAndName(ctx context.Context, namespace, name string) error {
```
simonpasquier, line 722: ditto

pkg/client/client.go
PhilipGough: yes, makes sense

juzhao:
testing passed:
/label qe-approved

simonpasquier (commented):
/lgtm

simonpasquier:
/label docs-approved
/label px-approved

Adding missing labels since https://issues.redhat.com/browse/MON-880 have been ack'ed already.

PR #1528, opened 2022-01-04
Bug 2035046: Recover Platform CPU by Reducing the Kubelet ServiceMonitor Scrape Interval

pixelsoccupied:
The change is required for core reduction in SNOs. 

Currently kubelet servicemonitor scrape interval is hard coded to 30 secs [here](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/control-plane/service-monitor-kubelet.yaml). With new Prometheus [feature](https://promlabs.com/blog/2021/09/14/whats-new-in-prometheus-2-30#controlling-scrape-intervals-and-timeouts-via-relabeling) we should able to change the value dynamically (preferably to 60 secs) and in-turn reduce CPU usage. Jira [link](https://issues.redhat.com/browse/CNF-2893)

- [x] I added CHANGELOG entry for this change.
- [ ] No user facing changes, so no entry in CHANGELOG was needed.

### DIFF
--- a/jsonnet/components/control-plane.libsonnet
+++ b/jsonnet/components/control-plane.libsonnet
@@ -106,6 +106,13 @@ function(params)
                       action: 'labeldrop',
                       regex: '__tmp_keep_metric',
                     },
+                    // allow kubelet to dynamically change its scrape interval during runtime
+                    {
+                      action: 'replace',
+                      regex: '(.+)',
+                      sourceLabels: ['__meta_kubernetes_service_annotation_openshift_io_scrape_interval'],
+                      targetLabel: '__scrape_interval__'
+                    }
                   ],
                 }
               else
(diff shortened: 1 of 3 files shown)


jsonnet/components/control-plane.libsonnet
```diff
@@ -106,6 +106,13 @@ function(params)
                       action: 'labeldrop',
                       regex: '__tmp_keep_metric',
                     },
+                    // allow kubelet to dynamically change its scrape interval during runtime
+                    {
+                      action: 'replace',
+                      regex: '(.+)',
+                      sourceLabels: ['__meta_kubernetes_service_annotation_openshift_io_scrape_interval'],
+                      targetLabel: '__scrape_interval__'
```
raptorsun, line 114: When using __meta_kubernetes_service_annotation_openshift_io_scrape_interval metric to set the scrape interval of the target, is there any mechanism to limit the maximum value for this interval?
Some rules require a minimum granularity of 60 seconds.

PR #1529, opened 2022-01-05
[MON-1630] Remove UI access to AlertManager

raptorsun:

### DIFF
--- a/jsonnet/components/alertmanager.libsonnet
+++ b/jsonnet/components/alertmanager.libsonnet
@@ -22,6 +22,7 @@ function(params)
         namespace: cfg.namespace,
       },
       spec: {
+        path: '/api/v2',
         to: {
           kind: 'Service',
           name: 'alertmanager-main',
--- a/jsonnet/components/alertmanager.libsonnet
+++ b/jsonnet/components/alertmanager.libsonnet
@@ -22,7 +22,7 @@ function(params)
         namespace: cfg.namespace,
       },
       spec: {
-        path: '/api/v2',
+        path: '/api',
         to: {
           kind: 'Service',
           name: 'alertmanager-main',
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -258,6 +258,19 @@ func (c *Client) ApiServersListWatchForResource(ctx context.Context, resource st
 	}
 }
 
+func (c *Client) ConsoleListWatch(ctx context.Context) *cache.ListWatch {
+	consoleInterface := c.oscclient.ConfigV1().Consoles()
+
+	return &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			return consoleInterface.List(ctx, metav1.ListOptions{})
+		},
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			return consoleInterface.Watch(ctx, options)
+		},
+	}
+}
+
 func (c *Client) AssurePrometheusOperatorCRsExist(ctx context.Context) error {
 	return wait.Poll(time.Second, time.Minute*5, func() (bool, error) {
 		_, err := c.mclient.MonitoringV1().Prometheuses(c.namespace).List(ctx, metav1.ListOptions{})
@@ -370,6 +383,10 @@ func (c *Client) GetInfrastructure(ctx context.Context, name string) (*configv1.
 
 func (c *Client) GetAPIServerConfig(ctx context.Context, name string) (*configv1.APIServer, error) {
 	return c.oscclient.ConfigV1().APIServers().Get(ctx, name, metav1.GetOptions{})
+}
+
+func (c *Client) GetConsoleConfig(ctx context.Context, name string) (*configv1.Console, error) {
+	return c.oscclient.ConfigV1().Consoles().Get(ctx, name, metav1.GetOptions{})
 }
 
 func (c *Client) GetConfigmap(ctx context.Context, namespace, name string) (*v1.ConfigMap, error) {
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/openshift/library-go/pkg/crypto"
 
+	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	securityv1 "github.com/openshift/api/security/v1"
 	"github.com/openshift/cluster-monitoring-operator/pkg/promqlgen"
@@ -294,6 +295,7 @@ type Factory struct {
 	proxy                 ProxyReader
 	assets                *Assets
 	APIServerConfig       *APIServerConfig
+	consoleConfig         *configv1.Console
 }
 
 // InfrastructureReader has methods to describe the cluster infrastructure.
@@ -309,7 +311,7 @@ type ProxyReader interface {
 	NoProxy() string
 }
 
-func NewFactory(namespace, namespaceUserWorkload string, c *Config, infrastructure InfrastructureReader, proxy ProxyReader, a *Assets, apiServerConfig *APIServerConfig) *Factory {
+func NewFactory(namespace, namespaceUserWorkload string, c *Config, infrastructure InfrastructureReader, proxy ProxyReader, a *Assets, apiServerConfig *APIServerConfig, consoleConfig *configv1.Console) *Factory {
 	return &Factory{
 		namespace:             namespace,
 		namespaceUserWorkload: namespaceUserWorkload,
@@ -318,6 +320,7 @@ func NewFactory(namespace, namespaceUserWorkload string, c *Config, infrastructu
 		proxy:                 proxy,
 		assets:                a,
 		APIServerConfig:       apiServerConfig,
+		consoleConfig:         consoleConfig,
 	}
 }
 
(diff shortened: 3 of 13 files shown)


jsonnet/components/alertmanager.libsonnet
```diff
@@ -22,6 +22,29 @@ function(params)
         namespace: cfg.namespace,
       },
       spec: {
+        path: '/api/v1',
```
prashbnair, line 25: Is there a need to have the /api/v1 path exposed?

raptorsun: Thanks for pointing out. 
Openshift console is not using v1 endpoint. Not sure if other components use v1.
I will update this PR limiting the route to v2 and see what happens in the tests.

raptorsun:
e2e-agnostic has a failling test probably related to this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1933144
test again
/test e2e-agnostic

simonpasquier:
/hold 

we need to change the external URL parameter:
https://github.com/openshift/cluster-monitoring-operator/blob/7a908ea2b0947dbc3a9bdd8e3db9d8422d2ce67b/pkg/manifests/manifests.go#L454
I suggest that we link to the OCP console
cc @jan--f

raptorsun:
> /hold
> 
> we need to change the external URL parameter:
> 
> https://github.com/openshift/cluster-monitoring-operator/blob/7a908ea2b0947dbc3a9bdd8e3db9d8422d2ce67b/pkg/manifests/manifests.go#L454
> 
> 
> I suggest that we link to the OCP console
> cc @jan--f

why shall we change that URL?

simonpasquier:
The back-link URL can be included in the notifications sent by Alertmanager (which is a way for users to jump quickly to the Alertmanager UI ) and the current value is the URL of the OpenShift route. Since we remove the access to the Alertmanager UI, we need to change the back-link URL too.

raptorsun:
> The back-link URL can be included in the notifications sent by Alertmanager (which is a way for users to jump quickly to the Alertmanager UI ) and the current value is the URL of the OpenShift route. Since we remove the access to the Alertmanager UI, we need to change the back-link URL too.

Now I see, thank you :) I will change it to the URL of the Openshift console.

pkg/tasks/alertmanager.go
```diff
@@ -178,6 +178,13 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 			return errors.Wrap(err, "initializing Alertmanager object failed")
 		}
 
+		// Use console URL as Alertmanager ExternalURL since UI access to AlertManager is disabled.
+		consoleRoute, err := t.client.GetRouteWithNameAndNameSpace(ctx, "console", "openshift-monitoring")
```
prashbnair, line 182: Should you not be using the openshift-console namespace? openshift-monitoring does not have a console route.

prashbnair, line 182: I also feel that the GetRouteURL method can be reused.

raptorsun: ooops, I will correct

pkg/tasks/alertmanager.go
```diff
@@ -178,6 +178,17 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
...
+		// Use console URL as Alertmanager ExternalURL since UI access to AlertManager is disabled.
+		consoleRoute, err := t.client.GetRouteWithNameAndNameSpace(ctx, "console", "openshift-console")
+		if err != nil {
+			return errors.Wrap(err, "getting console route failed")
+		}
+		consoleUrl, err := t.client.GetRouteURL(ctx, consoleRoute)
+		if err != nil {
+			return errors.Wrap(err, "getting console route URL failed")
+		}
+		a.Spec.ExternalURL = consoleUrl.String()
```
jan--f, line 190: Iiuc the backlink should contain a `monitoring` path component, like here https://github.com/openshift/cluster-monitoring-operator/pull/1532/files#diff-b0ca3a050a076f5a4007692cff8088723c9fe6f8c6734f8a8bc1bdd54f8d624cR105

pkg/tasks/alertmanager.go
```diff
@@ -178,6 +178,17 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 			return errors.Wrap(err, "initializing Alertmanager object failed")
 		}
 
+		// Use console URL as Alertmanager ExternalURL since UI access to AlertManager is disabled.
+		consoleRoute, err := t.client.GetRouteWithNameAndNameSpace(ctx, "console", "openshift-console")
+		if err != nil {
+			return errors.Wrap(err, "getting console route failed")
+		}
```
jan--f, line 185: This is probably superfluous? `GetRouteUrl` does a get with the passed route name and namespace, so its sufficient to pass a route object with those fields set I think.

jan--f (changes requested):

pkg/tasks/alertmanager.go
raptorsun: that's better, it redirects to the monitoring page. 👍

pkg/tasks/alertmanager.go
raptorsun: ok, It will be changed back to GetRouteUrl for simplicity.

raptorsun: I wonder which starts first, the CMO or console. The test show it fails to find the route console in namespace openshift-console.

pkg/operator/operator.go
```diff
@@ -276,6 +277,18 @@ func New(
 	})
 	o.informers = append(o.informers, informer)
 
+	informer = cache.NewSharedIndexInformer(
```
prashbnair, line 280: Do we need the watch and informer infra for this use case?

raptorsun: This informer is for the case where the console initialize its route, and trigger a change in the custom resource "console". So CMO can update immediately the externalURL for AlertManager using the URL to console.

raptorsun: Does the informer consume many resources and we may not need the externalURL of AlertManager set immediately when the console becomes available?

simonpasquier: Using an informer, the operator doesn't have to wait for the console operator to complete. I would expect that we watch only the cluster resource though like we do for [infrastructure config](https://github.com/openshift/cluster-monitoring-operator/blob/320ea1b465d8aa398ae1aefa05e67dc048fa33dc/pkg/operator/operator.go#L258-L261).

pkg/client/client.go
```diff
@@ -258,6 +258,10 @@ func (c *Client) ApiServersListWatchForResource(ctx context.Context, resource st
 	}
 }
 
+func (c *Client) ConsoleListWatch() *cache.ListWatch {
+	return cache.NewListWatchFromClient(c.kclient.CoreV1().RESTClient(), "consoles", "", fields.Everything())
```
simonpasquier, line 262: you need `c.oscclient.ConfigV1().Consoles()` rather than `c.kclient.CoreV1().RESTClient()`.

pkg/operator/operator.go
```diff
@@ -724,6 +743,20 @@ func (o *Operator) loadApiServerConfig(ctx context.Context) (*manifests.APIServe
 	return o.lastKnownApiServerConfig, nil
 }
 
+func (o *Operator) loadConsoleConfig(ctx context.Context) (*configv1.Console, error) {
+	config, err := o.client.GetConsoleConfig(ctx, "cluster")
+	if err != nil {
+		klog.Warningf("failed to get console config: %v", err)
+
+		if o.lastKnownConsoleConfig == nil {
```
simonpasquier, line 751: we don't want to fail but rather return an "empty" `configv1.Console`.

simonpasquier:
/lgtm
@juzhao for awareness :)

pkg/manifests/manifests.go
```diff
@@ -451,7 +454,9 @@ func (f *Factory) AlertmanagerMain(host string, trustedCABundleCM *v1.ConfigMap)
 
 	a.Spec.Image = &f.config.Images.Alertmanager
 
-	a.Spec.ExternalURL = f.AlertmanagerExternalURL(host).String()
+	if f.consoleConfig != nil {
+		a.Spec.ExternalURL = f.consoleConfig.Status.ConsoleURL
```
jan--f, line 458: Sorry I missed this last review: we want to append the path component `/monitoring`. See https://github.com/openshift/cluster-monitoring-operator/pull/1532#discussion_r785567207

pkg/manifests/manifests.go
```diff
@@ -451,7 +454,9 @@ func (f *Factory) AlertmanagerMain(host string, trustedCABundleCM *v1.ConfigMap)
```
jan--f, line 454: We no longer need the `host` argument to `AlertmanagerMain`

jan--f (changes requested):
Sorry, I have a last minute request.

lihongyan1:
Tested with PR:
Following case https://polarion.engineering.redhat.com/polarion/redirect/project/OSE/workitem?id=OCP-47419
Everything works well except the following, filed a premerge [bug ](https://bugzilla.redhat.com/show_bug.cgi?id=2045946)
curl -H "Authorization: Bearer $token" -k "https://$host/api/v2"   
<a href="/api/v2/">Moved Permanently</a>.

raptorsun:
> Tested with PR: Following case https://polarion.engineering.redhat.com/polarion/redirect/project/OSE/workitem?id=OCP-47419 Everything works well except the following issues https://bugzilla.redhat.com/show_bug.cgi?id=2045946 https://bugzilla.redhat.com/show_bug.cgi?id=2046013

Route to AlertManager is set to "/api" now, the same as Thanos' routes.

lihongyan1:
> 
verified and closed the related bug 2045975

raptorsun:
test comment

PR #1531, opened 2022-01-07
Bug 2037073: pkg/manifests: account for slow DNS resolutions

simonpasquier:


bison (approved):
I see this is WIP, but it looks good to me. I guess this means it could take 60 (20 + 4 * 10) seconds total?

EmilienM:
I tested this patch with OpenStack and it solved our issue, we managed to get AlertManager happy and no issue was found.

Thanks

simonpasquier:
> I guess this means it could take 60 (20 + 4 * 10) seconds total?

@bison yes

bison (approved):
/lgtm

PR #1532, opened 2022-01-07
MON-1631: prometheus: remove ui access

jan--f:

### DIFF
--- a/assets/prometheus-k8s/prometheus.yaml
+++ b/assets/prometheus-k8s/prometheus.yaml
@@ -161,6 +161,7 @@ spec:
         memory: 10Mi
   enableFeatures: []
   externalLabels: {}
+  externalURL: https://prometheus-k8s.openshift-monitoring.svc:9091
   image: quay.io/prometheus/prometheus:v2.32.1
   listenLocal: true
   nodeSelector:
--- a/jsonnet/components/prometheus.libsonnet
+++ b/jsonnet/components/prometheus.libsonnet
@@ -22,8 +22,8 @@ function(params)
       data: {},
     },
 
-    // OpenShift route to access the Prometheus UI.
-    route: {
+    // OpenShift route to access the Prometheus api.
+    apiRoute: {
       apiVersion: 'v1',
       kind: 'Route',
       metadata: {
@@ -35,6 +35,31 @@ function(params)
           kind: 'Service',
           name: 'prometheus-k8s',
         },
+        path: '/api',
+        port: {
+          targetPort: 'web',
+        },
+        tls: {
+          termination: 'Reencrypt',
+          insecureEdgeTerminationPolicy: 'Redirect',
+        },
+      },
+    },
+
+    // OpenShift route to access the Prometheus federate endpoint.
+    federateRoute: {
+      apiVersion: 'v1',
+      kind: 'Route',
+      metadata: {
+        name: 'prometheus-k8s-federate',
+        namespace: cfg.namespace,
+      },
+      spec: {
+        to: {
+          kind: 'Service',
+          name: 'prometheus-k8s',
+        },
+        path: '/federate',
         port: {
           targetPort: 'web',
         },
@@ -316,6 +341,7 @@ function(params)
           'kube-rbac-proxy',
           'metrics-client-certs',
         ],
+        externalURL: 'https://prometheus-k8s.openshift-monitoring.svc:9091',
         configMaps: ['serving-certs-ca-bundle', 'kubelet-serving-ca-bundle', 'metrics-client-ca'],
         probeNamespaceSelector: cfg.namespaceSelector,
         podMonitorNamespaceSelector: cfg.namespaceSelector,
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -115,7 +115,8 @@ var (
 	PrometheusK8sProxySecret                          = "prometheus-k8s/proxy-secret.yaml"
 	PrometheusRBACProxySecret                         = "prometheus-k8s/kube-rbac-proxy-secret.yaml"
 	PrometheusUserWorkloadRBACProxySecret             = "prometheus-user-workload/kube-rbac-proxy-secret.yaml"
-	PrometheusK8sRoute                                = "prometheus-k8s/route.yaml"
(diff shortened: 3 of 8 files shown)


assets/prometheus-k8s/prometheus.yaml
```diff
@@ -42,14 +42,16 @@ spec:
   - metrics-client-ca
   containers:
   - args:
+    - -request-logging=true
```
raptorsun, line 45: Why request-logging is needed in this case?

jan--f: It's just for debugging :) I can't figure out why oauth-proxy isn't authorizing the requests I expect it to.

jan--f:
/hold
Still investigating something

jsonnet/components/prometheus.libsonnet
```diff
@@ -366,7 +366,7 @@ function(params)
               '-upstream=http://localhost:9090',
               '-openshift-service-account=prometheus-k8s',
               '-openshift-sar={"resource": "namespaces", "verb": "get"}',
-              '-openshift-delegate-urls={"/": {"resource": "namespaces", "verb": "get"}}',
+              '-openshift-delegate-urls={"/api": {"resource": "namespaces", "verb": "get"},"/metrics": {"resource": "namespaces", "verb": "get"},"/federate": {"resource": "namespaces", "verb": "get"}}',
```
simonpasquier, line 369: i'm not sure that we need to declare `/metrics` since the metrics should be scraped from the kube-rbac-proxy container?

jsonnet/components/prometheus.libsonnet
```diff
@@ -366,7 +366,7 @@ function(params)
               '-upstream=http://localhost:9090',
               '-openshift-service-account=prometheus-k8s',
               '-openshift-sar={"resource": "namespaces", "verb": "get"}',
```
simonpasquier, line 368: Should we remove this line to disable login via OAuth?

jsonnet/components/prometheus.libsonnet
jan--f: I went by what the functional tests look for. https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus.go#L546 checks for `/metrics` on port 9091.

simonpasquier: Ok, can you add a TODO comment to be fixed once 4.11 opens?

jsonnet/components/prometheus.libsonnet
```diff
@@ -316,6 +341,7 @@ function(params)
           'kube-rbac-proxy',
           'metrics-client-certs',
         ],
+        externalURL: 'https://prometheus-k8s.openshift-monitoring.svc:9091',
```
simonpasquier, line 344: hmm I missed this detail :thinking:...
Ideally I would expect that we link back to the OCP console but the generated URL is hardcoded to `<externalURL>/graph?g0.expr=up&g0.tab=1` which wouldn't work for the console. Maybe @kyoto has some ideas?

jan--f: Not sure I follow...link back from where? Afaiu this is what the Prometheus UI uses internally. With this PR it should only be accessible through the Prometheus Service, so setting this as the external URL should suffice.

simonpasquier: Sorry for the missing details :) 
When Prometheus sends alerts to Alertmanager, the payload includes a `generatorURL` field that is a back-link to the Prometheus UI (see https://prometheus.io/docs/alerting/latest/clients/). The URL is constructed by concatenating `--web.external-url` and `/graph?g0.expr=<alert expression>&g0.tab=1` (see [here](https://github.com/prometheus/prometheus/blob/a5d57c926903677e0e16c29bce46de69dfca153f/cmd/prometheus/main.go#L1230) and [here](https://github.com/prometheus/prometheus/blob/a5d57c926903677e0e16c29bce46de69dfca153f/util/strutil/strconv.go#L26-L29)).
My thinking is that if the generator URL would link to the console route and the console redirects to the "Observe > Metrics" page... The other option would be to discuss upstream if it would be possible to have a full customization of the generator URL (not sure if it makes sense in general).

kyoto: Adding a Console redirect sounds like a reasonable workaround to me.

Can we have the generator URL go directly to `<Console URL>/monitoring` in the Console? Then, if I understand correctly, the constructed URL would be `<Console URL>/monitoring/graph?g0.expr=<alert expression>&g0.tab=1`, which Console would then redirect to `<Console URL>/monitoring/query-browser?query0=<alert expression>`?

jan--f: Looks good @kyoto, lets try it.

jan--f:
Rebased on #1529

pkg/manifests/manifests.go
```diff
@@ -1420,7 +1424,10 @@ func (f *Factory) PrometheusK8s(host string, grpcTLS *v1.Secret, trustedCABundle
 	}
 
 	p.Spec.Image = &f.config.Images.Prometheus
-	p.Spec.ExternalURL = f.PrometheusExternalURL(host).String()
+
+	if f.consoleConfig != nil {
+		p.Spec.ExternalURL = f.consoleConfig.Status.ConsoleURL + "/monitoring"
```
simonpasquier, line 1429: (nit) 
```suggestion
		p.Spec.ExternalURL = path.Join(f.consoleConfig.Status.ConsoleURL, "monitoring")
```

jan--f: :+1: Makes sense yes, any objections against a followup PR? The manifest creation for Alertmanager has the same code and could use the same fix.

simonpasquier: not at all

lihongyan1:
Test with pr, no issue found
/label qe-approved

jsonnet/components/prometheus.libsonnet
kyoto: PR for handling the Console redirect: https://github.com/openshift/console/pull/10963

simonpasquier:
/label docs-approved
/label px-approved
The parent epic (https://issues.redhat.com/browse/MON-1961) has the required acks.

PR #1533, opened 2022-01-10
Bug 2037891: Changing the grafana data source secret to be updatable. 

prashbnair:


prashbnair:
In 4.10 the structure of the grafana datasource secret has been changed to use secureJsonData. During the upgrade process, the secret was not getting updated. Because of this, when the cluster in upgraded to 4.10, the grafana api requests to prometheus were failing because the prometheus htpassword secret was not generated correctly.

PR #1534, opened 2022-01-10
Changed the severity of NodeClockNotSynchronising alert

lnguyen1401:
Changed from warning to critical


raptorsun:
Hello Linh, thank you very much for your contribution :) 
There are 2 issues to address before modifying the severity to "critical".
1. Critical alerts will probably wake up an administrator and it requires a runbook detailing the actions to take. Please refer to [this page](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md) for more details.
2. The file modified by this PR is in fact generated from a libsonnet file in vendor/, so we should either patch it in Cluster Monitoring Operator, or modify it in the upstream package (prometheus/node_exporter). To patch it , we can add the patch to [the list of patched rules](assets/node-exporter/prometheus-rule.yaml) and regenerate the yaml files by `make generate`. Then commit to the PR branch and push again.

PR #1536, opened 2022-01-17
OADP-22: Send Telemetry metrics on OADP

shubham-pampattiwar:
This PR adds two (allowlist):

1. `velero_backup_total` : measures the current number of existent backups
2. `velero_restore_total` : measures the current number of existent restores

Related PRs:
https://github.com/openshift/oadp-operator/pull/410
https://github.com/openshift/oadp-operator/pull/537

Related JIRA:
https://issues.redhat.com/browse/OADP-22


shubham-pampattiwar:
@bparees @bison @simonpasquier Can you please review the PR ?

simonpasquier (commented):
one suggestion: to reduce the cardinality of the metrics, it would be good to have a recording rule aggregating away dimensions like `instance`.

shubham-pampattiwar:
> one suggestion: to reduce the cardinality of the metrics, it would be good to have a recording rule aggregating away dimensions like `instance`.

@simonpasquier Can you please give an example of what needs to be done regarding your suggestion ? I am having a hard time understanding it.

simonpasquier:
@shubham-pampattiwar sorry for the late reply :-/
to explain it better, all samples scraped from the OADP operator have an `instance` label that is equal to `pod_ip:port_number`. So every time the pod gets redeployed, the value of this label will change which will create new timeseries (and increase the cardinality).
In practice, the OADP operator should deploy the [recording rules](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/) that provide the metrics without the `instance` label. Here is an example from the Node Tuning Operator: https://github.com/openshift/cluster-node-tuning-operator/blob/11f88d18a6273448a17c74374e8ee18b16fc4392/manifests/30-monitoring.yaml#L130-L131
In your case, you probably want 2 rules like:

```
    - expr: max without(instance) (velero_restore_total)
      record: cluster:velero_restore_total:max
    - expr: max without(instance) (velero_backup_total)
      record: cluster:velero_backup_total:max
```

shubham-pampattiwar:
> @shubham-pampattiwar sorry for the late reply :-/ to explain it better, all samples scraped from the OADP operator have an `instance` label that is equal to `pod_ip:port_number`. So every time the pod gets redeployed, the value of this label will change which will create new timeseries (and increase the cardinality). In practice, the OADP operator should deploy the [recording rules](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/) that provide the metrics without the `instance` label. Here is an example from the Node Tuning Operator: https://github.com/openshift/cluster-node-tuning-operator/blob/11f88d18a6273448a17c74374e8ee18b16fc4392/manifests/30-monitoring.yaml#L130-L131 In your case, you probably want 2 rules like:
> 
> ```
>     - expr: max without(instance) (velero_restore_total)
>       record: cluster:velero_restore_total:max
>     - expr: max without(instance) (velero_backup_total)
>       record: cluster:velero_backup_total:max
> ```

@simonpasquier Thank you for the explanation. I have raised a PR against oadp-operator with the suggested recording rules. Please review the same https://github.com/openshift/oadp-operator/pull/579

shubham-pampattiwar:
@simonpasquier The recording rule PR is [merged](https://github.com/openshift/oadp-operator/pull/579) and I have updated this PR with the new metric record names as well. PTAL at this PR.

simonpasquier (commented):
/lgtm

note that `velero_restore_total` and `velero_backup_total` are gauge metrics hence they shouldn't use the `_total` suffix which is reserved for counters (see https://prometheus.io/docs/instrumenting/writing_exporters/#naming). It would be good to communicate this issue to upstream.

shubham-pampattiwar:
@simonpasquier Thank you for the review. I will open an issue regarding the metrics naming convention at upstream Velero. Can we go ahead with the PR merge ?

simonpasquier:
/label px-approved
/label qe-approved
/label docs-approved

This increases the number of series by 2 per cluster deploying the operator which looks reasonable.

PR #1537, opened 2022-01-18
Bug 2039119: assets: let CVO manage the CMO Service resource

jan--f:
Problem: In
23c29ca52af9a41182e8ae16f9dcaa7a5551448c the
CMO service resource was changed to be managed by CVO (by creating it in
the asset/ subdir). See the commit/PR for motivation. The service
however was not removed from the assets that CMO should create itself,
leading to CVO and CMO competing over this resource

Solution: Remove the resource from the CMO-managed resources.

Issues: https://bugzilla.redhat.com/show_bug.cgi?id=2039119
https://bugzilla.redhat.com/show_bug.cgi?id=2016352


simonpasquier (commented):
/lgtm

PR #1539, opened 2022-01-19
Bug 2037891: Reverting the secureJsonData change for the grafana password

prashbnair:

### DIFF
--- a/assets/grafana/dashboard-datasources.yaml
+++ b/assets/grafana/dashboard-datasources.yaml
@@ -16,6 +16,7 @@ stringData:
             {
                 "access": "proxy",
                 "basicAuth": true,
+                "basicAuthPassword": "",
                 "basicAuthUser": "internal",
                 "editable": false,
                 "jsonData": {
@@ -23,9 +24,6 @@ stringData:
                 },
                 "name": "prometheus",
                 "orgId": 1,
-                "secureJsonData": {
-                    "basicAuthPassword": ""
-                },
                 "type": "prometheus",
                 "url": "https://prometheus-k8s.openshift-monitoring.svc:9091",
                 "version": 1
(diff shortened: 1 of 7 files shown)


prashbnair:
@simonpasquier PTAL

prashbnair:
This PR reverts https://github.com/openshift/cluster-monitoring-operator/pull/1363/ and https://github.com/openshift/cluster-monitoring-operator/pull/1533.

simonpasquier:
/lgtm

To be complete, the fix for https://bugzilla.redhat.com/show_bug.cgi?id=1999521 (#1363) introduced a regression for clusters upgrading to 4.10 (described in https://bugzilla.redhat.com/show_bug.cgi?id=2037891). #1533 tried to fix it but as a side effect, the Grafana task generates a new HTTP password on every sync that may never be propagated to the Prometheus oauth-proxy.
We've discussed the different options with Prashant and decided that reverting the 2 pull requests and marking BZ1999521 as WONTFIX is the safest solution. We could add more logic to CMO to keep the HTTP password in sync between Prometheus and Grafana but it would be more complicated and not worth it given that Grafana is slated for removal anyway.

arajkumar:
@prashbnair Looks like grafana is broken after this. Should we use [CreateOrUpdate](https://github.com/openshift/cluster-monitoring-operator/pull/1539/files#diff-bce1a4225ddd13f984c575a24020c0c873a9b2be1057f69c81224104744f17f9L118) here?

Screenshot 2022-01-19 at 10 20 44 PM

prashbnair:
@arajkumar I tested this using https://amd64.ocp.releases.ci.openshift.org/releasestream/4.10.0-0.nightly/release/4.10.0-0.nightly-2022-01-19-212639 and I do not see any issues.

Screenshot 2022-01-20 at 11 31 21

PR #1551, opened 2022-02-03
pkg: drop code removing the legacy Alertmanager service monitor

simonpasquier:
The Alertmanager service monitor was renamed from "alertmanager" to
"alertmanager-main" in 4.10 and the operator took care of removing the
old service monitor for clusters that upgraded from 4.9. Now that the
master branch targets 4.11, we can drop this code.


simonpasquier:
/skip
/label qe-approved
/label px-approved
/label docs-approved

This is only clean-up work.

slashpai (approved):

PR #1552, opened 2022-02-03
[MON-2151] remove the Oauth-proxy before AlertManager

raptorsun:

### DIFF
--- a/jsonnet/components/alertmanager.libsonnet
+++ b/jsonnet/components/alertmanager.libsonnet
@@ -10,8 +10,6 @@ function(params)
   };
 
   alertmanager(cfg) {
-    trustedCaBundle: generateCertInjection.trustedCNOCaBundleCM(cfg.namespace, 'alertmanager-trusted-ca-bundle'),
-
     // OpenShift route to access the Alertmanager UI.
 
     route: {
@@ -37,16 +35,7 @@ function(params)
       },
     },
 
-    // The ServiceAccount needs this annotation, to signify the identity
-    // provider, that when a users it doing the oauth flow through the oauth
-    // proxy, that it should redirect to the alertmanager-main route on
-    // successful authentication.
     serviceAccount+: {
-      metadata+: {
-        annotations+: {
-          'serviceaccounts.openshift.io/oauth-redirectreference.alertmanager-main': '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"alertmanager-main"}}',
-        },
-      },
       // automountServiceAccountToken is set to true as kube-rbac-proxy sidecar
       // requires connection to kubernetes API
       automountServiceAccountToken: true,
@@ -89,19 +78,6 @@ function(params)
         ],
         type: 'ClusterIP',
       },
-    },
-
-    // The proxy secret is there to encrypt session created by the oauth proxy.
-    proxySecret: {
-      apiVersion: 'v1',
-      kind: 'Secret',
-      metadata: {
-        name: 'alertmanager-main-proxy',
-        namespace: cfg.namespace,
-        labels: { 'app.kubernetes.io/name': 'alertmanager-main' },
-      },
-      type: 'Opaque',
-      data: {},
     },
 
     // In order for the oauth proxy to perform a TokenReview and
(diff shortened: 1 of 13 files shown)


jsonnet/components/alertmanager.libsonnet
```diff
@@ -156,6 +154,30 @@ function(params)
...
+      type: 'Opaque',
+      stringData: {
+        'config.yaml': std.manifestYamlDoc({
+          authorization: {
+            resourceAttributes:
+              {
+                apiGroup: 'monitoring.coreos.com',
+                resource: 'alertmanagers',
+                namespace: 'openshift-monitoring',
+                verbs: ['patch', 'create', 'get'],
```
simonpasquier, line 174: I suspect that we need `*` since delete isn't covered.

jsonnet/components/alertmanager.libsonnet
```diff
@@ -156,6 +154,30 @@ function(params)
...
+        labels: { 'app.kubernetes.io/name': 'alertmanager-main' },
+      },
+      type: 'Opaque',
+      stringData: {
+        'config.yaml': std.manifestYamlDoc({
+          authorization: {
+            resourceAttributes:
+              {
+                apiGroup: 'monitoring.coreos.com',
+                resource: 'alertmanagers',
```
simonpasquier, line 172: to be consistent with the tenant-aware [kube-rbac-proxy config](https://github.com/openshift/cluster-monitoring-operator/blob/38a487897d146715c734d11e80a2b60e7c374eb0/assets/alertmanager/kube-rbac-proxy-secret.yaml#L10-L17), it may be better to authorize based on the permissions to the `prometheusrules` resource.

jsonnet/components/cluster-monitoring-operator.libsonnet
```diff
@@ -281,11 +281,13 @@ function(params) {
...
+    rules: [
+      {
+        apiGroups: [''],
+        resources: ['namespaces'],
+        verbs: ['get'],
+      },
+    ],
   },
 
   // This role enables access to the Alertmanager APIs and UIs through OAuth proxy.
```
simonpasquier, line 293: this would need an update.

jsonnet/components/alertmanager.libsonnet
raptorsun: ok, the permission to manipulate prometheusrules is more logic than alertmanagers. I think the kube-rbac-proxy before thanos can  use this new RBAC settings, too.

arajkumar (commented):
1. ServiceAccount annotation[2] to support oauth proxy direction can be removed.
2. ProxySecret[2] related to OAuth Proxy shall be removed too


[1] https://github.com/openshift/cluster-monitoring-operator/blob/a507ad3afaa1d09c8a7555c8d9fad5524d3b7f2f/jsonnet/components/alertmanager.libsonnet#L45

[2] https://github.com/openshift/cluster-monitoring-operator/blob/a507ad3afaa1d09c8a7555c8d9fad5524d3b7f2f/jsonnet/components/alertmanager.libsonnet#L93

pkg/tasks/alertmanager.go
```diff
@@ -157,23 +167,8 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 	}
 
 	{
-		// Create trusted CA bundle ConfigMap.
```
arajkumar, line 160: Make sure that you delete the `trusted-ca-bundle` resource which would be coming from upgrade path. I had followed an approach which you can refer[1], but it might change based on the reviewer's opinion :)

[1] https://github.com/openshift/cluster-monitoring-operator/blob/edda4a35ff5a5b66f846b7e68b198380f3f393a7/pkg/tasks/prometheus.go#L390-L402

raptorsun: Thanks :) I forgot to remove this resource when upgrading.  Will correct it now.

pkg/tasks/alertmanager.go
```diff
@@ -236,6 +232,16 @@ func (t *AlertmanagerTask) destroy(ctx context.Context) error {
...
+	rsw, err := t.factory.AlertmanagerRBACProxyWebSecret()
+	if err != nil {
+		return errors.Wrap(err, "initializing Alertmanager RBAC proxy web Secret failed")
+	}
+
+	err = t.client.DeleteSecret(ctx, rsw)
+	if err != nil {
+		return errors.Wrap(err, "deleting Alertmanager RBAC proxy web Secret failed")
+	}
+
```
simonpasquier, line 244: do we need this since the `create()` function would delete the secret too?

pkg/tasks/alertmanager.go
```diff
@@ -136,14 +146,9 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 		return errors.Wrap(err, "reconciling Alertmanager ServiceAccount failed")
 	}
 
-	ps, err := t.factory.AlertmanagerProxySecret()
-	if err != nil {
-		return errors.Wrap(err, "initializing Alertmanager proxy Secret failed")
-	}
-
-	err = t.client.CreateIfNotExistSecret(ctx, ps)
+	err = t.client.DeleteSecretByNamespaceAndName(ctx, t.client.Namespace(), "alertmanager-trusted-ca-bundle")
```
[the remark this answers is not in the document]
raptorsun: yup, will move deletion to end of function with todo comments.

pkg/tasks/alertmanager.go
raptorsun: This secret is useful for alertmanager. It provides the certificate to kube-rbac-proxy serving REST API calls on port 9095.

PR #1553, opened 2022-02-04
Bug 2050707: test: account for pdb and Prometheus' staleness period

jan--f:
Problem: We recently introduced pod disruption budgets and improved
availability on Prometheus. This is reflected in the up metric for
Prometheus. By virtue of Prometheus staleness mechanism (it will return
samples up to 5m old for a series) the old query can appear to see 4
Prometheus instances.

Solution: Use the last_over_time aggregation with a 2m interval to only
look at the last sample in that interval. This effectively adjusts
Prometheus staleness period to 2m.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -84,7 +84,7 @@ func testMain(m *testing.M) error {
 			body []byte
 			v    int
 		)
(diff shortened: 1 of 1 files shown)


jan--f:
/retitle Bug 2050707: test: account for pdb and Prometheus' staleness period

PR #1554, opened 2022-02-04
Bug 2037513: Fix dashboards having container_fs* metrices in queries

slashpai:
container_fs_.* metrics doesn't have "container" label after
openshift/cluster-monitoring-operator/pull/1402 which caused regression
for dashboards using "container" label since it gives empty datapoints
for all queries using "container" label in queries containing
container_fs_.* metrices.
    
Inorder to fix the dashboards added containerfsSelector in
kubernetes-monitoring/kubernetes-mixin/pull/737 and overrided the containerfsSelector
default value from  container!="" to harmless id!="" in this commit.
We really don't need that selector since we dropped the container label
 in OCP for container_fs_.* metrices but we need to override the
 selector for the queries involving container_fs_.* to work. Dropped the dashboard 
**"Kubernetes / Compute Resources / Pod"** since Prometheus doesn't collect the 
per-container fs metrics (discussed in https://github.com/openshift/cluster-monitoring-operator/pull/1685#issuecomment-1149561025)

Had added deviceDeviceSelector in container_fs* queries as part of
kubernetes-monitoring/kubernetes-mixin#737 since diskDeviceSelector
includes only node_exporter label values updated disDeviceSelector
to include cadvisor devices too (/dev)

More Details:
[Bugzilla Comment 9](https://bugzilla.redhat.com/show_bug.cgi?id=2037513#c9)
[Bugzilla Comment 10](https://bugzilla.redhat.com/show_bug.cgi?id=2037513#c10)

cc: @arajkumar @PhilipGough 


slashpai:
Looks like https://github.com/openshift/cluster-monitoring-operator/pull/1554/commits/d72bec8bd7704728cc5c6725c29d211a66036b4f#diff-08daaad7f1c27b41ec86bee6cb6af0259066e38448b11e5389412c720813b255R25-R32 included more changes than expected

jsonnet/jsonnetfile.json
```diff
@@ -17,7 +17,7 @@
           "subdir": "jsonnet/kube-prometheus"
         }
       },
-      "version": "release-0.10"
```
arajkumar, line 20: Change all jsonnet packages to its main/master branch.

assets/prometheus-k8s/cluster-role.yaml
```diff
@@ -8,7 +8,6 @@ metadata:
     app.kubernetes.io/part-of: openshift-monitoring
     app.kubernetes.io/version: 2.32.1
   name: prometheus-k8s
-  namespace: openshift-monitoring
```
arajkumar, line 11: Why these namespace are removed?

assets/prometheus-user-workload/cluster-role-binding.yaml
```diff
@@ -8,7 +8,6 @@ metadata:
     app.kubernetes.io/part-of: openshift-monitoring
     app.kubernetes.io/version: 2.32.1
   name: prometheus-user-workload
-  namespace: openshift-user-workload-monitoring
```
arajkumar, line 11: ditto

assets/prometheus-k8s/cluster-role.yaml
slashpai: I didn't remove it explicitly like I mentioned in description after running generate it did a lot of changes unrelated to this bug. I am checking why it happend

slashpai: Ok looks like its due to https://github.com/prometheus-operator/kube-prometheus/pull/1566 we should override this in config to fix

jsonnet/jsonnetfile.json
PhilipGough: for the purpose of fixing this bug and reducing the change-set, perhaps it is better in this case to only make relevant changes, wdyt?

assets/prometheus-k8s/cluster-role.yaml
slashpai: But why do we need a `namespace` field for clusterrole and clusterrole binding since its not a namespaced resource

assets/prometheus-user-workload/cluster-role-binding.yaml
slashpai: https://github.com/openshift/cluster-monitoring-operator/pull/1554#discussion_r802525716

slashpai:
I will create a separate PR for kube-prometheus sync and generate to avoid confusion

assets/prometheus-k8s/cluster-role.yaml
arajkumar: Ahh, thats correct, We don't need a namespace for cluster role

slashpai:
cc: @arajkumar @PhilipGough PTAL when you get a chance
I have added description why we needed this change

slashpai:
Had added deviceDeviceSelector in container_fs* queries as part of [kubernetes-mixin PR](https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/737) since [diskDeviceSelector](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/config.libsonnet#L100-L101) includes only node_exporter label values figuring out what all need to be updated to make it work for cadvisor label values as well

slashpai:
Dropped the dashboard **"Kubernetes / Compute Resources / Pod"** as discussed in https://github.com/openshift/cluster-monitoring-operator/pull/1685#issuecomment-1149561025

slashpai:
> Dropped the dashboard **"Kubernetes / Compute Resources / Pod"** as discussed in [#1685 (comment)](https://github.com/openshift/cluster-monitoring-operator/pull/1685#issuecomment-1149561025)

cc: @juzhao for visibility

raptorsun (approved):
/lgtm

slashpai:
@juzhao we have dropped two rows in **Kubernetes / Compute Resources / Pod"** 'Storage IO - Distribution(Containers)' and 'Storage IO - Distribution'

PR #1557, opened 2022-02-10
MON-1632 Removing grafana from monitoring stack

prashbnair:


jsonnet/components/grafana.libsonnet
```diff
@@ -35,60 +35,18 @@ function(params)
...
-        },
-        port: {
-          targetPort: 'https',
-        },
-        tls: {
-          termination: 'Reencrypt',
-          insecureEdgeTerminationPolicy: 'Redirect',
-        },
-      },
+    config:: {
```
simonpasquier, line 38: It looks like we need to strip out most of the kube-prometheus jsonnet. Could we use github.com/brancz/kubernetes-grafana directly?

PhilipGough:
>  I have removed everything in this PR except the dashboard definitions which we need to retain. Are you suggesting another way of doing the same.

I could be wrong,  but I believe the comment refers to the need to delete (cleanup) existing objects on the upgrade path as well as just "not creating" them on a fresh install which is what this PR does IIUC.

prashbnair:
/retitle MON-1632 Removing grafana from monitoring stack

prashbnair:
@PhilipGough PTAL.

pkg/tasks/grafana.go
```diff
@@ -42,99 +39,14 @@ func NewGrafanaTask(client *client.Client, factory *manifests.Factory, config *m
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      grafanaLegacyDataSourceSecretName,
-			Namespace: sds.GetNamespace(),
-		},
-	})
-	if err != nil {
-		return errors.Wrap(err, "deleting legacy Grafana Datasources secret failed")
-	}
-
 	cmdds, err := t.factory.GrafanaDashboardDefinitions()
```
arajkumar, line 50: I don't think we even need this. https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_90_cluster-monitoring-operator_01-dashboards.yaml would serve the console dashboard needs which is be created here https://github.com/openshift/cluster-monitoring-operator/blob/master/hack/build-jsonnet.sh#L37

arajkumar (changes requested):

simonpasquier:
Can you add `TODO` comments for things that need to be removed in 4.12? This covers jsonnet/components/grafana.libsonnet, `pkg/tasks/grafana.go` and this line in `pkg/operator/operator.go`:

https://github.com/openshift/cluster-monitoring-operator/blob/38a487897d146715c734d11e80a2b60e7c374eb0/pkg/operator/operator.go#L609

pkg/tasks/grafana.go
prashbnair: Done

prashbnair:
> Can you add TODO comments for things that need to be removed in 4.12?

Done

pkg/tasks/grafana.go
```diff
@@ -14,12 +14,10 @@
...
+// TODO this file can be be removed in 4.12
 import (
 	"context"
 
-	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"github.com/openshift/cluster-monitoring-operator/pkg/client"
 	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
 	"github.com/pkg/errors"
```
arajkumar, line 23: Unused package local var grafanaLegacyDataSourceSecretName should be removed too.

prashbnair: Done

arajkumar (approved):

simonpasquier (commented):
/lgtm

PhilipGough (approved):

jan--f:
/label docs-approved
Adding doc approval as per agreemtent.
Ping @juzhao for qe approval.

juzhao:
@lihongyan1 please help to review and add approve

PR #1558, opened 2022-02-11
Bug 2043518: set degraded and available status based on Prometheus resource status

sthaha:
Bug 2043518: set degraded and available status based on Prometheus pod status

NOTE: This patch addresses 2 related Bugs:
   1. Bug 2043518: Better message in the CMO degraded/unavailable
      conditions when pods can't be scheduled
   2. Bug 2039411: Monitoring operator reports unavailable=true while
      one Prometheus pod is ready

The patch sets the Degraded and Availability status of ClusterOperator
by inspecting Prometheus Resource status. To achieve this, a new
PrometheusValidationTask is added that inspects the Prometheus resource status
and returns degraded and/or unavailable validation errors.

If at least one of the pod is running, the state will be marked as `Degraded`
but `Available` thus fixing the previous implementation which
incorrectly reported `Availability` as `False`.

Thus, given the following invalid monitoring config which uses a non-existent PVC

```yaml
  prometheusK8s:
    volumeClaimTemplate:
      metadata:
        name: prometheus-db
      spec:
        storageClassName: foo
        resources: { requests: {storage: 100Mi} }
```

It now correctly marks the ClusterOperator status as follows (formatted
for readability)

```
 {
    "type": "Degraded"
    "message": "<message from prometheus CR>",
    reason": "UpdatingPrometheusK8S",
    "status": "True",
  },
  {
    "type": "Available"
    "message": "<message from prometheus CR>",
    "reason": "UpdatingPrometheusK8S",
    "status": "False",
  }
]
```

---

[ ] I added CHANGELOG entry for this change.
[x] No user facing changes, so no entry in CHANGELOG was needed.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -16,6 +16,7 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"reflect"
 	"strings"
@@ -39,6 +40,7 @@ import (
 	"github.com/prometheus-operator/prometheus-operator/pkg/thanos"
 	thanosoperator "github.com/prometheus-operator/prometheus-operator/pkg/thanos"
 	admissionv1 "k8s.io/api/admissionregistration/v1"
+
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
@@ -49,6 +51,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	apiutilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
@@ -859,6 +863,90 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
 		return errors.Wrapf(err, "waiting for Prometheus %s/%s", p.GetNamespace(), p.GetName())
 	}
 	return nil
+}
+
+// validatePrometheusResource is a helper method for ValidatePrometheus.
+// NOTE: this function is refactored out of wait.Poll for testing
+func (c Client) validatePrometheusResource(ctx context.Context, prom types.NamespacedName) (bool, []error) {
+
+	p, err := c.mclient.MonitoringV1().Prometheuses(prom.Namespace).Get(ctx, prom.Name, metav1.GetOptions{})
+	if err != nil {
+		// failing to get Prometheus -> Degraded: Unknown & Unavailable: Unknown
+		klog.V(4).Info("validate prometheus failed to get prometheus: ", err)
+		return false, []error{
+			NewUnknownAvailabiltyError(err.Error()),
+			NewUnknownDegradedError(err.Error()),
+		}
+	}
+
+	avail, err := prometheusConditonForType(p.Status.Conditions, monv1.PrometheusAvailable)
+	if err != nil {
+		klog.V(4).Info("validate prometheus failed to get prometheus avail condition: ", err)
+		// failing to get Prometheus.Status.Condtion -> Degraded: Unknown & Unavailable: Unknown
+		return false, []error{
+			NewUnknownAvailabiltyError(err.Error()),
+			NewUnknownDegradedError(err.Error()),
+		}
+	}
+
+	if avail.Status == monv1.PrometheusConditionTrue {
+		// Prometheus is Available; check reconciled Condition as well
+		reconciled, err := prometheusConditonForType(p.Status.Conditions, monv1.PrometheusReconciled)
+		if err != nil {
+			klog.V(4).Info("validate prometheus failed to get prometheus reconciled condition: ", err)
+			// failing to get Prometheus.Status.Condtion -> Degraded: Unknown
+			return false, []error{NewUnknownDegradedError(err.Error())}
+
+		} else if reconciled.Status != monv1.PrometheusConditionTrue {
+			klog.V(4).Info("validate prometheus failed reconciled condition: ", reconciled.Status)
+			msg := fmt.Sprintf("%s: %s", reconciled.Reason, reconciled.Message)
+			return false, []error{NewDegradedError(msg)}
+		}
+
+		// terminate if available & reconciled
+		return true, nil
+	}
+
+	// return reason for failure as state-errors - Degraded: True & Unavailable: True
+	// since prometheus is Unavailable
+
+	msg := fmt.Sprintf("%s: %s", avail.Reason, avail.Message)
+	errs := []error{NewDegradedError(msg)}
+
+	if avail.Status == monv1.PrometheusConditionFalse {
+		// prometheus not available should result in a Degraded and Unavailable error
+		errs = append(errs, NewAvailabilityError(msg))
+	}
+	return false, errs
+}
+
+// ValidatePrometheus returns nil error if Prometheus is fully available.
+// Otherwise, it returns
+// Degraded(True) and Unavailable(True) if Prometheus is not running
+// Degraded(Unknown)  and Unavailable(Unknown) if it fails to retrieve Prometheus status
+func (c *Client) ValidatePrometheus(ctx context.Context, promNsName types.NamespacedName) error {
+	validationErrors := []error{}
+
+	pollErr := wait.Poll(10*time.Second, 5*time.Minute, func() (bool, error) {
+		stop, errs := c.validatePrometheusResource(ctx, promNsName)
+		validationErrors = errs
+		return stop, nil
+	})
+
+	if pollErr == wait.ErrWaitTimeout {
+		return apiutilerrors.NewAggregate(validationErrors)
+	}
+
+	return nil
+}
+
+func prometheusConditonForType(conditions []monv1.PrometheusCondition, t monv1.PrometheusConditionType) (monv1.PrometheusCondition, error) {
+	for _, c := range conditions {
+		if c.Type == t {
+			return c, nil
+		}
+	}
+	return monv1.PrometheusCondition{}, fmt.Errorf("prometheus: missing condition type - %s", t)
 }
 
 func (c *Client) WaitForAlertmanager(ctx context.Context, a *monv1.Alertmanager) error {
--- a/pkg/client/state_error.go
+++ b/pkg/client/state_error.go
@@ -1,0 +1,64 @@
+// Copyright 2022 The Cluster Monitoring Operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+import (
+	"fmt"
+)
+
+type State string
+
+const (
+	DegradedState    State = "degraded"
+	UnavailableState State = "unavailable"
+)
+
+// StateError indicates if an operation performed by the client has
+// resulted in an invalid state such as Degraded, Unavailable, Unknown
+type StateError struct {
+	State   State
+	Unknown bool
+	Reason  string
+}
+
+var _ error = (*StateError)(nil)
+
+func (se StateError) Error() string {
+	unknown := ""
+	if se.Unknown {
+		unknown = " (unknown)"
+	}
+	return fmt.Sprintf("%s%s: %s", se.State, unknown, se.Reason)
+}
+
+func NewDegradedError(reason string) *StateError {
+	return &StateError{State: DegradedState, Unknown: false, Reason: reason}
+}
+
+func NewAvailabilityError(reason string) *StateError {
+	return &StateError{State: UnavailableState, Unknown: false, Reason: reason}
+}
+
+func NewUnknownStateError(s State, reason string) *StateError {
+	return &StateError{State: s, Unknown: true, Reason: reason}
+}
+
+func NewUnknownAvailabiltyError(reason string) *StateError {
+	return NewUnknownStateError(UnavailableState, reason)
+}
+
+func NewUnknownDegradedError(reason string) *StateError {
+	return NewUnknownStateError(DegradedState, reason)
+}
--- a/pkg/client/status_reporter.go
+++ b/pkg/client/status_reporter.go
@@ -16,7 +16,6 @@ package client
 
 import (
 	"context"
-	"fmt"
 	"reflect"
 
 	cmostr "github.com/openshift/cluster-monitoring-operator/pkg/strings"
@@ -36,6 +35,32 @@ const (
 	UserAlermanagerConfigMisconfiguredMessage        = "Misconfigured Alertmanager:  Alertmanager for user-defined alerting is enabled in the openshift-monitoring/cluster-monitoring-config configmap by setting 'enableUserAlertmanagerConfig: true' field. This conflicts with a dedicated Alertmanager instance enabled in  openshift-user-workload-monitoring/user-workload-monitoring-config. Alertmanager enabled in openshift-user-workload-monitoring takes precedence over the one in openshift-monitoring, so please remove the 'enableUserAlertmanagerConfig' field in openshift-monitoring/cluster-monitoring-config."
 	UserAlermanagerConfigMisconfiguredReason         = "UserAlertmanagerMisconfigured"
 )
+
+// Status represents if the state being reported is known to be True, False, or Unknown.
+//
+// NOTE: redefined instead of reusing openshift/config/v1 to avoid coupling
+type Status string
+
+const (
+	TrueStatus    Status = "True"
+	FalseStatus   Status = "False"
+	UnknownStatus Status = "Unknown"
+)
+
+// StateInfo provides all information required by the StatusReporter to report
+// a condition's state such as Status, Reason, and Message
+type StateInfo interface {
+	Status() Status
+	Reason() string
+	Message() string
+}
+
+// StatesReport represents State Information about various States that the
+// StatusReporter will report on such as the Degraded and Available.
+type StatesReport interface {
+	Degraded() StateInfo
+	Available() StateInfo
+}
 
 type StatusReporter struct {
 	client                clientv1.ClusterOperatorInterface
@@ -172,20 +197,32 @@ func (r *StatusReporter) SetRollOutInProgress(ctx context.Context) error {
 	return r.setConditions(ctx, co, conditions)
 }
 
-func (r *StatusReporter) SetFailed(ctx context.Context, statusErr error, reason string) error {
+// ReportState sets the Degraded and the Availability condition as reported by the
+// StatesReport passed. If degraded or availability is nil, no changes are make
+// to the existing condition
+func (r *StatusReporter) ReportState(ctx context.Context, report StatesReport) error {
 	co, err := r.getOrCreateClusterOperator(ctx)
 	if err != nil {
 		return err
 	}
 
 	time := metav1.Now()
-	// The Reason should be upper case camelCase (PascalCase) according to the API docs.
-	reason = cmostr.ToPascalCase(reason)
 
 	conditions := newConditions(co.Status, r.version, time)
-	conditions.setCondition(v1.OperatorAvailable, v1.ConditionFalse, unavailableMessage, reason, time)
-	conditions.setCondition(v1.OperatorProgressing, v1.ConditionFalse, unavailableMessage, reason, time)
-	conditions.setCondition(v1.OperatorDegraded, v1.ConditionTrue, fmt.Sprintf("Failed to rollout the stack. Error: %v", statusErr), reason, time)
+
+	if degraded := report.Degraded(); degraded != nil {
+		// The Reason should be upper case camelCase (PascalCase) according to the API docs.
+		reason := cmostr.ToPascalCase(degraded.Reason())
+
+		conditions.setCondition(v1.OperatorDegraded, v1.ConditionStatus(degraded.Status()),
+			degraded.Message(), reason, time)
+	}
+
+	if available := report.Available(); available != nil {
+		reason := cmostr.ToPascalCase(available.Reason())
+		conditions.setCondition(v1.OperatorAvailable, v1.ConditionStatus(available.Status()),
+			available.Message(), reason, time)
+	}
 
 	return r.setConditions(ctx, co, conditions)
 }
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -31,7 +31,9 @@ import (
 	certapiv1 "k8s.io/api/certificates/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apiutilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/rest"
@@ -120,6 +122,7 @@ func (pc *ProxyConfig) NoProxy() string {
 
 const (
 	resyncPeriod = 15 * time.Minute
+	maxFailCount = 2
 
 	// see https://github.com/kubernetes/apiserver/blob/b571c70e6e823fd78910c3f5b9be895a756f4cbb/pkg/server/options/authentication.go#L239
 	apiAuthenticationConfigMap    = "kube-system/extension-apiserver-authentication"
@@ -600,7 +603,7 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 
 	config, err := o.Config(ctx, key)
 	if err != nil {
-		o.reportError(ctx, err, "InvalidConfiguration")
+		o.reportFailed(ctx, newRunReportForError("InvalidConfiguration", err))
 		return err
 	}
 	config.SetImages(o.images)
@@ -613,7 +616,7 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 	apiServerConfig, err = o.loadApiServerConfig(ctx)
 
 	if err != nil {
-		o.reportError(ctx, err, "APIServerConfigError")
+		o.reportFailed(ctx, newRunReportForError("APIServerConfigError", err))
 		return err
 	}
 
@@ -665,16 +668,10 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 		klog.Errorf("error occurred while setting status to in progress: %v", err)
 	}
 
-	taskErrors := tl.RunAll(ctx)
-	if len(taskErrors) > 0 {
-		var failedTask string
-		if len(taskErrors) == 1 {
-			failedTask = cmostr.ToPascalCase(taskErrors[0].Name + "Failed")
-		} else {
-			failedTask = "MultipleTasksFailed"
-		}
+	if taskErrors := tl.RunAll(ctx); len(taskErrors) > 0 {
 
-		o.reportError(ctx, taskErrors, failedTask)
+		report, failedTask := generateRunReportFromTaskErrors(taskErrors)
+		o.reportFailed(ctx, report)
 		return errors.Errorf("cluster monitoring update failed (reason: %s)", failedTask)
 	}
 
@@ -707,18 +704,19 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 	return nil
 }
 
-func (o *Operator) reportError(ctx context.Context, err error, failedTaskReason string) {
-	klog.Infof("ClusterOperator reconciliation failed (attempt %d), retrying. ", o.failedReconcileAttempts+1)
-	if o.failedReconcileAttempts >= 2 {
-		// Only update the ClusterOperator status after 3 retries have been attempted to avoid flapping status.
-		klog.Warningf("Updating ClusterOperator status to failed after %d attempts.", o.failedReconcileAttempts+1)
-
-		reportErr := o.client.StatusReporter().SetFailed(ctx, err, failedTaskReason)
-		if reportErr != nil {
-			klog.Errorf("error occurred while setting status to failed: %v", reportErr)
-		}
-	}
+func (o *Operator) reportFailed(ctx context.Context, report runReport) {
 	o.failedReconcileAttempts++
+	klog.Infof("ClusterOperator reportFailed (attempt %d).", o.failedReconcileAttempts)
+
+	if o.failedReconcileAttempts < maxFailCount {
+		klog.Infof("ClusterOperator reconciliation %d - skipping update", o.failedReconcileAttempts)
+		return
+	}
+
+	if err := o.client.StatusReporter().ReportState(ctx, report); err != nil {
+		klog.ErrorS(err, "failed to update cluster operator status")
+	}
+
 }
 
 func (o *Operator) loadInfrastructureConfig(ctx context.Context) *InfrastructureConfig {
@@ -988,4 +986,188 @@ func (o *Operator) workloadsToRebalance() []rebalancer.Workload {
 		)
 	}
 	return workloads
+}
+
+// stateErrorOrUnavailable converts an error to Unavailable & Degraded
+// StateErrors if it is not already a StateError.
+func stateErrorOrUnavailable(err error) []*client.StateError {
+
+	// unpack aggregate before converting to state errors
+	var serr *client.StateError
+	if errors.As(err, &serr) {
+		return []*client.StateError{serr}
+	}
+
+	// convert any generic error to 2 state-errors -> Unavailable & Degraded
+	return []*client.StateError{
+		client.NewAvailabilityError(err.Error()),
+		client.NewDegradedError(err.Error()),
+	}
+}
+
+func toStateErrors(err error) []*client.StateError {
+
+	serrs := []*client.StateError{}
+
+	var aggregate apiutilerrors.Aggregate
+	if errors.As(err, &aggregate) {
+		// unpack aggregate before converting to state errors
+		errs := apiutilerrors.Flatten(aggregate).Errors()
+		for _, err := range errs {
+			serrs = append(serrs, stateErrorOrUnavailable(err)...)
+		}
+	} else {
+		serrs = append(serrs, stateErrorOrUnavailable(err)...)
+	}
+	return serrs
+}
+
+func generateRunReportFromTaskErrors(tge tasks.TaskGroupErrors) (runReport, string) {
+
+	failedTask := cmostr.ToPascalCase(tge[0].Name) + "Failed"
+	if len(tge) > 1 {
+		failedTask = "MultipleTasksFailed"
+	}
+
+	degraded := &stateInfo{reason: failedTask, status: client.UnknownStatus}
+	available := &stateInfo{reason: failedTask, status: client.UnknownStatus}
+
+	for _, terr := range tge {
+
+		// each task can return a single or multiple errors (as an Aggregate)
+		// each error can be a StateError or a generic error (fmt.Errorf)
+
+		for _, serr := range toStateErrors(terr.Err) {
+
+			switch serr.State {
+			case client.DegradedState:
+				degraded.messages = append(degraded.messages, serr.Reason)
+				if !serr.Unknown {
+					degraded.status = client.TrueStatus
+				}
+
+			case client.UnavailableState:
+				available.messages = append(available.messages, serr.Reason)
+				if !serr.Unknown {
+					available.status = client.FalseStatus
+				}
+			}
+		}
+	}
+
+	rpt := runReport{}
+
+	if len(degraded.messages) != 0 {
+		rpt.degraded = degraded
+	} else {
+		rpt.degraded = asExpected(client.FalseStatus)
+	}
+
+	if len(available.messages) != 0 {
+		rpt.available = available
+	} else {
+		rpt.available = asExpected(client.TrueStatus)
+	}
+
+	return rpt, failedTask
+}
+
+// stateInfo satifies a client.StateInfo
+type stateInfo struct {
+	status   client.Status
+	reason   string
+	messages []string
+}
+
+var _ client.StateInfo = (*stateInfo)(nil)
+
+func (si stateInfo) Status() client.Status {
+	return si.status
+}
+
+func (si stateInfo) Message() string {
+	return strings.Join(si.messages, ", ")
+}
+
+func (si stateInfo) Reason() string {
+	return si.reason
+}
+
+// expectedStatus is a client.StateInfo which is returned when the state
+// of the system is as expected.
+type expectedStatus client.Status
+
+var _ client.StateInfo = (*expectedStatus)(nil)
+
+func (expectedStatus) Message() string {
+	return ""
+}
+
+func (s expectedStatus) Status() client.Status {
+	return client.Status(s)
+}
+
+func (expectedStatus) Reason() string {
+	return "AsExpected"
+}
+
+func asExpected(s client.Status) *expectedStatus {
+	ret := expectedStatus(s)
+	return &ret
+}
+
+type runReport struct {
+	degraded  client.StateInfo
+	available client.StateInfo
+}
+
+var _ client.StatesReport = (*runReport)(nil)
+
+func newRunReportForError(reason string, err error) runReport {
+	return runReport{
+		degraded: &stateInfo{
+			status:   client.TrueStatus,
+			reason:   reason,
+			messages: []string{err.Error()},
+		},
+		available: &stateInfo{
+			status:   client.FalseStatus,
+			reason:   reason,
+			messages: []string{err.Error()},
+		},
+	}
+}
+func (r runReport) isDegraded() bool {
+	if isNilOrAsExpected(r.degraded) {
+		return false
+	}
+	return true
+}
+
+func (r runReport) isUnavailable() bool {
+	if isNilOrAsExpected(r.available) {
+		return false
+	}
+	return true
+}
+
+func (r runReport) Available() client.StateInfo {
+	return r.available
+}
+
+func (r runReport) Degraded() client.StateInfo {
+	return r.degraded
+}
+
+func isNilOrAsExpected(s client.StateInfo) bool {
+	if s == nil {
+		return true
+	}
+
+	switch s.(type) {
+	case *expectedStatus:
+		return true
+	default:
+		return false
+	}
 }
--- a/pkg/tasks/prometheus.go
+++ b/pkg/tasks/prometheus.go
@@ -22,6 +22,7 @@ import (
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apiutilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
 )
 
@@ -40,6 +41,24 @@ func NewPrometheusTask(client *client.Client, factory *manifests.Factory, config
 }
 
 func (t *PrometheusTask) Run(ctx context.Context) error {
+
+	errs := []error{}
+
+	err := t.create(ctx)
+	if err != nil {
+		klog.V(4).ErrorS(err, "updation of prometheus failed")
+		errs = append(errs, err)
+	}
+
+	// NOTE: the validation task is run even if creation fails so that
+	// existing deployment is validated.
+	validation := NewPrometheusValidationTask(t.client, t.factory)
+	errs = append(errs, validation.Run(ctx))
+
+	return apiutilerrors.NewAggregate(errs)
+}
+
+func (t *PrometheusTask) create(ctx context.Context) error {
 	cacm, err := t.factory.PrometheusK8sServingCertsCABundle()
 	if err != nil {
 		return errors.Wrap(err, "initializing serving certs CA Bundle ConfigMap failed")
@@ -345,12 +364,6 @@ func (t *PrometheusTask) Run(ctx context.Context) error {
 		err = t.client.CreateOrUpdatePrometheus(ctx, p)
 		if err != nil {
 			return errors.Wrap(err, "reconciling Prometheus object failed")
-		}
-
-		klog.V(4).Info("waiting for Prometheus object changes")
-		err = t.client.WaitForPrometheus(ctx, p)
-		if err != nil {
-			return errors.Wrap(err, "waiting for Prometheus object changes failed")
 		}
 	}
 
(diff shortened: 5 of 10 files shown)


jan--f (changes requested):
Looks like a good start.
What is the motivation for pulling the validation step out of a tasks run? This seems to cause some awkwardness in the status reporting and adds a lot of code with no description in the commit or PR.

Some typos and questions inline. There are also several `TODO`s and `FIXME`s as well as code in comments still in there.

pkg/operator/operator.go
```diff
@@ -480,7 +484,7 @@ func (o *Operator) handleEvent(obj interface{}) {
 		return
 	}
 
-	klog.V(5).Infof("ConfigMap or Secret updated: %s", key)
+	klog.V(8).Infof("ConfigMap or Secret updated: %s", key)
```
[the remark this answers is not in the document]
sthaha: good catch! no good reason, other than that I changed that while debugging

pkg/operator/operator.go
```diff
@@ -632,16 +636,38 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 		klog.Errorf("error occurred while setting status to in progress: %v", err)
 	}
 
-	taskErrors := tl.RunAll(ctx)
```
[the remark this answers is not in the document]
sthaha: because there is a run - step and a validation step

pkg/client/client.go
```diff
@@ -761,6 +763,126 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
...
+	//
+	// Total Pods = Available + Unavailable
+	// Pods Running but NEEDS updation = Available - UpdatedReplicas (also returned as []pods)
+	// NeedsUpdation = Available - UpdatedReplicas
+
+	required := *p.Spec.Replicas
+	available := status.AvailableReplicas
+	uptoDate := status.UpdatedReplicas
+
+	needsUpdation := available - uptoDate
```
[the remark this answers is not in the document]
sthaha: Old habits die hard :D

pkg/client/client.go
```diff
@@ -761,6 +763,126 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
...
+		// All pods are upto date and nothing to do; see reason above
+		//   -- OR --
+		// all pods need to be updated but they are running, in which case, we
+		// treat it as valid and the next reconciliation loop will tell us if the
+		// states are okay or not
+		return nil
+	}
+
+	// some or all  pods are NOT running, find the reason
+	// FIXME(sthaha): this assumes that the p.Spec.Shard is nil which is the case in OCP
```
sthaha, line 815: @jan--f Fixing this requires upstream change to promoperator. Which I will so that can call `promoperator.Pods(prometheus)`

pkg/tasks/prometheus.go
```diff
@@ -384,18 +387,20 @@ func (t *PrometheusTask) Run(ctx context.Context) error {
...
 		if err != nil {
 			return errors.Wrap(err, "reconciling Prometheus object failed")
 		}
 
-		klog.V(4).Info("waiting for Prometheus object changes")
-		err = t.client.WaitForPrometheus(ctx, p)
-		if err != nil {
-			return errors.Wrap(err, "waiting for Prometheus object changes failed")
-		}
+		// TODO(sthaha): delete this
```
sthaha, line 398: need to delete this.

pkg/client/client.go
```diff
@@ -761,6 +763,126 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
...
+}
+
+var _ error = (*ValidationError)(nil)
+
+func (c *Client) validatePromethuesPodState(ctx context.Context, p *monv1.Prometheus) *ValidationError {
+
+	status, _, err := prometheusoperator.Status(ctx, c.kclient.(*kubernetes.Clientset), p)
+	if err != nil {
+		klog.V(4).ErrorS(err, "validatePrometheusPodState: failed to get Prometheus status")
+		// TODO(sthaha): do we want to not set Degraded in this case?
```
sthaha, line 786: @simonpasquier @jan--f Any thoughts if we want to set Degraded if prometheusoperator.Status returns err?

pkg/client/client.go
```diff
@@ -761,6 +763,126 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
...
+		// states are okay or not
+		return nil
+	}
+
+	// some or all  pods are NOT running, find the reason
+	// FIXME(sthaha): this assumes that the p.Spec.Shard is nil which is the case in OCP
+	pods, err := c.kclient.CoreV1().Pods(p.Namespace).List(ctx, prometheusoperator.ListOptions(p.Name))
+	if err != nil {
+		// TODO(sthaha): should we get Unavailable ?? and Degraded ??
+		return &ValidationError{
```
sthaha, line 819: Should we set `Degraded` and `Unavailable` if `List` fails?

sthaha:
> This seems to cause some awkwardness in the status reporting and adds a lot of code with no description in the commit or PR.

@jan--f I happy to add description to wherever necessary. Currently I have updated the commit and the PR description.

pkg/client/client.go
simonpasquier: `Unknown` might be the best choice?

pkg/client/client.go
```diff
@@ -761,6 +763,126 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
...
+	// Pods Running but NEEDS updation = Available - UpdatedReplicas (also returned as []pods)
+	// NeedsUpdation = Available - UpdatedReplicas
+
+	required := *p.Spec.Replicas
+	available := status.AvailableReplicas
+	uptoDate := status.UpdatedReplicas
+
+	needsUpdate := available - uptoDate
+
+	if uptoDate >= required || needsUpdate >= required {
```
simonpasquier, line 805: Comparing `available` and `required` directly should yield the same result?
```suggestion
	if available >= required {
```

pkg/client/client.go
simonpasquier: https://github.com/prometheus-operator/prometheus-operator/pull/4580 will make the upstream fix obsolete.

pkg/client/status_reporter.go
```diff
@@ -163,6 +164,37 @@ func (r *StatusReporter) SetRollOutInProgress(ctx context.Context) error {
...
+	time := metav1.Now()
+	// The Reason should be upper case camelCase (PascalCase) according to the API docs.
+	reason = cmostr.ToPascalCase(reason)
+
+	conditions := newConditions(co.Status, r.version, time)
+	conditions.setCondition(v1.OperatorAvailable, v1.ConditionFalse, statusErr.Error(), reason, time)
+
+	return r.setConditions(ctx, co, conditions)
+}
+func (r *StatusReporter) SetDegraded(ctx context.Context, statusErr error, reason string) error {
```
simonpasquier, line 181: (nit)
```suggestion
}

func (r *StatusReporter) SetDegraded(ctx context.Context, statusErr error, reason string) error {
```

pkg/tasks/prometheus.go
```diff
@@ -420,3 +418,13 @@ func (t *PrometheusTask) Run(ctx context.Context) error {
 
 	return nil
 }
+
+func (t *PrometheusTask) Validate(ctx context.Context) *client.ValidationError {
+	klog.V(4).Info("validate Prometheus object")
+	if t.prometheus == nil {
```
simonpasquier, line 424: this should never happen right? if so, I'd prefer a hard fail with panic()

pkg/tasks/tasks.go
```diff
@@ -112,7 +162,18 @@ type TaskSpec struct {
 	Task Task
 }
 
+type Validator interface {
+	Validate(ctx context.Context) *client.ValidationError
```
simonpasquier, line 166: It feels strange to return a struct from the `client` package.

pkg/tasks/prometheus.go
```diff
@@ -384,18 +387,20 @@ func (t *PrometheusTask) Run(ctx context.Context) error {
 		if err != nil {
 			return errors.Wrap(err, "initializing Prometheus object failed")
 		}
+		t.prometheus = p
```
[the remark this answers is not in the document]
simonpasquier: no blocker for me but maybe we could just store a copy of `p.ObjectMeta`?

pkg/tasks/prometheus.go
sthaha: I wasn't too sure about `panic` :).  Will change to use `panic`

pkg/tasks/tasks.go
sthaha: I think we should must return `TaskValidationError` instead. Good point! 🙇

pkg/client/client.go
sthaha: @simonpasquier do you mean we  add an `Unknown` as well to ValidationError?

pkg/client/client.go
sthaha: +1 since these pods are already running

pkg/tasks/tasks.go
sthaha: We can't return `TaskValidationError` which needs to know the `name` of the Task. Hence I am aliasing client.ValidationError to ValidationError

pkg/operator/operator.go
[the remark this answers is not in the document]
jan--f: I see. What is the motivation for separating these two steps? The current "validation" is part of the task run, wouldn't that work for the more extensive validation as well?

The reason I'm asking is that this does add additional code that is close to a duplication of existing code (error handling after `RunAll` and `ValidateAll`, and more concerning imho `ValidateAll` and `ValidateConcurrently` in the `task` module).
I would also argue that linking resource creation and subsequent validation make sense in the same task, since one should happen after the other. By running them independently we have to handle an additional error case (https://github.com/openshift/cluster-monitoring-operator/pull/1558/files#diff-b0ca3a050a076f5a4007692cff8088723c9fe6f8c6734f8a8bc1bdd54f8d624cR423).
I feel like I'm missing something.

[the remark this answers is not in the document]
sthaha: > The current "validation" is part of the task run, wouldn't that work for the more extensive validation as well?

That is the initial thought I had as well, but when Simon and I discussed, we felt this approach is the better one to take since it separates the concerns correctly. What we concluded was that there are 2 concerns here 

1. Create / update resources: These are non blocking and independent steps which are fire and forget and should be non-blocking. E.g. the task that applies the prometheus resources shouldn't need to validate (blocking `wait.Poll()` ). Its responsibility is to ensure that the resources can be created and updated properly, failing which means the operator is in a `degraded` state.

2. `Validate` if the system is in the right state and ideally these should can run independent of the `Run` albeit in this PR I do cache prometheus in `t.prometheus`


> The reason I'm asking is that this does add additional code that is close to a duplication of existing code (error handling after `RunAll` and `ValidateAll`, and more concerning imho `ValidateAll` and `ValidateConcurrently` in the `task` module).

I do think we can keep the code dry at the cost of increased complexity and since it's an internal implementation, we can change it anytime.

> I would also argue that linking resource creation and subsequent validation make sense in the same task, since one should happen after the other.

IMHO, validation of a system's state shouldn't require a create/update step to have been executed. I feel these are independent of one another.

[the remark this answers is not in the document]
jan--f: I have strong feelings about being able to maintain things in the future. Most other concerns should serve that in my opinion :)

Talking about an individual task (say `PrometheusOperator`) I agree that Create/Update has no dependency on a Validate step. I would however argue that Validate entirely depends on the state we want and the state we have. Validate needs an input to check the actual state against. In CMO this expected state is very predictable and static (we don't expect any other actors to make any changes), it is in fact the input to Create/Update.
As you say yourself:
> Validate if the system is in the right state and ideally these should can run independent of the Run albeit in this PR I do cache prometheus in t.prometheus

Validate needs an input to compare the actual state of the world against.

To accomplish what the PR title says we can clearly distinguish between errors in the Create/Update step and validation errors. This can probably be done with custom errors alone. If we have dependencies between tasks, we should run these tasks in different task groups (as we are doing already).

And just a pedantic point: `Run` is just named `Run` as part of the task runner implementation. What that task runs is what we are discussing. Imho a task should contain the Create and Validate steps. If there are good arguments for distinguishing these steps and running them independently, we should have create tasks and validation tasks (one for every task we currently have). I don't think the task runner needs to know what a validation is, that knowledge belongs into the validation task.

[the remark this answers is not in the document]
sthaha: > I have strong feelings about being able to maintain things in the future. Most other concerns should serve that in my opinion :)

Well, the feeling is mutual, but IMHO, the premise feels faulty. It already starts with an assumption  that the code is unmaintainable rather than providing arguments that leads the me to understand why it is so  and leaves me to defend why it is more maintainable  😂

IOW, it fails to make the argument that making the `validation` separate from create/update leads to an unmaintainable state while providing reasons why keeping them close leads to more maintainable code. 

> Talking about an individual task (say PrometheusOperator) I agree that Create/Update has no dependency on a Validate step. 

Thinking about it, there is a actually a dependency that the validation "needs to know" the state expected by the `run` step; i.e. it should have the knowledge of what `run` step intended the system's state to be for it to validate **correctly** and therefore these aren't strictly independent. However this can be rectified by allowing the operator that makes both calls (Run, Validate) to consolidate the errors. 


### Scenario

Given: there already exists a `prometheus` CR and its pods are running fine.

#### Using the current implementation 

1. Now CMO runs  `prometheus` Task ( increase replica count for instance) 
2. the `run` step fails to apply `promethues` CR (for whatever reason). 
3. It sets the state as `degraded` 
4. Operator waits for all task `run`s to finish 
5. Operator runs `validate` step
6. The validate step gets the existing (but old) `prometheus` CR
7. It checks if the pods are running fine (which they are) 
8. Validates reports that system is `{ Degraded: false, Available: true }`.

This contradicts what the `run` step intents the system's state to be. 

In the current implementation, since the `run` step returned `degraded` and the `validation` step returned `fine`, CMO would report `{Degraded: true, Available true}`

NOTE: There is  inconsistency between what `run` step reports and what `validate` reports but it is consolidated by the `operator` and reports the state correctly.

Also note that the current structure ensures that all Tasks are validated and isn't the responsibility of an individual task to ensure that validation is performed in case of a failure to create/update. 

Now let's take the implementation that is being suggested and compare it to the approach in this PR. 

#### Combine Create and Validation into `Run`

1. CMO runs  `prometheus` Task ( increase replica count for instance) 
2. the `run` step failed to apply `promethues` CR (for whatever reason). 
4. It returns the state as `degraded`. ( see unless... )
5. The validate step won't even run, so the state isn't validated 

CMO would report `{Degraded: true, Available: ???}`

#### Unless every `Run` Task is structured differently

```go
type TaskError struct {
  Unknown bool
  Unavailable bool
  Degraded bool

}

func (t Task) Run() TaskError {
   var creationError = t.create()
   var validationError = t.validate()
   return combine(creationError, validationError)
}
```

Issues: 
Here the task instead of the `operator` ensures that the validation is run after create and combines the error but the knowledge that the `validation` **has to be run after `create`** is lost. 

Advantages: 
* Can Piggy back on current implementation of `Task.Run` albeit all `Run` need change to return the `TaskError` 

I am happy to take this approach but I will wait for a consensus before reorganising the code. 

cc: @jan--f @simonpasquier

juzhao:
cc: @lihongyan1

pkg/operator/operator.go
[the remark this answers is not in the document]
jan--f: Thank you for elaborating! But me apologize first.

> IOW, it fails to make the argument that making the validation separate from create/update leads to an unmaintainable state while providing reasons why keeping them close leads to more maintainable code.

By no means did I intent to say this code is unmaintainable, only that I think we can improve it and make the life of future developer easier. For example duplicate code will likely drift in future iterations and thus increase complexity. A `Task`, that so far only knows how to `Run` something, now learns about an individual tasks internals (that it has a validation). This increases the context a developer has to keep when thinking about tasks, a step that imho should not be taken lightly.

Your `Task` restructuring proposal makes perfect sense to me. This keeps with how the tasks are structured today but with better error reporting and improved validation, as your PR proposes. I'm not sure I follow the issue you point out. The knowledge that validation has to happen after create is not lost, its encoded in the task at hand. I think this is the right place for it too.
The operator however encodes how task groups depend on each other. When two tasks depend on each other they should be run in two subsequent task groups.
The current implementation blurs this a bit though. By running _all_ create tasks before validation we would introduce some noise into the status reporting, e.g. creating a Prometheus CR will have no effect if creating the initial prometheus-operator deployment failed, yet both would be reported.

pkg/client/state_error.go
```diff
@@ -0,0 +1,178 @@
+package client
```
sthaha, line 1: missing header

pkg/operator/operator.go
```diff
@@ -638,16 +642,9 @@ func (o *Operator) sync(ctx context.Context, key string) error {
...
-	if len(taskErrors) > 0 {
-		var failedTask string
-		if len(taskErrors) == 1 {
-			failedTask = cmostr.ToPascalCase(taskErrors[0].Name + "Failed")
-		} else {
-			failedTask = "MultipleTasksFailed"
-		}
-
-		o.reportError(ctx, taskErrors, failedTask)
+	runErrors := tl.RunAll(ctx)
```
sthaha, line 645: change this back to taskErrors

pkg/operator/operator.go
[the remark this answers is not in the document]
sthaha: Hopefully, this change capture what we discussed 🤞

pkg/tasks/tasks.go
```diff
@@ -112,24 +113,65 @@ type TaskSpec struct {
...
 
 func (tge TaskGroupErrors) Error() string {
 	if len(tge) == 0 {
 		return ""
 	}
+
 	messages := make([]string, 0, len(tge))
-	for _, err := range tge {
-		messages = append(messages, fmt.Sprintf("%v: %v", strings.ToLower(err.Name), err.Err))
+	for _, tve := range tge {
```
sthaha, line 156: revert name change

pkg/tasks/tasks.go
```diff
@@ -112,24 +113,65 @@ type TaskSpec struct {
...
+func (tge TaskGroupErrors) ToConsolidatedTaskError() *TaskErr {
+	if len(tge) == 0 {
+		return nil
+	}
+
+	if len(tge) == 1 {
+		return &tge[0]
+	}
+
+	re := client.NewStateError()
```
sthaha, line 171: better name?

jan--f (changes requested):
Nice, this definitely improves things.

One thing I don't quite understand is the `StateError` type. It tracks `State`s by type, though the `State` also has a type (which seems unused). It can track all types of errors, but we don't actually seem to use that feature, i.e. we use either `se.State(Degraded).AddReasons` or `se.State(Unavailable).AddReasons`, but never both.
Wouldn't it be sufficient (and simpler) to have define an error interface for tasks and distinguish the 2 or 3 error types we are interested in. Then a task can just return a list of errors. Looking at the BZ we are interested in the condition an error causes and a message what causes it.

We talked about splitting creation and validation into separate tasks. A create task can then create a validation task and pass the state to be validated. This would allow us to run a validation separately and tasks would no longer need to track any resource specs. We can also leave that for later though, right now we aren't running validations separately afaiu.

pkg/tasks/prometheus.go
```diff
@@ -36,7 +38,21 @@ func NewPrometheusTask(client *client.Client, factory *manifests.Factory, config
...
+
+	return client.MergeStateErrors(createErr, validationErr)
+}
+
+func (t *PrometheusTask) RunX(ctx context.Context) *StateError {
+	createErr := degradedError(t.create(ctx))
+	validationErr := t.validate(ctx)
+
+	return mergeErrors(createErr, validationErr)
+}
```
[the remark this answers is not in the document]
sthaha: yes. I thought the refactor or the Tasks's validation and creation as 2 separate tasks can be another refactor PR instead of piggy backing on this bug. WDYT?

sthaha:
> Nice, this definitely improves things.
> 
> One thing I don't quite understand is the `StateError` type. It tracks `State`s by type, though the `State` also has a type (which seems unused).
Let me remove the `Type` in State albeit I have hunch that it may be useful. 

>  It can track all types of errors, but we don't actually seem to use that feature, i.e. we use either `se.State(Degraded).AddReasons` or `se.State(Unavailable).AddReasons`, but never both.

Oh!! Good catch! 
You are right! but actually it should have been used!! This was lost in the refactor I did from the previous version. 
When prometheus is unavailable. We should also set the condition `Degraded`. 

```go
	if available == 0 {
		se.State(Degraded).AddReasons("None of the prometheus pods are running")
		addReason = se.State(Unavailable).AddReasons
	}
```


>  Wouldn't it be sufficient (and simpler) to have define an error interface for tasks and distinguish the 2 or 3 error types we are interested in.  Then a task can just return a list of errors. Looking at the BZ we are interested in the condition an error causes and a message what causes it.

We could do that however my preference would be to return a single error instead of many since we have to now manage / merge these errors like how the  `[] TaskError ` is abstracted into `TaskGroupError`

The current pattern allows for all `Degraded` or `Unavailble` reasons to be collated into one unit. 

```
func foo() *StateError {
   err := NewStateError()

  if !bar {
    err.State(Degraded).AddReason(" for failure")
  }
  if !bzzz {
    err.State(Unavailable).AddReason(" for failure")
  }
  if !xyz {
    err.State(Degraded).AddReason("for failure")
  }

 return err.Cleanup() // << returns nil if err object is empty
}

// Usage is 

if err := foo(); err != nil {
  fmt.Println("%v", err)
}
```

As opposed to ... 

```
type StateErrors []StateError

func (se StateErrors) Error() string {
   // iterate through unvailable first and collate the errors
   // iterate degraded errors and collote those
  
}

func foo() StateErrors {
   errs := StateErrors{}

  if !bar {
    errs := append(errs, DegradedError(" for failure")) 
  }
  if !bzzz {
    errs := append(errs, UnavailableError(" for failure"))
  }
  if !xyz {
    errs := append(errs, DegradedError(" for failure"))
  }

 return errs
}

// Usage is 

if errs := foo(); len(errs) != 0 {
  // handle err
}
```
I prefer the first version to the  second version because just like in TaskGroupErrors,
1.  We have to create another abstraction which is leaky because the callers has to know that StateErrors is `[]StateError` inorder to use ` errs := append(errs, DegradedError(" for failure")) ` 
2. We have to iterate over the errors and collate those

3. it allows for usage that you see in `Prometheus Validation`  which returns a `single` error for prometheus validation failure instead of many.

```
err := NewStateError()
err.State(Degraded).AddReason("No prometheus pods are running")
for each pod {
  err.State(Unavailable).AddReason("... based on pod state")
}
```

> 
> We talked about splitting creation and validation into separate tasks. A create task can then create a validation task and pass the state to be validated. This would allow us to run a validation separately and tasks would no longer need to track any resource specs. We can also leave that for later though, right now we aren't running validations separately afaiu.

You are right and I thought we can do the refactor as a follow up patch. Currently this PR does not separate out validation so that the fix for the bug can be delivered first, tested. WDYT?

sthaha:
@jan--f How about we keep the validation refactor discussions here on #1605 so that this fix can progress if the implementation addresses the bug?

jan--f:
I assumed with the status subresource of the Prometheus CR we'll rely on that instead of checking individual pod states.
Is there anything missing in the new status subresource?

sthaha:
> I assumed with the status subresource of the Prometheus CR we'll rely on that instead of checking individual pod states. Is there anything missing in the new status subresource?

@jan--f You are right, the prometheus status  is what we should rely on, however at the time this was submitted (IIRC), we didn't have the new PO merged or there was a revert patch. Now that the latest PO is merged, I am going to use that to validate prometheus status.

Could you please review if the rest of the implementation (using Aggregate) looks good enough?

jan--f:
> Could you please review if the rest of the implementation (using Aggregate) looks good enough?

Yup I think this is a good approach. I'm wondering whether we should think about replacing `TaskGroupErrors` with an `Aggregate` as well, but happy to tackle this in a separate PR.

sthaha:
> replacing TaskGroupErrors with an Aggregate as well, but happy to tackle this in a separate PR.

Yes, lets do that in  separate PR :)

pkg/client/client.go
```diff
@@ -864,6 +889,59 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
 	return nil
 }
 
+func (c *Client) ValidatePrometheus(ctx context.Context, promNsName types.NamespacedName) error {
+	var validationError error
+
+	pollErr := wait.Poll(10*time.Second, 5*time.Minute, func() (bool, error) {
```
sthaha, line 895: is 10s poll too aggressive and 5m timeout too much :) ?

pkg/client/client.go
```diff
@@ -864,6 +889,59 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
...
+			errs = append(errs, NewDegradedError(msg))
+			newStateError = NewUnavailableError
+		}
+
+		errs = append(errs, newStateError(msg))
+		validationError = apiutilerrors.NewAggregate(errs)
+		return false, nil
+	})
+
+	if pollErr == wait.ErrWaitTimeout {
```
jan--f, line 929: When we timeout retrieving the prometheus object, shouldn't this create an UnknownStateError?

sthaha: Good catch!! 
I think it makes sense to mark both Degraded and Available as `Unknown` WDYT? 
cc: @simonpasquier

pkg/operator/operator.go
```diff
@@ -977,3 +1020,152 @@ func (o *Operator) workloadsToRebalance() []rebalancer.Workload {
 	}
 	return workloads
 }
+
+// toStateError converts an error to Degraded StateError if it is not
+// already a StateError.
+func toStateError(err error) *client.StateError {
```
sthaha, line 1026: I am not happy with the func name - any suggestions?

pkg/client/client.go
```diff
@@ -864,6 +889,58 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
...
+
+	pollErr := wait.Poll(3*time.Second, 15*time.Second, func() (bool, error) {
+
+		p, err := c.GetPrometheusByNsName(ctx, promNsName)
+		if err != nil {
+			klog.V(4).Info("validate prometheus failed to get prometheus: ", err)
+			return false, err
+		}
+
+		avail, err := prometheusConditonForType(p.Status.Conditions, monv1.PrometheusAvailable)
```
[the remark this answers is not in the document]
jan--f: I think similarly to failure to retrieve prometheus `Unkown` for the available condition makes sense, doesn't it?

pkg/client/client.go
```diff
@@ -864,6 +889,59 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
...
+
+		// return reason for failure as state-errors - Degraded/Unavailable
+		errs := []error{}
+		msg := fmt.Sprintf("%s: %s", avail.Reason, avail.Message)
+
+		newStateError := NewDegradedError
+		if avail.Status == monv1.PrometheusConditionFalse {
+			// prometheus not available is a Degraded and Unavailable error
+			errs = append(errs, NewDegradedError(msg))
+			newStateError = NewUnavailableError
```
jan--f, line 921: Not sure I follow this. If the available condition is false (or unavailable), to that is a stronger statement then degraded. Degraded tells me that it _is_ available, just not at the level specified.

jan--f: hm I think I misunderstood. In p-o the `Available` condition can be `Degraded`.
I'm not sure we want to set the `Degraded` condition here. Reconciliation might have worked fine and Prometheus might be unavailable for other reasons.

pkg/client/client.go
jan--f: I think that makes sense yes. Though based on my comment further up, maybe it makes sense to change this function to check for the `Available` and `Reconciled` conditions of Prometheus separately and deduce from that which errors we want to pass?

pkg/tasks/prometheus.go
```diff
@@ -38,6 +39,26 @@ func NewPrometheusTask(client *client.Client, factory *manifests.Factory, config
...
+		klog.V(4).ErrorS(err, "updation of prometheus failed")
+		errs = append(errs, err)
+	}
+
+	// NOTE: the validation task is run even if creation fails so that existing
+	// deployment is validated.
+	validation := NewPrometheusValidationTask(t.client, t.factory)
+	if err = validation.Run(ctx); err != nil {
+		errs = append(errs, err)
+	}
```
jan--f, line 56: Since we create an `Aggregate` from this list, here is no need to check for nil, `NewAggregate` will do that.

jan--f:
I like how this passes errors down from the tasks. I do have some difficulty following the reporting code though.

Right now the code iterates over the `TaskGroupErrors` to convert them to `[]StateError`. Right after we iterate over the this list of `StateError` to aggregate the reasons in a `RunReport`. This is then passed to `ReportStatus()`. Would it make sense to do the aggregation in `ReportStatus()` itself and avoid the intermediate representation? Unless I'm missing something this should be fairly straight forward, even more so once we refactor `TaskGroupErrors` to an `Aggregate` as well. 

Somewhat unrelated to this: iiuc if one `StateError` that comes back `Unknown` is reported, we report an unknown status for the respective condition. So if we have `[StateError{State: DegradedState, Unknown: false, Reason: reason}, StateError{State: DegradedState, Unknown: true, Reason: reason}` we'd report the Degraded condition as unknown..?
Should this not report Degraded as true, since we know at least one component is positively degraded?

pkg/client/client.go
sthaha: This actually aligns with our earlier discussions that Prometheus becoming unavailable should set `Degraded` condition of CMO to `true`.

pkg/client/client.go
sthaha: I feel that  from the point of view of CMO, failure to get the state information of `prometheus` should be communicated to back to the user as  `Degraded` and `Available` (of CMO) as `unknown`
WDYT?

pkg/client/client.go
raptorsun: This is 0.1 Qps, I think it is not significant traffic to the Prometheus.

sthaha:
> Right now the code iterates over the TaskGroupErrors to convert them to []StateError. Right after we iterate over the this list of StateError to aggregate the reasons in a RunReport. This is then passed to ReportStatus(). 

I agree this isn't easy to follow and should simplify when everything moves to using only Aggregate and StateErrors (no TaskGroupError, TaskErr) until then, this might help 😇

Here is the flow (to provide some clarity to other reviewers)

* Tasks returns `error` it can be any error - (Aggregate, StateError, fmt.Errorf...)
* TaskGroup (a group of `Task`s) returns TaskGroupErrors (plural) which is `[]TaskErr` and each error returned from a `Task` is `embedded` in  `TaskErr.Err`
* `TaskRunners`'s `RunAll` func returns TaskGroupErrors


The `StatusReporter` has no knowledge (no tight coupling) of these errors ( StateErrors / TaskErr / Aggregates )  by design and defines `StatesReport` (interface) which it can report using the `ReportState` function. 

`StateReport` requires implementor to provide (satisfy `StateInfo` interface) State Information about `Degraded` and `Available` states. A `StateInfo` contains `Status - True|False|Unknown`, `Message: string`, `Reason` . This design makes implementation of StatusReporter quite simple and easy to understand.  E.g. see `asExpected` 

---

> Would it make sense to do the aggregation in ReportStatus() itself and avoid the intermediate representation? Unless I'm missing something this should be fairly straight forward, even more so once we refactor TaskGroupErrors to an Aggregate as well.

tl;dr: The intermediate representation is needed to satisfy ( or implement) the interface `StatesReport`.  The runReport provides information about the `Degraded` and `Available` state. 


The confusion arises when we turn all the errors (`tasks * errors` > len(task)) mentioned above into **`a`** a single StateReport and rules are as follows 

* A StateError contains all information needed to report the state properly, so use it as is 
* Any `error` that isn't a `StateError` will be converted to a degraded state error. 
* All reasons need to be combined into a single reason

Since it is the operator that runs the Tasks, gets a `TaskGroupError ( = [ ]TaskErr)`, it naturally follows that the operator aggregates all these errors, create a `StateReport` and pass it to the `StatusReporter` to report the state. 

It does this in the following method (edited)  where all errors are collated into a `runReport` that satisfies the `StateInfo`

```go
func runReportForTaskGroupErrors(tge tasks.TaskGroupErrors) *runReport {

	rpt := &runReport{name: tge[0].Name}

	degraded := &client.StateError{State: client.DegradedState}
	unavailable := &client.StateError{State: client.UnavailableState}

	degradedReasons, unavailableReasons  := []string{}, []string{}

	for _, err := range taskGroupErrorsToStateErrors(tge) {
		switch err.State {
		case client.DegradedState:
			degradedReasons = append(degradedReasons, err.Reason)
			// if any of the errors are marked as unknown, retain that in the report
			degraded.Unknown = degraded.Unknown || err.Unknown

		case client.UnavailableState:
			unavailableReasons = append(unavailableReasons, err.Reason)
			unavailable.Unknown = unavailable.Unknown || err.Unknown
		}
	}

	degraded.Reason = strings.Join(degradedReasons, ", ")
	rpt.degraded = degraded

	unavailable.Reason = strings.Join(unavailableReasons, ", ")
	rpt.unavailable = unavailable

	return rpt
}
```
NOTE: `taskGroupErrorsToStateErrors` is responsible for converting errors to `StateError` so that it can be identified as `Degraded` - This method should go away when we switch over to using `Aggregate` 


Now, why does the `runReport` (re)use StateError for `available` and `degraded` ?
It is only because the `StateError` hold all the details needed to provide a State information - `Status,  Reason` it actually satisfies `StateInfo`
---



> Somewhat unrelated to this: iiuc if one StateError that comes back Unknown is reported, we report an unknown status for the respective condition. So if we have [StateError{State: DegradedState, Unknown: false, Reason: reason}, StateError{State: DegradedState, Unknown: true, Reason: reason} we'd report the Degraded condition as unknown..?
Should this not report Degraded as true, since we know at least one component is positively degraded?

Fair point  and I agree with you, it even simplifies the implementation a bit :) 
@simonpasquier WDYT?

jan--f:
Thanks for the clarifying comments. I wonder if the naming could be improved to make this a bit clearer. I'll think about this some more.

in any case I still don't see the point of iterating over the error list twice. To reuse your code example:
``` golang
func runReportForTaskGroupErrors(tge tasks.TaskGroupErrors) *runReport {

	rpt := &runReport{name: tge[0].Name}

	degraded := &client.StateError{State: client.DegradedState}
	unavailable := &client.StateError{State: client.UnavailableState}

	degradedReasons, unavailableReasons  := []string{}, []string{}

        // taskGroupErrorsToStateErrors iterates over all task group errors, transforms them
        // to a state error and adds them individually to a list. This list is returned below, only
        // to be iterated over for information extraction. shouldn't this be done in the same iteration?
	for _, err := range taskGroupErrorsToStateErrors(tge) {
		switch err.State {
		case client.DegradedState:
			degradedReasons = append(degradedReasons, err.Reason)
			// if any of the errors are marked as unknown, retain that in the report
			degraded.Unknown = degraded.Unknown || err.Unknown

		case client.UnavailableState:
			unavailableReasons = append(unavailableReasons, err.Reason)
			unavailable.Unknown = unavailable.Unknown || err.Unknown
		}
	}

	degraded.Reason = strings.Join(degradedReasons, ", ")
	rpt.degraded = degraded

	unavailable.Reason = strings.Join(unavailableReasons, ", ")
	rpt.unavailable = unavailable

	return rpt
}

**Comment by sthaha** [comment]
> in any case I still don't see the point of iterating over the error list twice. To reuse your code example:

Any suggestion? the hierarchy is like this 

```
- [] Task Error
          + Error -> Aggregate (list) | StateError | error 
                              + - error | StateError 
```
Given multiple Task Errors, with each Task Error may contain multiple Errors, you have a list of a list and requires 2 loops.

**Comment by sthaha** [comment]
> I wonder if the naming could be improved to make this a bit clearer. 

@jan--f  Happy to change to a better name. My logic was that the function produces a  (task) `run` `report` `for` given `Task Errors`.

**Comment by jan--f** [comment]
> Given multiple Task Errors, with each Task Error may contain multiple Errors, you have a list of a list and requires 2 loops.

I understand that, and this code does have two for loops to take care of that: https://github.com/openshift/cluster-monitoring-operator/pull/1558/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR1041-R1053

But has additionally a third loop which iterates the flattened list yet again: https://github.com/openshift/cluster-monitoring-operator/pull/1558/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR1071

Why not extract the the state error information in the first iteration when the generic error is converted to a StateError?

**Comment by sthaha** [comment]
> But has additionally a third loop which iterates the flattened list yet again: https://github.com/openshift/cluster-monitoring-operator/pull/1558/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR1071

> Why not extract the the state error information in the first iteration when the generic error is converted to a StateError?

Given the number of items in the list, I feel this is clearer to maintain than the version suggested. I can try a benchmark to validate that.

**Comment by jan--f** [comment]
> Given the number of items in the list, I feel this is clearer to maintain than the version suggested. This will be akin to using aggregate.Flatten in future which returns the hierarchy of errors as a single list of errors.

So far I'm trying to outline what I find confusing, with the assumption that future readers will also stumble over this. But its a fair point that that wasn't very constructive, so here is a suggestion.
I think one source of confusion for me is the re-use of `StateError` to contain information for multiple errors with the same state. By making this use case explicit the type itself (simply by change `Reason string` to `Reasons []string` and calling `strings.Join` at the time the report is prepared for humans) results in clearer code in my mind. 
 https://github.com/sthaha/cluster-monitoring-operator/compare/availability-degraded...jan--f:cluster-monitoring-operator:review-1558?expand=1
 
> I can try a benchmark to validate that.

As I attempt to reason above, this is not an efficiency issue per se. The version above is probably not more efficient then the original (maybe even much less efficient), but shorter and easier to follow (at least to me).

File: pkg/client/client.go (hunk 22 above)
Comment (sthaha) [reply pkg/client/client.go:919]: @jan--f Marking this as resolved. Please reopen if you want to discuss the comment above further. 

> This actually aligns with our earlier discussions that Prometheus becoming unavailable should set Degraded condition of CMO to true.

File: pkg/client/client.go (hunk 19 above)
Comment (sthaha) [reply pkg/client/client.go:927]: I have updated the `validationError` to return appropriate state errors, so returning `validationError` should suffice.

File: pkg/client/client.go (hunk 21 above)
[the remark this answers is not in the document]
Comment (sthaha) [reply pkg/client/client.go:902]: I have changed it to return `Available` is `Unknown`

File: pkg/client/client.go (hunk 19 above)
Comment (sthaha) [reply pkg/client/client.go:928]: Will add degraded as well to the list of errors if we timeout and change the decision based on feedback later.

File: pkg/client/status_reporter.go [hunk 24]
```diff
@@ -163,6 +188,33 @@ func (r *StatusReporter) SetRollOutInProgress(ctx context.Context) error {
...
+	reason = cmostr.ToPascalCase(available.Reason())
+	conditions.setCondition(v1.OperatorAvailable, v1.ConditionStatus(available.Status()),
+		available.Message(), reason, time)
+
+	return r.setConditions(ctx, co, conditions)
+}
+
 func (r *StatusReporter) SetFailed(ctx context.Context, statusErr error, reason string) error {
 	co, err := r.getOrCreateClusterOperator(ctx)
 	if err != nil {
```
sthaha, line 220:
```suggestion
```

pkg/client/client.go
sthaha: Added - reconciled condition check as well.

lihongyan1:
For senario in bug https://bugzilla.redhat.com/show_bug.cgi?id=2039411, test with pr failed, suppose Degraded is true, but it's false.
% oc -n openshift-monitoring get pod|grep prometheus-k8s
prometheus-k8s-0                                         6/6     Running   0          33m
prometheus-k8s-1                                         0/6     Pending   0          13m
% oc get co monitoring
NAME         VERSION                                                   AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
monitoring   4.12.0-0.ci.test-2022-08-25-065719-ci-ln-pjcbf32-latest   True        False         False      33m

lihongyan1:
Message when alert manager has wrong pvc is wrong, may the issue can be fixed in a new pr, then the pr can be approved and merged
% oc -n openshift-monitoring describe pod alertmanager-main-0 |tail -n 10
QoS Class:                   Burstable
Node-Selectors:              kubernetes.io/os=linux
Tolerations:                 node.kubernetes.io/memory-pressure:NoSchedule op=Exists
                             node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
  Type     Reason            Age   From               Message
  ----     ------            ----  ----               -------
  Warning  FailedScheduling  13m   default-scheduler  0/6 nodes are available: 6 pod has unbound immediate PersistentVolumeClaims. preemption: 0/6 nodes are available: 6 Preemption is not helpful for scheduling.
  Warning  FailedScheduling  13m   default-scheduler  0/6 nodes are available: 6 pod has unbound immediate PersistentVolumeClaims. preemption: 0/6 nodes are available: 6 Preemption is not helpful for scheduling.
hongyli@hongyli-mac Downloads % oc get co monitoring -o jsonpath='{.status.conditions}' | jq 'map(select(.type=="Degraded" or .type=="Available"))'
[
  {
    "lastTransitionTime": "2022-08-25T08:51:55Z",
    "reason": "AsExpected",
    "status": "True",
    "type": "Available"
  },
  {
    "lastTransitionTime": "2022-08-25T08:51:55Z",
    "message": "waiting for Alertmanager object changes failed: waiting for Alertmanager openshift-monitoring/main: expected 2 replicas, got 0 updated replicas",
    "reason": "UpdatingAlertmanagerFailed",
    "status": "True",
    "type": "Degraded"
  }
]

sthaha:
@lihongyan1 Could you please provide the `status` of the `Alertmanager` ? 
This PR will only look at the status reported by Prometheus Operator and not the pods. 

```
oc -n openshift-monitoring get alertmanager -ojsonpath=status
```

lihongyan1:
/label qe-approved
The fix makes co monitoring reflect status of prometheus pos/prometheus operator now, will file additional bug to track alert manager issue

lihongyan1 (commented):
/label qe-approved

pkg/client/client.go
```diff
@@ -861,6 +887,95 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
...
+	newStateError := NewDegradedError
+	if avail.Status == monv1.PrometheusConditionFalse {
+		// prometheus not available should result in a Degraded and Unavailable error
+		errs = []error{NewDegradedError(msg)}
+		newStateError = NewAvailabilityError
+	}
+
+	errs = append(errs, newStateError(msg))
+	return false, errs
+}
```
jan--f, line 948:
```suggestion
	msg := fmt.Sprintf("%s: %s", avail.Reason, avail.Message)

        errs := []error{NewDegradedError(msg)}
	if avail.Status == monv1.PrometheusConditionFalse {
		// prometheus not available should result in a Degraded and Unavailable error
		errs = append(errs, NewAvailabilityError(msg))
	}

	return false, errs
}
```

pkg/client/client.go
```diff
@@ -861,6 +887,95 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
...
+
+// ValidatePrometheus returns nil error if Prometheus is fully avaialble.
```
jan--f, line 950:
```suggestion
// ValidatePrometheus returns nil error if Prometheus is fully available.
```

pkg/client/client.go
```diff
@@ -861,6 +887,95 @@ func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) err
 	return nil
 }
 
+// validatePrometheusResource is a helper method for ValidatePrometheus.
+// NOTE: this function is refactored out of wait.Poll for testing
+func (c Client) validatePrometheusResource(ctx context.Context, prom types.NamespacedName) (bool, []error) {
+
+	p, err := c.mclient.MonitoringV1().Prometheuses(prom.Namespace).Get(ctx, prom.Name, metav1.GetOptions{})
+	// p, err := c.GetPrometheusByNsName(ctx, prom)
```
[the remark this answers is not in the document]
sthaha: 🤦 , thank you! I should have seen this

pkg/client/client.go
sthaha: Nice!

sthaha:
@jan--f Thank you for the review. I have addressed the comments 🤞

jan--f (approved):
Nice, thanks! This
/lgtm
Just adding a hold to maybe get @simonpasquier eyes on this as well.
/hold

test/e2e/config_test.go
```diff
@@ -59,9 +59,9 @@ func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
 	}
 	f.MustCreateOrUpdateConfigMap(t, cm)
 
-	t.Log("asserting that CMO goes degraded after an invalid configuration is pushed")
+	t.Log("asserting that CMO goes degraded and availability is unknown after an invalid configuration is pushed")
 	f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionTrue)(t)
-	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionFalse)(t)
+	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionUnknown)(t)
```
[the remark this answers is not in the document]
sthaha: IIRC, we had decided to set the Availability as `Unknown` when operator becomes degraded. Should we change `reportDegraded` to preserve the old `Availability` status ?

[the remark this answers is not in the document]
sthaha: I will change the behaviour to not modify the `Availability` status when operator reports degraded and resubmit

pkg/client/client.go
```diff
@@ -824,6 +828,28 @@ func (c *Client) DeleteSecret(ctx context.Context, s *v1.Secret) error {
 	return err
 }
 
+func (c *Client) GetPrometheusByNsName(ctx context.Context, prom types.NamespacedName) (*monv1.Prometheus, error) {
```
simonpasquier, line 831: (nit) it seems to be unused

simonpasquier (commented):
IIUC only when Prometheus gets unavailable, CMO will report available=false in the cluster operator conditions. It means that a fully unavailable Alertmanager would be turned into (available=true,degraded=true) which isn't great from a user perspective. I suggest that for all task expect those based on Prometheus status, CMO keeps reporting (available=false,degraded=true) if an error is returned. WDYT?

pkg/operator/operator.go
```diff
@@ -707,18 +704,20 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 	return nil
 }
 
-func (o *Operator) reportError(ctx context.Context, err error, failedTaskReason string) {
-	klog.Infof("ClusterOperator reconciliation failed (attempt %d), retrying. ", o.failedReconcileAttempts+1)
-	if o.failedReconcileAttempts >= 2 {
-		// Only update the ClusterOperator status after 3 retries have been attempted to avoid flapping status.
-		klog.Warningf("Updating ClusterOperator status to failed after %d attempts.", o.failedReconcileAttempts+1)
+func (o *Operator) reportFailed(ctx context.Context, report runReport) {
+	report.isDegraded()
```
simonpasquier, line 708: not needed?

simonpasquier (commented):
/lgtm

pkg/operator/operator.go
sthaha: You are right :). Happened during our pair programming session :D

jan--f:
/lgtm
Thanks for the hard work everyone!

sthaha:
Thanks a lot @simonpasquier & @jan--f 🙇 
 😌

PR #1560, opened 2022-02-15
Use service ca beta annotation

simonpasquier:


simonpasquier:
/label qe-approved
/label px-approved
/label docs-approved

This has no functional impact and it is safe to be merged given the green signal from the CI.

PR #1562, opened 2022-02-22
manifest: use path module to construct externalURL

jan--f:
path.Join will ensure the result is a proper path.
Follow up from review on 7d46ea0bf5b2ebd299aa100deee68c93fdd38ff9.


jan--f:
Adding the following labels since this is an internal cleanup and has no docs, user or QE impact.
/label docs-approved
/label qe-approved
/label px-approved

raptorsun (approved):

PR #1563, opened 2022-02-22
Bug 2057025: fix init-config-reloader resource requests

simonpasquier:


bison (approved):
/lgtm

PR #1564, opened 2022-02-23
Bug 2057403: jsonnet: Give CMO explicit get permissions for ReplicaSets

bison:
The cluster-monitoring-operator reads the owner reference for its pod,
and fetches that ReplicaSet as part of setting up event recording.
None of the mixed-in permissions end up allowing GET actions on these,
so the fetch fails and an error is reported in the logs.  This adds
the permission explicitly.

### DIFF
--- a/jsonnet/components/cluster-monitoring-operator.libsonnet
+++ b/jsonnet/components/cluster-monitoring-operator.libsonnet
@@ -121,6 +121,16 @@ function(params) {
       },
     },
     rules: [
+      // The permissions mixed-in in main.jsonnet don't seem to include GET
+      // access on these, but the operator needs them when fetching
+      // OwnerReferences.
+      //
+      // See: https://bugzilla.redhat.com/show_bug.cgi?id=2057403
+      {
+        apiGroups: ['apps'],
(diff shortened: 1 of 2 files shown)


bison:
Not seeing any more warnings with this:

```console
$ kubectl -n openshift-monitoring logs -l 'app=cluster-monitoring-operator' -c cluster-monitoring-operator  | grep -c '^W'
0
```

slashpai (approved):

simonpasquier:
/lgtm
I can also see that the events have the correct links now.

```
$ jq -cr '.items | map(select(.involvedObject.name == "cluster-monitoring-operator")) | map(.message) | join("\n")' < events.json 
Scaled up replica set cluster-monitoring-operator-6b4fcd45b8 to 1
No valid client certificate for OpenShiftMonitoringClientCertRequester is found: unable to parse certificate: data does not contain any valid RSA or ECDSA certificates
A csr "system:openshift:openshift-monitoring-5sg7d" is created for OpenShiftMonitoringClientCertRequester
A new client certificate for OpenShiftMonitoringClientCertRequester is available
```

PR #1573, opened 2022-03-01
add verticalpodautoscalers to kube-state-metrics

matthyx:

### DIFF
--- a/assets/kube-state-metrics/deployment.yaml
+++ b/assets/kube-state-metrics/deployment.yaml
@@ -50,6 +50,7 @@ spec:
           ^kube_pod_container_status_running$,
           ^kube_pod_completion_time$,
           ^kube_pod_status_scheduled$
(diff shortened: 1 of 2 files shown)


matthyx:
@fpetkovski I would like to add vpa metrics to OpenShift in order to use existing dashboards:

The documentation is here: https://github.com/kubernetes/kube-state-metrics/blob/master/docs/verticalpodautoscaler-metrics.md#examples

Do you see how this could integrate within the Operator? Should we detect [an existing VPA](https://docs.openshift.com/container-platform/4.8/nodes/pods/nodes-pods-vertical-autoscaler.html) and configure `kube-state-metrics` accordingly?
Or maybe this should be an option in the [ConfigMap](https://docs.openshift.com/container-platform/4.8/monitoring/configuring-the-monitoring-stack.html#creating-cluster-monitoring-configmap_configuring-the-monitoring-stack)?

Thanks in advance for your guidance.

matthyx:
@PhilipGough can you have a look?

PhilipGough (commented):
Hey @matthyx - I understand where you are coming from with this PR, but in short, we generally don't accept patches like this in this project from external contributors. 

The process here generally requires prioritisation from a PM perspective etc so if this is a requirement for you, I suggest opening a support ticket which can become an RFE.

matthyx:
@PhilipGough it's https://issues.redhat.com/browse/RFE-2009

PhilipGough:
@matthyx - thanks for both the link and your willingness to help on the implementation.

It won't make 4.10 because it is already cut but let's see if we can get some eyes on the RFE for 4.11

cc @jan--f

matthyx:
Thanks 👍

PR #1574, opened 2022-03-02
Bug 2050120: Sanitize all regex allow/denylist used in KSM component

arajkumar:
Similar to https://github.com/prometheus-operator/kube-prometheus/pull/1614, but for allow/deny list passed from CMO to KSM.

### DIFF
--- a/assets/kube-state-metrics/deployment.yaml
+++ b/assets/kube-state-metrics/deployment.yaml
@@ -35,7 +35,10 @@ spec:
         - --port=8081
         - --telemetry-host=127.0.0.1
         - --telemetry-port=8082
-        - --metric-denylist=kube_secret_labels,kube_.*_annotations
+        - |
+          --metric-denylist=
+          ^kube_secret_labels$,
+          ^kube_.+_annotations$
         - --metric-labels-allowlist=pods=[*],nodes=[*],namespaces=[*],persistentvolumes=[*],persistentvolumeclaims=[*],poddisruptionbudgets=[*],poddisruptionbudget=[*]
         - |
           --metric-denylist=
--- a/jsonnet/components/kube-state-metrics.libsonnet
+++ b/jsonnet/components/kube-state-metrics.libsonnet
@@ -138,7 +138,11 @@ function(params)
                   else
                     c {
                       args+: [
-                        '--metric-denylist=kube_secret_labels,kube_.*_annotations',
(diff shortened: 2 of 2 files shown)


assets/kube-state-metrics/deployment.yaml
```diff
@@ -35,7 +35,10 @@ spec:
         - --port=8081
         - --telemetry-host=127.0.0.1
         - --telemetry-port=8082
-        - --metric-denylist=kube_secret_labels,kube_.*_annotations
+        - |
+          --metric-denylist=
+          ^kube_secret_labels$,
+          ^kube_.+_annotations$
```
sthaha, line 41: doesn't this create  `"--metric-denylist=\n^kube_secret_labels$,\n^kube_.+_annotations$\n"` ?

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -138,7 +138,11 @@ function(params)
                   else
                     c {
                       args+: [
-                        '--metric-denylist=kube_secret_labels,kube_.*_annotations',
+                        |||
+                          --metric-denylist=
+                          ^kube_secret_labels$,
+                          ^kube_.+_annotations$
+                        |||,
```
sthaha, line 145: This is verbatim, IIUC, it should insert \n which may not be what we want.

assets/kube-state-metrics/deployment.yaml
arajkumar: yes, it does but KSM seem to be interpreting correctly, The same pattern has been used in [upstream mixins too](https://github.com/prometheus-operator/kube-prometheus/blob/main/jsonnet/kube-prometheus/addons/ksm-lite.libsonnet#L18).

Also I had verified these metrics are not emitted by KSM.

sthaha: I see that is how it is applied upstream  but I wonder if `addArgs` fixes the `\n` .. 
Are you able to add a test to https://github.com/kubernetes/kube-state-metrics/blob/ef8bede4064f04472a8c2fe8c281e6bf97de4329/pkg/options/types_test.go#L165 to see if that really works? my experience says otherwise

sthaha (changes requested):

sthaha (approved):
lgtm

PR #1575, opened 2022-03-02
Bug 2060083: React to changes in clusteroperators

alvaroaleman:
There are already eventhandlers for the various clusteroperators
configured, but the actual handlefunc drops events from them.

I manually deployed a CMO with this and change the consoles .status.URL, which successfully triggered a reconcile:
```
I0302 16:39:30.018897       1 operator.go:507] Triggering an update due to cluster resource: *v1.Console
```

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -475,6 +475,12 @@ func (o *Operator) handleEvent(obj interface{}) {
 		return
 	}
 
+	if _, ok := obj.(*configv1.Console); ok {
+		klog.Info("Triggering update due to a console update")
+		o.enqueue(cmoConfigMap)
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
Another option would be to catch the object earlier in the function like we do already for Infrastructure, APIServer and PersistentVolumeClaim. It would be more consistent with what already exists though the logic could definitely be improved.

https://github.com/openshift/cluster-monitoring-operator/blob/04992cc379d313fa4077d1b873386615d9d54322/pkg/operator/operator.go#L460-L477

alvaroaleman:
@simonpasquier updated to handle it more explicitly as suggested

PR #1576, opened 2022-03-02
Bug 2060091: Properly deal with an empty console URL

alvaroaleman:
Currently, when the console object exist but has an empty url (happens
e.G. during cluster creation), an invalid alertmanager statefulset is
produced that has an argument of `--web.external-url=monitoring`.

This is made worse by the fact that we wait up to five minutes for the
alertmanager pods to become ready, which will never happen with this config,
resulting in a lot of startup delay if the CMO happens to be quicker
than the console operator.

This also fixes this for Prometheus, where it isn't strictly required as
prometheus seems to accept `monitoring` as a valid `external-url`. Odds
are this validation will be added there too, so fix that while we are at
it.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -448,7 +448,7 @@ func (f *Factory) AlertmanagerMain(trustedCABundleCM *v1.ConfigMap) (*monv1.Aler
 
 	a.Spec.Image = &f.config.Images.Alertmanager
 
-	if f.consoleConfig != nil {
+	if f.consoleConfig != nil && f.consoleConfig.Status.ConsoleURL != "" {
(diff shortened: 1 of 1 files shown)


jan--f:
/lgtm
@alvaroaleman I assume we should backport this to 4.10, correct?

alvaroaleman:
> @alvaroaleman I assume we should backport this to 4.10, correct?

@jan--f Yes, that would be awesome! When I go to https://bugs.ci.openshift.org/clones?ID=2060091 the only target version I can select for cloning is 4.10.0, it needs to be for 4.10.z or not?

jan--f:
/cherry-pick release-4.10

@alvaroaleman yes you are correct, the 4.10 z-stream was opened yesterday just after your attempt.

PR #1578, opened 2022-03-03
MON-2245: manifest: Add cluster_id label to remote_write configurations

jan--f:
Problem: In many cases cluster admins configure
remote_write to target a centralized system that is used by multiple
clusters. When this centralized metric store is queried, cluster
specific queries need to distinguish metrics by a label. Cluster admins
can already add such a label in their remote_write configuration. This
however scales poorly and requires attention for every cluster.

Solution: Each OpenShift cluster has a unique cluster ID. We already use
this in our telemetry service to query for individual cluster metrics.
The same ID should be available for cluster admins to use in their
custom remote_write configurations. This commits adds a label
`__tmp_openshift_cluster_id__` to every user supplied remote_write
configuration as the first relabel_config entry. Cluster admins can then
use this metric as input to their own label name. If they don't the
label is automatically dropped (as denoted by the `__` prefix), which
keeps this backwards compatible.

Issues: https://issues.redhat.com/browse/MON-2245

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1458,6 +1458,7 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 	}
 
 	telemetryEnabled := f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.IsEnabled()
+	clusterID := f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.ClusterID
 	if telemetryEnabled && f.config.RemoteWrite {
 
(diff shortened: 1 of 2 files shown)


simonpasquier:
lgtm but I think that it deserves an e2e test, WDYT?
Upstream we've talked about using Prometheus as a remote write receiver (https://github.com/prometheus-operator/prometheus-operator/pull/4337) which enables more thorough checks (e.g. verifying that the received metrics are as expected).

jan--f:
> lgtm but I think that it deserves an e2e test, WDYT?

I think that is a fantastic idea. We wanted to change our existing remote_write e2e test anyway. Any objections to doing that in a separate PR?

simonpasquier:
> Any objections to doing that in a separate PR?

not at all

/lgtm

jan--f:
/label docs-approved
Adding docs label as per agreemtent. I'll provide doc input via https://issues.redhat.com/browse/MON-2246
Ping @juzhao for qe approval.

PR #1579, opened 2022-03-03
MON-2193: pkg/manifests: Expose retention size settings for Platform Prometheus

slashpai:
[MON-2216](https://issues.redhat.com/browse/MON-2216): Expose retention size settings for Platform Prometheus in the CMO configuration

Introduce a new "retentionSize" field under the prometheusK8s key
When neither retention nor retentionSize are defined, a default of 15d for time retention applies.
When "retentionSize" is defined and "retention" isn't, only size retention applies.
When "retentionSize" isn't defined and "retention" is, only time retention applies
When both time and size retention fields are defined, both apply.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1442,6 +1442,10 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 		p.Spec.Retention = f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.Retention
 	}
 
+	if f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RetentionSize != "" {
+		p.Spec.RetentionSize = f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RetentionSize
+	}
+
 	p.Spec.Image = &f.config.Images.Prometheus
 
 	if f.consoleConfig != nil && f.consoleConfig.Status.ConsoleURL != "" {
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -1676,6 +1676,74 @@ func TestPrometheusQueryLogFileConfig(t *testing.T) {
 	}
 }
 
+func TestPrometheusRetentionConfigs(t *testing.T) {
+	for _, tc := range []struct {
+		name                  string
+		retention             string
+		retentionSize         string
+		expectedRetention     string
+		expectedRetentionSize string
+	}{
+		{
+			name:                  "both retention and retentionSize defined",
+			retention:             "30d",
+			retentionSize:         "15GiB",
+			expectedRetention:     "30d",
+			expectedRetentionSize: "15GiB",
+		},
+		{
+			name:                  "only retention defined",
+			retention:             "45d",
+			expectedRetention:     "45d",
+			expectedRetentionSize: "",
+		},
+		{
+			name:                  "only retentionSize defined",
+			retentionSize:         "25GB",
+			expectedRetention:     "",
+			expectedRetentionSize: "25GB",
+		},
+		{
+			name:                  "both retention and retentionSize empty",
+			expectedRetention:     "15d",
+			expectedRetentionSize: "",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			c := NewDefaultConfig()
+			c.ClusterMonitoringConfiguration.PrometheusK8sConfig.Retention = tc.retention
+			c.ClusterMonitoringConfiguration.PrometheusK8sConfig.RetentionSize = tc.retentionSize
+
+			f := NewFactory("openshift-monitoring", "openshift-user-workload-monitoring", c, defaultInfrastructureReader(), &fakeProxyReader{}, NewAssets(assetsPath), &APIServerConfig{}, &configv1.Console{})
+
+			p, err := f.PrometheusK8s(
+				&v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+				&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+			)
+
+			if err != nil {
+				t.Fatalf("Unexpected error occured %v", err)
+				return
+			}
+
+			if tc.retention == "" && tc.retentionSize == "" {
+				if p.Spec.Retention != "15d" && p.Spec.RetentionSize != "" {
+					t.Fatal("Default Retention is not configured correctly")
+				}
+				return
+			}
+
+			if p.Spec.Retention != tc.expectedRetention {
+				t.Fatal("Retention is not configured correctly")
+			}
+
+			if p.Spec.RetentionSize != tc.expectedRetentionSize {
+				t.Fatal("RetentionSize is not configured correctly")
+			}
+		})
+	}
+}
+
 func TestPrometheusK8sAdditionalAlertManagerConfigsSecret(t *testing.T) {
 	testCases := []struct {
 		name           string
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -118,6 +118,7 @@ func TestClusterMonitorPrometheusK8Config(t *testing.T) {
 	data := fmt.Sprintf(`prometheusK8s:
   logLevel: debug
   retention: 10h
+  retentionSize: 15GB
   queryLogFile: /tmp/test.log
   tolerations:
     - operator: "Exists"
@@ -156,6 +157,7 @@ func TestClusterMonitorPrometheusK8Config(t *testing.T) {
 					expectMatchingRequests(podName, containerName, mem, cpu),
 					expectContainerArg("--log.level=debug", containerName),
 					expectContainerArg("--storage.tsdb.retention.time=10h", containerName),
+					expectContainerArg("--storage.tsdb.retention.size=15GB", containerName),
 				},
 			),
 		},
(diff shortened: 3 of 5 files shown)


pkg/manifests/manifests.go
```diff
@@ -1425,6 +1429,10 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 		p.Spec.Retention = f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.Retention
 	}
 
+	if !zeroRetentionSizeConfig(f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RetentionSize) {
```
PhilipGough, line 1432: I wonder should the previous check `if f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.Retention != ""` also check for zero values? 

If yes then perhaps we can rename `zeroRetentionSizeConfig` to `isUnsetRetentionConfig` or something similar

test/e2e/config_test.go
```diff
@@ -151,6 +151,7 @@ func TestClusterMonitorPrometheusK8Config(t *testing.T) {
 	data := fmt.Sprintf(`prometheusK8s:
   logLevel: debug
   retention: 10h
+	retention: 15G
```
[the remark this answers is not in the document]
slashpai: ya mistyped

pkg/manifests/manifests.go
slashpai: Trying to cover the scenario If the retention value is "", 0 or "0", it means that the value isn't defined, so `isUnsetRetentionConfig` name makes sense

slashpai:
@PhilipGough @simonpasquier Please review when you get a chance

PhilipGough (approved):
/lgtm

slashpai:
cc: @openshift/team-monitoring-qe-approvers

pkg/manifests/manifests_test.go
```diff
@@ -1034,6 +1034,7 @@ func TestPrometheusK8sRemoteWrite(t *testing.T) {
 func TestPrometheusK8sConfiguration(t *testing.T) {
 	c, err := NewConfigFromString(`prometheusK8s:
   retention: 25h
+  retentionSize: 10Gi
```
simonpasquier, line 1037: `10Gi` isn't a valid value.

```suggestion
  retentionSize: 10GB
```

pkg/manifests/manifests.go
```diff
@@ -1425,6 +1429,10 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 		p.Spec.Retention = f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.Retention
 	}
 
+	if !isUnsetRetentionConfig(f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RetentionSize) {
```
simonpasquier, line 1432: I feel that `if !isUnsetRetentionConfig(...)` can be confusing to read with the double negation ("unless retention is not set"). I'd be more inclined to remove the intermediate function.

```suggestion
	if f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RetentionSize != "" && f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RetentionSize != "0" {
```

simonpasquier (commented):
Eventually CMO should run more validations on the user inputs to avoid pushing bad configuration but it's something that we should address globally so I'm mostly fine with this PR.

slashpai:
> Eventually CMO should run more validations on the user inputs to avoid pushing bad configuration but it's something that we should address globally so I'm mostly fine with this PR.

Once we add validating webhook in Prometheus operator for the fields it should be enough? Or should we add validations for user inputs on top of it?

pkg/manifests/manifests_test.go
slashpai: sure I updated it to 10GiB I missed it earlier

lihongyan1:
After rentionSize unit issue is fixed: Still Two things are not as expected.
Detail see bug https://bugzilla.redhat.com/show_bug.cgi?id=2062510

1. When set retentionSize only, default retention time applies. Expected only retentionSize applies.

2. When retentionSize is set 0 or "0", which means the value isn't defined
   When rentention Time is set 0 or "0", 0 will be the valid value applied, suppose this is not correct

slashpai:
> After rentionSize unit issue is fixed: Still Two things are not as expected. Detail see bug https://bugzilla.redhat.com/show_bug.cgi?id=2062510
> 
> 1. When set retentionSize only, default retention time applies. Expected only retentionSize applies.
> 2. When retentionSize is set 0 or "0", which means the value isn't defined
>    When rentention Time is set 0 or "0", 0 will be the valid value applied, suppose this is not correct

1st issue will be fixed after https://github.com/openshift/prometheus-operator/pull/162 merged

2nd issue havn't touched retention part on this so whatever value set for retention will be used even if it is `0`, I think we should add condition similar as retentionSize for retention too

@simonpasquier wdyt?

simonpasquier:
> When retentionSize is set 0 or "0", which means the value isn't defined
When rentention Time is set 0 or "0", 0 will be the valid value applied, suppose this is not correct

This is a very specific edge case :) but Prometheus would handle it by defaulting to a retention period of 15 days.

https://github.com/prometheus/prometheus/blob/1635deed7637bc5b3616c66f4de9c2add5973eaa/cmd/prometheus/main.go#L477-L480

I'm not sure we should be doing something to prevent this configuration in CMO.  As @slashpai said, I believe that it is the same behavior in all previous OCP versions.

jan--f:
/retitle MON-2193: pkg/manifests: Expose retention size settings for Platform Prometheus

slashpai:
@lihongyan1 can you test this again now?

pkg/manifests/manifests.go
```diff
@@ -1430,6 +1430,10 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 		p.Spec.Retention = f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.Retention
 	}
 
+	if f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RetentionSize != "" && f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RetentionSize != "0" {
```
simonpasquier, line 1433: Going back to this I'm not sure that we need to treat "0" in a special way.

simonpasquier (commented):
`Config.applyDefaults()` would need an update too: DefaultRetentionValue should only be applied when both time and size retentions are empty.

/hold

pkg/manifests/manifests_test.go
```diff
@@ -1230,6 +1230,7 @@ func TestPrometheusK8sRemoteWriteURLs(t *testing.T) {
 func TestPrometheusK8sConfiguration(t *testing.T) {
```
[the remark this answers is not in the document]
slashpai: added

slashpai:
@simonpasquier I updated code and added new unit test as well

@lihongyan1 could you please test again? :)

pkg/manifests/manifests.go
slashpai: I removed the check for 0

slashpai:
@simonpasquier Can you review again? :)

lihongyan1:
Filed a bug https://bugzilla.redhat.com/show_bug.cgi?id=2073972

slashpai:
> Filed a bug https://bugzilla.redhat.com/show_bug.cgi?id=2073972

We had added openapi validation at CRD level to retentionSize [pkg/apis/monitoring/v1: Add validations for size settings at CRD level](https://github.com/prometheus-operator/prometheus-operator/pull/4661) in prometheus-operator

That change is synced to cmo as well as part of [kube-prometheus update](https://github.com/openshift/cluster-monitoring-operator/pull/1615/files).  So RetentionSize now is validated by the API server. If the value is invalid CMO throws an error when applying the Prometheus spec, since its rejected prometheus-operator never even gets to see the value hence no logging is done there.

Once https://github.com/prometheus-operator/prometheus-operator/pull/4684 is merged you will see same effect for retention as well

cc: @simonpasquier

pkg/manifests/manifests_test.go
```diff
@@ -1557,6 +1557,74 @@ func TestPrometheusQueryLogFileConfig(t *testing.T) {
...
+			c.ClusterMonitoringConfiguration.PrometheusK8sConfig.Retention = tc.retention
+			c.ClusterMonitoringConfiguration.PrometheusK8sConfig.RetentionSize = tc.retentionSize
+
+			f := NewFactory("openshift-monitoring", "openshift-user-workload-monitoring", c, defaultInfrastructureReader(), &fakeProxyReader{}, NewAssets(assetsPath), &APIServerConfig{}, &configv1.Console{})
+
+			p, err := f.PrometheusK8s(
+				&v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+				&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+			)
+
```
JoaoBraveCoding, line 1604: nit, wdyt about moving the `err != nil`  from lines `1620-1623` to here? Makes the test easier to read no?

JoaoBraveCoding (commented):
Apart from small comment it looks good 👍

pkg/manifests/manifests_test.go
slashpai: ya definitely :)

I have updated the code

slashpai:
> Apart from small comment it looks good 👍

Thank you :), I have updated the test

sthaha (commented):
/lgtm 

I think string comparison `== ""` or " != ""` should make use of Trim(). Applicable also when you set it. 
```go
RetentionSize = strings.Trim(config.RetentionSize)
```

slashpai:
> /lgtm
> 
> I think string comparison `== ""` or " != ""` should make use of Trim(). Applicable also when you set it.
> 
> ```go
> RetentionSize = strings.Trim(config.RetentionSize)
> ```

Good suggestion, I will update :)

slashpai:
> > /lgtm
> > I think string comparison `== ""` or " != ""` should make use of Trim(). Applicable also when you set it.
> > ```go
> > RetentionSize = strings.Trim(config.RetentionSize)
> > ```
> 
> Good suggestion, I will update :)

Also we have other fields as well which will be good to use trim, can do that in separate PR?

PR #1582, opened 2022-03-04
Bug 2060756: Properly deal with an empty console URL

jan--f:
Currently, when the console object exist but has an empty url (happens
e.G. during cluster creation), an invalid alertmanager statefulset is
produced that has an argument of `--web.external-url=monitoring`.

This is made worse by the fact that we wait up to five minutes for the
alertmanager pods to become ready, which will never happen with this config,
resulting in a lot of startup delay if the CMO happens to be quicker
than the console operator.

This also fixes this for Prometheus, where it isn't strictly required as
prometheus seems to accept `monitoring` as a valid `external-url`. Odds
are this validation will be added there too, so fix that while we are at
it.

(cherry picked from commit 2d53fe15a5e7073d8796fccb7934aae0293ee94d)

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"net"
 	"net/url"
+	"path"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


juzhao:
tested with openshift/cluster-monitoring-operator#1585,openshift/cluster-monitoring-operator#1582 together, CMO reacts to changes in clusteroperator console, and pods are Running, no error
/label cherry-pick-approved

PR #1587, opened 2022-03-11
Bug 2063047: Added support for full-path query log file

JoaoBraveCoding:


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Note: This CHANGELOG is only for the monitoring team to track all monitoring related changes. Please see OpenShift release notes for official changes.
 
+## 4.11
+- [#1587](https://github.com/openshift/cluster-monitoring-operator/pull/1589) Added support for full-path query log file
```
simonpasquier, line 4: We can omit the changelog entry because the Bugzilla doc text will take care of it.

JoaoBraveCoding: Done ✅

pkg/manifests/manifests.go
```diff
@@ -1659,6 +1662,30 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 	return p, nil
 }
 
+func (f *Factory) queryLogFileConfig(p *monv1.Prometheus, queryLogFile string) (string, error) {
+	if strings.HasPrefix(queryLogFile, "/") && !(strings.HasPrefix(queryLogFile, "/dev") || strings.HasPrefix(queryLogFile, "/prometheus")) {
+		if strings.Count(queryLogFile, "/") == 1 {
+			return "", errors.Wrap(ErrConfigValidation, "attempting to mount query log file on root \"/\"")
+		}
```
arajkumar, line 1669: nit:

Refactor a bit for readability :)

```suggestion
                dirPath := filepath.Dir(queryLogFile)
		if dirPath == "/" {
		  return "", errors.Wrap(ErrConfigValidation, `attempting to mount query log file on root "/"`)
		}
		if dirPath == "/prometheus" || dirPath = "/dev"{
		  return queryLogFile, nil
		}
		... volume mount ..
```

pkg/manifests/manifests.go
```diff
@@ -1659,6 +1663,33 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 	return p, nil
 }
 
+func (f *Factory) queryLogFileConfig(p *monv1.Prometheus, queryLogFile string) (string, error) {
```
simonpasquier, line 1666: The function can take care of setting `p.spec.queryLogFile` directly since it already modifies the p argument.

```suggestion
func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) (error) {
```

in which case the caller side would look like:

```
if err != f.setupQueryLogFile(p) {
    return nil, err
}
```

and `setupQueryLogFile()` can exit early if queryLogFile is empty.

pkg/manifests/manifests.go
```diff
@@ -1659,6 +1663,33 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
...
+	dirPath := filepath.Dir(queryLogFile)
+	if dirPath == "/" {
+		return "", errors.Wrap(ErrConfigValidation, `attempting to mount query log file on root "/"`)
```
simonpasquier, line 1669:
```suggestion
		return "", errors.Wrap(ErrConfigValidation, `query log file can't be stored on the root directory`)
```

pkg/manifests/manifests_test.go
```diff
@@ -1194,6 +1194,74 @@ ingress:
 	}
 }
 
+func TestPrometheusQueryLogFileConfig(t *testing.T) {
+	testCases := []struct {
+		name             string
+		queryLogFilePath string
+		expected         string
+		errString        string
```
simonpasquier, line 1202: It's less brittle to check the error presence rather than matching the error message.

```suggestion
		errExpected        bool
```

pkg/manifests/manifests_test.go
```diff
@@ -1194,6 +1194,74 @@ ingress:
 	}
 }
 
+func TestPrometheusQueryLogFileConfig(t *testing.T) {
```
simonpasquier, line 1197: the test can also check the presence/absence of the volume + volume mount.

pkg/manifests/manifests_test.go
```diff
@@ -1194,6 +1194,74 @@ ingress:
...
+	}{
+		{
+			name:             "basic config",
+			queryLogFilePath: "/tmp/query.log",
+			expected:         "/tmp/query.log",
+			errString:        "",
+		},
+		{
+			name:             "basic config",
+			queryLogFilePath: "/dev/query.log",
```
simonpasquier, line 1212: I'm not sure that Prometheus would start successfully in this case :thinking:

pkg/manifests/manifests_test.go
```diff
@@ -1194,6 +1194,74 @@ ingress:
...
+		{
+			name:             "basic config",
+			queryLogFilePath: "/query.log",
+			expected:         "",
+			errString:        "attempting to mount query log file on root",
+		},
+	}
+	for _, tt := range testCases {
+		// We need to initialize the volumeClaimTemplate otherwise we will
+		// get a SIGSEGV when creating the EmptyDir volume
```
simonpasquier, line 1231: can you elaborate on this issue?

pkg/manifests/manifests_test.go
```diff
@@ -1194,6 +1194,74 @@ ingress:
...
+			errString:        "",
+		},
+		{
+			name:             "basic config",
+			queryLogFilePath: "/dev/query.log",
+			expected:         "/dev/query.log",
+			errString:        "",
+		},
+		{
+			name:             "basic config",
```
simonpasquier, line 1217:
```suggestion
			name:             "query log file on the TSDB storage volume",
```

pkg/manifests/manifests_test.go
```diff
@@ -1194,6 +1194,74 @@ ingress:
...
+			queryLogFilePath: "/prometheus/query.log",
+			expected:         "/prometheus/query.log",
+			errString:        "",
+		},
+		{
+			name:             "basic config",
```
simonpasquier, line 1223:
```suggestion
			name:             "invalid path",
```

pkg/manifests/manifests_test.go
JoaoBraveCoding: Yes, so if I initialize the config with just `c := NewDefaultConfig()` then once the test run I get the error:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xcf96f5]
```
The error happens on `manifests.go:1682` the line were we see
```
			VolumeSource: v1.VolumeSource{
				EmptyDir: p.Spec.Storage.EmptyDir,
			},
```

pkg/manifests/manifests.go
JoaoBraveCoding: Implemented suggestion

pkg/manifests/manifests_test.go
JoaoBraveCoding: Added tests to volumes

pkg/manifests/manifests_test.go
JoaoBraveCoding: Very true, implemented suggestion

pkg/manifests/manifests_test.go
JoaoBraveCoding: I'm also not sure, I'm spinning up a cluster-bot to test. If it's doesn't we can add that to the tests and allow "/dev/stdout" but forbid anything other than know outputs: /dev/null /dev/stdout

pkg/manifests/manifests.go
```diff
@@ -1659,6 +1662,36 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
...
+	if dirPath == "/prometheus" || dirPath == "/dev" {
+		return nil
+	}
+
+	p.Spec.Volumes = append(
+		p.Spec.Volumes,
+		v1.Volume{
+			Name: "query-log-file",
+			VolumeSource: v1.VolumeSource{
+				EmptyDir: p.Spec.Storage.EmptyDir,
```
simonpasquier, line 1682: we want a default emptyDir here:

```suggestion
				EmptyDir: &v1.EmptyDirVolumeSource{},
```

pkg/manifests/manifests_test.go
simonpasquier: ack, I've added a comment about emptyDir above that should make it possible to remove the code here.

pkg/manifests/manifests.go
```diff
@@ -1659,6 +1662,43 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 	return p, nil
 }
 
+func (f *Factory) setupQueryLogFile(p *monv1.Prometheus) error {
+	queryLogFile := f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.QueryLogFile
```
simonpasquier, line 1666: we can handle the empty value in the function directly

```suggestion
func (f *Factory) setupQueryLogFile(p *monv1.Prometheus) error {
	queryLogFile := f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.QueryLogFile
	if queryLogFile == "" {
		return nil
	}
```

pkg/manifests/manifests_test.go
```diff
@@ -1194,6 +1195,89 @@ ingress:
...
+			name:             "invalid file under dev",
+			queryLogFilePath: "/dev/query.log",
+			expected:         "",
+			errExpected:      true,
+		},
+	}
+	for _, tt := range testCases {
+		// We need to initialize the volumeClaimTemplate otherwise we will
+		// get a SIGSEGV when creating the EmptyDir volume
+		c, err := NewConfigFromString(`prometheusK8s:
```
simonpasquier, line 1239: this can be removed now?

pkg/manifests/manifests_test.go
```diff
@@ -1194,6 +1195,89 @@ ingress:
...
+			&v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+			&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+		)
+		if err != nil && !tt.errExpected {
+			t.Fatal(err)
+		}
+		if err == nil && tt.errExpected {
+			t.Fatal("Expected query log file ", tt.queryLogFilePath, " to give an error, but err is nil")
+		}
+		if !tt.errExpected {
```
simonpasquier, line 1262: A bit easier to read IMHO (handling error case first and return early if possible)

```suggestion
		if err != nil {
			if !tt.errExpected {
				t.Fatalf("expecting no error but got: %v", err)
			}
			return
		}
		if tt.errExpected {
			t.Fatal("Expected query log file ", tt.queryLogFilePath, " to return an error but got none")
		}
```

pkg/manifests/manifests_test.go
```diff
@@ -1194,6 +1195,89 @@ ingress:
...
+			if p.Spec.QueryLogFile != tt.expected {
+				t.Fatal("Prometheus query log is not configured correctly")
+			}
+
+			dirPath := filepath.Dir(p.Spec.QueryLogFile)
+			if dirPath != "/prometheus" && dirPath != "/dev" {
```
simonpasquier, line 1268: It would be better to encode the check in the test case struct (e.g. `if tc.expectVolume { ... }`

JoaoBraveCoding:
@simonpasquier yes indeed, I think the problem might have been fixed with my last commit https://github.com/openshift/cluster-monitoring-operator/pull/1587/commits/6983bfcc8c38885dd7d17f6f4e2386544da5fb85

JoaoBraveCoding:
not clear to me why e3e-agnostic-operator is failing... going to simply re-test /test e2e-agnostic-operator

pkg/manifests/manifests.go
```diff
@@ -1659,6 +1660,50 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
...
+			VolumeSource: v1.VolumeSource{
+				EmptyDir: &v1.EmptyDirVolumeSource{},
+			},
+		})
+
+	p.Spec.VolumeMounts = append(
+		p.Spec.VolumeMounts,
+		v1.VolumeMount{
+			Name:      "query-log-file",
+			MountPath: queryLogFile,
```
simonpasquier, line 1702:
```suggestion
			MountPath: dirPath,
```

pkg/manifests/manifests.go
```diff
@@ -1659,6 +1660,50 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
...
+		if base != "stdout" && base != "stderr" && base != "null" {
+			return errors.Wrap(ErrConfigValidation, `query log file can't be stored on a new file on the dev directory`)
+		}
+		return nil
+	}
+
+	p.Spec.Volumes = append(
+		p.Spec.Volumes,
+		v1.Volume{
+			Name: "query-log-file",
```
simonpasquier, line 1692:
```suggestion
			Name: "query-log",
```

pkg/manifests/manifests.go
```diff
@@ -1659,6 +1660,50 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
...
+			VolumeSource: v1.VolumeSource{
+				EmptyDir: &v1.EmptyDirVolumeSource{},
+			},
+		})
+
+	p.Spec.VolumeMounts = append(
+		p.Spec.VolumeMounts,
+		v1.VolumeMount{
+			Name:      "query-log-file",
```
simonpasquier, line 1701:
```suggestion
			Name:      "query-log",
```

pkg/manifests/manifests.go
```diff
@@ -1659,6 +1660,50 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
...
+	if queryLogFile == "" {
+		return nil
+	}
+	dirPath := filepath.Dir(queryLogFile)
+	if dirPath == "/" {
+		return errors.Wrap(ErrConfigValidation, `query log file can't be stored on the root directory`)
+	}
+
+	// /prometheus it is not necesssary to mount a volume since this is
+	// the location where prometheus will store TSDB
```
simonpasquier, line 1673:
```suggestion
	// /prometheus is where Prometheus will store the TSDB so it is 
	// already mounted inside the pod (either from a persistent volume claim or from an empty dir).
```

pkg/manifests/manifests.go
```diff
@@ -1659,6 +1660,50 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
...
+	p.Spec.QueryLogFile = queryLogFile
+	if dirPath == "/prometheus" {
+		return nil
+	}
+
+	// /dev it is not necesssary to mount a volume if the user configured
+	// the query log file to be one of the preexisting linux output streams
```
simonpasquier, line 1680:
```suggestion
	// It is not necesssary to mount a volume if the user configured
	// the query log file to be one of the preexisting linux output streams.
```

simonpasquier (commented):
minor comments only, you can wait for the CI to pass before fixing them :)

pkg/manifests/manifests_test.go
```diff
@@ -1194,6 +1194,85 @@ ingress:
...
+		},
+		{
+			name:             "query log file on the TSDB storage volume",
+			queryLogFilePath: "/prometheus/query.log",
+			expected:         "/prometheus/query.log",
+			errExpected:      false,
+			volumeExpected:   false,
+		},
+		{
+			name:             "query log file on the TSDB storage volume",
```
simonpasquier, line 1220:
```suggestion
			name:             "log to stdout",
```

PR #1588, opened 2022-03-14
WIP: test/e2e: Update query log path

slashpai:
prometheus-operator now enforces a root read-only filesystem for
prometheus container. Hence you either can use queryLog without
specifying a full file path as prometheus-operator will mount the file
into an emptyDir volume at `/var/log/prometheus` or If a full path is
provided, e.g. /tmp/query.log, you must mount a volume in the specified
directory and it must be writable.

This is a work in progress PR

We can upgrade [prometheus-operator to 0.55](https://github.com/openshift/prometheus-operator/pull/162) once we update CMO in respective places where we use query log

**TODO:**

- Update documentation for queryLog usage

cc: @simonpasquier 

### DIFF
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -528,7 +528,7 @@ func TestUserWorkloadMonitorPrometheusK8Config(t *testing.T) {
   enforcedTargetLimit: 10
   logLevel: debug
   retention: 10h
-  queryLogFile: /tmp/test.log
+  queryLogFile: test.log
   tolerations:
     - operator: "Exists"
   externalLabels:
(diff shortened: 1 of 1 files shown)


arajkumar:
@slashpai I didn't your PR :'( I too created a PR to fix the same problem with different approach, not sure which one is better :)

https://github.com/openshift/cluster-monitoring-operator/pull/1589

arajkumar:
Just found @JoaoBraveCoding's PR on the same topic. https://github.com/openshift/cluster-monitoring-operator/pull/1587

slashpai:
> @slashpai I didn't your PR :'( I too created a PR to fix the same problem with different approach, not sure which one is better :)
> 
> #1589

No problem :). I was just looking to upgrade prometheus-operator soon to proceed with retentionSize https://github.com/openshift/cluster-monitoring-operator/pull/1579

We would require to update docs as well since this is a breaking change

I will wait for @simonpasquier to review :)

slashpai:
I will close this PR since this is being addressed in https://github.com/openshift/cluster-monitoring-operator/pull/1587
/close

PR #1590, opened 2022-03-14
WIP: Add support for user-defined alerting rules

bison:
This adds a controller that watches a new `AlertingRule` type with the `monitoring.openshift.io/v1alpha1` GroupVersion. This allows for user-defined alerting rules in the `openshift-monitoring` namespace, while disallowing creation of recording rules.

Each `AlertingRule` instance results in the generation of a corresponding `PrometheusRule` in the `openshift-monitoring` namespace. Each individual rule is labeled with `openshift_io_user_alert=true` to identify it as being a user-defined platform alerting rule.

This is a partial proof-of-concept for the [User-Defined Alerting and Relabeling Rules](https://github.com/openshift/enhancements/pull/1030) enhancement proposal.


pkg/alert/rule_controller.go
```diff
@@ -0,0 +1,465 @@
...
+		cache.Indexers{},
+	)
+
+	// The namespace watcher keeps an up-to-date list of platform namespaces,
+	// i.e. namespaces labeled with `openshift.io/cluster-monitoring=true` so that
+	// the controller can ignore resources in non-platform namespaces.
+	//
+	// TODO(bison): Should we add event handlers to the namespace watcher? If the
+	// platform-monitoring label is removed from a namespace, but AlertingRule
+	// objects still exist, what is the correct action?
```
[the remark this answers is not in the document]
jan--f: :+1: for a follow-up. While this is an edge-case, a situation like that would be tricky to debug iiuc.

pkg/alert/rule_controller.go
```diff
@@ -0,0 +1,465 @@
...
+
+	rule.APIVersion = osmv1alpha1.GroupVersion.String()
+	rule.Kind = "AlertingRule"
+
+	// Generate the new or updated PrometheusRule object.
+	promRule := &monv1.PrometheusRule{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      rc.promRuleName(namespace, name),
+			Namespace: rc.client.Namespace(),
+			OwnerReferences: []metav1.OwnerReference{
```
simonpasquier, line 303: Now I remember that owner reference can't cross namespaces as explained in https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications
:thinking: not sure how to solve this

pkg/alert/rule_controller.go
```diff
@@ -0,0 +1,465 @@
...
+			return
+		}
+	}
+
+	rc.queue.Add(key)
+}
+
+// inPlatformNamespace returns true if the given object, which must satisfy the
+// metav1.Object interface, is in a platform monitoring namespace.
+func (rc *RuleController) inPlatformNamespace(obj interface{}) bool {
```
jan--f, line 171: Any reason not to use `metav1.Object` in the signature directly?

jan--f: Oh I see...scratch that :) Better check here then in all `handle*` functions.

bison:
> This looks very nice! This needs manifest files for the CRDs, some tests....something else? I'd love to give it a spin.

Thanks! Yeah, the CRDs are actually here: `vendor/github.com/openshift/api/monitoring/v1alpha1/*.crd.yaml` -- I guess I should put them in the `manifests` directory.

pkg/alert/rule_controller.go
```diff
@@ -0,0 +1,465 @@
...
+				Annotations: rule.Annotations,
+			}
+
+			if monv1Rule.Labels == nil {
+				monv1Rule.Labels = make(map[string]string)
+			}
+
+			// Set a static label indicating this rule comes from
+			// platform-monitoring, but is user defined.
+			monv1Rule.Labels[alertSourceLabel] = alertSourceValue
```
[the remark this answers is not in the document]
bison: I realized this would be overwritten by the default relabel rules anyway. I've changed it to `openshift_io_user_alert=true` for now, but of course I'm open to suggestions.

bison:
@simonpasquier, I actually think we can avoid the problems with the `OwnerReference` and the name collisions altogether by only allowing `AlertingRule` resources in the `openshift-monitoring` namespace. I've updated this to do that, and to drop the namespace watcher thing altogether.

The generated `PrometheusRule` resources are now named with the name of the owning `AlertingRule` object, plus part of a hash of the namespace and name of that owning object.

pkg/alert/rule_controller.go
bison: Responded [here](https://github.com/openshift/cluster-monitoring-operator/pull/1590#issuecomment-1073784960).

pkg/alert/rule_controller.go
```diff
@@ -0,0 +1,465 @@
...
+	rule.Status.ObservedGeneration = rule.Generation
+	rule.Status.PrometheusRule = fmt.Sprintf("%s/%s", promRule.Namespace, promRule.Name)
+
+	return rc.client.UpdateAlertingRuleStatus(ctx, rule)
+}
+
+// promRuleName returns the name of PrometheusRule to be generated for the given
+// AlertingRule namespace and name.
+func (rc *RuleController) promRuleName(namespace, name string) string {
+	return fmt.Sprintf("alerts-%s-%s", namespace, name)
```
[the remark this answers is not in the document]
bison: Responded [here](https://github.com/openshift/cluster-monitoring-operator/pull/1590#issuecomment-1073784960).

bison:
@jan--f, I added a commit copying the CRD to the `manifests` directory. There's also a very basic example resource in `examples/alerting-rule.yaml`, so it should be pretty easy to test this out either with cluster-bot or manually.

jan--f:
/close
Superseded by #1675

PR #1592, opened 2022-03-17
Makefile: add tools to path for run-local target

jan--f:
Problem: The run-local target requires gojsontoyaml, which is installed
to tmp/bin. Without gojsontoyaml installed or tmp/bin in the PATH, this
target fails.

Solution: add tmp/bin to PATH as part of the run-local target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ clean:
 
 .PHONY: run-local
 run-local: build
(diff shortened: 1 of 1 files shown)


jan--f:
/label docs-approved
/label qe-approved
/label px-approved
Setting approved labels, since this is a change purely related to a development environment.

PR #1595, opened 2022-03-18
Bug 2065076: manifests: advertise public urls without path component

jan--f:
Problem: When we removed some UIs (see
7d46ea0bf5b2ebd299aa100deee68c93fdd38ff9,
f85d2c15a8d686ab44de695d7617c1ad618ccd6d and
b8b6a503e82cf7e7b7c7f0f197184e047953de36), some Routes gained a path constraint.
We expose the Route URLs to the `monitoring-shared-config` configmap,
using `url.String()`. By using `url.String()` any path constraint on a
Route is added to the exposed URLs. For users who use those exposed URL
and add a path component, this causes issues.

Solution: Only expose <scheme>://<host> of the respective Route URLs.

Issues: https://bugzilla.redhat.com/show_bug.cgi?id=2065076

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1380,15 +1380,15 @@ func (f *Factory) SharingConfig(promHost, amHost, thanosHost *url.URL) *v1.Confi
 	// Configmap keys need to include "public" to indicate that they are public values.
 	// See https://bugzilla.redhat.com/show_bug.cgi?id=1807100.
 	if promHost != nil {
-		data["prometheusPublicURL"] = promHost.String()
+		data["prometheusPublicURL"] = fmt.Sprintf("%s://%s", promHost.Scheme, promHost.Host)
 	}
 
 	if amHost != nil {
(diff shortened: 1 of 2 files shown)


jan--f:
/bugzilla refresh
cc @kyoto @spadgett A 4.10 backport is needed, right?

simonpasquier (commented):
it might be worth updating `TestSharingConfig()` with a new test case?

JoaoBraveCoding:
Looks to me, agree that an improvement to `TestSharingConfig()` would be nice!

jan--f:
> it might be worth updating `TestSharingConfig()` with a new test case?

Good point, thanks! Updated.

simonpasquier (commented):
/lgtm

kyoto:
> cc @kyoto @spadgett A 4.10 backport is needed, right?

I think so. The issue only impacts the dev environment, I suppose it's not technically needed, but not backporting would inhibit our ability to make any monitoring UI fixes for 4.10.

PR #1596, opened 2022-03-18
WIP: Add support for user-defined alert relabel configs

bison:
This adds a controller that watches a new `AlertRelabelConfig` type with the `monitoring.openshift.io/v1alpha1` GroupVersion. This allows users to create alert relabel rules in a way that can be combined with our default config.

On any change to an `AlertRelabelConfig` object in the `openshift-monitoring` namespace, all `AlertRelabelConfig` objects are read from the cache and combined in lexicographical order to generate an `alert-relabel-configs` secret, which the platform Prometheus instance is configured to read for additional alert relabel configs.

This is a partial proof-of-concept for the [User-Defined Alerting and Relabeling Rules](https://github.com/openshift/enhancements/pull/1030) enhancement proposal.


jsonnet/components/prometheus.libsonnet
```diff
@@ -215,26 +215,24 @@ function(params)
...
-      apiVersion: 'v1',
-      kind: 'Secret',
+    // The default AlertRelabelConfig allows us to identify alerts that are coming
+    // from platform monitoring.
+    defaultAlertRelabelConfigs: {
+      apiVersion: 'monitoring.openshift.io/v1alpha1',
+      kind: 'AlertRelabelConfig',
       metadata: {
-        name: 'alert-relabel-configs',
+        name: 'zz99-default',
```
simonpasquier, line 224: I'm assuming that you've chose the `zz99-` prefix to ensure that it would be appended last in the final config. What if the relabel controller had this rule hardcoded. As far as I can tell, we went with a static YAML manifest because it needed no customization which isn't the case anymore with your PR.

bison: We can definitely do that. I actually wasn't sure if it should come first or last though.

simonpasquier: I'd say last since we want to enforce this label. Now it's also dependent on the discussion regarding `openshift_io_alert_source` :)

bison: True. Also, I guess we may have to use a different label, not just a different value anyway, because this would overwrite the static label added to user-defined alerts.

bison: I dropped all the static bits from the assets. The controller now defines the default config, and always appends it to the list. I've tested that it all seems to work even when the secret doesn't exist at Prometheus startup. The controller creates it shortly after, and it's automatically picked up.

jan--f:
/close
superseded by #1676

PR #1597, opened 2022-03-18
Bug 2065682: manifest: explicitly drop the temporary cluster id label

jan--f:
Problem: In commit 461653b460267b2246a18781169df92a9572ec94 we added the tmeporary label
__tmp_openshift_cluster_id__ label as the first entry to all
remote_write relabel configs. The assumption was that all __-prefixed
labels are dropped at the end of the relabeling step. This is however
not true for write_relabel_configs.
See also https://github.com/prometheus/prometheus/discussions/9416

Solution: Explicitly drop the temprary label by adding a drop step at
the end of all write_relabel_configs.

Issues: https://bugzilla.redhat.com/show_bug.cgi?id=2065682

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -56,6 +56,8 @@ const (
 
 	htpasswdArg = "-htpasswd-file=/etc/proxy/htpasswd/auth"
 	clientCAArg = "--client-ca-file=/etc/tls/client/client-ca.crt"
+
+	tmpClusterIDLabelName = "__tmp_openshift_cluster_id__"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/manifests/manifests.go
```diff
@@ -4165,19 +4167,26 @@ func (f *Factory) HashSecret(secret *v1.Secret, data ...string) (*v1.Secret, err
...
 		},
 	}
+	tmpRelabelDrop := monv1.RelabelConfig{
+		Regex:  tmpClusterIDLabelName,
+		Action: "labeldrop",
+	}
 
 	for _, target := range rwTargets {
+		// prepend our temporary cluster id label
+		writeRelabelConfigs := append(clusterIDRelabelConfig, target.WriteRelabelConfigs...)
```
simonpasquier, line 4181: (nit) it might be more readable to group 

```suggestion
		// prepend our temporary cluster id label and drop it eventually
		writeRelabelConfigs := append(clusterIDRelabelConfig, target.WriteRelabelConfigs...)
		writeRelabelConfigs := append(writeRelabelConfigs, tmpRelabelDrop)
```

simonpasquier (commented):
/lgtm

PR #1602, opened 2022-03-21
MON-2269: test: deploy prometheus as remote_write receiver

jan--f:
Problem: Until this commit we used
quay.io/coreos/instrumented-sample-app as the target for our
remote_write test. This sample app doesn't actually ingest metrics, it
only offers mtls support and accepts most http requests. We need to grep
its logs in order to confirm remote_write works as expected.

Solution: Instead deploy a prometheus instance and use that with the
remote-write receive feature. This also supports mtls and we can query
this instance to test for any metric we expect.

Issues: https://issues.redhat.com/browse/MON-2269

### DIFF
--- a/test/e2e/framework/prometheus.go
+++ b/test/e2e/framework/prometheus.go
@@ -1,0 +1,131 @@
+// Copyright 2022 The Cluster Monitoring Operator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package framework
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+
+	routev1 "github.com/openshift/api/route/v1"
+)
+
+func (f Framework) MakePrometheusWithWebTLSRemoteReceive(name, tlsSecretName string) *monitoringv1.Prometheus {
+	// This is not required in the Prometheus spec, but we inspect that value in
+	// WaitForPrometheus. Omitting it causes this code to derefence a nil.
+	replicas := int32(1)
+	return &monitoringv1.Prometheus{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        name,
+			Namespace:   f.Ns,
+			Annotations: map[string]string{},
+			Labels: map[string]string{
+				E2eTestLabelName: E2eTestLabelValue,
+			},
+		},
+		Spec: monitoringv1.PrometheusSpec{
+			Replicas:           &replicas,
+			ServiceAccountName: "prometheus-k8s",
+			Secrets:            []string{tlsSecretName},
+			EnableFeatures:     []string{"remote-write-receiver"},
+			Web: &monitoringv1.WebSpec{
+				TLSConfig: &monitoringv1.WebTLSConfig{
+					ClientCA: monitoringv1.SecretOrConfigMap{
+						Secret: &v1.SecretKeySelector{
+							LocalObjectReference: v1.LocalObjectReference{
+								Name: tlsSecretName,
+							},
+							Key: "ca.crt",
+						},
+					},
+					Cert: monitoringv1.SecretOrConfigMap{
+						Secret: &v1.SecretKeySelector{
+							LocalObjectReference: v1.LocalObjectReference{
+								Name: tlsSecretName,
+							},
+							Key: "server.crt",
+						},
+					},
+					KeySecret: v1.SecretKeySelector{
+						LocalObjectReference: v1.LocalObjectReference{
+							Name: tlsSecretName,
+						},
+						Key: "server.key",
+					},
+					ClientAuthType: "VerifyClientCertIfGiven",
+				},
+			},
+		},
+	}
+}
+
+func (f Framework) MakePrometheusService(ns, name, group string, serviceType v1.ServiceType) *v1.Service {
+	return &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("prometheus-%s", name),
+			Labels: map[string]string{
+				"group":          group,
+				E2eTestLabelName: E2eTestLabelValue,
+			},
+			Namespace: ns,
+		},
+		Spec: v1.ServiceSpec{
+			Type: serviceType,
+			Ports: []v1.ServicePort{
+				{
+					Name:       "http",
+					Port:       80,
+					TargetPort: intstr.FromInt(9090),
+				},
+				{
+					Name:       "https",
+					Port:       443,
+					TargetPort: intstr.FromInt(9090),
+				},
+			},
+			Selector: map[string]string{
+				"prometheus": name,
+			},
+		},
+	}
+}
+
+func (f Framework) MakePrometheusServiceRoute(svc *v1.Service) *routev1.Route {
+	return &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			// naming this after the service would make sense but route
+			// names have limitations (length, character set), so
+			// its hardcoded
+			Name: "prometheusreceiver",
+			Labels: map[string]string{
+				E2eTestLabelName: E2eTestLabelValue,
+			},
+			Namespace: f.Ns,
+		},
+		Spec: routev1.RouteSpec{
+			To: routev1.RouteTargetReference{
+				Kind: "Service",
+				Name: svc.Name,
+			},
+			TLS: &routev1.TLSConfig{
+				Termination: routev1.TLSTerminationPassthrough,
+			},
+		},
+	}
+}
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -15,19 +15,15 @@
 package e2e
 
 import (
-	"bytes"
 	"context"
 	"fmt"
-	"strings"
 	"testing"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 
 	osConfigv1 "github.com/openshift/api/config/v1"
-	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
 	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
 )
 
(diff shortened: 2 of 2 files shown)


jan--f:
/label docs-approved
/label qe-approved
/label px-approved
Adding labels as this PR only adds e2e test cases.

test/e2e/prometheus_test.go
```diff
@@ -94,152 +90,175 @@ func TestAntiAffinity(t *testing.T) {
...
+	route := f.MakePrometheusServiceRoute(svc)
+	if err := f.OperatorClient.CreateRouteIfNotExists(ctx, route); err != nil {
 		t.Fatal(err)
 	}
-	for _, scenario := range []struct {
-		name   string
-		port   string
-		rwSpec string
+
+	url, err := f.OperatorClient.GetRouteURL(ctx, route)
```
JoaoBraveCoding, line 150: Can't we use here the function `framework.NewPrometheusClientFromRoute`?

test/e2e/prometheus_test.go
```diff
@@ -250,69 +269,21 @@ spec:
...
-func remoteWriteCheckMetrics(ctx context.Context, t *testing.T) {
-	time.Sleep(1 * time.Minute)
-	for _, pod := range []string{
-		"prometheus-k8s-0",
-		"prometheus-k8s-1",
-	} {
-		f.ThanosQuerierClient.WaitForQueryReturn(
-			t, 1*time.Minute, fmt.Sprintf(`ceil(delta(prometheus_remote_storage_samples_pending{pod="%s"}[1m]))`, pod),
+func remoteWriteCheckMetrics(ctx context.Context, t *testing.T, promClient *framework.PrometheusClient, tests []remoteWriteTest) {
+	time.Sleep(2 * time.Minute)
```
JoaoBraveCoding, line 279: Why do we have a sleep here?

test/e2e/prometheus_test.go
```diff
@@ -250,69 +269,21 @@ spec:
...
+	for _, test := range tests {
+		promClient.WaitForQueryReturn(
+			t, 1*time.Minute, test.query,
 			func(v int) error {
-				if v == 0 {
-					return fmt.Errorf("prometheus_remote_storage_samples_pending indicates no remote write progress, expected a continuously changing delta")
+				if v == test.expectedValue {
```
JoaoBraveCoding, line 284: Shouldn't it be `!=` here?

test/e2e/prometheus_test.go
jan--f: The `sleep` here give the in-cluster Prometheus a chance to send some samples to the `remote_write` target. We don't want to fail by querying the target immediately when no samples have been send yet.

test/e2e/prometheus_test.go
jan--f: Yes good catch, it should be. That also means I need to change the query `ceil(delta(prometheus_remote_storage_samples_pending{pod="%[1]s",prometheus_replica="%[1]s"}[1m]))` as this is a success when `!= 0`.

test/e2e/prometheus_test.go
jan--f: Good point, I think that's a left-over from when I tried to get away without creating a `Route`.

test/e2e/prometheus_test.go
jan--f: I'll add a comment to that effect.

test/e2e/framework/prometheus.go
```diff
@@ -0,0 +1,156 @@
...
+			},
+		},
+		KeySecret: v1.SecretKeySelector{
+			LocalObjectReference: v1.LocalObjectReference{
+				Name: sec.ObjectMeta.Name,
+			},
+			Key: "server.key",
+		},
+		ClientAuthType: "VerifyClientCertIfGiven",
+		// ClientAuthType: "NoClientCert",
```
JoaoBraveCoding, line 153:
```suggestion

```

test/e2e/prometheus_test.go
```diff
@@ -94,152 +90,178 @@ func TestAntiAffinity(t *testing.T) {
...
+      action: replace
+`,
+			expected: []remoteWriteTest{
+				{
+					query:       `count(prometheus_remote_storage_samples_pending{cluster_id!=""})`,
+					expected:    func(v int) bool { return v == 4 },
+					description: "Expected to find 4 time series of metric prometheus_remote_storage_samples_pending with the cluster_id label",
+				},
+				{
+					query:       `absent(prometheus_remote_storage_samples_pending{__tmp_openshift_cluster_id__=~".+"})`,
```
JoaoBraveCoding, line 257: I think you might have forgotten to update this, no?

jan--f: I'm afraid I need more hints :) 
The description looks a bit weird but is actually correct, if that is what you mean. 
> absent(v instant-vector) returns an empty vector if the vector passed to it has any elements and a 1-element vector with the value 1 if the vector passed to it has no elements.

JoaoBraveCoding: Sorry, my mistake! Please ignore.

test/e2e/framework/prometheus.go
```diff
@@ -0,0 +1,155 @@
...
+			ServiceMonitorSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"group": group,
+				},
+			},
+			PodMonitorSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"group": group,
+				},
+			},
```
simonpasquier, line 53: To ensure that the instance acts as a remote-write receiver only (e.g. no scrape targets and rules), we could set all selector fields to nil.

```suggestion
			ServiceMonitorSelector: nil,
			ServiceMonitorNamespaceSelector: nil,
			PodMonitorSelector: nil,
			PodMonitorNamespaceSelector: nil,
```

test/e2e/framework/prometheus.go
```diff
@@ -0,0 +1,155 @@
...
+			RuleSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"role": "rulefile",
+				},
+			},
+			Resources: v1.ResourceRequirements{
+				Requests: v1.ResourceList{
+					v1.ResourceMemory: resource.MustParse("400Mi"),
+				},
+			},
```
simonpasquier, line 64: I'm not sure we need to set any explicit resource requests.

test/e2e/framework/prometheus.go
```diff
@@ -0,0 +1,155 @@
...
+			Name:        name,
+			Namespace:   ns,
+			Annotations: map[string]string{},
+			Labels: map[string]string{
+				E2eTestLabelName: E2eTestLabelValue,
+			},
+		},
+		Spec: monitoringv1.PrometheusSpec{
+			Replicas: &replicas,
+			Version:  operator.DefaultPrometheusVersion,
```
simonpasquier, line 43: if we let it empty, the operator will default to the correct version (e.g. the Prometheus version that was defined as being default for that particular version of the operator).

test/e2e/framework/prometheus.go
```diff
@@ -0,0 +1,155 @@
...
+				MatchLabels: map[string]string{
+					"group": group,
+				},
+			},
+			ServiceAccountName: "prometheus-k8s",
+			RuleSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"role": "rulefile",
+				},
+			},
```
simonpasquier, line 59:
```suggestion
			RuleSelector: nil,
```

test/e2e/framework/prometheus.go
```diff
@@ -0,0 +1,155 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus-operator/prometheus-operator/pkg/operator"
+
+	routev1 "github.com/openshift/api/route/v1"
+)
+
+func (f *Framework) MakeBasicPrometheus(ns, name, group string, replicas int32, secrets []string) *monitoringv1.Prometheus {
```
simonpasquier, line 31: Since we're only using this function to setup a remote write receiver, we could merge it with `AddRemoteReceiveWithWebTLSToPrometheus`?

```suggestion
func (f *Framework) MakeRemoteReceiverPrometheus(name string, tlsSecretName string) *monitoringv1.Prometheus {
```

test/e2e/prometheus_test.go
simonpasquier: we could also extend the timeout passed to `promClient.WaitForQueryReturn` to avoid hard-waiting for 2 minutes?

test/e2e/prometheus_test.go
```diff
@@ -250,69 +271,22 @@ spec:
...
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	if strings.Contains(promLogs.String(), `msg="Failed to send batch, retrying`) {
-		t.Fatal("unexpected prometheus log message, failed to send batch to remote write endpoint")
-	}
-	if strings.Contains(rwEndpointLogs, "remote error: tls: bad certificate") {
-		t.Fatal("remote write tls endpoint sees bad or no certificate")
+		t.Run(tc.name, func(t *testing.T) {
```
simonpasquier, line 274: t.Run could be moved at the top of the code block (L263)?

test/e2e/prometheus_test.go
```diff
@@ -250,69 +271,22 @@ spec:
...
+			remoteWriteCheckMetrics(ctx, t, prometheusReceiveClient, tc.expected)
+		})
```
simonpasquier, line 276: wondering if we shouldn't reset the remote write config and recycle the Prometheus receiver instance after each sub-test? Otherwise the metric check might be polluted by the previous tests.

jan--f:
@simonpasquier Thanks for the review, all very good points!. `test/e2e/framework/prometheus.go` didn't get enough attention after lifting most of the code from prometheus-operator.

Updated as per you review.

simonpasquier:
/retest

_sorry for the fat fingers, it's Monday morning_

test/e2e/framework/prometheus.go
```diff
@@ -0,0 +1,143 @@
...
+			Labels: map[string]string{
+				E2eTestLabelName: E2eTestLabelValue,
+			},
+		},
+		Spec: monitoringv1.PrometheusSpec{
+			Replicas:                        &replicas,
+			ServiceMonitorSelector:          nil,
+			ServiceMonitorNamespaceSelector: nil,
+			PodMonitorSelector:              nil,
+			PodMonitorNamespaceSelector:     nil,
```
simonpasquier, line 50: In fact I realize that the selectors would be nil by default so no need to set them explicitly.

```suggestion
```

test/e2e/framework/prometheus.go
```diff
@@ -0,0 +1,143 @@
...
+	routev1 "github.com/openshift/api/route/v1"
+)
+
+func (f *Framework) MakePrometheusWithWebTLSRemoteReceive(name, tlsSecretName string) *monitoringv1.Prometheus {
+	p := f.MakeBasicPrometheus(name, 1, []string{tlsSecretName})
+	f.AddRemoteReceiveWithWebTLSToPrometheus(p, tlsSecretName)
+	return p
+}
+
+func (f *Framework) MakeBasicPrometheus(name string, replicas int32, secrets []string) *monitoringv1.Prometheus {
```
simonpasquier, line 35: (nit) since we don't seem to use `MakeBasicPrometheus` elsewhere, it might be clearer move everything into `MakePrometheusWithWebTLSRemoteReceive()`.

test/e2e/framework/prometheus.go
jan--f: I was wondering about that, though failed to check myself.

test/e2e/framework/prometheus.go
jan--f: Yeah I considered that based on your first review. I stuck with the 2 extra functions since the decomposition makes a lot of sense I find.
But you're right, we can always add that when we need it.

PR #1603, opened 2022-03-21
Bug 2065577:  CMO now creates by default an empty CM for UWM

JoaoBraveCoding:

### DIFF
--- a/assets/cluster-monitoring-operator/user-workload-config-edit-role.yaml
+++ b/assets/cluster-monitoring-operator/user-workload-config-edit-role.yaml
@@ -11,4 +11,8 @@ rules:
   resources:
   - configmaps
   verbs:
-  - '*'
+  - get
+  - list
+  - watch
+  - patch
+  - update
--- a/assets/prometheus-user-workload/config-map.yaml
+++ b/assets/prometheus-user-workload/config-map.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+data: {}
+kind: ConfigMap
+metadata:
+  name: user-workload-monitoring-config
+  namespace: openshift-user-workload-monitoring
--- a/jsonnet/components/cluster-monitoring-operator.libsonnet
+++ b/jsonnet/components/cluster-monitoring-operator.libsonnet
@@ -354,7 +354,7 @@ function(params) {
       apiGroups: [''],
       resourceNames: ['user-workload-monitoring-config'],
       resources: ['configmaps'],
-      verbs: ['*'],
+      verbs: ['get', 'list', 'watch', 'patch', 'update'],
     }],
   },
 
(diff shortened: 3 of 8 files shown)


sthaha (commented):
/lgtm

I wonder if this behaviour has to validated using a test

juzhao:
@JoaoBraveCoding from assets/cluster-monitoring-operator/user-workload-config-edit-role.yaml file, we've added detailed `verbs` for configmaps, then it's no need to create an empty CM for UWM, see file: assets/prometheus-user-workload/config-map.yaml
FYI: in a fresh OCP cluster, cluster-monitoring-config cofigmap is not created by default, for UWM, we should not create user-workload-monitoring-config confimap either.
@simonpasquier WDYT?

simonpasquier:
> @JoaoBraveCoding from assets/cluster-monitoring-operator/user-workload-config-edit-role.yaml file, we've added detailed `verbs` for user-workload-monitoring-config-edit role, then it's no need to create an empty CM for UWM, see file: assets/prometheus-user-workload/config-map.yaml FYI: in a fresh default OCP cluster, cluster-monitoring-config cofigmap is not created by default, for UWM, we should not create user-workload-monitoring-config confimap either. @simonpasquier WDYT?

IIUC the `create` verb can't be restrictd to a particular resource. From https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources

>  Note: You cannot restrict create or deletecollection requests by their resource name. For create, this limitation is because the name of the new object may not be known at authorization time. If you restrict list or watch by resourceName, clients must include a metadata.name field selector in their list or watch request that matches the specified resourceName in order to be authorized. For example, kubectl get configmaps --field-selector=metadata.name=my-configmap 

so we have to the following options
* CMO creates the `user-workload-monitoring-config` configmap (with no data) if it doesn't exist
* Allow the `user-workload-monitoring-config-edit` role to create configmaps in the `openshift-user-workload-monitoring` namespace.

The latter option means that a user could create as many configmaps as they want which isn't something that we want?

For consistency, CMO may create an empty `cluster-monitoring-config` configmap if missing but I feel that it's outside the scope of this PR.

juzhao:
> so we have to the following options
> 
> * CMO creates the `user-workload-monitoring-config` configmap (with no data) if it doesn't exist
> * Allow the `user-workload-monitoring-config-edit` role to create configmaps in the `openshift-user-workload-monitoring` namespace.
Thanks, I thought any option would fix the issue, after checking, we need both of them together to fix the issue

juzhao:
tested with PR, CMO now creates by default an empty CM for UWM and user with user-workload-monitoring-config-edit role could edit user-workload-monitoring-config configmap
/label qe-approved

PR #1273, opened 2021-07-09
Cleanup telemetry owners

simonpasquier:
No functional change.

* [ ] I added CHANGELOG entry for this change.
* [X] No user facing changes, so no entry in CHANGELOG was needed.

PR #1606, opened 2022-03-23
Documentation/data-collection: Assign cluster-version metrics to Cincinnati team

wking:
The installer assignments are from a7146706b4 (#1273).  Not clear to me why
that assignment was chosen then, but [the Cincinnati team][1] (the GitHub name of the OpenShift-core updates team) is in charge of the cluster-version operator, and [these are all metrics served by the CVO][2].

`cluster_installer` leans heavily on installer-supplied data, so I'm leaving that one co-owned by the installer team.

Also from a7146706b4 was the two `owners` lines for `cluster_installer`.  I don't think OLM was intended to be an owner for this metric (and they're separately listed as a consumer), so I've dropped their owner line.

[1]: https://github.com/orgs/openshift/teams/openshift-team-cincinnati/members
[2]: https://github.com/openshift/cluster-version-operator/blob/fc7d096f5fcde2758dafd1ce968601cd774479f4/pkg/cvo/metrics.go#L65-L105

### DIFF
--- a/manifests/0000_50_cluster-monitoring-operator_04-config.yaml
+++ b/manifests/0000_50_cluster-monitoring-operator_04-config.yaml
@@ -47,7 +47,7 @@ data:
     # flake.
     - '{__name__="count:up1"}'
     #
-    # owners: (@openshift/openshift-team-installer)
+    # owners: (@openshift/openshift-team-cincinnati)
     #
     # cluster_version reports what payload and version the cluster is being
     # configured to and is used to identify what versions are on a cluster that
@@ -56,7 +56,7 @@ data:
     # consumers: (@openshift/openshift-team-cluster-manager)
     - '{__name__="cluster_version"}'
     #
-    # owners: (@openshift/openshift-team-installer)
+    # owners: (@openshift/openshift-team-cincinnati)
     #
     # cluster_version_available_updates reports the channel and version server
     # the cluster is configured to use and how many updates are available. This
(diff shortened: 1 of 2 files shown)


simonpasquier:
>  Not clear to me why that assignment was chosen then

I can't remember how I came to the conclusion that it should be `@openshift-team-installer` so happy to fix it :)

/lgtm
/label px-approved
/label docs-approved
/label qe-approved
/skip

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -91,9 +91,8 @@ data:
     # consumers: (@openshift/openshift-team-cluster-manager)
     - '{__name__="cluster_version_payload"}'
     #
-    # owners: (@openshift/openshift-team-installer)
+    # owners: (@openshift/openshift-team-installer, @openshift/openshift-team-cincinnati)
```
jan--f, line 94:
```suggestion
    # owners: (@openshift/openshift-team-cincinnati)
```
Did you want to keep one instance of `@openshift/openshift-team-installer`?

jan--f:
/hold
Just to make sure the single `openshift-team-installer` mention left was intentional.

wking:
> Just to make sure the single openshift-team-installer mention left was intentional.

Yup.  Per my initial comment and the commit message:

> `cluster_installer` leans heavily on installer-supplied data, so I'm leaving that one co-owned by the installer team.

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
wking: Yup. Per my initial comment and the commit message:

> `cluster_installer` leans heavily on installer-supplied data, so I'm leaving that one co-owned by the installer team.

PR #1607, opened 2022-03-23
Documentation/data-collection: Collect cluster_version_capability

wking:
Builds on #1606; possibly review that one first.

The cluster-version operator is growing this new metric in openshift/cluster-version-operator#755, and @bparees wants it included in Telemetry so we collect it for [clusters that have Telemetry enabled, even if Insights is disabled][2].  Cardinality is expected to be very small, with a single label ([`name`][1]) and a restricted value set (only three registered capabilities expected for 4.11).

[1]: https://github.com/openshift/cluster-version-operator/pull/755
[2]: https://github.com/openshift/enhancements/pull/922#discussion_r742287363


jan--f:
/lgtm
/label docs-approved
/label px-approved
/label qe-approved
/hold
until #1606 is merged.

wking:
/retest

@jan--f , looks like your label requests didn't stick?  Can you add them again?

wking:
I guess with no `lgtm` label yet to blow away, I might as well rebase on master to show that we're down to one new commit here, so I've done that with f0a80517 -> ebdc1aea.

PR #1608, opened 2022-03-23
Bug 2063047: Add condition to check for relative paths in query log file path

JoaoBraveCoding:
Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2063047

problem: prometheus-operator does not support relative paths for the query
log file path

solution: add condition to not allow the field queryLogFilePath to start
with a '.'

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1657,7 +1657,7 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
 		return nil
 	}
 	dirPath := filepath.Dir(queryLogFile)
-	if strings.HasPrefix(dirPath, ".") {
+	if dirPath != "." && strings.HasPrefix(dirPath, ".") {
 		return errors.Wrap(ErrConfigValidation, `relative paths to query log file are not supported`)
 	}
 	if dirPath == "/" {
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1657,7 +1657,8 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
 		return nil
 	}
 	dirPath := filepath.Dir(queryLogFile)
-	if dirPath != "." && strings.HasPrefix(dirPath, ".") {
+	// queryLogFile is not an absolute path nor a simple filename
+	if !filepath.IsAbs(queryLogFile) && dirPath != "." {
 		return errors.Wrap(ErrConfigValidation, `relative paths to query log file are not supported`)
 	}
 	if dirPath == "/" {
@@ -1666,8 +1667,10 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
 
 	// /prometheus is where Prometheus will store the TSDB so it is
 	// already mounted inside the pod (either from a persistent volume claim or from an empty dir).
+	// When queryLogFile is a simple filename the prometheus-operator will take
+	// care of mounting an emptyDir under /var/log/prometheus
 	p.Spec.QueryLogFile = queryLogFile
-	if dirPath == "/prometheus" {
+	if dirPath == "/prometheus" || dirPath == "." {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/manifests/manifests.go
```diff
@@ -1657,6 +1657,9 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
 		return nil
 	}
 	dirPath := filepath.Dir(queryLogFile)
+	if strings.HasPrefix(dirPath, ".") {
```
simonpasquier, line 1660: `query.log` is valid (the Prometheus operator will automatically mount an emptyDir volume at `/var/log/prometheus`) but `filepath.Dir()` would return `.` for it.

```suggestion
	if dirPath != "." && strings.HasPrefix(dirPath, ".") {
```

Also worth adding a test case.

JoaoBraveCoding: Ohhh I didn't knew that! Thank you~

JoaoBraveCoding:
\retest

pkg/manifests/manifests_test.go
```diff
@@ -1433,30 +1433,36 @@ func TestPrometheusQueryLogFileConfig(t *testing.T) {
...
 		},
-	}
-	for _, tt := range testCases {
+		{
+			name:             "invalid path, relative path",
+			queryLogFilePath: "./dev/query.log",
+			expected:         "",
+			errExpected:      true,
+			volumeExpected:   false,
+		},
```
simonpasquier, line 1442: we need another test case when `queryLogFile` is a filename without a directory :)

```suggestion
		{
			name:             "invalid path, relative path",
			queryLogFilePath: "./dev/query.log",
			expected:         "",
			errExpected:      true,
			volumeExpected:   false,
		},
		{
			name:             "filename only",
			queryLogFilePath: "query.log",
			expected:         "query.log",
			errExpected:      false,
			volumeExpected:   false,
		},
```

pkg/manifests/manifests.go
```diff
@@ -1657,6 +1657,9 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
 		return nil
 	}
 	dirPath := filepath.Dir(queryLogFile)
+	if dirPath != "." && strings.HasPrefix(dirPath, ".") {
```
arajkumar, line 1660: Can we use [filepath#Abs](https://pkg.go.dev/path/filepath#Abs)?
```suggestion
	if filepath.Abs(queryLogFile) != queryLogFile {
```

pkg/manifests/manifests_test.go
```diff
@@ -1433,30 +1433,43 @@ func TestPrometheusQueryLogFileConfig(t *testing.T) {
...
+		{
+			name:             "filename only",
+			queryLogFilePath: "query.log",
+			expected:         "query.log",
+			errExpected:      false,
```
arajkumar, line 1447: Isn't `query.log` a relative path?

simonpasquier: if `queryLogFile` doesn't contain a directory (absolute or relative), the Prometheus operator will add and automatically mount an emptyDir volume (see https://github.com/prometheus-operator/prometheus-operator/pull/4566).

pkg/manifests/manifests_test.go
```diff
@@ -1433,30 +1433,43 @@ func TestPrometheusQueryLogFileConfig(t *testing.T) {
...
+			volumeExpected:   false,
+		},
+	} {
```
simonpasquier, line 1450: You missed to wrap the test with `t.Run()` and I think that it swallows a failing sub-test...

```suggestion
	} {
		t.Run(t.name, func(t *testing.T) {
```

simonpasquier (commented):
/hold

pkg/manifests/manifests.go
JoaoBraveCoding: Ended up using a combination of this with `dirPath != "."` since we have to account for the case where the `queryLogFile` its just a filename

pkg/manifests/manifests.go
```diff
@@ -1657,14 +1657,20 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
 		return nil
 	}
 	dirPath := filepath.Dir(queryLogFile)
+	// queryLogFile is not an absolute path nor a simple filename
+	if !filepath.IsAbs(queryLogFile) && dirPath != "." {
```
simonpasquier, line 1661: :+1:

simonpasquier (commented):
/lgtm
/hold cancel

PR #1611, opened 2022-03-23
Bug 2048333: Adds readinessProbe and livenessProbe to prometheus-adapter jsonnet

JoaoBraveCoding:
Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2048333

problem: Currently the prometheus-adapter pods are restarted at the same
time even though the deployment is configured with strategy RollingUpdate.
This happens because the kubelet does not know when the prometheus-adapter
pods are ready to start receiving requests.

solution: Add both readinessProbe and livenessProbe to the
prometheus-adapter, this way the kubelet will know when either the pod
stoped working and should be restarted or simply when it ready to start
receiving requests.


simonpasquier (commented):
I think that the change should go [upstream](https://github.com/prometheus-operator/kube-prometheus/blob/main/jsonnet/kube-prometheus/components/prometheus-adapter.libsonnet) first. WDYT?

JoaoBraveCoding:
Yes, indeed!

JoaoBraveCoding:
Done https://github.com/prometheus-operator/kube-prometheus/pull/1696

Shall I close this PR then?

simonpasquier:
> Shall I close this PR then?

yes, we'll pull from upstream once it's merged.

/close

PR #1617, opened 2022-03-29
[MON-2208] Add Oauth2 settings to prometheusK8s.remoteWrite config

raptorsun:
This PR addes Oauth2 settings to PrometheusK8s.remoteWrite config, allowing prometheus remoteWrite use client secret to access the remote write server. An example config can be:
```yaml
prometheusK8s:
  remoteWrite:
    - url: https://test.remotewrite.com/api/write
      remoteTimeout: 30s
      oauth2:
        clientId:
          secret:
            name: oauth2-credentials
            key: id
        clientSecret:
          name: oauth2-credentials
          key: secret
        tokenUrl: https://example.com/oauth2/token
        scopes:
          - scope1
          - scope2
        endpointParams:
          param1: value1
          param2: value2
```

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -164,6 +164,8 @@ type RemoteWriteSpec struct {
 	QueueConfig *monv1.QueueConfig `json:"queueConfig,omitempty"`
 	// MetadataConfig configures the sending of series metadata to remote storage.
 	MetadataConfig *monv1.MetadataConfig `json:"metadataConfig,omitempty"`
+	// OAuth2 configures OAuth2 authentication for remote write.
+	OAuth2 *monv1.OAuth2 `json:"oauth2,omitempty"`
 }
 
 type PrometheusK8sConfig struct {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [1578](https://github.com/openshift/cluster-monitoring-operator/pull/1578) Add temporary cluster id label to remote write relabel configs.
 - [#1350](https://github.com/openshift/cluster-monitoring-operator/pull/1350) Support label scrape limits in user-workload monitoring
 - [#1601](https://github.com/openshift/cluster-monitoring-operator/pull/1601) Expose the /federate endpoint of UWM Prometheus as a service
+- [#1617](https://github.com/openshift/cluster-monitoring-operator/pull/1617) Add Oauth2 setting to PrometheusK8s remoteWrite config
 
 ## 4.10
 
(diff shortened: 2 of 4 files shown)


raptorsun:
The new commit resolved merge conflict in `CHANGELOG.md`.

pkg/manifests/config.go
```diff
@@ -164,6 +164,8 @@ type RemoteWriteSpec struct {
 	QueueConfig *monv1.QueueConfig `json:"queueConfig,omitempty"`
 	// MetadataConfig configures the sending of series metadata to remote storage.
 	MetadataConfig *monv1.MetadataConfig `json:"metadataConfig,omitempty"`
+	// OAuth2 configs. Requires Prometheus versions 2.27.0+.
```
simonpasquier, line 167:
```suggestion
	// OAuth2 configures OAuth2 authentication for remote write.
```

JoaoBraveCoding:
Ping @juzhao for qe approval.

pkg/manifests/config.go
```diff
@@ -164,6 +164,8 @@ type RemoteWriteSpec struct {
 	QueueConfig *monv1.QueueConfig `json:"queueConfig,omitempty"`
 	// MetadataConfig configures the sending of series metadata to remote storage.
 	MetadataConfig *monv1.MetadataConfig `json:"metadataConfig,omitempty"`
+	// OAuth2 configures OAuth2 authentication for remote write.
+	OAuth2 *monv1.OAuth2 `json:"oauth2,omitempty"`
```
juzhao, line 168: @raptorsun for `monv1.OAuth2`, checked in https://github.com/openshift/cluster-monitoring-operator/blob/release-4.11/vendor/github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1/types.go#L1171-L1183
the `OAuth2 struct` do not have `tls_config` field as listed in
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#oauth2

raptorsun: That's a good point. I'm going to submit a PR to add this field to the upstream Prometheus Operator 👍
In OAuth2 the necessary configurations are Client ID, Client Secret and TokenURL. 
TLSConfig sets TLS for requesting token from TokenURL, which is not required on all TLS configs, so this still works on most systems. We can live without that field :)

juzhao:
bug: https://bugzilla.redhat.com/show_bug.cgi?id=2073371
remoteWrite.oauth2 setting is not loaded to prometheus config file /etc/prometheus/config_out/prometheus.env.yaml

PR #1619, opened 2022-03-30
Bug 2030698: KubePodCrashLooping may fire when pod is not in CrashLoopBackOff

arajkumar:
This change pickups https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/721


juzhao:
tested with the PR, KubePodCrashLooping could be fired, and the expr is continuous in graph, no gaps, see the picture.
/label cherry-pick-approved

PR #1620, opened 2022-03-30
[MON-1773] doc: add Testing section to CONTRIBUTING.md

raptorsun:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,7 @@ This document outlines some of the conventions on development workflow, commit m
 
 # Certificate of Origin
 
-By contributing to this project you agree to the Developer Certificate of Origin (DCO). 
+By contributing to this project you agree to the Developer Certificate of Origin (DCO).
 This document was created by the Linux Kernel community and is a simple statement that you, as a contributor, have the legal right to make the contribution. See the [DCO](DCO) file for details.
 
 # Notes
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -110,6 +110,29 @@ At this point, you should follow a standard git workflow:
...
 
+## Testing
+
+Supposing $KUBECONFIOG is set to the config file of a Kubernetes cluster running the codes to test. We can run all tests by using this command `make test`.
+
+The testing consist of 3 aspects:
+- unit tests, can be run seperately by `make test-unit`.
+- Prometheus rule tests, can be run seperately by `make test-rules`.
+- end to end tests, can be run seperately by `make test-e2e`.
+Unit testing can be run by using `make test-unit` command.
```
JoaoBraveCoding, line 121: This phrase can be removed no? Since the first bullet point already tell us this
```suggestion
```

CONTRIBUTING.md
```diff
@@ -110,6 +110,29 @@ At this point, you should follow a standard git workflow:
...
+go test -v -timeout=120m ./test/e2e/ --kubeconfig $KUBECONFIG -run TestBodySizeLimit
+```
+Attention that we have to pass a valid $KUBECONFIG explicitly `--kubeconfig $KUBECONFIG` even if ~/.kube/config exists.
+
+To run a specific test case of unit tests, we can refer to these commands.
+```bash
+go test -v -timeout=120m ./pkg/manifests -run TestImageParsing
+go test -v -timeout=120m ./pkg/client -run TestCreateOrUpdateService
+```
+Make sure change the root directory of the Golang package, the test function have to be named as `TestXXXXXX`.
```
JoaoBraveCoding, line 134: I'm not sure I fully understood  the first part of this phrase. Do you mean that to run the commands above we should be at the root of the golang package? In this case, inside the folder `cluster-monitoring-opetator`

JoaoBraveCoding (commented):
Thank you for this 👍 really great to have

CONTRIBUTING.md
raptorsun: I should rephrase this line. I want to say to replace ./pkg/client to the root directory of the Go package (where we declare `package "XXX"` in the source code files).

raptorsun: Just rephrased the line with an example. Does the new version offer a more clear explanation?

JoaoBraveCoding: Super clear now, thank you :)

PR #1469, opened 2021-11-08
MON-1261: exposing topology spread constraints through config

prashbnair:
This PR exposes the [TopologySpreadConstraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/) through the CMO config. 

* [ ] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #1624, opened 2022-03-31
MON-1261: exposing topology spread constraints through config

jan--f:
This PR exposes the [TopologySpreadConstraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/) through the CMO config.

This is just #1469 rebased on master.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -676,6 +676,11 @@ func (f *Factory) AlertmanagerMain(trustedCABundleCM *v1.ConfigMap) (*monv1.Aler
 		a.Spec.Tolerations = f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Tolerations
 	}
 
+	if len(f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.TopologySpreadConstraints) > 0 {
+		a.Spec.TopologySpreadConstraints =
+			f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.TopologySpreadConstraints
+	}
+
 	for i, c := range a.Spec.Containers {
 		switch c.Name {
 		case "alertmanager-proxy":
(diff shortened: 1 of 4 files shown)


jan--f:
/hold
for approved labels

pkg/manifests/manifests.go
```diff
@@ -534,6 +534,11 @@ func (f *Factory) AlertmanagerMain(trustedCABundleCM *v1.ConfigMap) (*monv1.Aler
 		a.Spec.Tolerations = f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Tolerations
 	}
 
+	if len(f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.TopologySpreadConstraints) > 0 {
```
sthaha, line 537: should we check the `len` ? Wouldn't `a.Spec.TopologySpreadConstraints =		 f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.TopologySpreadConstraints` yield the same result?

sthaha (commented):
/lgtm 

A nit that you may want to consider addressing.

faust64:
Any news?
This would be quite useful.
When I last upgraded OpenShift clusters to 4.10, I got a few clusters stuck running pre-upgrade checks. Upgrade would not start, if all pods from an alertmanager, prometheus or thanos statefulset are located on the same node.
Which I could not believe: there isn't even a kubernetes.io/hostname anti-affinity rule? Nor any way, yet, to configure affinities for cluster-monitoring components?

jan--f:
@faust64 please follow https://issues.redhat.com/browse/MON-1261. I'll update the code and get it merged soon. However this will only make it into the 4.12 release due to documentation and QE work.

pkg/manifests/manifests.go
jan--f: We do check `len` in other places in CMO too. Iiuc what this does is, if an empty slice is passed from the user, it gets normalized to `nil`. Maybe that provides a small optimization for the resulting json? The nil value can be omitted instead of having to encode and empty slice?
Not sure, but I'd opt to stay consistent with the other places here.

simonpasquier: yes the `len(...) > 0` check is to handle nil and empty slices altogether (e.g. `topologySpreadConstraints: []` and `topologySpreadConstraints: nil`).

jan--f:
/unhold
/label docs-approved
Adding doc approval as per agreement.
Ping @juzhao for qe approval.
https://issues.redhat.com/browse/MON-1261

PR #1625, opened 2022-03-31
Added a KubeCronJobFailed alert

aliceh:
In this commit, I added the KubeCronJobFailed alert per [OSD-8667](https://issues.redhat.com/browse/OSD-8667).

* [ X] No user facing changes, so no entry in CHANGELOG was needed.

### DIFF
--- a/assets/control-plane/prometheus-rule.yaml
+++ b/assets/control-plane/prometheus-rule.yaml
@@ -184,6 +184,17 @@ spec:
       for: 12h
       labels:
         severity: warning
+    - alert: KubeCronJobFailed
+      annotations:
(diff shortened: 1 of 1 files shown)


aliceh:
@dgrisonnet , @bison could someone please help me understand what is missing?

aliceh:
@bison, thanks for the info! I wrote this alert is for catching failed cron jobs.  Yes, as you corretly stated: 
completion time - the start time < 0, AND the job is not active, alert after 15 minutes. 
I will close this PR and submit correctly.

simonpasquier:
I guess that the alert will be useful for upstream too (e.g. https://github.com/kubernetes-monitoring/kubernetes-mixin). cc @arajkumar

PR #1630, opened 2022-04-05
MON-2193: pkg/manifests: Expose retention size settings for UWM Prometheus

slashpai:
WIP: Need to wait till #1579 gets merged and rebase on top of that

[MON-2217](https://issues.redhat.com/browse/MON-2217): Expose retention size settings for UWM Prometheus in the CMO configuration

Introduce a new "retentionSize" field under the prometheus key

Signed-off-by: Jayapriya Pai [janantha@redhat.com](mailto:janantha@redhat.com)

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1762,6 +1762,10 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret) (*monv1.Prometheus,
 		p.Spec.Retention = f.config.UserWorkloadConfiguration.Prometheus.Retention
 	}
 
+	if f.config.UserWorkloadConfiguration.Prometheus.RetentionSize != "" {
+		p.Spec.RetentionSize = f.config.UserWorkloadConfiguration.Prometheus.RetentionSize
+	}
+
 	p.Spec.Image = &f.config.Images.Prometheus
 
 	if f.config.UserWorkloadConfiguration.Prometheus.Resources != nil {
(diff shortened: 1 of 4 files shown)


simonpasquier (commented):
/lgtm

pkg/manifests/manifests.go
```diff
@@ -1746,6 +1746,10 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret) (*monv1.Prometheus,
 		p.Spec.Retention = f.config.UserWorkloadConfiguration.Prometheus.Retention
 	}
 
+	if f.config.UserWorkloadConfiguration.Prometheus.RetentionSize != "" && f.config.UserWorkloadConfiguration.Prometheus.RetentionSize != "0" {
```
sthaha, line 1749: I think it may be better to trim before comparing strings.

simonpasquier: I've put a comment on the other PR #1579: I'd be fine if we don't treat "0" as a special case.

> Isn't 0 an acceptable value?

it is a valid value and means the same as "".

> is "0GB" acceptable?

yes.

slashpai: ya 0 is valid for [parse2bytes](https://github.com/alecthomas/units/blob/b94a6e3cc13755c0a75fffecbb089eb346fc4289/bytes.go#L31-L37)  used by PO validate function , there was sort of confusion when `0` was specified to avoid in jira. I have updated pr https://github.com/openshift/cluster-monitoring-operator/pull/1579 will update this one too :)

slashpai:
@simonpasquier Could you please another look at this?

simonpasquier (commented):
/lgtm

PR #1632, opened 2022-04-11
WIP StateErrors -> error

sthaha:

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -16,6 +16,7 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"reflect"
 	"strings"
@@ -33,10 +34,12 @@ import (
 	"github.com/prometheus-operator/prometheus-operator/pkg/alertmanager"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	monitoring "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+	"github.com/prometheus-operator/prometheus-operator/pkg/k8sutil"
 	prometheusoperator "github.com/prometheus-operator/prometheus-operator/pkg/prometheus"
 	"github.com/prometheus-operator/prometheus-operator/pkg/thanos"
 	thanosoperator "github.com/prometheus-operator/prometheus-operator/pkg/thanos"
 	admissionv1 "k8s.io/api/admissionregistration/v1"
+
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
@@ -729,10 +732,13 @@ func (c *Client) DeleteSecret(ctx context.Context, s *v1.Secret) error {
 	return err
 }
 
-func (c *Client) WaitForPrometheus(ctx context.Context, p *monv1.Prometheus) error {
+func (c *Client) WaitForPrometheusByNsName(ctx context.Context, metadata *metav1.ObjectMeta) (*monv1.Prometheus, error) {
 	var lastErr error
+	var p *monv1.Prometheus
+
 	if err := wait.Poll(time.Second*10, time.Minute*5, func() (bool, error) {
-		p, err := c.mclient.MonitoringV1().Prometheuses(p.GetNamespace()).Get(ctx, p.GetName(), metav1.GetOptions{})
+		var err error
+		p, err = c.mclient.MonitoringV1().Prometheuses(metadata.GetNamespace()).Get(ctx, metadata.GetName(), metav1.GetOptions{})
 		if err != nil {
 			lastErr = err
 			klog.V(4).ErrorS(err, "WaitForPrometheus: failed to get Prometheus object")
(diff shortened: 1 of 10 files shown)


jan--f:
Based on the [various](https://github.com/openshift/cluster-monitoring-operator/pull/1632/files#diff-2895f30116602d8fe6b0545c186d4f81247fa4eaab049dab174c6c91c0036e08R874) [`TODO`](https://github.com/openshift/cluster-monitoring-operator/pull/1632/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR122) items, [unused functions](https://github.com/openshift/cluster-monitoring-operator/pull/1632/files#diff-e54a75645d7ceee957e904aca41e423870d944f5392f671dd88c5dfeb63d2788R162) and [section of commented out code](https://github.com/openshift/cluster-monitoring-operator/pull/1632/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR624-R633) I didn't add a detailed review yet, but I think its on a good path. The taks separation and prometheus validation look good.
Two higher level oberservations I would appreciate a discussion about:
- Consistent naming. This PR currently uses `State` and `Status` for the same (or at least very similar) concept(s). Since this aims to improve reporting on the status of resources using `Status` consistently makes most sense to me. I.e. mostly `StateError` -> `StatusError`. If there are good arguments to use `State` then lets do so consistently.
- The `StateError` type is encapsulating what we need in order to improve our reporting really well. `StateErrors` and `StateErrorBuilder` on the other hand add a lot of code that we maybe don't need. [The apimachinery errors module with its Aggregate](https://pkg.go.dev/k8s.io/apimachinery@v0.23.5/pkg/util/errors#Aggregate) type looks like it does everything we need in terms of error aggregation. The `Is` and `FilterOut` can do much of the heavy lifting needed for reporting, i.e. identifying what errors were reported and separating them in order to aggregate the reasons that we want to report.
Maybe that's not the case and I'm being dense here. In that case some clarifying comments would be appreciated.

sthaha:
> Consistent naming. This PR currently uses State and Status for the same (or at least very similar) concept(s). Since this aims to improve reporting on the status of resources using Status consistently makes most sense to me. I.e. mostly StateError -> StatusError. If there are good arguments to use State then lets do so consistently.

Agreed, let's keep it consistent.

> The StateError type is encapsulating what we need in order to improve our reporting really well. StateErrors and StateErrorBuilder on the other hand add a lot of code that we maybe don't need.

I am happy to strip away code that is unused but do note that these were added during the development when there was an actual need for the functions to exists. E.g. ValidationTask will **always** return StateErrors, so `MustAdd` were useful ... however, I later decided that we return just any error. IMHO, tasks returning State/StatusError is way better than return any error but thats a different discussion. 


In relation to use of Aggregate I think it solves a different problem than what we have but please feel free to correct.

 I prefer `error-builder` since it handles the `list of errors` creation gracefully. The caller doesn't have to worry about manually maintaining a list and check if `len(errs) == 0 -> return nil`, `len(errs) == 1; return -> errs[0]`  as compared to Aggregate that already expects actually  list of errors to be passed to a NewAggregate.

 The Builder simplifies and flattens errors without having to call Flatten since it already maintains errors as a slice and there aren't any hierarchies.

The intended usage is something like ...:
```go
func xyz() error {

  b := StatusErrorsBuilder{}
  err := foo()
  b.AddError(err, DegradedStatus)

  err2 := bar()
  b.AddError(err2, UnavailableStatus)

  return b.ToError() // can return nil, StatusError or []StatusError 
}

if err := xyz(); err != nil { // can check if `err == nil `
   panic(...)
}
```

But like I mentioned please feel free to correct, it would be nice to see how Aggregate could be a better fit.

jan--f:
[`NewAggregate` will do the length check](https://github.com/kubernetes/apimachinery/blob/276a8a7530a30f42c82b1f0fa2261a905130bf97/pkg/util/errors/errors.go#L46) for you. So the usage pattern is the same, create a container, add errors if there are any and return the wrapper that behaves like an error.

`Flatten` is only fr use cases where you have an aggregate of aggregates. Unless we create those, there is no need to call flatten. However that functionality does open the path to replace `TaskErrors` with `Aggregate` in the future.

So the usage stays pretty much the same and we can just import the code instead of maintaining it ourselves.

sthaha:
> [`NewAggregate` will do the length check](https://github.com/kubernetes/apimachinery/blob/276a8a7530a30f42c82b1f0fa2261a905130bf97/pkg/util/errors/errors.go#L46) for you. So the usage pattern is the same, create a container, add errors if there are any and return the wrapper that behaves like an error.
> 
```
func xyz() error {

  //   b := StatusErrorsBuilder{}
  var  errs []error

  err := foo()
 //   b.AddError(err, DegradedStatus)

  // convert err to StateError if not already ... 
  if !errors.Is(err, &serr) {
     errs = append(errs, StateError{ reason: err.Error(), DegradedState})
  } else {
    errs = append(errs, err) 
  } 

  err2 := bar()
  // b.AddError(err2, UnavailableStatus)
  errs = append(errs, err2) // same as above 

 //  return b.ToError() // can return nil, StatusError or []StatusError 
  return NewAggregate(errs)
}

if err := xyz(); err != nil { // can check if `err == nil `
   panic(...)
}
```

sthaha:
closed in favour of https://github.com/openshift/cluster-monitoring-operator/pull/1558 which uses Aggregates instead of StateErrorBuilder

PR #1634, opened 2022-04-11
Bug 2074084: CMO metrics not visible in the OCP webconsole UI

raptorsun:
This PR corrects the selector of Service Monitor of CMO, thus correct the bug.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -124,6 +124,7 @@ func TestTargetsUp(t *testing.T) {
 		"prometheus-k8s",
 		"prometheus-operator",
 		"alertmanager-main",
+		"cluster-monitoring-operator",
 	}
 
 	for _, target := range targets {
--- a/manifests/0000_50_cluster-monitoring-operator_03-service.yaml
+++ b/manifests/0000_50_cluster-monitoring-operator_03-service.yaml
@@ -7,7 +7,7 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
   labels:
-    app: cluster-monitoring-operator
+    app.kubernetes.io/name: cluster-monitoring-operator
   name: cluster-monitoring-operator
   namespace: openshift-monitoring
 spec:
@@ -17,4 +17,4 @@ spec:
     port: 8443
     targetPort: https
   selector:
-    app: cluster-monitoring-operator
+    app.kubernetes.io/name: cluster-monitoring-operator


raptorsun:
> From my side looks good, I'm just not sure if we should remove the old label or simply add the new one.
> 
> Also for my education, how does fixing this label fixes the bug itself? I'm not seeing the connection

The cause of the problem is Prometheus not scraping the target of the cluster monitoring operator. This scrape target endpoint is configured by this [ServiceMonitor](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/cluster-monitoring-operator/service-monitor.yaml#L18). This ServiceMonitor uses the label `app.kubernetes.io/name: cluster-monitoring-operator` as its selector.  So we should make the label in Service and ServiceMonitor match each other. Kubernetes convention prefers the label `app.kubernetes.io/name: cluster-monitoring-operator`, so I change it to that label.

raptorsun:
The 2 missing metrics show up as expected.
image

JoaoBraveCoding:
> > From my side looks good, I'm just not sure if we should remove the old label or simply add the new one.
> > Also for my education, how does fixing this label fixes the bug itself? I'm not seeing the connection
> 
> The cause of the problem is Prometheus not scraping the target of the cluster monitoring operator. This scrape target endpoint is configured by this [ServiceMonitor](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/cluster-monitoring-operator/service-monitor.yaml#L18). This ServiceMonitor uses the label `app.kubernetes.io/name: cluster-monitoring-operator` as its selector. So we should make the label in Service and ServiceMonitor match each other. Kubernetes convention prefers the label `app.kubernetes.io/name: cluster-monitoring-operator`, so I change it to that label.
> 
> Still testing :)

Thank you a lot for the explanation! Now I understand :)

simonpasquier:
can you add an e2e test to ensure we don't regress in the future?
/hold

raptorsun:
> can you add an e2e test to ensure we don't regress in the future? /hold

E2E test has been added to check presence of the 2 metrics `cluster_monitoring_operator_last_reconciliation_successful` and `cluster_monitoring_operator_reconcile_attempts_total`.

lihongyan1:
test with pr, the issue is fixed, that is, can see cluster-monitoring-opertator related metrics
```
oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- curl -k -H "Authorization: Bearer $token" 'https://thanos-querier.openshift-monitoring.svc:9091/api/v1/label/__name__/values'|jq|grep cluster_monitoring_operator
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 75354    0 75354    0     0  1711k      0 --:--:-- --:--:-- --:--:-- 1711k
    "cluster_monitoring_operator_last_reconciliation_successful",
    "cluster_monitoring_operator_reconcile_attempts_total",
```

simonpasquier (commented):
Instead of a new test you could extend `TestTargetsUp` to include `cluster-monitoring-operator`.

raptorsun:
New E2E test is moved to `TestTargetsUp`, checking target `cluster-monitoring-operator` and its 2 metrics.

test/e2e/main_test.go
```diff
@@ -134,6 +135,15 @@ func TestTargetsUp(t *testing.T) {
...
 
+	// Make sure CMO metrics are up.
+	f.ThanosQuerierClient.WaitForQueryReturnOne(
+		t, 5*time.Minute, "cluster_monitoring_operator_last_reconciliation_successful",
+	)
+
+	f.ThanosQuerierClient.WaitForQueryReturnGreaterEqualOne(
+		t, 5*time.Minute, "cluster_monitoring_operator_reconcile_attempts_total",
+	)
+
```
simonpasquier, line 146: this isn't needed, what we want to ensure is that CMO is scraped successfully by Promehteus and the query on the up metric already checks this. 

```suggestion
```

simonpasquier:
`manifests/0000_50_cluster-monitoring-operator_05-deployment-ibm-cloud-managed.yaml` needs an update too but it should have been detected by the CI :thinking:

simonpasquier (commented):
/lgtm

PR #1638, opened 2022-04-12
MON-2206: Adds sigv4 settings for remote write

JoaoBraveCoding:
Issue https://issues.redhat.com/browse/MON-2206

Problem: Prometheus and Prometheus operator already support sigv4 authentication for
remote write. This should be possible to configure the same in the CMO configuration

Solution: Add to the RemoteWriteSpec struct the Sigv4 field so users can
specify Sigv4 configuration in the CMO ConfigMap, pass this field to the
Prometheus CRD

### DIFF
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -1344,6 +1344,48 @@ func TestRemoteWriteAuthorizationConfig(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "sigv4 authentication configuration",
+			config: `prometheusK8s:
+  remoteWrite:
+  - url: "https://authorization.remotewrite.com/api/write"
+    sigv4:
+      region: eu
+      accessKey:
+        name: aws-credentials
+        key: access
+      secretKey:
+        name: aws-credentials
+        key: secret
+      profile: "SomeProfile"
+      roleArn: "SomeRoleArn"
+`,
+			checkFn: []func(*testing.T, monv1.RemoteWriteSpec){
+				func(t *testing.T, target monv1.RemoteWriteSpec) {
+					if target.Sigv4.Region != "eu" {
+						t.Fatalf("Region field not correct in section RemoteWriteSpec.Sigv4 expected 'eu', got %s", target.Sigv4)
+					}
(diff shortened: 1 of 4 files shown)


simonpasquier (commented):
/lgtm
/retest

JoaoBraveCoding:
Ping @juzhao for qe approval.

JoaoBraveCoding:
/label docs-approved
Adding docs label as per agreement.

pkg/manifests/manifests_test.go
```diff
@@ -1344,6 +1344,44 @@ func TestRemoteWriteAuthorizationConfig(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "sigv4 authentication configuration",
+			config: `prometheusK8s:
+  remoteWrite:
+  - url: "https://authorization.remotewrite.com/api/write"
+    sigv4:
+      accessKey:
```
juzhao, line 1353: from https://github.com/openshift/cluster-monitoring-operator/blob/release-4.11/vendor/github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1/types.go#L754-L765
there is a `region` field, we'd better test it

pkg/manifests/manifests_test.go
```diff
@@ -1344,6 +1344,48 @@ func TestRemoteWriteAuthorizationConfig(t *testing.T) {
...
+    sigv4:
+      region: eu
+      accessKey:
+        name: aws-credentials
+        key: access
+      secretKey:
+        name: aws-credentials
+        key: secret
+      profile: "SomeProfile"
+      roleArn: "SomeRoleArn"
```
juzhao, line 1361: FYI
RoleArn has minimum filed size of 20, otherwise, would see error in prometheus-k8s pod logs
```
ts=2022-04-14T11:45:44.525Z caller=dedupe.go:112 component=remote level=warn remote_name=a42715 url=https://authorization.remotewrite.com/api/write msg="Failed to send batch, retrying" err="Post \"https://authorization.remotewrite.com/api/write\": failed to sign request: InvalidParameter: 1 validation error(s) found.\n- minimum field size of 20, AssumeRoleInput.RoleArn.\n"
```

PR #1639, opened 2022-04-12
Bug 1933144: Extend KubeAggregatedAPIDown alert "for" to 15m

pacevedom:
Per discussion in this PR thread, and also given the fact that [BZ 1933144](https://bugzilla.redhat.com/show_bug.cgi?id=1933144) is being solved with https://github.com/kubernetes/kubernetes/pull/110039, this PR raises the KubeAggregatedAPIDown `for` duration to 15m instead of 5m.

SNO deployments will see some downtime on upgrades and cert rotations, but this new duration allows for enough time to perform those operations.

### DIFF
--- a/assets/control-plane/prometheus-rule.yaml
+++ b/assets/control-plane/prometheus-rule.yaml
@@ -461,7 +461,7 @@ spec:
         summary: Kubernetes aggregated API is down.
       expr: |
         (1 - max by(name, namespace, cluster)(avg_over_time(aggregator_unavailable_apiservice[10m]))) * 100 < 85
-      for: 5m
+      for: 15m
       labels:
         severity: warning
     - alert: KubeAPIDown
--- a/jsonnet/utils/sanitize-rules.libsonnet
+++ b/jsonnet/utils/sanitize-rules.libsonnet
@@ -252,6 +252,15 @@ local patchedRules = [
     ],
   },
   {
+    name: 'kubernetes-system-apiserver',
+    rules: [
+      {
+        alert: 'KubeAggregatedAPIDown',
+        'for': '15m',
+      },
+    ],
+  },
+  {
     name: 'prometheus',
     rules: [
       {


pkg/manifests/manifests.go
```diff
@@ -2878,6 +2878,39 @@ func (f *Factory) ControlPlanePrometheusRule() (*monv1.PrometheusRule, error) {
 		}
 		r.Spec.Groups = groups
 	}
+	if !f.infrastructure.HighlyAvailableInfrastructure() {
```
omertuc, line 2881: the control plane topology is also interesting and should be `||` in this condition, I think some of the aggregated APIs are scheduled on the control plane nodes

pkg/manifests/manifests.go
```diff
@@ -2878,6 +2878,39 @@ func (f *Factory) ControlPlanePrometheusRule() (*monv1.PrometheusRule, error) {
...
+			if group.Name == "kubernetes-system-apiserver" {
+				for j := range group.Rules {
+					rule := &group.Rules[j]
+					if rule.Alert == "KubeAggregatedAPIDown" {
+						rule.Labels["severity"] = "info"
+						break
+					}
+				}
+				group.Rules = append(group.Rules, monv1.Rule{
+					Alert: "KubeAggregatedAPIUnavailable",
```
omertuc, line 2899: `s/Unavailable/Offline`

omertuc, line 2899:
```suggestion
					Alert: "KubeAggregatedAPIOffline",
```

pkg/manifests/manifests.go
```diff
@@ -2878,6 +2878,39 @@ func (f *Factory) ControlPlanePrometheusRule() (*monv1.PrometheusRule, error) {
...
+					Annotations: map[string]string{
+						"description": `Kubernetes aggregated API {{ $labels.name }}
+/{{ $labels.namespace }} has been only {{ $value | humanize }}% available over the
+last 20m.`},
+					Expr: intstr.FromString("(1 - max by(name, namespace)(avg_over_time(aggregator_unavailable_apiservice[20m]))) * 100 < 85"),
```
omertuc, line 2904: When prometheus first starts collecting metrics on SNO, it sometimes begins while some of the aggregated APIs are still down for a brief period. So the metrics look something like this:

```
<blank>-<blank>-<blank>-<blank>-<blank>-<blank>-<blank>-<blank>-<down>-<down>-<up>-<up>-<up>-...
```

If:
1) blanks are considered "down" by this expression, then we're surely gonna trigger this alert
2) blanks are ignored by this expression and the window is just cut short, then we're also surely gonna trigger this alert because in the first 5m the window is growing from 0m to 5m and the spike of <down>s we have at the beginning is likely gonna make up a >85% portion of it

In case of (1) -
Is there anything we can do in this query to pretend that `<blank>`s are actually `<up>`?

In case of (2) - Can we make sure that the expression is only true when there's actually enough data to fill a 20m window?

simonpasquier (commented):
Ideally we'd want a BZ to track this work.

In general I avoid tweaking the static manifests in the operator code as it is harder to track down where the rule comes from. But I'm not sure how we can do otherwise in this case.

Another point for discusion (not really specific to this PR but important for the future) is whether the alert shouldn't be owned by cluster-kube-apiserver-operator eventually? cc @dgrisonnet if you have any input on this.

pkg/manifests/manifests.go
simonpasquier: Another option would be keep the existing `KubeAggregatedAPIDown` but extend the for duration in case of SNO.

pkg/manifests/manifests.go
pacevedom: Added a new function to check for this.

pkg/manifests/manifests.go
pacevedom: @omertuc Done.
@simonpasquier Would that be acceptable for a multi node deployment? We would be changing the behavior of this alert for already existing releases. Problem with SNO is that downtime is expected, so you know it's going to happen and that's normal behavior. In multi node this is not the case.

dgrisonnet:
Yeah, that is what I was also thinking @simonpasquier, this alert should definitely be moved to the cluster-kube-apiserver-operator.

> To overcome this, the KubeAggregatedAPIDown alert is not a warning anymore (now info, we want to get notified but not take any measures yet) as it is expected to happen, and a new KubeAggregatedAPIUnavailable alert is added to catch long-lasting downtime anomalies.

I don't think that is the correct way to solve the problem. An aggregated API being is very problematic no matter the setup of the cluster. In OpenShift we have a lot of aggregated APIs and some have very important roles. For example, you have the oauth-apiserver responsible for the authentications in the cluster, the prometheus-adapter responsible for autoscaling, the openshift-apiserver responsible for widely used resources, .... . If any of these are down, this is a problem, and the admin will want to be notified. However, if the normal alert is too noisy for SNO, instead of updating its severity you could update its `for` clause to be more resilient to downtime.

Today in OpenShift this alert can fire after 6m30 to 15m of disruption. Maybe this could be extended to 30m or even an hour in SNO case. Warning alerts are not meant to be prioritized right away by the SREs anyways, so waiting a bit more before making the alert fire isn't really a problem.

pkg/manifests/manifests.go
pacevedom: @simonpasquier misunderstood, I actually like this idea better. Thanks!

pacevedom:
/hold
Testing some different numbers for `for`

pacevedom:
Thanks for the comments.

I finally increased the `for` to 15m, as empirical measures for failed tests show that these typically take up to (or slightly over) 10m.

pkg/manifests/manifests.go
```diff
@@ -2898,6 +2899,26 @@ func (f *Factory) ControlPlanePrometheusRule() (*monv1.PrometheusRule, error) {
 		}
 		r.Spec.Groups = groups
 	}
+	if !f.infrastructure.HighlyAvailableInfrastructure() || !f.infrastructure.HighlyAvailableControlPlane() {
+		// We are in SNO. Aggregated API may go down at times since there is no high
```
omertuc, line 2903: >// We are in SNO.

nit: non-highly-available infrastructure doesn't imply SNO

omertuc, line 2903:
```suggestion
		// Aggregated API may go down at times since there is no high
```

pacevedom: Committed your suggestion, thanks!

pkg/manifests/manifests.go
```diff
@@ -2898,6 +2899,26 @@ func (f *Factory) ControlPlanePrometheusRule() (*monv1.PrometheusRule, error) {
...
+				for j := range group.Rules {
+					rule := &group.Rules[j]
+					if rule.Alert == "KubeAggregatedAPIDown" {
+						rule.For = "15m"
+						break
+					}
+				}
+				break
+			}
+		}
```
dgrisonnet, line 2920: @simonpasquier I don't think we need this change to be SNO specific, waiting 15 minutes instead of 5 for a warning level alert should be fine for all use-cases.

If I remember correctly, in the past people from TRT/apiserver wanted this alert to be really sensitive in order to catch issues in CI, but I don't think that is the correct direction to go with. Alerts and tests should be dissociated and that's what we've tried to start doing more in origin since CI testing is relying a lot on alerts which is preventing everyone from making them more resilient and less noisy in general. So I would be in favor of making this alert more resilient in general and adding a new test in origin that would perform a more sensitive check.

dgrisonnet (commented):
@simonpasquier I will move this alert to the cluster-kube-apiserver-operator once this change is done.

pkg/manifests/manifests.go
pacevedom: Do we know how SRE's are responding to this alert? Do we have any SLO? For SNO we can expect downtime since there is no high availability, hence you need to be a bit more tolerant, but not so sure about multi node. Is there a requirement for uptime?
If you need help for the CI count on me.

simonpasquier: I would ask @openshift/sre-alert-sme how they feel about the `for` duration being extended to 15 minutes from 5 minutes for the `KubeAggregatedAPIDown` alert.
And ideally this alert (as well as `KubeAggregatedAPIErrors`) should be moved to the cluster-kube-apiserver-operator project.

dofinn: No issues from me expanding this to 15m for a warning alert. Given the amount of alerts around the apiserver - this is contextual as a warning.

dgrisonnet: Sounds like we have an agreement that setting `for` to 15m for the general case is fine.

dgrisonnet:
Since we've come to the agreement that the `for` clause needed to be increased, I am putting a hold on this PR until a test is added to origin to still require the same amount of availability for aggregated APIs in CI. I will look into that in the upcoming days.

/hold

pacevedom:
A quick update:
- This [upstream PR](https://github.com/kubernetes/kubernetes/pull/110039) was created to fix the wardle API flakes.
- This [downstream PR](url) picked up on it and is finally merged.
- Need to wait until https://github.com/openshift/origin/pull/27380 is merged
- This PR is finally moving the `for` to 15m as agreed before.

eggfoobar:
@dgrisonnet @simonpasquier Can we get another look at this PR with the latest updates?

dgrisonnet:
I am fine with this change.

/lgtm

dgrisonnet:
Unholding since I created https://issues.redhat.com/browse/API-1482 to take care of adding a test in origin and I don't think that the fact that CI relies heavily on alerts should prevent us from making alerts more resilient.

/unhold

jan--f:
/retitle Bug 1933144: Extend KubeAggregatedAPIDown alert "for" to 15m

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - [#1467](https://github.com/openshift/cluster-monitoring-operator/pull/1467) Add bodysize limit for metric scraping
 - [#1661](https://github.com/openshift/cluster-monitoring-operator/pull/1661) Support deployment of dedicated Alertmanager for user-defined alerts.
 - [#1682](https://github.com/openshift/cluster-monitoring-operator/pull/1682) Support AlertmanagerConfig v1beta1.
+- [#1639](https://github.com/openshift/cluster-monitoring-operator/pull/1639) Increase "for" duration to 15min in KubeAggregatedAPIDown alert.
```
jan--f, line 24: Let remove the changelog entry, we usually track feature addition here. The bug fix will be added to the release notes by the doc team.

jan--f (changes requested):
One nit, other this looks good.

raptorsun (approved):

jan--f (approved):
/lgtm

PR #1646, opened 2022-04-21
Bug 2057832: Updates recording rule cluster:telemetry_selected_series:count

JoaoBraveCoding:
Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2057832

Problem: Current recording rule expression isn't completely accurate since it may count more series than what is effectively sent to telemeter.

Solution: Use instead metrics from telemeter-client which capture
exactly the number of samples being sent.

Relates to:
https://github.com/openshift/telemeter/pull/411
https://github.com/openshift/cluster-monitoring-operator/pull/1647

### DIFF
--- a/assets/telemeter-client/prometheus-rule.yaml
+++ b/assets/telemeter-client/prometheus-rule.yaml
@@ -1,0 +1,11 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: telemetry
+  namespace: openshift-monitoring
+spec:
+  groups:
+  - name: telemeter.rules
+    rules:
+    - expr: max(federate_samples - federate_filtered_samples)
+      record: cluster:telemetry_selected_series:count
--- a/jsonnet/components/telemeter-client.libsonnet
+++ b/jsonnet/components/telemeter-client.libsonnet
@@ -21,6 +21,7 @@ function(params) {
   },
 
   // Remapping everything as this is the only way I could think of without refactoring imported library
+  prometheusRule: tc.telemeterClient.prometheusRule,
   clusterRoleBindingView: tc.telemeterClient.clusterRoleBindingView,
   clusterRoleBinding: tc.telemeterClient.clusterRoleBinding,
   clusterRole: tc.telemeterClient.clusterRole,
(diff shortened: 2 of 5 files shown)


JoaoBraveCoding:
@simonpasquier I'm not sure if I should also remove almost all instances of the variable `telemetryMatches` in the operator apart from the assignment to the Prometheus CR and I'm also not sure if the test `TestGenerateTelemeterWhitelistRec` makes sense or not to have still.

Documentation/sample-metrics.md
```diff
@@ -13,7 +13,7 @@ return the full set of metrics that the Telemeter client captures:
...
+max(federate_samples - federate_filtered_samples)
```
juzhao, line 16: the metrics only show the total number of the metrics, don't know what metrics are collected by the Telemeter service
```
# token=`oc sa get-token prometheus-k8s -n openshift-monitoring`
# oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- curl -k -H "Authorization: Bearer $token" 'https://thanos-querier.openshift-monitoring.svc:9091/api/v1/query?' --data-urlencode 'query=max(federate_samples - federate_filtered_samples)' | jq
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {},
        "value": [
          1653271388.85,
          "520"
        ]
      }
    ]
  }
}
```
https://github.com/openshift/cluster-monitoring-operator/blob/39a4e00c94de39d436f2a2a605e06c892a2fe809/Documentation/sample-metrics.md

JoaoBraveCoding: Yes, I originally updated this wrongfully.

simonpasquier:
/lgtm
/hold

IIUC during upgrade, the old prometheusrule object that was computed dynamically will be replaced by the "static" version because both have the same name? Feel free to unhold if my assumption is correct :)

JoaoBraveCoding:
> /lgtm /hold
> 
> IIUC during upgrade, the old prometheusrule object that was computed dynamically will be replaced by the "static" version because both have the same name? Feel free to unhold if my assumption is correct :)

Correct, that is my expectation!
/unhold

PR #1652, opened 2022-04-29
MON-2480: Double the ServiceMonitor intervals for SNO

tremes:
This is the first PR for the MON-2384. This automatically doubles the ServiceMonitor interval values in non-highavailable infrastructure.

### DIFF
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/openshift/library-go/pkg/crypto"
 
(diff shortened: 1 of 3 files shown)


jan--f:
/retitle MON-2480: Double the ServiceMonitor intervals for SNO

simonpasquier (commented):
I'm wondering if we shouldn't have an upper limit? The max recommended scrape interval is 2m (to deal with failed scrapes and the 5m staleness window) and it probably makes sense to ensure that CMO doesn't increase the scrape interval beyond this value.

tremes:
@simonpasquier Yes I think it's a good idea so do you suggest to limit the max value of the interval to 5m (or 2m?)?

jan--f:
> @simonpasquier Yes I think it's a good idea so do you suggest to limit the max value of the interval to 5m (or 2m?)?

Lets set 2m as the upper limit.

pkg/manifests/manifests_test.go
```diff
@@ -3292,6 +3293,153 @@ func TestNonHighlyAvailableInfrastructure(t *testing.T) {
...
+				t.Errorf("Unexpected error when parsing %s: %v", noHAEndpoints[i].Interval, err)
+			}
+			HAInt, err := time.ParseDuration(HAEndpoints[i].Interval)
+			if err != nil {
+				t.Errorf("Unexpected error when parsing %s: %v", HAEndpoints[i].Interval, err)
+			}
+
+			if HAInt*2 >= 2*time.Minute {
+				if noHAInt != 2*time.Minute {
+					t.Errorf("Unexpected value. %d could be max 2 minutes", noHAInt)
```
jan--f, line 3434:
```suggestion
					t.Errorf("Unexpected value. %d should be max 2 minutes", noHAInt)
```
Just a nit.

jan--f (commented):
One more nit. And could you please a an entry to CHANGELOG.md? Something along the lines of `Double scrape interval for all CMO controlled ServiceMonitors on single node deployments` and a link to the respective [epic](https://issues.redhat.com/browse/MON-2384).

jan--f:
/label docs-approved
/label px-approved
Adding approvals according to epic.
Ping @juzhao for qe approval.

juzhao:
@jan--f tested with PR, only one concern
 I see `scrape_interval: 1m0s` and  `scrape_interval: 2m0s`， `scrape_interval: 1m` and  `scrape_interval: 2m` are fine
```
# oc get node
NAME                                 STATUS   ROLES           AGE   VERSION
ci-ln-nsqz6zk-72292-kqffg-master-0   Ready    master,worker   98m   v1.24.0+bb9c2f1

# oc -n openshift-monitoring get servicemonitor alertmanager-main -oyaml
...
spec:
  endpoints:
  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
    bearerTokenSecret:
      key: ""
    interval: 1m0s
...
# oc -n openshift-monitoring get servicemonitor openshift-state-metrics -oyaml
...
spec:
  endpoints:
  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
    bearerTokenSecret:
      key: ""
    honorLabels: true
    interval: 2m0s
    port: https-main
    scheme: https
    scrapeTimeout: 2m
...
  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
    bearerTokenSecret:
      key: ""
    interval: 2m0s
    port: https-self
    scheme: https
    scrapeTimeout: 2m
```
see all targets in cluster, http://pastebin.test.redhat.com/1056321

jan--f:
I think an intervals like `1m0s` should be fine. This is what `time.Duration.String()` returns and it is a valid value for the `Interval` field in an `Endpoint`. It is specifically allowed by the [validation pattern for this field](https://github.com/prometheus-operator/prometheus-operator/blob/a346e572072ae4949d1212586c50aef0ab1e29e5/pkg/apis/monitoring/v1/types.go#L384-L385) (note the `(([0-9]+)s)?` part).

I agree that it seems superfluous, but I'd rather stick with what the go standard library produces here instead of massaging the string the the minimal version.
@juzhao wdyt?

PR #1655, opened 2022-05-02
WIP Bug 2069068: Refactors CreateUpdateDeployment to use controllerruntime util

JoaoBraveCoding:
Issue https://bugzilla.redhat.com/show_bug.cgi?id=2069068

problem: when reconciling the prometheus deployment sometimes the
operator would get an API error and it would delete and create
a deployment in a short span of time without giving time to the API
server to reconcile.

solution: use common library to CreateOrUpdate deployment instead of
trying to do it ourselves


pkg/client/client.go
```diff
@@ -930,6 +931,29 @@ func (c *Client) WaitForDeploymentRollout(ctx context.Context, dep *appsv1.Deplo
...
+		if err != nil && !apierrors.IsNotFound(err) {
+			klog.V(4).ErrorS(err, "WaitForDeploymentDeletion: failed to get Deployment")
+			return false, nil
+		}
+		if d.Status.AvailableReplicas != 0 {
+			lastErr = errors.Errorf("got %d available replicas",
+				d.Status.AvailableReplicas)
+			return false, nil
+		}
+		return true, nil
```
simonpasquier, line 947: It could be easier to read if we checked first that the returned error is "not found"?

```suggestion
		if !apierrors.IsNotFound(err) {
			return true, nil
		}
		
		if err != nil {
			lastErr = err
		} else {
			lastErr = errors.Errorf("got %d available replicas",
				d.Status.AvailableReplicas)
		}
		
		return false, nil
```

simonpasquier: no strong feeling for me. By the way my initial suggestion was buggy! I've modified it to be more in-lined with what you mentioned.

JoaoBraveCoding: Yes, I also noticed ;) This way is better I'll update

simonpasquier (commented):
/lgtm

PR #1640, opened 2022-04-13
*: add standalone admission webhook

simonpasquier:
Before this commit, the validating webhook configurations for
PrometheusRule and AlertmanagerConfig resources leveraged the webhook
service of Prometheus operator. Because the operator doesn't implement
any leader election, it has to run as a single replica meaning that the
webhook service couldn't be highly-available (especially during
upgrades).

This change deploys the admission webhook as a standalone service with 2
replicas + hard anti-affinity + pod disruption budget. A follow-up PR
will switch the failure policy of the validating webhook configurations
for PrometheusRule and AlertmanagerConfig resources from 'Ignore' to
'Fail'.

JIRA ref: https://issues.redhat.com/browse/MON-2221

* [x] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #1659, opened 2022-05-05
jsonnet: switch webhook failure policy to 'Fail'

simonpasquier:
After #1640, we can change the failure policy for admission webhooks
from 'Ignore' to 'Fail' since the webhook service is running with 2
replicas + anti-affinity + pod disruption budget.


simonpasquier:
/hold

Based on https://bugzilla.redhat.com/show_bug.cgi?id=2088842 and https://bugzilla.redhat.com/show_bug.cgi?id=1949711, using the `Fail` policy would prevent CVO from recreating the monitoring objects after the cluster admin deletes the `openshift-monitoring` namespace. We need to investigate how we can workaround this issue: ownerReferences aren't an option because webhook configuration aren't namespaced objects so we're probably left with finalizers.

PR #1660, opened 2022-05-05
Bug 2079292: Adds necessary SecurityContext settings to UWM deployments

JoaoBraveCoding:
Issue https://bugzilla.redhat.com/show_bug.cgi?id=2079292

Problem: In Kubernetes v1.23 the PodSecurity feature is enabled by
default, OpenShift is making use of this feature and by default some
expectations are already set [1], these expectations end up
triggering a lot of alarms in the logs of CMO and workload running
on the UWM namespace.
[1] https://github.com/openshift/cluster-kube-apiserver-operator/pull/1308

Solution: Add PodSecurity namespace labels to allow workloads running
on UWM  to have full privileges which will silence the alerts


simonpasquier:
/skip
/hold
/lgtm

pinging @s-urbaniak for a confirmation that it's the correct approach. IIUC https://github.com/openshift/cluster-monitoring-operator/pull/1413 allowed to run privileged pods but only in the `openshift-monitoring` namespace. Extending the permission to the `openshift-user-workload-monitoring` namespace should be ok.

s-urbaniak:
@jan--f @simonpasquier the only thing you should consider whether `privileged` is really needed. your failure shows:

```
W0427 09:52:28.128970       1 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (containers "prometheus-operator", "kube-rbac-proxy" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (containers "prometheus-operator", "kube-rbac-proxy" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or containers "prometheus-operator", "kube-rbac-proxy" must set securityContext.runAsNonRoot=true), seccompProfile (pod or containers "prometheus-operator", "kube-rbac-proxy" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

It might be enough to annotate the manifests like in https://github.com/openshift/cluster-samples-operator/pull/425. That way your namespace can stay restricted.

In doubt you can use this tool https://github.com/stlaz/psachecker to evaluate your workloads.

jan--f:
/hold
might be worth considering Sergs comment, but could also be done in a follow up.

simonpasquier:
thanks for the details @s-urbaniak I agree that if we can make `openshift-user-workload-monitoring` comply with the default pod security admission level, it's even better. I see that @JoaoBraveCoding already applied the suggestion so let's see how it goes :)

jan--f:
Looks like the test fails to roll out the UWM prometheus-operator. Maybe some test resources need a change too?

assets/prometheus-user-workload/prometheus.yaml
```diff
@@ -211,6 +231,8 @@ spec:
     fsGroup: 65534
     runAsNonRoot: true
     runAsUser: 65534
+    seccompProfile:
```
JoaoBraveCoding, line 234: @simonpasquier @jan--f So after a lot of debugging I've concluded that this is the problematic part. If we set a `seccompProfile` then we cannot set `runAsUser` with `65534` since "`must be in the ranges: [1000440000, 1000449999]`". I see this error in the event of the UWM namespace with an error for a failed creation of a pod. 

I thought about also removing the `runAsUser` but from what I understood in [this comment](https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/components/prometheus-user-workload.libsonnet#L139-L146), it's necessary and cannot be removed.

Sooo now I'm thinking we should either set the Pod Security Admission labels to `baseline` on the namespace (`baseline` does not enforce `seccompProfile`). Or use a User in the range??

jan--f: Afaiu there nothing technical special about the `65534` uid, its just a convention that typically means `nobody`. So using a uid in the required range would be the ideal solution imho.
I might be missing some history here though, so curious to see what @simonpasquier can add here.

simonpasquier: I'd be wary to change this setting. The full history behind it is in https://bugzilla.redhat.com/show_bug.cgi?id=1868976 and https://github.com/openshift/cluster-monitoring-operator/pull/981 but the TL;DR is that to be immune to cluster admins adding new SCCs, workloads running with persistent storage need to be associated with the `nonroot` SCC so they get assigned predictable uid and fsGroup values. I would assume that `baseline` is acceptable since monitoring is a core OCP component.

JoaoBraveCoding: After some more debugging I found out that everything works correctly if we use `nonroot-v2` instead of `nonroot`. I'm not quite sure why k8s is not taking into account `nonroot` however...

simonpasquier: `nonroot-v2` is something new to me. Can you shed some light on how it differs from `nonroot`? 

> Now my question is, is it supported by OCP that a user changes the restricted SCC? Couldn't the user use instead the nonroot SCC?

IIRC though it's strongly advised not to modify the default SCCs [1] but the issue was also happening when users created new SCCs (check [comment 34](https://bugzilla.redhat.com/show_bug.cgi?id=1868976#c34)).

[1] https://docs.openshift.com/container-platform/4.10/authentication/managing-security-context-constraints.html#default-sccs_configuring-internal-oauth

simonpasquier (commented):
/lgtm

jsonnet/components/prometheus-user-workload.libsonnet
```diff
@@ -142,7 +142,7 @@ function(params)
           // to be immune against UID changes.
           apiGroups: ['security.openshift.io'],
           resources: ['securitycontextconstraints'],
-          resourceNames: ['nonroot'],
+          resourceNames: ['nonroot-v2'],
```
simonpasquier, line 145: maybe update the comment above and explain a bit about why we use `nonroot-v2` rather than `nonroot`?

JoaoBraveCoding: Yes, 100% makes sense, I've added some lines to make it more clear

PR #1662, opened 2022-05-06
Adding a usage metric for Openshift Sandboxed Containers 

littlejawa:
Adding a metric from the Openshift Sandboxed Container operator.
The goal is to have some usage information on top of installation number.


littlejawa:
> /skip do you have somewhere (JIRA, Google doc) the approval for the new metric?

Yes. We have it tracked under https://issues.redhat.com/browse/KATA-1392
This is a broader request. This metric is focusing on the first aspect of it: provide the number of running sandboxes per install.

Other items might be needed to complete this JIRA, but my investigation, and the exiting metrics we expose, make me think that it would be better achieved using alerts rather than raw metrics, so it will be done on a later release.

simonpasquier:
> Other items might be needed to complete this JIRA, but my investigation, and the exiting metrics we expose, make me think that it would be better achieved using alerts rather than raw metrics, so it will be done on a later release.

FWIW alerts are already pushed to Telemetry.

simonpasquier:
/hold
until https://github.com/openshift/sandboxed-containers-operator/pull/189 merges

littlejawa:
@simonpasquier thanks for reviewing this

https://github.com/openshift/sandboxed-containers-operator/pull/189 is merged now.
I also pushed the corresponding change to the telemetry server config

simonpasquier:
/label px-approved
/label qe-approved
/label docs-approved

Approved for 1 new metric.

PR #1664, opened 2022-05-10
Bug 2009352: IR-254: Collecting registry and image stream usage

ricardomaraschini:
As an effort to evaluate how often customers are using Image Streams and
the internal registry this commit introduces two new metrics to be
collected:

- The amount of Image Stream Tags in the cluster
- The amount of pushes and pulls executed against the registry


ricardomaraschini:
For team validation:

/assign @dmage @flavianmissi

ricardomaraschini:
Opened a new PR for this: https://github.com/openshift/cluster-monitoring-operator/pull/1668

PR #1665, opened 2022-05-10
Bug 2077722: Adjust NodeFilesystemSpaceFillingUp thresholds according default kubelet GC behavior

arajkumar:
Previously[1] we attempted to do the same, but there was a
misunderstanding about the GC behavior and it caused the alert to be
fired even before GC comes into play.

According to[2][3] kubelet GC kicks in only when imageGCHighThresholdPercent is hit which is set to 85% by default. However NodeFilesystemSpaceFillingUp is set to fire as soon as 80% usage is hit.

This commit changes the fsSpaceFillingUpWarningThreshold to 15% so
that we give ample time to GC to reclaim unwanted images. This commit
also changes fsSpaceFillingUpCriticalThreshold to 10% which gives more time to admins to react to warning before sending critical alert.

[1] https://github.com/prometheus-operator/kube-prometheus/pull/1357
[2] https://docs.openshift.com/container-platform/4.10/nodes/nodes/nodes-nodes-garbage-collection.html#nodes-nodes-garbage-collection-images_nodes-nodes-configuring
[3] https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2077722

### DIFF
--- a/assets/node-exporter/prometheus-rule.yaml
+++ b/assets/node-exporter/prometheus-rule.yaml
@@ -23,7 +23,7 @@ spec:
         summary: Filesystem is predicted to run out of space within the next 24 hours.
       expr: |
         (
-          node_filesystem_avail_bytes{job="node-exporter",fstype!=""} / node_filesystem_size_bytes{job="node-exporter",fstype!=""} * 100 < 20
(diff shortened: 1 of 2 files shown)


jan--f:
/lgtm
/backport-risk-assessed

Thanks Arun!

arajkumar:
@jan--f Can you add backport-risk-assessed label? Thanks

PR #1671, opened 2022-05-17
Bug 2084079: Refactors CreateRouteIfNotExists to CreateOrUpdateRoute

JoaoBraveCoding:
Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2084079

Problem: on OpenShift 4.11 the Prometheus route was updated with
a subpath /api, CMO was only able to create routes and not update them.
When users would update from a version before 4.11 to 4.11 the route
would not be updated.

Solution: allow CMO to Create or Update Routes similar to how it
interects with other k8s resources.


pkg/client/client.go
```diff
@@ -346,14 +346,27 @@ func (c *Client) CreateOrUpdateSecurityContextConstraints(ctx context.Context, s
...
+	existing, err := rclient.Get(ctx, r.GetName(), metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		_, err := rclient.Create(ctx, r, metav1.CreateOptions{})
 		return errors.Wrap(err, "creating Route object failed")
 	}
-	return nil
+	if err != nil {
+		return errors.Wrap(err, "retrieving Route object failed")
+	}
+	if reflect.DeepEqual(r.Spec, existing.Spec) {
```
simonpasquier, line 359: I think that the `spec.host` will always be different because it's automatically chosen if unset.

simonpasquier (commented):
Great! minor comments only :)

pkg/client/client.go
jan--f: @simonpasquier is [correct](https://github.com/openshift/api/blob/a99ffa1cac6709edf8f502b16890b16f9a557e00/route/v1/types.go#L75-L80).

Dropping this code path probably makes most sense?

simonpasquier (commented):
/lgtm

PR #1672, opened 2022-05-18
Ignore tunbr NodeNetworkInterfaceFlapping alert

jluhrsen:
tunbr is an ephemeral bridge interface that can go up an down depending
on if the ephemeral network is created/deleted, we shouldnt care if that
happens. Same as we dont care if a veth interface goes up/down as that
is going to happen with CNI pod events.

https://issues.redhat.com/browse/SDN-3008

Signed-off-by: Jamo Luhrsen <jluhrsen@gmail.com>


jluhrsen:
this is not right. I think we need a PR in https://github.com/prometheus-operator/kube-prometheus to add
tunbr to the ignore list, wherever that is. then when openshift updates to a release that has that change we can
do a 'make generate' in this repo to get what we want. looking in to that.

PR #1674, opened 2022-05-24
Bug 2089199: Removes etcd related dashboards from CMO

JoaoBraveCoding:
Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2089199

Problem: Currently CMO deploys a set of dashboards that report
information about etcd, these however do not make sense if etcd is not
deployed in the cluster (hypershift).

Solution: Move these dashboard to the etcd operator repo so that these
dashboards are only deployed to the cluster when etcd itself is running
there.


jsonnet/components/grafana.libsonnet
```diff
@@ -17,25 +17,6 @@ function(params)
 
   grafana(cfg) {
 
-    consoleDashboardDefinitions: {
```
simonpasquier, line 20: we don't want to drop all dashboards but only the etcd one. Instead of `std.map()` we need [`std.filterMap()`](https://jsonnet.org/ref/stdlib.html#filterMap).

JoaoBraveCoding: Indeed! Should be good now

simonpasquier (commented):
/lgtm
/skip
/hold

holding it until https://github.com/openshift/cluster-etcd-operator/pull/837 gets approved too.

Elbehery:
@simonpasquier @JoaoBraveCoding  any updates here ? 

Shall I remove the hold from https://github.com/openshift/cluster-etcd-operator/pull/837

JoaoBraveCoding:
Hello @Elbehery, yes both PR look OK so from my side we can try and merge both today. We will merge this one as soon as the etcd one is merged.
Pinging @simonpasquier  in case he wants to add something.

simonpasquier:
@Elbehery @JoaoBraveCoding the plan works for me.

PR #1679, opened 2022-05-26
Bug 2089574: Removes master node selector from PO when running with HostedControlPlane external

JoaoBraveCoding:
Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2089574

Problem: When UWM is run with HostedControlPlane external (HyperShift)
this makes it so the prometheus operator pods cannot start because of
the default node selector we set, where PO pods have to run on master

Solution: Add new default nodeSelector for UWM Prometheus Operator
when HostedControlPlane is external

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -58,6 +58,8 @@ const (
 	clientCAArg = "--client-ca-file=/etc/tls/client/client-ca.crt"
 
 	tmpClusterIDLabelName = "__tmp_openshift_cluster_id__"
+
+	nodeSelectorMaster = "node-role.kubernetes.io/master"
 )
 
 var (
@@ -2384,7 +2386,6 @@ func (f *Factory) PrometheusOperatorDeployment() (*appsv1.Deployment, error) {
 	}
 
 	if len(f.config.ClusterMonitoringConfiguration.PrometheusOperatorConfig.NodeSelector) > 0 {
-
 		d.Spec.Template.Spec.NodeSelector = f.config.ClusterMonitoringConfiguration.PrometheusOperatorConfig.NodeSelector
 	}
 
@@ -3378,6 +3379,10 @@ func (f *Factory) NewDeployment(manifest io.Reader) (*appsv1.Deployment, error) 
 
 	if d.GetNamespace() == "" {
 		d.SetNamespace(f.namespace)
+	}
+
+	if f.infrastructure.HostedControlPlane() {
+		delete(d.Spec.Template.Spec.NodeSelector, nodeSelectorMaster)
 	}
 
 	if !f.infrastructure.HighlyAvailableInfrastructure() {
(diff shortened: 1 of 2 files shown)


csrwng:
Thank you @JoaoBraveCoding !
We should do the same for the PrometheusOperator deployment:
https://github.com/openshift/cluster-monitoring-operator/pull/1679/files#diff-1f9d2d18d4cabd8262e585b932330183f7cb9b544ac9f27c8bcb3ef9c8d14376L2386
(either here or in a follow-up).
With that, we can stop reconciling a config into the guest cluster in the hypershift code. This will preserve the configuration strictly as something that's entirely user-driven just as it is in standalone ocp.

JoaoBraveCoding:
@csrwng I'll add the missing code to do that however if I'm not missing anything I think you would still have to reconcile the ConfigMap just to enable user workload monitoring, no?

csrwng:
We don't need to enable uwm by default. It's up to the user of the cluster if they want to enable it, just as in standalone OCP.

pkg/manifests/manifests.go
```diff
@@ -2384,12 +2384,13 @@ func (f *Factory) PrometheusOperatorDeployment() (*appsv1.Deployment, error) {
 	}
 
 	if len(f.config.ClusterMonitoringConfiguration.PrometheusOperatorConfig.NodeSelector) > 0 {
-
 		d.Spec.Template.Spec.NodeSelector = f.config.ClusterMonitoringConfiguration.PrometheusOperatorConfig.NodeSelector
 	}
 
 	if len(f.config.ClusterMonitoringConfiguration.PrometheusOperatorConfig.Tolerations) > 0 {
 		d.Spec.Template.Spec.Tolerations = f.config.ClusterMonitoringConfiguration.PrometheusOperatorConfig.Tolerations
+	} else if f.infrastructure.HostedControlPlane() {
```
csrwng, line 2392: This should go in the previous if block (this if is for tolerations)

simonpasquier (commented):
To be a bit more generic, I'd recommend to modify `NewDeployment()` instead and remove the `node-role.kubernetes.io/master` key from the node selector if/when the control place is hosted.

pkg/manifests/manifests.go
```diff
@@ -3380,6 +3379,10 @@ func (f *Factory) NewDeployment(manifest io.Reader) (*appsv1.Deployment, error)
 		d.SetNamespace(f.namespace)
 	}
 
+	if f.infrastructure.HostedControlPlane() {
+		delete(d.Spec.Template.Spec.NodeSelector, "node-role.kubernetes.io/master")
```
simonpasquier, line 3383: (nit) maybe declare a `const`?

pkg/manifests/manifests_test.go
```diff
@@ -3622,6 +3622,38 @@ enableUserWorkload: true
 	}
 }
 
+func TestPrometheusOperatorConfigurationHyperShift(t *testing.T) {
```
simonpasquier, line 3625: Not mandatory but a single test function with table-driven test cases would be neat :)
We could also check that setups without hosted control plane preserve the node selector.

simonpasquier (commented):
/lgtm

PR #1680, opened 2022-05-26
control plane: ignore flapping host interface 'tunbr'

dcbw:
It's an ephemeral interface like veths are; it's not a host NIC.
Thus it's not something we should be tracking as a host level
problem as it's managed by other components like secondary
network plugins.

https://issues.redhat.com/browse/SDN-3008

Replaces https://github.com/openshift/origin/pull/27157

### DIFF
--- a/jsonnet/main.jsonnet
+++ b/jsonnet/main.jsonnet
@@ -342,7 +342,7 @@ local inCluster =
           ruleLabels: $.values.common.ruleLabels,
           _config+: {
             diskDeviceSelector: $.values.nodeExporter.mixin._config.diskDeviceSelector,
-            hostNetworkInterfaceSelector: 'device!~"veth.+"',
+            hostNetworkInterfaceSelector: 'device!~"veth.+|tunbr"',
             kubeSchedulerSelector: 'job="scheduler"',
             namespaceSelector: $.values.common.mixinNamespaceSelector,
             cpuThrottlingSelector: $.values.common.mixinNamespaceSelector,


jluhrsen:
I made the same comment [here](https://github.com/openshift/origin/pull/27157#issuecomment-1138771095), but wondering how the [prom rules yaml file](https://github.com/openshift/cluster-monitoring-operator/blob/6f8b1dc6d572959174a65b277dd3cc8feb8216db/assets/cluster-monitoring-operator/prometheus-rule.yaml#L431) comes in to play here? does this tunbr not
need to make it there?

JoaoBraveCoding:
@jluhrsen @dcbw There is actually a bug in the jsonnet I've opened a PR to fix it https://github.com/prometheus-operator/kube-prometheus/pull/1771

dcbw:
@JoaoBraveCoding yeah, just figured out it's pulling defaults from `jsonnet/vendor/github.com/prometheus-operator/kube-prometheus/jsonnet/kube-prometheus/components/mixin/custom.libsonnet` and not from our `main.jsonnet`.

Is your PR there the fix for that?

dcbw:
@JoaoBraveCoding patching your fix into the vendored custom.jsonnet doesn't make it use stuff from `main.jsonnet`; but I don't know anything about this so maybe I'm just Doing It Wrong.

JoaoBraveCoding:
@dcbw yes... seems I was wrong... 😖

jluhrsen:
@JoaoBraveCoding @dcbw , so is my theory right that since we need something in prom-operator/kube-prom before we can get it working in tihs repo it will take some time?

If so, we can push to get [this PR](https://github.com/openshift/origin/pull/27157) merged so we don't fail in CI for now. I'll keep a jira open assigned to revert it once these other pieces get done.

juzhao:
I am afraid we also need to update the expr in prometheusrules cluster-monitoring-operator-prometheus-rules 
```
# oc -n openshift-monitoring get prometheusrules cluster-monitoring-operator-prometheus-rules -oyaml
...
    - alert: NodeNetworkInterfaceFlapping
      annotations:
        description: Network interface "{{ $labels.device }}" changing its up status
          often on node-exporter {{ $labels.namespace }}/{{ $labels.pod }}
        summary: Network interface is often changing its status
      expr: |
        changes(node_network_up{job="node-exporter",device!~"veth.+"}[2m]) > 2
      for: 2m
      labels:
        severity: warning
```

simonpasquier:
My feeling is that the `hostNetworkInterfaceSelector` field should be passed here:

https://github.com/openshift/cluster-monitoring-operator/blob/6f8b1dc6d572959174a65b277dd3cc8feb8216db/jsonnet/main.jsonnet#L96-L99

The following line can probably be deleted since it has no effect on the generated assets:

https://github.com/openshift/cluster-monitoring-operator/blob/6f8b1dc6d572959174a65b277dd3cc8feb8216db/jsonnet/main.jsonnet#L345

dcbw:
Close in favor of https://github.com/openshift/cluster-monitoring-operator/pull/1681

PR #1684, opened 2022-06-02
Bug 2094704: remove verbose output from kube-rbac-proxy in Prometheus-k8s pod

raptorsun:

When reading must-gather logs I find the logs of the kube-rbac-proxy in prometheus-k8s pods are too verbose for production use, so I suggest we remove the --v=10 argument for it.


simonpasquier:
/lgtm

Sounds legit to me :) 
Can you create a BZ for this?

PR #1685, opened 2022-06-03
Bug 2037513: Revert dropping container label and update diskDeviceSelector

slashpai:
**Why this change?**

container_fs_.* metrics doesn't have "container" label after
https://github.com/openshift/cluster-monitoring-operator/pull/1402 which caused regression
for dashboards using `container` label since it gives empty datapoints
for all queries using "container" label in queries containing
container_fs_.* metrices.

Had added deviceDeviceSelector in container_fs* queries as part of
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/737 since diskDeviceSelector
includes only node_exporter label values updated disDeviceSelector
to include cadvisor devices too (/dev)

**Tried alternative solution but didnt work**

We tried to fix this by introducing new selector in kubernetes-mixin and overriding that in cmo in https://github.com/openshift/cluster-monitoring-operator/pull/1554 but that won't solve the problem since many queries  use  `container` label in [`sum()` queries](https://github.com/openshift/cluster-monitoring-operator/pull/1554/files#diff-40e9476ea123a816ecc6418eb0bba6bf1527808cb904bd42f5bbfb58e1e30342R12337)

cc: @simonpasquier @jan--f 

/hold

### DIFF
--- a/assets/control-plane/service-monitor-kubelet.yaml
+++ b/assets/control-plane/service-monitor-kubelet.yaml
@@ -86,12 +86,6 @@ spec:
       sourceLabels:
       - __name__
       targetLabel: __tmp_keep_metric
-    - action: drop
-      regex: ;(container_fs_.*);.+
-      sourceLabels:
-      - __tmp_keep_metric
-      - __name__
-      - container
     - action: labeldrop
       regex: __tmp_keep_metric
     path: /metrics/cadvisor
--- a/assets/grafana/dashboard-definitions.yaml
+++ b/assets/grafana/dashboard-definitions.yaml
@@ -5767,7 +5767,7 @@ items:
                           "steppedLine": false,
                           "targets": [
                               {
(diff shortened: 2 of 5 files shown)


simonpasquier:
The goal of #1402 was to contain the label cardinality of metrics collected from kubelet. Reverting it would increase the load on Prometheus again...

simonpasquier:
one solution might be to pass `containerfsSelector: 'container=""'` to the mixins configuration?

slashpai:
> one solution might be to pass `containerfsSelector: 'container=""'` to the mixins configuration?

I tried with overriding containerfsSelector in https://github.com/openshift/cluster-monitoring-operator/pull/1554 by replacing it with `id~=""`. It fixed dashboards except *"Kubernetes / Compute Resources / Pod"* since it uses `container` in [sum() query](https://bugzilla.redhat.com/show_bug.cgi?id=2037513#c15)

simonpasquier:
Since Prometheus doesn't collect the per-container fs metrics, I would be in favor of dropping the per-container storage dashboards. @jan--f WDYT?

jan--f:
> Since Prometheus doesn't collect the per-container fs metrics, I would be in favor of dropping the per-container storage dashboards. @jan--f WDYT?

Indeed yes, the dashboard panels just show `No datapoints found`, we might as well get rid of them. cc @kyoto @jgbernalp for visibility.

slashpai:
> > Since Prometheus doesn't collect the per-container fs metrics, I would be in favor of dropping the per-container storage dashboards. @jan--f WDYT?
> 
> Indeed yes, the dashboard panels just show `No datapoints found`, we might as well get rid of them. cc @kyoto @jgbernalp for visibility.

Thank you. I can address that in https://github.com/openshift/cluster-monitoring-operator/pull/1554

jan--f:
/close
The bot doesn't like additional test after commands ;)

PR #1687, opened 2022-06-07
[do not merge] Test performance without verbose log in kube-rbac-proxy in prometheus-k8s pod

raptorsun:


raptorsun:
close this PR since it is the same as:
https://github.com/openshift/cluster-monitoring-operator/pull/1684

PR #1690, opened 2022-06-15
Bug 2100860: Pass user-defined Alertmanager service in shared configmap

simonpasquier:
The address of the Alertmanager's tenant-aware + cluster services is dependent on the user-defined Alertmanager being enabled or not. CMO should provide the service locations to the console operator for proper configuration of the console service.

https://issues.redhat.com/browse/MON-2289

Required by https://github.com/openshift/console-operator/pull/658
Required by https://github.com/openshift/console/pull/11712

### DIFF
--- a/pkg/tasks/configsharing.go
+++ b/pkg/tasks/configsharing.go
@@ -16,11 +16,14 @@ package tasks
 
 import (
 	"context"
+	"fmt"
 	"net/url"
+
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
 
 	"github.com/openshift/cluster-monitoring-operator/pkg/client"
 	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
-	"github.com/pkg/errors"
 )
 
 type ConfigSharingTask struct {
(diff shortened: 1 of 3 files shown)


kyoto:
Basic approach of passing these URLs through the shared ConfigMap LGTM.

simonpasquier:
/hold

until we got agreement on https://github.com/openshift/console-operator/pull/658 and https://github.com/openshift/console/pull/11712

slashpai (approved):

simonpasquier:
/label docs-approved
/label px-approved

this PR has no direct user impact.

pkg/tasks/configsharing.go
```diff
@@ -71,7 +74,27 @@ func (t *ConfigSharingTask) Run(ctx context.Context) error {
...
+		svc, err = t.factory.AlertmanagerUserWorkloadService()
+		if err != nil {
+			return errors.Wrap(err, "initializing Alertmanager User Workload Service failed")
+		}
+	} else {
+		svc, err = t.factory.AlertmanagerService()
+		if err != nil {
+			return errors.Wrap(err, "initializing Alertmanager Service failed")
+		}
+	}
```
jan--f, line 88: I can't figure out why we use the Alertmanager Service if UWM alertmanger is disabled. Might be worth leaving a comment for similarly dense future developers, i.e. me in 6 month..?

jan--f (commented):
Code looks good, just one questions that could maybe be addressed with a comment.

pkg/tasks/configsharing.go
simonpasquier: good call, I've also realized that the port numbers could vary so I've updated the patch anyway.

sthaha (commented):
/lgtm

PR #1692, opened 2022-06-17
Bug 2091902: Improve performance of Prometheus Adapter

raptorsun:
This PR set the query backend back to Prometheus for Prometheus Adapter, improving its performance to serve the metrics API.

This reverts https://github.com/openshift/cluster-monitoring-operator/pull/1417 (commit https://github.com/openshift/cluster-monitoring-operator/commit/421a65c5b32779c4cc3752732e8c4e6d03124ab5).

A fix to [bug#2091902](https://bugzilla.redhat.com/show_bug.cgi?id=2091902).

Some users are experience performance problems with metrics API, queries are queueing before Thanos querier when having many HPA active.

### DIFF
--- a/assets/prometheus-adapter/configmap-prometheus.yaml
+++ b/assets/prometheus-adapter/configmap-prometheus.yaml
@@ -5,18 +5,18 @@ data:
     clusters:
     - cluster:
         certificate-authority: /etc/ssl/certs/service-ca.crt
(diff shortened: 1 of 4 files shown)


simonpasquier:
@wangke19 please disregard the review request, the bug fix is in the cluster monitoring operator, nothing to do with the API team.

PR #1693, opened 2022-06-17
Pin Jsonnet versions for release 4.11

raptorsun:

This PR pins down versions of Jsonnet dependencies, preparing release 4.11.


jsonnet/jsonnetfile.json
```diff
@@ -8,7 +8,7 @@
           "subdir": "contrib/mixin"
         }
       },
-      "version": "main"
+      "version": "release-3.5"
```
simonpasquier, line 11: IMHO we should make an exception for etcd otherwise we would lose improvements on the etcd dashboards. AFAICT the [release-4.10 branch](https://github.com/openshift/cluster-monitoring-operator/blob/a9fcd1a056d08a09911f0f909ba79f43bd092103/jsonnet/jsonnetfile.json#L4-L12) uses `main` too + we have a plan to move the etcd dashboards to the cluster-etcd operator anyway.

simonpasquier (commented):
one small nit otherwise lgtm

simonpasquier:
/label docs-approved
/label qe-approved
/label px-approved

This is locking our jsonnet dependencies to the release branches and introduces no functional change to CMO assets.

PR #1701, opened 2022-06-23
Bug 2100472: start alert controllers only when techpreview

simonpasquier:
When the TechPreview feature gate isn't enabled, the AlertingRule and
AlertRelabelConfig CRDs don't exist which triggers warning logs when the
operator wants to start the respective informers.


slashpai (approved):
/lgtm

PR #1704, opened 2022-07-01
Bug 2095719: Updates CreateOrUpdateServiceAccounts

JoaoBraveCoding:
Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2095719

Problem: previously SA had a problem where they couldn't be updated
or they would always generate a new token. Since k8s 1.24 service
accounts don't generate a token automatically anymore, so we can update
them

Solution: start updating service accounts

### DIFF
--- a/jsonnet/components/cluster-monitoring-operator.libsonnet
+++ b/jsonnet/components/cluster-monitoring-operator.libsonnet
@@ -240,7 +240,7 @@ function(params) {
       {
         apiGroups: [''],
         resources: ['services', 'serviceaccounts', 'configmaps'],
-        verbs: ['create', 'get', 'list', 'watch', 'update', 'delete'],
+        verbs: ['create', 'get', 'list', 'watch', 'update', 'patch', 'delete'],
       },
       {
         apiGroups: ['apps'],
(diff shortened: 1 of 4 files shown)


JoaoBraveCoding:
> the original comment said that the update operation would be coarse grained as it would generate new tokens while nothing had changed. My assumption is that we want to use patch instead and merge the service account's metatadata + maybe the `automountServiceAccountToken` field (e.g. keep `secrets` and `imagePullSecrets` untouched).

Indeed, however, that comment dated the times back from when CMO was open-sourced so I'm not 100% sure it holds true, furthermore with the recent changes to SA accounts I wanted to test if we still observed this behaviour or not.

raptorsun:
> the original comment said that the update operation would be coarse grained as it would generate new tokens while nothing had changed. My assumption is that we want to use patch instead and merge the service account's metatadata + maybe the `automountServiceAccountToken` field (e.g. keep `secrets` and `imagePullSecrets` untouched).

Will [retainKeys ](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#use-strategic-merge-patch-to-update-a-deployment-using-the-retainkeys-strategy)help keeping these fields?

JoaoBraveCoding:
So it turns out we still see the same behaviour nowadays 😅 so I've modified the change

PR #1703, opened 2022-07-01
Bug 2103033: fix CMO e2e tests

simonpasquier:
* [ ] I added CHANGELOG entry for this change.
* [X] No user facing changes, so no entry in CHANGELOG was needed.

PR #1707, opened 2022-07-04
Bug 2100472: fix alert controllers when not in techpreview

simonpasquier:
When the TechPreview feature gate isn't enabled, the AlertingRule
and AlertRelabelConfig CRDs don't exist which triggers warning logs when
the operator wants to start the respective informers.

When not in techpreview mode, the controllers initialize the informers
with fake lister+watcher that return no items.


simonpasquier:
/hold

need https://github.com/openshift/cluster-monitoring-operator/pull/1703 to be merged first

PR #1710, opened 2022-07-06
Add telemetry metrics for HyperShift

csrwng:
Initial set of metrics to be reported by the HyperShift operator through telemetry


simonpasquier:
/assign @jan--f 
/assign @bill3tt

A recording rule aggregating away the less-useful labels like `instance`, `pod`, ... would be appreciated. For example:

```
  - name: ...
    rules:
    - record: platform:hypershift_hostedclusters:max
      expr: max by(platform) (hypershift_hostedclusters)
    - record: platform:hypershift_nodepools:max
      expr: max by(platform) (hypershift_nodepools)
```

csrwng:
@simonpasquier I switched the name of the metrics to the name of the recording rules. The rules are now laid down when installing hypershift -> https://github.com/openshift/hypershift/pull/1544

PR #1712, opened 2022-07-08
Add cluster infra nodes to cluster:capacity_cpu_cores:sum

rphillips:
Add infra nodes to cluster:capacity_cpu_cores:sum.

image

### DIFF
--- a/assets/cluster-monitoring-operator/prometheus-rule.yaml
+++ b/assets/cluster-monitoring-operator/prometheus-rule.yaml
@@ -57,11 +57,13 @@ spec:
     - expr: max without(endpoint, instance, job, pod, service) (kube_node_labels and
         on(node) kube_node_role{role="infra"})
       labels:
+        label_node_role_kubernetes_io: infra
         label_node_role_kubernetes_io_infra: "true"
       record: cluster:infra_nodes
     - expr: max without(endpoint, instance, job, pod, service) (cluster:master_nodes
         and on(node) cluster:infra_nodes)
       labels:
+        label_node_role_kubernetes_io: infra,master
         label_node_role_kubernetes_io_infra: "true"
         label_node_role_kubernetes_io_master: "true"
       record: cluster:master_infra_nodes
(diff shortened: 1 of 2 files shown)


assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -57,11 +57,13 @@ spec:
     - expr: max without(endpoint, instance, job, pod, service) (kube_node_labels and
         on(node) kube_node_role{role="infra"})
       labels:
+        label_node_role_kubernetes_io: infra
         label_node_role_kubernetes_io_infra: "true"
```
jan--f, line 61: Is there a functional difference between these two lines?

assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -87,6 +89,14 @@ spec:
...
             )
           )
+          or on(node)
+          (
+            cluster:infra_nodes
+            * on(node) group_left() max by(node)
+            (
+              kube_node_status_capacity{resource="cpu",unit="core"}
+            )
+          )
```
jan--f, line 99: I was going to ask whether we really want to include infra nodes in `cluster:capacity_cpu_cores:sum`, as I would interpret that as user available capacity. But I see master nodes are already counted toward this, so I suppose infra nodes should be counted as well.

assets/cluster-monitoring-operator/prometheus-rule.yaml
rphillips: Other queries in the file use label_node_role_kubernetes_io instead of label_node_role_kubernetes_io_infra.

assets/cluster-monitoring-operator/prometheus-rule.yaml
eparis: I think the 'user available capacity' is typically seen in the \*workload\* metrics.  Though Ryan could correct me if I'm wrong.

rphillips:
rebased and regenerated

PR #1714, opened 2022-07-12
MON-2658: Add security context to additional containers in e2e tests

juanrh:

### DIFF
--- a/test/e2e/alertmanager_helpers.go
+++ b/test/e2e/alertmanager_helpers.go
@@ -109,7 +109,7 @@ func setupWebhookReceiver(t *testing.T, f *framework.Framework, namespace string
 								"--log.level=debug",
 								`--id.template={{ .CommonLabels.alertname }}_{{ .CommonLabels.namespace }}`,
 							},
-							SecurityContext: getSecurityContext(),
+							SecurityContext: getSecurityContextRestrictedProfile(),
 						},
 					},
 				},
--- a/test/e2e/alertmanager_helpers.go
+++ b/test/e2e/alertmanager_helpers.go
@@ -84,8 +84,6 @@ func setupWebhookReceiver(t *testing.T, f *framework.Framework, namespace string
 		return nil, err
 	}
 
-	allowPrivilegeEscalation := false
-	runAsNonRoot := true
 	if err := f.OperatorClient.CreateOrUpdateDeployment(ctx, &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: webhookReceiverService,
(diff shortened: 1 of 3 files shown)


test/e2e/alertmanager_helpers.go
```diff
@@ -109,6 +109,7 @@ func setupWebhookReceiver(t *testing.T, f *framework.Framework, namespace string
 								"--log.level=debug",
 								`--id.template={{ .CommonLabels.alertname }}_{{ .CommonLabels.namespace }}`,
 							},
+							SecurityContext: getSecurityContext(),
```
JoaoBraveCoding, line 112: I'm surprised that you can specify both `allowPrivilegeEscalation` and `capabilities` here and not have to do it in the container section 🤔 since `oc explain deployment.spec.template.spec.securityContext` doesn't mention them...

test/e2e/utils.go
```diff
@@ -141,3 +141,19 @@ func createSelfSignedMTLSArtifacts(s *v1.Secret) error {
 
 	return nil
 }
+
+func getSecurityContext() *v1.SecurityContext {
```
JoaoBraveCoding, line 145: Do you think this would be an improvement? I'm a bit split between just letting it be or giving it a name more meaningful 🤔 
```suggestion
func getSecurityContextRestrictedProfile() *v1.SecurityContext {
```

JoaoBraveCoding (commented):
Great job! Just two comments but they aren't at all blockers

test/e2e/alertmanager_helpers.go
juanrh: `deployment.spec.template.spec.securityContext` is [PodSecurityContext v1 core](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#podsecuritycontext-v1-core), but this change uses `deployment.spec.template.spec.containers.securityContext` that is [SecurityContext v1 core](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#securitycontext-v1-core), because:

- Some fields like `allowPrivilegeEscalation`, that appears in the warning, appear in SecurityContext but not in PodSecurityContext
- As documented in [PodSecurityContext v1 core](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#podsecuritycontext-v1-core), "field values of container.securityContext take precedence over field values of PodSecurityContext" so this way the change is more stable to future modifications

test/e2e/utils.go
juanrh: That's a good idea, I'll change it to that

jan--f:
/label docs-approved
/label px-approved
/label qe-approved
Adding labels an internal test change.

PR #1716, opened 2022-07-15
Bug 2090988: Set HA convention on admission-webhook

slashpai:
Set `spec.rollingUpdate.maxUnavailable: 1` explicitly on prometheus-operator admission webhook


jan--f:
/lgtm
ty!

PR #1718, opened 2022-07-18
Use odf metrics instead of ceph

weirdwiz:
This commit renames some ceph metrics to odf metrics, to include other
components that report the metric.

Jira Issue: https://issues.redhat.com/browse/RHSTOR-3473


simonpasquier (commented):
I think we need to keep the "old" metrics around otherwise they might be removed from the allow-list of the telemeter server.

weirdwiz:
The `odf_system` metrics are just using the old metrics[1], so I believe adding the old ones are redundant. 

[1] : https://github.com/red-hat-storage/ocs-operator/blob/e8e012943edc12d42c4cc67d67fba519cd7e79de/metrics/deploy/prometheus-ocs-rules.yaml#L34-L39

mulbc:
I would be ok if we keep the old metrics in the list, but add the new ones if that makes it easier to merge. The content will be very similar though

jan--f:
/label docs-approved
/label px-approved
/label qe-approved
For a telemetry addition.

weirdwiz:
@simonpasquier The LGTM label got removed due to rebasing.

PR #1719, opened 2022-07-19
Bug 2083226: increase alertmanager startupProbe failure threshold

sthaha:
The actual issue in the Bug[1] seems to be DNS related where
alertmanager fails to establish a quorum since its peers can be reached.
The  stop gap fix is to increase failure threshold of alertmanager's
startupProbe.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2083226


sthaha:
closing this PR to submit one against 4.12

PR #1720, opened 2022-07-19
Bug 2083226: increase alertmanager startupProbe failure threshold

sthaha:
On some clusters, during the initial boot, DNS resolution of services
may fail and this can cause Alertmanager pods to fail, which in turn
causes CMO to be in degraded state.

This patch fixes the issue by increasing the startupProbe's tolerance.

NOTE: The actual issue in the [bug] seems to be DNS related where
alertmanager fails to establish a quorum since its peers can't be
reached.

[bug]: https://bugzilla.redhat.com/show_bug.cgi?id=2083226

---

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -582,9 +582,13 @@ func (f *Factory) AlertmanagerUserWorkload(trustedCABundleCM *v1.ConfigMap) (*mo
 // returns (20s is twice the time that Alertmanager waits before
 // declaring that it can start sending notfications).
 //
-// We also account for slow DNS resolvers by retrying for 40 seconds in
-// case the endpoint isn't ready after 20s (see
-// https://bugzilla.redhat.com/show_bug.cgi?id=2037073 for details).
+// We also account for slow DNS resolvers by retrying for 400 seconds
+// (PeriodSeconds x FailureThreshold) thus giving AlertManager a total
+// of 7 minutes (420 seconds) in case the endpoint isn't ready after 20s.
+//
+// See bugs below for details:
+//  - https://bugzilla.redhat.com/show_bug.cgi?id=2037073
+//  - https://bugzilla.redhat.com/show_bug.cgi?id=2083226
 func setupStartupProbe(a *monv1.Alertmanager) {
 	if a.Spec.Storage != nil {
 		return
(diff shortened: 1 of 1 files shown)


jsonnet/components/alertmanager.libsonnet
```diff
@@ -233,6 +233,19 @@ function(params)
           },
         },
         containers: [
+          {
+            name: 'alertmanager-main',
+            startupProbe: {
+              exec: {
+                command: ['sh', '-c', 'exec curl http://localhost:9093/-/ready'],
```
sthaha, line 240: differs from upstream; missing `--fail`
```
    startupProbe:
      exec:
        command:
        - sh
        - -c
        - exec curl --fail http://localhost:9093/-/ready
      failureThreshold: 4
      initialDelaySeconds: 20
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 3

```

sthaha: Resolved now

assets/alertmanager/alertmanager.yaml
```diff
@@ -23,6 +23,18 @@ spec:
         - openshift-monitoring
         topologyKey: kubernetes.io/hostname
   containers:
+  - name: alertmanager-main
```
sthaha, line 26: For ref the current startup probe in nightly is 

```yaml
   startupProbe:
      exec:
        command:
        - sh
        - -c
        - exec curl --fail http://localhost:9093/-/ready
      failureThreshold: 4
      initialDelaySeconds: 20
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 3
```

simonpasquier:
/retitle Bug 2083226: increase alertmanager startupProbe failure threshold

pkg/manifests/manifests.go
```diff
@@ -607,9 +610,9 @@ func setupStartupProbe(a *monv1.Alertmanager) {
 						},
 					},
 				},
-				InitialDelaySeconds: 20,
+				InitialDelaySeconds: 30,
```
simonpasquier, line 613: reading again the PR, I would keep the initial delay of 20s so it doesn't slow down startup for clusters where DNS is working as expected.

From the original bug, the issue resolved after updating to the following values

```
initialDelaySeconds: 120
periodSeconds: 30
```

Though I wouldn't change the period and initial delay, it's probably worth increasing `FailureThreshold` to have a 7 minutes wait time (e.g. `FailureThreshold: 40`)?

simonpasquier (commented):
/lgtm

:+1:

PR #1721, opened 2022-07-19
Whitelist metrics for MCG addon

rexagod:
Whitelist metrics for MCG addon in order for them to be federated.

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>


simonpasquier:
> However I'm not sure if there's some automated process in place that will automatically get these metrics in that repository once this PR is merged? If that's the case, I'll close it.

Once this PR merges, you need to run `make whitelisted_metrics` in rhobs/configuration, commit the changes and push.

simonpasquier:
I've looked at the recording rules and I've got a few comments... For instance,

```
       - expr: |
           NooBaa_num_buckets_claims{namespace="redhat-data-federation"}
         labels:
           origin: data-federation-service
         record: noobaa_num_buckets_claims
```

First `expr` doens't reduce the cardinality of the metric since it's only a rename, I would expect `max(NooBaa_num_buckets_claims{namespace="redhat-data-federation"})`. The recorded metric name should follow Prometheus [conventions](https://prometheus.io/docs/practices/rules/#naming-and-aggregation) to make it obvious that it comes from a recording rule (e.g. `cluster:noobaa_num_buckets_claims:max`).


I've got also comments on the alerting rules but I've posted them to https://github.com/red-hat-storage/mcg-osd-deployer/pull/61.

rexagod:
Hello @simonpasquier, thank you for your response. I have a few follow-up questions I'd like to ask.

> First expr doens't reduce the cardinality of the metric since it's only a rename

So earlier most of these metrics were returning outputs from Noobaa deployed by ODF as well(and other Noobaa deployements, but mostly (~80%) through ODF) in it's namespace (`openshift-storage`). Now it's just limited to `redhat-data-federation`. I think that reduces the cardinality, no?

> I would expect max(NooBaa_num_buckets_claims{namespace="redhat-data-federation"})

But that modifies the rule beyond its original purpose, no? We do not want to see the maximum OBCs at an instant, but all the OBCs present at a time in all the clusters in the addon's namespace, in order to provide the desired visualization on the telemeter dashboard. Same goes for all other metrics as well.

> The recorded metric name should follow Prometheus [conventions](https://prometheus.io/docs/practices/rules/#naming-and-aggregation) to make it obvious that it comes from a recording rule (e.g. cluster:noobaa_num_buckets_claims:max).

ACK, will make these changes once we have clarified the aforementioned questions.

Thanks!

simonpasquier:
> Hello @simonpasquier, thank you for your response. I have a few follow-up questions I'd like to ask.
> 
> > First expr doens't reduce the cardinality of the metric since it's only a rename
> 
> So earlier most of these metrics were returning outputs from Noobaa deployed by ODF as well (and other Noobaa deployements, but mostly (~80%) through ODF) in it's namespace (`openshift-storage`). Now it's just limited to `redhat-data-federation`. I think that reduces the cardinality, no?

Ack I wasn't correct then.

> 
> > I would expect max(NooBaa_num_buckets_claims{namespace="redhat-data-federation"})
> 
> But that modifies the rule beyond its original purpose, no? We do not want to see the maximum OBCs at an instant, but all the OBCs present at a time in all the clusters in the addon's namespace, in order to provide the desired visualization on the telemeter dashboard. Same goes for all other metrics as well.

My understanding was that for a given cluster, `NooBaa_num_buckets_claims{namespace="redhat-data-federation"}` returns only one item (series). The max() function is only there to remove the labels like instance, pod, container, etc. which aren't useful when queried in telemeter. In particular the "instance" label may change over time and would increase the cardinality.
On telemeter side, you'll be able to query for all clusters, just that every cluster will return only 1 row. Does it make sense?

rexagod:
FYI https://ci.int.devshift.net/job/service-managed-tenants-gl-pr-check/5856/console, but https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels suggests otherwise. Need to ping MTSRE on this.

simonpasquier (commented):
Not sure why you have unrelated changes. Could you rebase on top of master?

cc @jan--f, this adds 9 new series to telemetry.

rexagod:
FYI this was fixed from MTSRE side: https://gitlab.cee.redhat.com/service/managed-tenants/-/merge_requests/2789/.

Refer: https://coreos.slack.com/archives/C01L46M0FQC/p1659427170686569.

jan--f:
The metrics added here look alright to me.
One question for my understanding though: I skimmed https://github.com/red-hat-storage/mcg-osd-deployer and it looks like you already deploy your own Prometheus instance and send metrics to RHOBS (add in https://github.com/red-hat-storage/mcg-osd-deployer/pull/65)
The [ServiceMonitor](https://github.com/red-hat-storage/mcg-osd-deployer/blob/main/config/prometheus/monitor.yaml) is not deployed in an `openshift-*` namespace, so iiuc won't be scraped by the in-cluster stack.
So I wonder, does this change here actually have any effect and of yes, won't the metrics end up being shipped to telemeter twice?

rexagod:
Hello @jan--f! 👋🏼 

> The [ServiceMonitor](https://github.com/red-hat-storage/mcg-osd-deployer/blob/main/config/prometheus/monitor.yaml?rgh-link-date=2022-08-08T08%3A21%3A51Z) is not deployed in an openshift-* namespace, so iiuc won't be scraped by the in-cluster stack.

The `ServiceMonitor` for federation is created by CMO in `redhat-monitoring-mcg-osd`. The `ServiceMonitor` for the TLS service you're referring to is created in `redhat-data-federation`. Do you mean that the federation won't happen if the `ServiceMonitor` for the TLS service (and the service itself) is in a non-`openshift` namespace?

```
  - apiVersion: monitoring.coreos.com/v1
    kind: ServiceMonitor
    metadata:
      name: federated-sm-mcg-osd
      namespace: redhat-monitoring-mcg-osd
    spec:
      endpoints:
      - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
        honorLabels: true
        interval: 30s
        params:
          match[]:
          - ALERTS{alertstate="firing"}
          - '{__name__="NooBaa_num_buckets_claims"}'
          - '{__name__="NooBaa_num_namespace_buckets"}'
          - '{__name__="NooBaa_num_unhealthy_bucket_claims"}'
          - '{__name__="NooBaa_num_unhealthy_namespace_buckets"}'
          - '{__name__="noobaa_accounts_num"}'
          - '{__name__="noobaa_total_usage"}'
          - '{__name__="odf_system_health_status"}'
          - '{__name__="odf_system_iops_total_bytes"}'
          - '{__name__="odf_system_throughput_total_bytes"}'
        path: /federate
        port: https
        scheme: https
        tlsConfig:
          caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
          serverName: prometheus.redhat-data-federation.svc
      namespaceSelector:
        matchNames:
        - redhat-data-federation
      selector:
        matchLabels:
          app: managed-mcg
```

> So I wonder, does this change here actually have any effect and of yes, won't the metrics end up being shipped to telemeter twice?

Not sure about this happening twice. The remote-write will happen to the ODFMS's RHOBS tenant which is on a separate RHOBS instance, unlike the telemeter client that's present in-cluster, so the data on telemetry will be written only once (by reverse-federation of metrics through CMO to telemeter-client), no?

PLMK if I'm missing anything, thanks.

jan--f:
Can you point us to the code where this is set up? By default CMO only considers ServiceMonitors in a namespace with the `openshift-` prefix. Or does this use user-workload-monitoring?
If you have a running cluster to look at, that would be even better.

Writing the metrics to two different tenants is a fair point, though it still duplicates the data. I wonder what the motivation for this data replication is.

rexagod:
@simonpasquier @jan--f Is there anything else that needs to be done for this PR from my side for the `tide` step to pass?

jan--f:
/label docs-approved
/label px-approved
/label qe-approved
Adding labels for an addition to the telemetry metrics list.

PR #1724, opened 2022-07-20
Bug 2099939: Sets status when UserAlertmanagerConfig is missconfigured

JoaoBraveCoding:
Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2099939

Problem: When users enable in the platform configuration the flag
enableUserAlertmanagerConfig and at the same time enable the user
dedicated Alermanager instance, then if they do not enable in the UWM
configmap enableAlertmanagerConfig, user alertmanager configuration will
neither be consumed by the main alertmanager instance nor by the user
one.

Solution: Report in the status of the operator this inconsistent state.

### DIFF
--- a/pkg/client/status_reporter.go
+++ b/pkg/client/status_reporter.go
@@ -27,10 +27,12 @@ import (
 )
 
 const (
-	unavailableMessage          string = "Rollout of the monitoring stack failed and is degraded. Please investigate the degraded status error."
-	asExpectedReason            string = "AsExpected"
-	StorageNotConfiguredMessage        = "Prometheus is running without persistent storage which can lead to data loss during upgrades and cluster disruptions. Please refer to the official documentation to see how to configure storage for Prometheus: https://docs.openshift.com/container-platform/4.8/monitoring/configuring-the-monitoring-stack.html"
-	StorageNotConfiguredReason         = "PrometheusDataPersistenceNotConfigured"
+	unavailableMessage                        string = "Rollout of the monitoring stack failed and is degraded. Please investigate the degraded status error."
+	asExpectedReason                          string = "AsExpected"
+	StorageNotConfiguredMessage                      = "Prometheus is running without persistent storage which can lead to data loss during upgrades and cluster disruptions. Please refer to the official documentation to see how to configure storage for Prometheus: https://docs.openshift.com/container-platform/4.8/monitoring/configuring-the-monitoring-stack.html"
+	StorageNotConfiguredReason                       = "PrometheusDataPersistenceNotConfigured"
+	UserAlermanagerConfigMisconfiguredMessage        = "Misconfigured Alertmanager:  Alertmanager for user-defined alerting is enabled in the openshift-monitoring/cluster-monitoring-config configmap by setting 'enableUserAlertmanagerConfig: true' field. This conflicts with a dedicated Alertmanager instance enabled in  openshift-user-workload-monitoring/user-workload-monitoring-config. Alertmanager enabled in openshift-user-workload-monitoring takes precedence over the one in openshift-monitoring, so please remove the 'enableUserAlertmanagerConfig' field in openshift-monitoring/cluster-monitoring-config."
+	UserAlermanagerConfigMisconfiguredReason         = "UserAlertmanagerMisconfigured"
 )
 
 type StatusReporter struct {
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -66,6 +66,21 @@ func (c Config) IsStorageConfigured() bool {
 	return prometheusK8sConfig.VolumeClaimTemplate != nil
 }
 
+func (c Config) HasInconsistentAlertmanagerConfigurations() bool {
+	if c.ClusterMonitoringConfiguration == nil || c.UserWorkloadConfiguration == nil {
+		return false
+	}
+
+	amConfig := c.ClusterMonitoringConfiguration.AlertmanagerMainConfig
+	uwmConfig := c.UserWorkloadConfiguration.Alertmanager
+
+	if amConfig == nil || uwmConfig == nil {
+		return false
+	}
+
+	return amConfig.EnableUserAlertManagerConfig && uwmConfig.Enabled
+}
+
 // AdditionalAlertmanagerConfigsForPrometheusUserWorkload returns the alertmanager configurations for
 // the User Workload Monitoring Prometheus instance.
 // If no additional configurations are specified, GetPrometheusUWAdditionalAlertmanagerConfigs returns nil.
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-monitoring-operator/pkg/client"
 	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
 	"github.com/pkg/errors"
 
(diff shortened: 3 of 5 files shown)


pkg/manifests/config.go
```diff
@@ -66,6 +66,20 @@ func (c Config) IsStorageConfigured() bool {
 	return prometheusK8sConfig.VolumeClaimTemplate != nil
 }
 
+func (c Config) IsUserAlermanagerConfigMisconfigured() bool {
+	if c.ClusterMonitoringConfiguration != nil && c.UserWorkloadConfiguration != nil {
+		alertmanagerMainConfig := c.ClusterMonitoringConfiguration.AlertmanagerMainConfig
+		alertmanagerUWMConfig := c.UserWorkloadConfiguration.Alertmanager
+		if alertmanagerMainConfig != nil && alertmanagerUWMConfig != nil {
+			if alertmanagerMainConfig.EnableUserAlertManagerConfig && alertmanagerUWMConfig.Enabled &&
+				!alertmanagerUWMConfig.EnableAlertmanagerConfig {
```
simonpasquier, line 75: I would even consider that enabling AlertmanagerConfig resources in CMO configmap and user-defined Alertmanager in UWM configmap is enough to notify the user.

```suggestion
			if alertmanagerMainConfig.EnableUserAlertManagerConfig && alertmanagerUWMConfig.Enabled {
```

pkg/manifests/config.go
```diff
@@ -66,6 +66,20 @@ func (c Config) IsStorageConfigured() bool {
 	return prometheusK8sConfig.VolumeClaimTemplate != nil
 }
 
+func (c Config) IsUserAlermanagerConfigMisconfigured() bool {
```
simonpasquier, line 69:
```suggestion
func (c Config) InconsistentAlermanagerConfigurations() bool {
```

test/e2e/config_test.go
```diff
@@ -71,6 +72,107 @@ func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
...
+		storage = "2Gi"
+	)
+	for _, tc := range []struct {
+		name               string
+		config             string
+		userWorkloadConfig string
+		assertion          func(t *testing.T)
+	}{
+		{
+			name:               "default config, no persistent storage",
```
simonpasquier, line 86: neat!

pkg/manifests/config.go
JoaoBraveCoding: Indeed! Especially because the default of `enableAlertmanagerConfig` is false!

test/e2e/config_test.go
```diff
@@ -71,6 +72,107 @@ func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
...
+			assertion: func(t *testing.T) {
+				f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionTrue)(t)
+				f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionFalse)(t)
+				f.AssertOperatorConditionReason(configv1.OperatorDegraded, "")
+				f.AssertOperatorConditionMessage(configv1.OperatorDegraded, "")
+			},
+		},
+		{
+			name: "default config with presistent storage but with UserAlermanagerConfig missconfiguration",
+			config: fmt.Sprintf(`alertmanagerMain:
```
JoaoBraveCoding, line 115: @simonpasquier I think I finally figured out the issue with this! It's here, I specified the CMO config but forgot to enable UWM 🙈 this somehow would make the test pass but it would put CMO tests in a weird state...

pkg/client/status_reporter.go
```diff
@@ -27,10 +27,12 @@ import (
...
+	asExpectedReason                           string = "AsExpected"
+	StorageNotConfiguredMessage                       = "Prometheus is running without persistent storage which can lead to data loss during upgrades and cluster disruptions. Please refer to the official documentation to see how to configure storage for Prometheus: https://docs.openshift.com/container-platform/4.8/monitoring/configuring-the-monitoring-stack.html"
+	StorageNotConfiguredReason                        = "PrometheusDataPersistenceNotConfigured"
+	UserAlermanagerConfigMissconfiguredMessage        = "AlertmanagerConfig resources for user-defined alerting are enabled in the openshift-monitoring/cluster-monitoring-config configmap with the 'enableUserAlertmanagerConfig: true' field but a dedicated Alertmanager instance in openshift-user-workload-monitoring/user-workload-monitoring-config is also enabled which takes precedence. Please remove the 'enableUserAlertmanagerConfig' field in openshift-monitoring/cluster-monitoring-config and check if 'enableAlertmanagerConfig' is set to true in openshift-user-workload-monitoring/user-workload-monitoring-config."
```
sthaha, line 34: typo: UserAler**t**managerConfigMis~~s~~configuredMessage

```suggestion
	UserAlertmanagerConfigMisconfiguredMessage        = "Misconfigured Alertmanager:  Alertmanager for user-defined alerting is enabled in the openshift-monitoring/cluster-monitoring-config configmap by setting 'enableUserAlertmanagerConfig: true' field. This conflicts with a dedicated Alertmanager instance enabled in  openshift-user-workload-monitoring/user-workload-monitoring-config. Alertmanager enabled in openshift-user-workload-monitoring takes precedence over the one in openshift-monitoring, so please remove the 'enableUserAlertmanagerConfig' field in openshift-monitoring/cluster-monitoring-config."
```

Does the above sound better?

pkg/client/status_reporter.go
```diff
@@ -27,10 +27,12 @@ import (
...
+	UserAlermanagerConfigMissconfiguredReason         = "UserAlermanagerConfigMissconfigured"
```
sthaha, line 35:
```suggestion
	UserAlertmanagerConfigMisconfiguredReason         = "UserAlertmanagerMisconfigured"
```
2 Changes 
- typo
- Doesn't `UserAlertmanagerMisconfigured` sound better?

pkg/manifests/config.go
```diff
@@ -66,6 +66,19 @@ func (c Config) IsStorageConfigured() bool {
 	return prometheusK8sConfig.VolumeClaimTemplate != nil
 }
 
+func (c Config) InconsistentAlermanagerConfigurations() bool {
```
sthaha, line 69:
```suggestion
func (c Config) HasInconsistentAlertmanagerConfigurations() bool {
```

pkg/manifests/config.go
```diff
@@ -66,6 +66,19 @@ func (c Config) IsStorageConfigured() bool {
...
+	if c.ClusterMonitoringConfiguration != nil && c.UserWorkloadConfiguration != nil {
+		alertmanagerMainConfig := c.ClusterMonitoringConfiguration.AlertmanagerMainConfig
+		alertmanagerUWMConfig := c.UserWorkloadConfiguration.Alertmanager
+		if alertmanagerMainConfig != nil && alertmanagerUWMConfig != nil {
+			if alertmanagerMainConfig.EnableUserAlertManagerConfig && alertmanagerUWMConfig.Enabled {
+				return true
+			}
+		}
+	}
```
sthaha, line 78: We can avoid nesting by exiting early. e.g. 

```go

if c.ClusterMonitoringConfiguration == nil || c.UserWorkloadConfiguration == nil {
  return false
}

mainConfig := c.ClusterMonitoringConfiguration.AlertmanagerMainConfig
uwmConfig := c.UserWorkloadConfiguration.Alertmanager

if mainConfig == nil || uwmConfig == nil {
  return false
}

return mainConfig.EnableUserAlertManagerConfig && uwmConfig.Enabled

File: test/e2e/framework/assertions.go [hunk 9]
```diff
@@ -368,6 +368,54 @@ func (f *Framework) AssertOperatorCondition(conditionType configv1.ClusterStatus
...
+	}
+}
+
+func (f *Framework) AssertOperatorConditionMessage(conditionType configv1.ClusterStatusConditionType, conditionMessage string) func(t *testing.T) {
+	return func(t *testing.T) {
+		reporter := f.OperatorClient.StatusReporter()
+		err := Poll(time.Second, 5*time.Minute, func() error {
+			co, err := reporter.Get(ctx)
+			if err != nil {
+				t.Fatal(err)
```
sthaha, line 401: `return err` should suffice, doesn't it?

pkg/client/status_reporter.go
JoaoBraveCoding: Yes it sounds much better! Thank you!

pkg/manifests/config.go
JoaoBraveCoding: This looks wayy better! Thank you!

test/e2e/framework/assertions.go
JoaoBraveCoding: Yes, I had used a similar function, so I've also fixed it

pkg/manifests/config.go
```diff
@@ -66,6 +66,21 @@ func (c Config) IsStorageConfigured() bool {
...
 
+func (c Config) HasInconsistentAlertmanagerConfigurations() bool {
+	if c.ClusterMonitoringConfiguration == nil && c.UserWorkloadConfiguration == nil {
+		return false
+	}
+
+	alertmanagerMainConfig := c.ClusterMonitoringConfiguration.AlertmanagerMainConfig
+	alertmanagerUWMConfig := c.UserWorkloadConfiguration.Alertmanager
+
+	if alertmanagerMainConfig == nil && alertmanagerUWMConfig == nil {
```
sthaha, line 77: Shouldn't this be `||` instead of `&&` 
What is main is enabled by there is no uwm ? 
main == nil => false && uwm == nil => true  ==> false

now the code below uwm.Enabled will crash, won't it?

pkg/manifests/config.go
```diff
@@ -66,6 +66,21 @@ func (c Config) IsStorageConfigured() bool {
 	return prometheusK8sConfig.VolumeClaimTemplate != nil
 }
 
+func (c Config) HasInconsistentAlertmanagerConfigurations() bool {
+	if c.ClusterMonitoringConfiguration == nil && c.UserWorkloadConfiguration == nil {
+		return false
+	}
+
+	alertmanagerMainConfig := c.ClusterMonitoringConfiguration.AlertmanagerMainConfig
```
sthaha, line 74: nit: `main` IMHO is verbose enough :) https://github.com/golang/go/wiki/CodeReviewComments#variable-names 

Again, it is just a nitpick

pkg/manifests/config.go
JoaoBraveCoding: Ahhh good catch, when I changed it I forgot to change the && to || 😅 
I've also added another test to cover that case 😉

pkg/manifests/config.go
JoaoBraveCoding: No, when reviewing my code feel free to always nitpick, I also agree it look better!

sthaha (commented):
/lgtm

PR #1725, opened 2022-07-21
[MON-2695] Unpin Jsonnet dependencies.

raptorsun:

[JIRA Task](https://issues.redhat.com/browse/MON-2695)

### DIFF
--- a/assets/prometheus-k8s/service-account.yaml
+++ b/assets/prometheus-k8s/service-account.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-automountServiceAccountToken: false
+automountServiceAccountToken: true
 kind: ServiceAccount
 metadata:
   annotations:
--- a/assets/prometheus-k8s/service-account.yaml
+++ b/assets/prometheus-k8s/service-account.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-automountServiceAccountToken: true
+automountServiceAccountToken: false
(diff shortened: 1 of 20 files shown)


assets/prometheus-k8s/service-account.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-automountServiceAccountToken: false
+automountServiceAccountToken: true
```
simonpasquier, line 2: I think that we should override this in the CMO jsonnet since it worked fine for us (IIUC it's needed when using a version of Prometheus operator that doesn't set `automountServiceAccountToken: true` at the statefulset level).

assets/prometheus-user-workload/service-account.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-automountServiceAccountToken: false
+automountServiceAccountToken: true
```
simonpasquier, line 2: ditto

assets/prometheus-k8s/service-account.yaml
raptorsun: ok, will overide this option at CMO level.

simonpasquier:
/label px-approved
/label docs-approved
/label qe-approved

This is as usual after a new release branch is cut, we switch back tracking the master/main branches for each upstream project.

PR #1733, opened 2022-08-05
MON-2807: Use bearer token file for remote write authentication with telemeter

simonpasquier:
cc @saswatamcode

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -62,6 +62,8 @@ const (
 
 	platformAlertmanagerService     = "alertmanager-main"
 	userWorkloadAlertmanagerService = "alertmanager-user-workload"
+
+	telemetryTokenSecretKey = "token"
 )
 
 var (
@@ -145,6 +147,7 @@ var (
 	PrometheusK8sThanosSidecarServiceMonitor      = "prometheus-k8s/service-monitor-thanos-sidecar.yaml"
 	PrometheusK8sTAlertmanagerRoleBinding         = "prometheus-k8s/alertmanager-role-binding.yaml"
 	PrometheusK8sPodDisruptionBudget              = "prometheus-k8s/pod-disruption-budget.yaml"
+	PrometheusK8sTelemetry                        = "prometheus-k8s/telemetry-secret.yaml"
 
 	PrometheusUserWorkloadServingCertsCABundle        = "prometheus-user-workload/serving-certs-ca-bundle.yaml"
 	PrometheusUserWorkloadServiceAccount              = "prometheus-user-workload/service-account.yaml"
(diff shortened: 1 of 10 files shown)


pkg/manifests/manifests.go
```diff
@@ -1657,13 +1677,12 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 					Replacement:  "alerts",
 				},
 			},
+			MetadataConfig: &monv1.MetadataConfig{
```
JoaoBraveCoding, line 1680: This change seems to not be related, was it intentional?

JoaoBraveCoding (commented):
Looks good apart from a small comment, also do we have a Jira issue or a BZ with more content for the change?

JoaoBraveCoding:
For traceability I'm also linking this here since AFAIK these two PRs are related https://github.com/openshift/telemeter/pull/430

pkg/manifests/manifests.go
simonpasquier: Yes but it should be commented. Basically we don't want/need to forward metadata information to the telemeter server.

It comes from https://github.com/openshift/cluster-monitoring-operator/pull/1416 which was the previous PR that exercised the receive path of telemeter.

simonpasquier:
@jan--f I'm wondering if we shouldn't add an end-to-end test to validate the /receive path. Even though we don't support it yet in OCP, it would help detecting regressions on the server side. WDYT?

pkg/manifests/manifests.go
saswatamcode: Yes, Telemeter forwards rw requests to Thanos, and Thanos doesn't have metadata ingestion yet. We can still send it, but it would just be dropped. 🙂

test/e2e/telemeter_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// TestTelemeterRemoteWrite verifies that the monitoring stack can send data to
+// the telemeter server using the native Prometheus remote write endpoint.
+func TestTelemeterRemoteWrite(t *testing.T) {
```
JoaoBraveCoding, line 32: Small curiosity questions: 
1. Where does CMO get the config/info that Prom is supposed to remote-write to `https://infogw.api.openshift.com.+`? I see we are enabling remote-write but I don't get how it knows it's supposed to send them there.
2. Why are we modifying the deployment instead of using the ConfigMap

JoaoBraveCoding (commented):
Apart from the questions looks good 👌

test/e2e/telemeter_test.go
saswatamcode: I think it's hardcoded in the `SetRemoteWrite` function https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/manifests/config.go#L449. This is called in `Operator.sync` https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/operator/operator.go#L608.

JoaoBraveCoding: Indeed, thank you 👍

simonpasquier: yes, we have no way to turn on native remote write without setting CMO into unmanaged state and hacking the deployment args.

pkg/manifests/manifests.go
```diff
@@ -1604,23 +1629,18 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 		return nil, err
 	}
 
-	telemetryEnabled := f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.IsEnabled()
 	clusterID := f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.ClusterID
-	if telemetryEnabled && f.config.RemoteWrite {
-
+	if telemetrySecret != nil {
 		selectorRelabelConfig, err := promqlgen.LabelSelectorsToRelabelConfig(f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.TelemetryMatches)
```
jan--f, line 1634: Discussed separately: Future readers of this code would benefit from checking the actual config condition here (`if telemetryEnabled && f.config.RemoteWrite`) and assuming the secret was created instead of coupling this to the secret creating.

jan--f:
/retitle: MON-2807: Use bearer token file for remote write authentication with telemeter

jan--f:
Once change request discussed offline added here as a comment, otherwise this looks good.

jan--f:
/retitle MON-2807: Use bearer token file for remote write authentication with telemeter

pkg/manifests/manifests.go
jan--f: Did you still want to add a comment for this @simonpasquier ?

jan--f:
/lgtm
I think the commend about `MetadataConfig` would still be useful, but I can live without it.

On a side note, I was wondering about the `cluster_id` field in the secret...seems like that has been there since day 1 of this feature. Does anyone have an idea what the purpose is?

jan--f:
/label docs-approved
/label px-approved
/label qe-approved
Adding labels for an addition to our e2e test suite.

PR #1737, opened 2022-08-08
Bug 2116382: Give precedence to CMO config map proxy config 

juanrh:
Problem: Customers cannot configure a proxy for telemeter
client without setting up the cluster wide proxy.

Solution: Give precedence to CMO config map proxy config
over cluster proxy config for the proxy configuration,
otherwise cluster proxy config is always used, because
openshift.io/v1/Proxy/cluster is always present.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2116382

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -573,6 +573,22 @@ func (o *Operator) enqueue(obj interface{}) {
 	o.queue.Add(key)
 }
 
+type proxyConfigSupplier func(context.Context) (*ProxyConfig, error)
+
+func getProxyReader(ctx context.Context, config *manifests.Config, proxyConfigSupplier proxyConfigSupplier) manifests.ProxyReader {
+	if config.HTTPProxy() != "" || config.HTTPSProxy() != "" || config.NoProxy() != "" {
+		return config
+	}
+
+	clusterProxyConfig, err := proxyConfigSupplier(ctx)
+	if err != nil {
+		klog.Warningf("Proxy config in CMO configmap is empty and fallback to cluster proxy config failed - no proxy will be used: %v", err)
+		return config
+	}
+
+	return clusterProxyConfig
+}
+
 func (o *Operator) sync(ctx context.Context, key string) error {
 	// The operator may have left some nodes as unschedulable during a previous
 	// sync in an attempt to rebalance workloads.
@@ -591,12 +607,7 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 	config.SetTelemetryMatches(o.telemetryMatches)
 	config.SetRemoteWrite(o.remoteWrite)
 
-	var proxyConfig manifests.ProxyReader
-	proxyConfig, err = o.loadProxyConfig(ctx)
-	if err != nil {
-		klog.Warningf("using proxy config from CMO configmap: %v", err)
-		proxyConfig = config
-	}
+	var proxyConfig = getProxyReader(ctx, config, o.loadProxyConfig)
 
 	var apiServerConfig *manifests.APIServerConfig
 	apiServerConfig, err = o.loadApiServerConfig(ctx)
--- a/pkg/operator/operator_test.go
+++ b/pkg/operator/operator_test.go
@@ -20,6 +20,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-monitoring-operator/pkg/client"
+	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
 	"github.com/openshift/cluster-monitoring-operator/pkg/rebalancer"
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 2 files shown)


pkg/operator/operator.go
```diff
@@ -573,6 +573,22 @@ func (o *Operator) enqueue(obj interface{}) {
 	o.queue.Add(key)
 }
 
+type proxyConfigSupplier func() (*ProxyConfig, error)
+
+func getProxyReader(proxyConfigSupplier proxyConfigSupplier, config *manifests.Config) manifests.ProxyReader {
+	var proxyConfig manifests.ProxyReader = config
+	if proxyConfig.HTTPProxy() == "" && proxyConfig.HTTPSProxy() == "" && proxyConfig.NoProxy() == "" {
```
simonpasquier, line 580: I find it easier to read if we return early in the simple case.

```suggestion
	if proxyConfig.HTTPProxy() != "" || proxyConfig.HTTPSProxy() != "" || proxyConfig.NoProxy() != "" {
		return config
	}
```

pkg/operator/operator.go
```diff
@@ -591,12 +607,7 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 	config.SetTelemetryMatches(o.telemetryMatches)
 	config.SetRemoteWrite(o.remoteWrite)
 
-	var proxyConfig manifests.ProxyReader
-	proxyConfig, err = o.loadProxyConfig(ctx)
-	if err != nil {
-		klog.Warningf("using proxy config from CMO configmap: %v", err)
-		proxyConfig = config
-	}
+	var proxyConfig = getProxyReader(func() (*ProxyConfig, error) { return o.loadProxyConfig(ctx) }, config)
```
simonpasquier, line 610: rather than a new function we could extend `loadProxyConfig()` to read proxy configuration from the configmap (if it exists) and fallback to the cluster-wide proxy settings?

```
proxyConfig, err = o.loadProxyConfig(ctx, config)
```

juanrh: A downside of that is that then `loadProxyConfig` would have 2 concerns: fetching the cluster wide config, which is already somewhat complex due to state handling for `o.lastKnowProxyConfig`; and deciding the precedence between those two configurations. 
Also `loadProxyConfig` is difficult to test because the version of the openshift client in `vendor` doesn't have support for faking for the config clientset:

```

$ find vendor/github.com/openshift/client-go -type d -wholename '*/versioned/fake'
vendor/github.com/openshift/client-go/route/clientset/versioned/fake
vendor/github.com/openshift/client-go/security/clientset/versioned/fake
$
```

which explains why there is no unit test for `loadProxyConfig`. Even though this is a small issue, it has significant impact in the functionality, so it would be nice to have it covered by a unit test to prevent (or at least difficult) future regressions.

I was thinking on renaming `loadProxyConfig` to `loadClusterProxyConfig` to make all more explicit. What do you think? 

In any case I'm not adamant with this, and I'm ok with moving the new logic to `loadProxyConfig` if you think that would make the code simpler, and it is ok not adding a unit test for this

simonpasquier: > Also loadProxyConfig is difficult to test because the version of the openshift client in vendor doesn't have support for faking for the config clientset.

You should be able to import "github.com/openshift/client-go/config/clientset/versioned/fake" from the test file then run `go mod vendor` to update the vendor/ directory.

Totally agree that we need unit tests to avoid regressoins. I'm also not stuck on my recommendation so do as you feel :)

sthaha, line 610: Just a thought 
```suggestion
	var proxyConfig = getProxyReader(ctx, o.loadProxyConfig, config)
```
The code can read "simpler" if context is passed to `getProxyReader` which then passes the `ctx` to `proxySupplier` that accepts a `context`.

pkg/operator/operator.go
```diff
@@ -573,6 +573,22 @@ func (o *Operator) enqueue(obj interface{}) {
...
+type proxyConfigSupplier func() (*ProxyConfig, error)
+
+func getProxyReader(proxyConfigSupplier proxyConfigSupplier, config *manifests.Config) manifests.ProxyReader {
+	if config.HTTPProxy() != "" || config.HTTPSProxy() != "" || config.NoProxy() != "" {
+		return config
+	}
+
+	clusterProxyConfig, err := proxyConfigSupplier()
+	if err != nil {
+		klog.Warningf("Error falling back to cluster proxy config from an empty proxy config in CMO configmap, no proxy will be used: %v", err)
```
sthaha, line 585:
```suggestion
		klog.Warningf("Proxy config in CMO configmap is empty and fallback to cluster proxy config failed - no proxy will be used: %v", err)
```

pkg/operator/operator.go
```diff
@@ -591,12 +607,7 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 	config.SetTelemetryMatches(o.telemetryMatches)
 	config.SetRemoteWrite(o.remoteWrite)
 
-	var proxyConfig manifests.ProxyReader
-	proxyConfig, err = o.loadProxyConfig(ctx)
-	if err != nil {
-		klog.Warningf("using proxy config from CMO configmap: %v", err)
-		proxyConfig = config
-	}
+	var proxyConfig = getProxyReader(ctx, config, func(ctx context.Context) (*ProxyConfig, error) { return o.loadProxyConfig(ctx) })
```
sthaha, line 610:
```suggestion
	var proxyConfig = getProxyReader(ctx, config, o.loadProxyConfig)
```
Will this ☝️ work?

sthaha (commented):
/lgtm

jan--f:
/retitle OCPBUGSM-47783: Give precedence to CMO config map proxy config

jan--f:
/retitle OCPBUGSM-47331: Give precedence to CMO config map proxy config

PR #1738, opened 2022-08-08
Add client certificate and key to service monitor

xueqzhan:

Description of the change:
Add client certificate and key to service monitor so that prometheus scraper does not depend on API server availability.

Motivation for the change:
https://github.com/deads2k/openshift-enhancements/blob/master/enhancements/monitoring/client-cert-scraping.md

### DIFF
--- a/assets/kube-state-metrics/service-monitor.yaml
+++ b/assets/kube-state-metrics/service-monitor.yaml
@@ -34,6 +34,8 @@ spec:
     scrapeTimeout: 1m
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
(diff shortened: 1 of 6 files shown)


jan--f (changes requested):
Hey, the files under `assets/` are generated by jsonnet, so you'll have to change the respecitve files in `jsonnet/components` and run `make generate`. 

But please note that some service monitors weren't changed in the original PR adding mTLS auth for metric scraping: https://github.com/openshift/cluster-monitoring-operator/pull/1282.

I'd expect these issues persist.

jan--f:
/label docs-approved
/label px-approved
Adding doc and px approval as this has no impact on either.
Ping @juzhao for qe approval.

PR #1740, opened 2022-08-10
Bug 2100312: go.mod: update openshift-api to current release-4.12 tip

jan--f:


jan--f:
/label docs-approved
/label px-approved
Ping @juzhao for qe approval.

PR #1745, opened 2022-08-22
  Bug 2095719: Updates CreateOrUpdateServiceAccounts 

JoaoBraveCoding:
Issue: https://bugzilla.redhat.com/show_bug.cgi?id=2095719

Problem: When a SA is updated if we do not use the existing resource
this will make it so new secrets are created for such service account.

Solution: When updating a SA we should instead updating with the new SA
we should use the referance for the existing SA and update it with the
necessary updates, this will make it so the old secrets are preserved


pkg/client/client.go
```diff
@@ -63,6 +64,7 @@ const (
 	deploymentCreateTimeout = 5 * time.Minute
 	deploymentDeleteTimeout = 5 * time.Minute
 	metadataPrefix          = "monitoring.openshift.io/"
+	cmoFieldManager         = "cluster-monitoring-operator"
```
juanrh, line 67: it looks like this is not used

pkg/client/client.go
```diff
@@ -1526,42 +1528,27 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
 
 func (c *Client) CreateOrUpdateServiceAccount(ctx context.Context, sa *v1.ServiceAccount) error {
 	sClient := c.kclient.CoreV1().ServiceAccounts(sa.GetNamespace())
-	_, err := sClient.Get(ctx, sa.GetName(), metav1.GetOptions{})
+	existing, err := sClient.Get(ctx, sa.GetName(), metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		_, err := sClient.Create(ctx, sa, metav1.CreateOptions{})
 		return errors.Wrap(err, "creating ServiceAccount object failed")
```
juanrh, line 1534: Shouldn't we be checking the new value of `err` instead of always assuming creation failed?
And then we should assign the new client to `existing`, to be able to use it later

pkg/client/client_test.go
```diff
@@ -952,6 +952,100 @@ func TestCreateOrUpdateService(t *testing.T) {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        "prometheus-k8s",
+					Namespace:   ns,
+					Labels:      tc.initialLabels,
+					Annotations: tc.initialAnnotations,
+				},
+			}
+
+			c := Client{
+				kclient: fake.NewSimpleClientset(sa.DeepCopy()),
```
juanrh, line 1020: What about adding a test case for when there is no existing service account?

juanrh (changes requested):

pkg/client/client.go
JoaoBraveCoding: It's in this case we don't need to check it since `Wrap` already does it for us ;)

> Wrap returns an error annotating err with a stack trace
> at the point Wrap is called, and the supplied message.
> If err is nil, Wrap returns nil.

pkg/client/client.go
```diff
@@ -1526,42 +1527,27 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
-	// patchBytes, patchErr := strategicpatch.CreateTwoWayMergePatch(oldData, newData, required)
-	// if patchErr != nil {
-	// 	return errors.Wrap(patchErr, "creating ServiceAccount two way merge patch failed")
-	// }
-	// _, err = sClient.Patch(ctx, required.Name, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
-	// return errors.Wrap(err, "patching ServiceAccount object failed")
+	if err != nil {
+		return errors.Wrap(err, "retrieving ServiceAccount object failed")
+	}
+	modified := resourcemerge.BoolPtr(false)
```
juanrh, line 1538: Nitpick: wouldn't it be simpler to just use the following?:

```go
modified := false
...
resourcemerge.EnsureObjectMeta(&modified, &existing.ObjectMeta, sa.ObjectMeta)
...
	if !modified {
```

pkg/client/client.go
```diff
@@ -1526,42 +1527,27 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+	existingCopy := existing.DeepCopy()
```
juanrh, line 1539: why not just using `existing` directly? `existing` is not used after this line

juanrh (changes requested):

pkg/client/client.go
JoaoBraveCoding: This could also work, I took inspiration from a go library that deals with the same problem https://github.com/openshift/library-go/blob/84c00ba9649a8c718023a43ace173e972656afee/pkg/operator/resource/resourceapply/core.go#L227

Same for the comment bellow 😅

juanrh:
removing lgmt label so Jan and Simon can take a look

pkg/client/client.go
jan--f: I think Juans suggestion would improve readbility.

pkg/client/client.go
jan--f: :+1:

jan--f (changes requested):
It would be great to have more details in the commit message about why `Patch` doesn't work for us in this context and we have to fall back to an additional dependecy.

pkg/client/client.go
simonpasquier: I wonder if we should use `ApplyServiceAccountImproved()` directly as duplicating the same logic seems error-prone?

JoaoBraveCoding: > I think Juans suggestion would improve readbility.

I'm okay with it, I just didn't do it to try to be as close as possible to the original code. 

> I wonder if we should use ApplyServiceAccountImproved() directly as duplicating the same logic seems error-prone?


Yes, I wanted to use it but the function requires a `recorder events.Recorder` which if I'm not mistaken we don't have in CMO

pkg/operator/operator.go
```diff
@@ -336,6 +336,9 @@ func New(
 		controllerRef,
 	)
 
+	// Set event recorder
```
JoaoBraveCoding, line 339: @simonpasquier I did it like this since we need a k8s client to be able to instantiate an event recorder. Let me know if it's OK or if there's a better alternative.

simonpasquier (commented):
My suggestion would be to modify the signature of NewForConfig() to allow passing options. It will be more obvious in the operator package where the event recorder comes from.

```
pkg/client/client.go
func NewForConfig(cfg *rest.Config, version string, namespace, userWorkloadNamespace string, options ...Option) (*Client, error) {
...
}

pkg/operator/operator.go
func New(
...
) {
    eventRecorder := ...
    c, err := client.NewForConfig(config, version, namespace, namespaceUserWorkload, client.EventRecorder(eventRecorder))
...
```

pkg/client/client_test.go
```diff
@@ -952,6 +953,109 @@ func TestCreateOrUpdateService(t *testing.T) {
...
+					Labels:      tc.initialLabels,
+					Annotations: tc.initialAnnotations,
+				},
+			}
+
+			var c Client
+			if tc.initialAnnotations == nil && tc.initialLabels == nil {
+				c = Client{
+					kclient:       fake.NewSimpleClientset(),
+					eventRecorder: events.NewInMemoryRecorder("cluster-monitoring-operator"),
```
simonpasquier, line 1025: the event recorder could be created in advance.

pkg/client/client_test.go
```diff
@@ -952,6 +953,109 @@ func TestCreateOrUpdateService(t *testing.T) {
...
+			},
+			expectedLabels: map[string]string{
+				"app.kubernetes.io/name": "app",
+			},
+			expectedAnnotations: map[string]string{
+				"monitoring.openshift.io/foo": "bar",
+			},
+		},
+		{
+			name: "label/annotation merge and spec change",
```
simonpasquier, line 984:
```suggestion
			name: "label/annotation merge",
```

pkg/client/client_test.go
```diff
@@ -952,6 +953,109 @@ func TestCreateOrUpdateService(t *testing.T) {
 	}
 }
 
+func TestCreateOrUpdateServiceAccount(t *testing.T) {
+	ctx := context.Background()
+	testCases := []struct {
```
simonpasquier, line 958: could you add a test ensuring that no update is triggered when the metadata doesn't change but the SA secrets have changed?

JoaoBraveCoding:
> My suggestion would be to modify the signature of NewForConfig() to allow passing options. It will be more obvious in the operator package where the event recorder comes from.
> 
> ```
> pkg/client/client.go
> func NewForConfig(cfg *rest.Config, version string, namespace, userWorkloadNamespace string, options ...Option) (*Client, error) {
> ...
> }
> 
> pkg/operator/operator.go
> func New(
> ...
> ) {
>     eventRecorder := ...
>     c, err := client.NewForConfig(config, version, namespace, namespaceUserWorkload, client.EventRecorder(eventRecorder))
> ...
> ```

With this approach, there is still the problem of needing a k8s client to initialize the eventRecorder... from my POV there are three approaches:
1. We use a setter like I was doing originally
2. We create the k8s client outside of the client package and we pass it as options or even as an argument, snippet below
3. We move the eventRecorder to the client package, version committed (I used a ctx.TODO, not sure if it's ok or if I should pass ctx)

Option 2
```
	kclient, err := kubernetes.NewForConfig(config)
	if err != nil {
		return nil, errors.Wrap(err, "creating kubernetes clientset client")
	}

	controllerRef, err := events.GetControllerReferenceForCurrentPod(ctx, kclient, namespace, nil)
	if err != nil {
		klog.Warningf("unable to get owner reference (falling back to namespace): %v", err)
	}

	eventRecorder := events.NewKubeRecorderWithOptions(
		kclient.CoreV1().Events(namespace),
		events.RecommendedClusterSingletonCorrelatorOptions(),
		"cluster-monitoring-operator",
		controllerRef,
	)

	c, err := client.NewForConfig(config, version, namespace, namespaceUserWorkload, client.KubernetesClient(kclient), client.EventRecorder(eventRecorder))
	if err != nil {
		return nil, err
	}
```

raptorsun:
it looks good to me, the new version is even simpler than the old one :)

simonpasquier (commented):
/lgtm

simonpasquier (commented):
/lgtm

PR #1747, opened 2022-08-24
Bug 2114721: Adds telemeter token hash to Deployment annotation

JoaoBraveCoding:
Problem: Currently after a pull secret is changed/updated the telemeter
client starts failing to authenticate with telemeter server. This
happens because the client does not load the new secret unless it is
restarted. The salt in the secret is being continuously updated which is
not correct.

Solution: Add a hash of the secret to the deployment so that when the
secret is rottated the deployment is restarted. Only update the salt
in the secret when the token is updated

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -16,6 +16,7 @@ package manifests
 
 import (
 	"crypto/md5"
+	"crypto/sha256"
 	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
@@ -3807,10 +3808,16 @@ func (f *Factory) TelemeterClientPrometheusRule() (*monv1.PrometheusRule, error)
 // TelemeterClientDeployment generates a new Deployment for Telemeter client.
 // If the passed ConfigMap is not empty it mounts the Trusted CA Bundle as a VolumeMount to
 // /etc/pki/ca-trust/extracted/pem/ location.
-func (f *Factory) TelemeterClientDeployment(proxyCABundleCM *v1.ConfigMap) (*appsv1.Deployment, error) {
+func (f *Factory) TelemeterClientDeployment(proxyCABundleCM *v1.ConfigMap, s *v1.Secret) (*appsv1.Deployment, error) {
 	d, err := f.NewDeployment(f.assets.MustNewAssetReader(TelemeterClientDeployment))
 	if err != nil {
 		return nil, err
+	}
+
+	// Set annotation on deployment to trigger redeployments
+	if s != nil {
+		hash := sha256.New()
+		d.Spec.Template.Annotations["telemeter-token-hash"] = string(hash.Sum(s.Data["token"]))
 	}
 
 	for i, container := range d.Spec.Template.Spec.Containers {
@@ -3896,7 +3903,6 @@ func (f *Factory) TelemeterClientServiceAccount() (*v1.ServiceAccount, error) {
 	return s, nil
 }
 
-// TelemeterClientSecret generates a new Secret for Telemeter client.
 func (f *Factory) TelemeterClientSecret() (*v1.Secret, error) {
 	s, err := f.NewSecret(f.assets.MustNewAssetReader(TelemeterClientSecret))
 	if err != nil {
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -16,6 +16,7 @@ package manifests
 
 import (
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"net/url"
@@ -698,7 +699,7 @@ func TestUnconfiguredManifests(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	_, err = f.TelemeterClientDeployment(nil)
+	_, err = f.TelemeterClientDeployment(nil, nil)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -3277,7 +3278,7 @@ func TestTelemeterConfiguration(t *testing.T) {
 		t.Fatal(err)
 	}
 	f := NewFactory("openshift-monitoring", "openshift-user-workload-monitoring", c, defaultInfrastructureReader(), &fakeProxyReader{}, NewAssets(assetsPath), &APIServerConfig{}, &configv1.Console{})
-	d, err := f.TelemeterClientDeployment(&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "foo"}})
+	d, err := f.TelemeterClientDeployment(&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "foo"}}, &v1.Secret{Data: map[string][]byte{"token": []byte("test")}})
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -3300,6 +3301,15 @@ func TestTelemeterConfiguration(t *testing.T) {
 		}
 	}
 
+	hash := sha256.New()
+	expectedTokenHash := string(hash.Sum([]byte("test")))
+
+	if tokenHash, ok := d.Spec.Template.Annotations["telemeter-token-hash"]; !ok {
+		t.Fatalf("telemeter-token-hash annotation not set in telemeter-client deployment")
+	} else if expectedTokenHash != tokenHash {
+		t.Fatalf("incorrect token hash on telemeter-token-hash annotation, \n got %s, \nwant %s", tokenHash, expectedTokenHash)
+	}
+
 	expectedKubeRbacProxyTLSCipherSuitesArg := fmt.Sprintf("%s%s",
 		KubeRbacProxyTLSCipherSuitesFlag,
 		strings.Join(crypto.OpenSSLToIANACipherSuites(APIServerDefaultTLSCiphers), ","))
@@ -3313,6 +3323,87 @@ func TestTelemeterConfiguration(t *testing.T) {
 	if expectedKubeRbacProxyMinTLSVersionArg != kubeRbacProxyMinTLSVersionArg {
 		t.Fatalf("incorrect TLS version \n got %s, \nwant %s", kubeRbacProxyMinTLSVersionArg, expectedKubeRbacProxyMinTLSVersionArg)
 	}
+}
+
+func TestTelemeterClientSecret(t *testing.T) {
+	for _, tc := range []struct {
+		name                 string
+		config               string
+		existingData         map[string][]byte
+		expectedData         map[string][]byte
+		updateToSaltExpected bool
+	}{
+		{
+			name: "No existing secret",
+			config: `telemeterClient:
+  token: mySecretToken
+`,
+			existingData: map[string][]byte{},
+			expectedData: map[string][]byte{
+				"token": []byte("mySecretToken"),
+			},
+			updateToSaltExpected: true,
+		},
+		{
+			name: "Existing secret, salt gets deleted",
+			config: `telemeterClient:
+  token: mySecretToken
+`,
+			existingData: map[string][]byte{
+				"token": []byte("mySecretToken"),
+			},
+			expectedData: map[string][]byte{
+				"token": []byte("mySecretToken"),
+			},
+			updateToSaltExpected: true,
+		},
+		{
+			name: "Existing secret, secret changes",
+			config: `telemeterClient:
+  token: myNewSecretToken
+`,
+			existingData: map[string][]byte{
+				"token": []byte("mySecretToken"),
+				"salt":  []byte("1234456789ABCDEF"),
+			},
+			expectedData: map[string][]byte{
+				"token": []byte("myNewSecretToken"),
+			},
+			updateToSaltExpected: true,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			c, err := NewConfigFromString(tc.config)
+			if err != nil {
+				t.Fatal(err)
+			}
+			c.UserWorkloadConfiguration = NewDefaultUserWorkloadMonitoringConfig()
+			f := NewFactory("openshift-monitoring", "openshift-user-workload-monitoring", c, defaultInfrastructureReader(), &fakeProxyReader{}, NewAssets(assetsPath), &APIServerConfig{}, &configv1.Console{})
+			generatedS, err := f.TelemeterClientSecret()
+			if err != nil {
+				t.Fatal(err)
+			}
+			byteT, exists := generatedS.Data["token"]
+			newToken := string(byteT)
+			if !exists {
+				t.Fatalf("generated TelemeterClientSecret does not contain a token")
+			}
+			byteS, exists := generatedS.Data["salt"]
+			newSalt := string(byteS)
+			if !exists {
+				t.Fatalf("generated TelemeterClientSecret does not contain a salt")
+			}
+			if string(tc.expectedData["token"]) != newToken {
+				t.Fatalf("generated token is different from expected, expected %s, got %s", tc.expectedData["token"], newToken)
+			}
+			if tc.updateToSaltExpected && string(tc.existingData["salt"]) == newSalt {
+				t.Fatalf("generated salt remain the same expected it to be different, got %s", newSalt)
+			} else if !tc.updateToSaltExpected && string(tc.expectedData["salt"]) != newSalt {
+				t.Fatalf("generated salt is different from expected, expected %s, got %s", tc.expectedData["salt"], newSalt)
+			}
+		})
+	}
+
 }
 
 func TestThanosRulerConfiguration(t *testing.T) {
--- a/pkg/tasks/telemeter.go
+++ b/pkg/tasks/telemeter.go
@@ -21,6 +21,7 @@ import (
 	"github.com/openshift/cluster-monitoring-operator/pkg/client"
 	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
 	"github.com/openshift/cluster-monitoring-operator/pkg/promqlgen"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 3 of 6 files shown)


pkg/client/client.go
```diff
@@ -1674,6 +1674,17 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
 	return int(podCapacityTotal), nil
 }
 
+func (c *Client) GetTelemeterSecretToken(ctx context.Context) (string, error) {
+	currentSecret, err := c.GetSecret(ctx, c.namespace, "telemeter-client")
```
juanrh, line 1678: `func (c *Client) CreateOrUpdateSecret` uses `s.GetNamespace()` that corresponds to the namespace in `assets/telemeter-client/secret.yaml`. Currently `c.namespace` and that namespace are both `openshift-monitoring`, but that could get out of sync at some point.

pkg/manifests/manifests.go
```diff
@@ -3897,22 +3904,24 @@ func (f *Factory) TelemeterClientServiceAccount() (*v1.ServiceAccount, error) {
...
 	if f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token != "" {
 		s.Data["token"] = []byte(f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token)
 	}
 
+	if f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token != secretToken {
+		salt, err := GeneratePassword(32)
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate Telemeter client salt: %v", err)
+		}
+		s.Data["salt"] = []byte(salt)
```
juanrh, line 3922: This only adds salt to the secret if the token has changed. In the [deployment of telemeter client](https://github.com/openshift/telemeter/blob/master/manifests/client/deployment.yaml) the secret is mounted and passed as `--anonymize-salt-file=/etc/telemeter/salt`, and I understand this would lead to a file not found error [here](https://github.com/openshift/telemeter/blob/master/pkg/forwarder/forwarder.go#L126). The Pod could be restarted due to some crash or reschedule, so I think the salt should always be defined in the secret. For example we could modify `(c *Client) GetTelemeterSecretToken` to return both the current token and the current salt and pass them both to this method

pkg/manifests/manifests.go
```diff
@@ -3807,12 +3808,18 @@ func (f *Factory) TelemeterClientPrometheusRule() (*monv1.PrometheusRule, error)
...
 // If the passed ConfigMap is not empty it mounts the Trusted CA Bundle as a VolumeMount to
 // /etc/pki/ca-trust/extracted/pem/ location.
-func (f *Factory) TelemeterClientDeployment(proxyCABundleCM *v1.ConfigMap) (*appsv1.Deployment, error) {
+func (f *Factory) TelemeterClientDeployment(proxyCABundleCM *v1.ConfigMap, s *v1.Secret) (*appsv1.Deployment, error) {
 	d, err := f.NewDeployment(f.assets.MustNewAssetReader(TelemeterClientDeployment))
 	if err != nil {
 		return nil, err
 	}
 
+	// Set annotation on deplyment to trigger redeployments
```
juanrh, line 3817: Nitpick: typo in "deplyment"

pkg/manifests/manifests.go
```diff
@@ -3807,12 +3808,18 @@ func (f *Factory) TelemeterClientPrometheusRule() (*monv1.PrometheusRule, error)
...
+	if s != nil {
+		hash := sha256.New()
+		d.Spec.Template.Annotations["telemeter-token-hash"] = string(hash.Sum(s.Data["token"]))
```
juanrh, line 3820: For what I see in [the documentation for the flags](https://github.com/openshift/telemeter/blob/master/cmd/telemeter-client/main.go#L59) both the salt and the token are sensitive information, so it would be best not to make any of them public, even obfuscated with a hash function. One idea would be to use some kind of version number for this annotation, and increment it when the token has changed. In `func (f *Factory) TelemeterClientSecret` we already identify the token change in `f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token != secretToken`, and we could increment the counter there. One option is storing that number in the secret as `s.Data["revision"]`, which is already available here. That way we are just exposing a counter of how many the token has changed. What do you think?

pkg/client/client.go
JoaoBraveCoding: Correct however I'm not so worried about the namespace as I am with the probability of the name of the secret changing. However, I have a hard time finding a more suitable solution 🤔

pkg/manifests/manifests.go
JoaoBraveCoding: I don't think there is such a huge risk of exposing a secret if it's hashed, specially sha256, however, to try and make it more challenging I've also added the salt to the hash, this way it will make it basically impossible for an attacker to get the secret.

pkg/manifests/manifests.go
JoaoBraveCoding: So I've changed this slightly so that the salt is generated if the token changes or if the salt gets erased since a customer might do it by accident.

juanrh: But if the token has not changed and the salt exists, then `s.Data` will not contain the salt. Then `func (t *TelemeterClientTask) create` will do `err = t.client.CreateOrUpdateSecret(ctx, s)` which will replace the current secret with `s` after doing `mergeMetadata` with the existing secret which takes care of `ObjectMeta` but not the `Data` field. As a result, the secret in the k8s API won't have an entry `Data["salt"]`, which will be a problem if the telemeter pod is restarted. 
As seen in `vendor/k8s.io/client-go/kubernetes/typed/core/v1/secret.go` we have `func (c *secrets) Update` using `c.client.Put()` that is the [API for replacing a Secret object](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#replace-secret-v1-core), this is not updating selected fields as with PATCH

JoaoBraveCoding: You are 💯 correct, I'm sorry I was overlooking this particularity 🤦

pkg/client/client.go
juanrh: What about the following?

- In pkg/manifests/manifests.go

```go
func (f *Factory) TelemeterClientSecretTemplate() (*v1.Secret, error) {
	return f.NewSecret(f.assets.MustNewAssetReader(TelemeterClientSecret))
}

// TelemeterClientSecret generates a new Secret for Telemeter client.
func (f *Factory) TelemeterClientSecret(secretData map[string][]byte) (*v1.Secret, error) {
	secretTemplate, _ := f.TelemeterClientSecretTemplate() // FIXME error handling
	return f.TelemeterClientSecretFromTemplate(secretTemplate, secretData)
}

// TelemeterClientSecret generates a new Secret for Telemeter client.
func (f *Factory) TelemeterClientSecretFromTemplate(secretTemplate *v1.Secret, secretData map[string][]byte) (*v1.Secret, error) {
	/*
	s, err := f.NewSecret(f.assets.MustNewAssetReader(TelemeterClientSecret))
	if err != nil {
		return nil, err
	}*/

	s := secretTemplate
...
```

- In pkg/client/client.go

```go
func (c *Client) GetTelemeterSecretToken(ctx context.Context, secretTemplate *v1.Secret) (map[string][]byte, error) {
	currentSecret, err := c.GetSecret(ctx, secretTemplate.Namespace, "telemeter-client")
...
```

- In pkg/tasks/telemeter.go

```go
	secretTemplate, _ := t.factory.TelemeterClientSecretTemplate() // FIXME error handling
	secretData, err := t.client.GetTelemeterSecretToken(ctx, secretTemplate)
	if err != nil {
		return errors.Wrap(err, "geting Telemeter Client Secret")
	}
...
```

pkg/client/client.go
```diff
@@ -1674,6 +1674,17 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
 	return int(podCapacityTotal), nil
 }
 
+func (c *Client) GetTelemeterSecretToken(ctx context.Context) (map[string][]byte, error) {
```
juanrh, line 1677: At this point `GetTelemeterSecretData` sounds like a more descriptive name, what do you think?

pkg/manifests/manifests.go
jan--f: This won't expose any sensitive information. Since 'crypto/sha256` notes that "Package sha256 implements the SHA224 and SHA256 hash algorithms as defined in FIPS 180-4". So this code here exposes a cryptographic hash sum of the token. This hash sum is not reversible to the original data by any means currently known.

pkg/manifests/manifests.go
```diff
@@ -3896,21 +3903,31 @@ func (f *Factory) TelemeterClientServiceAccount() (*v1.ServiceAccount, error) {
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to generate Telemeter client salt: %v", err)
 	}
 	s.Data["salt"] = []byte(salt)
 
-	if f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token != "" {
-		s.Data["token"] = []byte(f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token)
+	oldToken := string(secretData["token"])
+	oldSalt, saltExists := secretData["salt"]
+	if f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token != oldToken && saltExists {
```
juanrh, line 3929: This is using the old salt when the token is updated, but what we want is "Only update the salt
in the secret when the token is updated", that is:

- if the token is updated then use the new salt: with this code `f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token != oldToken && saltExists` would be `true`, so the old salt would override the new one
- if the token is not updated then use the old salt: with this code `f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token != oldToken && saltExists` would be `false`, so we would use the previous assignment `s.Data["salt"] = []byte(salt)` that employs the new salt

Can we add a unit test for the salt update logic?

pkg/client/client.go
```diff
@@ -1674,6 +1674,17 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
 	return int(podCapacityTotal), nil
 }
 
+func (c *Client) GetTelemeterSecretData(ctx context.Context, s *v1.Secret) (map[string][]byte, error) {
```
jan--f, line 1677:
```suggestion
func (c *Client) GetCurrentTelemeterSecretData(ctx context.Context, s *v1.Secret) (map[string][]byte, error) {
```
Maybe a small name improvement? The current name suggests to me that it just extracts the data field of the passed secret. Really just a nit though.

pkg/tasks/telemeter.go
```diff
@@ -112,7 +112,17 @@ func (t *TelemeterClientTask) create(ctx context.Context) error {
...
 
-	s, err := t.factory.TelemeterClientSecret()
+	sTemplate, err := t.factory.TelemeterClientSecretTemplate()
+	if err != nil {
+		return errors.Wrap(err, "getting Telemeter client Secret template")
+	}
+	
+	sData, err := t.client.GetTelemeterSecretData(ctx, sTemplate)
+	if err != nil {
+		return errors.Wrap(err, "geting Telemeter Client Secret")
```
jan--f, line 122:
```suggestion
		return errors.Wrap(err, "getting Telemeter Client Secret failed")
```

pkg/tasks/telemeter.go
```diff
@@ -112,7 +112,17 @@ func (t *TelemeterClientTask) create(ctx context.Context) error {
 		return errors.Wrap(err, "reconciling Telemeter client Service failed")
 	}
 
-	s, err := t.factory.TelemeterClientSecret()
+	sTemplate, err := t.factory.TelemeterClientSecretTemplate()
+	if err != nil {
+		return errors.Wrap(err, "getting Telemeter client Secret template")
```
jan--f, line 117:
```suggestion
		return errors.Wrap(err, "getting Telemeter client Secret template failed")
```

pkg/manifests/manifests.go
jan--f: I agree with Juan, the update logic seems off.
:+1: if we can add a test for this.

jan--f (changes requested):
Just a few nits.

pkg/manifests/manifests.go
JoaoBraveCoding: I've added tests 👌

pkg/manifests/manifests_test.go
```diff
@@ -3315,6 +3325,102 @@ func TestTelemeterConfiguration(t *testing.T) {
...
+			if !exists {
+				t.Fatalf("generated TelemeterClientSecret does not contain a token")
+			}
+			byteS, exists := generatedS.Data["salt"]
+			newSalt := string(byteS)
+			if !exists {
+				t.Fatalf("generated TelemeterClientSecret does not contain a salt")
+			}
+			if string(tc.expectedData["token"]) != newToken {
+				t.Fatalf("generated token is diferent from expected, expected %s, got %s", tc.expectedData["token"], newToken)
```
juanrh, line 3412: nit: "diferent" -> "different"

pkg/manifests/manifests_test.go
```diff
@@ -3315,6 +3325,102 @@ func TestTelemeterConfiguration(t *testing.T) {
...
+			}
+			if tc.tokenChanged && string(tc.existingData["salt"]) == newSalt {
+				t.Fatalf("generated salt remain the same expected it to be different, got %s", newSalt)
+			} else if !tc.tokenChanged && string(tc.expectedData["salt"]) != newSalt {
+				t.Fatalf("generated salt is diferent from expected, expected %s, got %s", tc.expectedData["salt"], newSalt)
```
juanrh, line 3417: nit: "diferent" -> "different"

pkg/manifests/manifests_test.go
```diff
@@ -3315,6 +3325,102 @@ func TestTelemeterConfiguration(t *testing.T) {
...
+			config: `telemeterClient:
+  token: mySecretToken
+`,
+			existingData: map[string][]byte{
+				"token": []byte("mySecretToken"),
+			},
+			expectedData: map[string][]byte{
+				"token": []byte("mySecretToken"),
+			},
+			tokenChanged: true,
```
juanrh, line 3373: Nit: I understand you need this field because the salt is randomly generated so you cannot add the expected value of the salt in the test data. But it is a bit misleading having `tokenChanged = true` here because the token hasn't changed in this case. Renaming `tokenChanged` to something like `saltUpdateExpected` would be more readable

jan--f:
/lgtm
Thanks Joao!

pkg/client/client.go
```diff
@@ -1674,6 +1674,17 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
 	return int(podCapacityTotal), nil
 }
 
+func (c *Client) GetCurrentTelemeterSecretData(ctx context.Context, s *v1.Secret) (map[string][]byte, error) {
```
simonpasquier, line 1677: there's nothing really specific to the telemeter secret here. I'd advise to either have a more generic name and return a *v1.Secret object or use the existing GetSecret() method and move the `apierrors.IsNotFound` test to the task.

pkg/manifests/manifests.go
simonpasquier: still relevant

pkg/manifests/manifests.go
```diff
@@ -3896,21 +3903,31 @@ func (f *Factory) TelemeterClientServiceAccount() (*v1.ServiceAccount, error) {
...
 // TelemeterClientSecret generates a new Secret for Telemeter client.
-func (f *Factory) TelemeterClientSecret() (*v1.Secret, error) {
-	s, err := f.NewSecret(f.assets.MustNewAssetReader(TelemeterClientSecret))
+func (f *Factory) TelemeterClientSecret(secretData map[string][]byte) (*v1.Secret, error) {
+	s, err := f.TelemeterClientSecretTemplate()
 	if err != nil {
 		return nil, err
 	}
 
+	if f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token != "" {
```
simonpasquier, line 3917: I'd rather move the check to the task and decide there whether we need to update the secret or not (e.g. if the existing token value is equal to the configured token then do nothing otherwise update the secret). This would leave this method untouched.

simonpasquier (commented):
/hold

pkg/client/client.go
JoaoBraveCoding: Yes, it makes sense, after the reviews, it changed a lot. I've ended up just using `GetSecret`

pkg/manifests/manifests_test.go
```diff
@@ -3315,6 +3325,87 @@ func TestTelemeterConfiguration(t *testing.T) {
 	}
 }
 
+func TestTelemeterClientSecret(t *testing.T) {
```
JoaoBraveCoding, line 3328: I've maintained these tests but with the new e2e tests I'm not so sure it's worth it to test this function so much. Let me know if you want me to remove it

pkg/manifests/manifests.go
simonpasquier, line 3817:
```suggestion
	// Set annotation on deployment to trigger redeployments
```

pkg/tasks/telemeter.go
```diff
@@ -112,11 +113,25 @@ func (t *TelemeterClientTask) create(ctx context.Context) error {
...
+
 	s, err := t.factory.TelemeterClientSecret()
 	if err != nil {
 		return errors.Wrap(err, "initializing Telemeter client Secret failed")
 	}
 
+	if string(oldS.Data["token"]) == t.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token {
+		s = oldS
+	}
+
```
simonpasquier, line 134: It would be more concise to get the existing secret and copy the `Data` field from the actual secret to the generated one. We also need to check for nil values.

```suggestion
	s, err := t.factory.TelemeterClientSecret()
	if err != nil {
		return errors.Wrap(err, "initializing Telemeter client Secret failed")
	}
	
	oldS, err := t.client.GetSecret(ctx, s.Namespace, s.Name)
	if err != nil && !apierrors.IsNotFound(err) {
		return errors.Wrap(err, "getting Telemeter Client Secret failed")
	}

	if oldS != nil && string(oldS.Data["token"]) == t.config.ClusterMonitoringConfiguration.TelemeterClientConfig.Token {
		s.Data = oldS.Data
	}

```

test/e2e/config_test.go
```diff
@@ -329,6 +329,61 @@ func TestClusterMonitorTelemeterClientConfig(t *testing.T) {
...
 
+func TestTelemeterClientSecret(t *testing.T) {
+	for _, tc := range []struct {
+		name      string
+		oldC      string
+		newC      string
+		assertion framework.AssertionFunc
+	}{
+		{
+			name: "No existing secret",
```
simonpasquier, line 340: The secret would already exist since it's populated from the global pull secret. Not sure how this test is different from `Existing Secret, new token`.

pkg/manifests/manifests_test.go
simonpasquier: It's fine for me to keep the unit tests since they existed before the e2e test.

simonpasquier (commented):
/lgtm

PR #1752, opened 2022-08-26
OCPBUGS-1364: Dedicated kubelet ServiceMonitor for prometheus-adapter

jan--f:
This adds an optional kubelet service monitor for prometheus-adapter, that sets `honorTimestamps: true` in order to improve `oc adm top pod` and HPA data consistency.
Please refer to the commit messages for more details.

https://issues.redhat.com/browse/MON-1949

### DIFF
--- a/assets/control-plane/service-monitor-kubelet-prometheus-adapter.yaml
+++ b/assets/control-plane/service-monitor-kubelet-prometheus-adapter.yaml
@@ -1,0 +1,60 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/name: kubelet
+    app.kubernetes.io/part-of: openshift-monitoring
+    k8s-app: kubelet
+  name: kubelet-for-pa
+  namespace: openshift-monitoring
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    honorLabels: true
+    honorTimestamps: true
+    interval: 30s
+    metricRelabelings:
+    - action: drop
+      regex: container_(network_tcp_usage_total|network_udp_usage_total|tasks_state|cpu_load_average_10s)
+      sourceLabels:
+      - __name__
+    - action: drop
+      regex: (container_spec_.*|container_file_descriptors|container_sockets|container_threads_max|container_threads|container_start_time_seconds|container_last_seen);;
+      sourceLabels:
+      - __name__
+      - pod
+      - namespace
+    - action: drop
+      regex: (container_blkio_device_usage_total);.+
+      sourceLabels:
+      - __name__
+      - container
+    - action: keep
+      regex: container_cpu_usage_seconds_total|container_memory_working_set_bytes
+      sourceLabels:
+      - __name__
+    - action: replace
+      replacement: pa_$1
+      sourceLabels:
+      - __name__
+      targetLabel: __name__
+    path: /metrics/cadvisor
+    port: https-metrics
+    relabelings:
+    - sourceLabels:
+      - __metrics_path__
+      targetLabel: metrics_path
+    scheme: https
+    scrapeTimeout: 30s
+    tlsConfig:
+      caFile: /etc/prometheus/configmaps/kubelet-serving-ca-bundle/ca-bundle.crt
+      certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
+      insecureSkipVerify: false
+      keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
+  jobLabel: k8s-app
+  namespaceSelector:
+    matchNames:
+    - kube-system
+  selector:
+    matchLabels:
+      k8s-app: kubelet
--- a/assets/control-plane/service-monitor-kubelet-prometheus-adapter.yaml
+++ b/assets/control-plane/service-monitor-kubelet-prometheus-adapter.yaml
@@ -14,21 +14,6 @@ spec:
     honorTimestamps: true
     interval: 30s
     metricRelabelings:
-    - action: drop
-      regex: container_(network_tcp_usage_total|network_udp_usage_total|tasks_state|cpu_load_average_10s)
-      sourceLabels:
-      - __name__
-    - action: drop
-      regex: (container_spec_.*|container_file_descriptors|container_sockets|container_threads_max|container_threads|container_start_time_seconds|container_last_seen);;
-      sourceLabels:
-      - __name__
-      - pod
-      - namespace
-    - action: drop
-      regex: (container_blkio_device_usage_total);.+
-      sourceLabels:
-      - __name__
-      - container
     - action: keep
       regex: container_cpu_usage_seconds_total|container_memory_working_set_bytes
       sourceLabels:
@@ -38,7 +23,7 @@ spec:
       sourceLabels:
       - __name__
       targetLabel: __name__
-    path: /metrics/cadvisor
+    path: /metrics/resource
     port: https-metrics
     relabelings:
     - sourceLabels:
--- a/jsonnet/components/control-plane.libsonnet
+++ b/jsonnet/components/control-plane.libsonnet
@@ -164,6 +164,7 @@ function(params)
             ,
             function(e)
               e {
+                path: '/metrics/resource',
                 tlsConfig+: {
                   caFile: '/etc/prometheus/configmaps/kubelet-serving-ca-bundle/ca-bundle.crt',
                   insecureSkipVerify: false,
@@ -175,7 +176,7 @@ function(params)
                 // because the kubelet metric endpoints might take more than the default
                 // 10 seconds to reply.
                 scrapeTimeout: '30s',
-                metricRelabelings+: [
+                metricRelabelings: [
                   {
                     sourceLabels: ['__name__'],
                     action: 'keep',
--- a/jsonnet/components/control-plane.libsonnet
+++ b/jsonnet/components/control-plane.libsonnet
@@ -142,7 +142,7 @@ function(params)
     },
 
     // This adds a kubelet ServiceMonitor for special use with
-    // prometheus-adapter
+    // prometheus-adapter if enabled by the configuration of the cluster monitoring operator.
     serviceMonitorKubeletResourceMetrics: self.serviceMonitorKubelet {
       metadata+: {
         name: 'kubelet-resource-metrics',
@@ -161,8 +161,9 @@ function(params)
                   {
                     sourceLabels: ['__name__'],
                     action: 'keep',
-                    regex: 'container_cpu_usage_seconds_total|container_memory_working_set_bytes',
+                    regex: 'container_cpu_usage_seconds_total|container_memory_working_set_bytes|scrape_error',
                   },
+                   // To avoid clashes with the cAdvisor metrics, the resource metrics are prefixed with a distinct identifier.
                   {
                     sourceLabels: ['__name__'],
                     targetLabel: '__name__',
--- a/jsonnet/components/control-plane.libsonnet
+++ b/jsonnet/components/control-plane.libsonnet
@@ -156,17 +156,7 @@ function(params)
             function(e)
               e {
                 path: '/metrics/resource',
-                tlsConfig+: {
-                  caFile: '/etc/prometheus/configmaps/kubelet-serving-ca-bundle/ca-bundle.crt',
-                  insecureSkipVerify: false,
-                  certFile: '/etc/prometheus/secrets/metrics-client-certs/tls.crt',
-                  keyFile: '/etc/prometheus/secrets/metrics-client-certs/tls.key',
-                },
                 honorTimestamps: true,
-                // Increase the scrape timeout to match the scrape interval
-                // because the kubelet metric endpoints might take more than the default
-                // 10 seconds to reply.
-                scrapeTimeout: '30s',
                 metricRelabelings: [
                   {
                     sourceLabels: ['__name__'],
--- a/jsonnet/components/control-plane.libsonnet
+++ b/jsonnet/components/control-plane.libsonnet
@@ -143,20 +143,11 @@ function(params)
 
     // This adds a kubelet ServiceMonitor for special use with
     // prometheus-adapter
-    serviceMonitorKubeletPrometheusAdapter: super.serviceMonitorKubelet + {
+    serviceMonitorKubeletResourceMetrics: self.serviceMonitorKubelet {
       metadata+: {
-        labels+: {
-          'k8s-app': 'kubelet',
-        },
-        name: 'kubelet-for-pa',
+        name: 'kubelet-resource-metrics',
       },
       spec+: {
-        jobLabel: 'k8s-app',
-        selector: {
-          matchLabels: {
-            'k8s-app': 'kubelet',
-          },
-        },
         endpoints:
           std.filterMap(
             function(e)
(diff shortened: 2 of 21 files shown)


jan--f:
To test this, deploy the change to an openshift cluster and create a pod with a defined load, like:
``` yaml
kind: Pod
apiVersion: v1
metadata:
  name: cpu-demo
  namespace: default
spec:
  containers:
    - name: cpu-demo-ctr
      image: vish/stress
      args:
        - '-cpus'
        - '2'
      resources:
        limits:
          cpu: '1'
        requests:
          cpu: 500m
```
In a separate terminal watch the `oc adm top pod` output
``` bash
watch -n1 oc adm top pod
```
Note how the output fluctuates around the target value of 1.

Then configure the new feature by applying this config fragment to the `cluster-monitoring-config`:
``` yaml
kind: ConfigMap
apiVersion: v1
metadata:
data:
  config.yaml: |-
    k8sPrometheusAdapter:
      dedicatedServiceMonitors:
        enabled: true
...(additional config)
```

Once everything is reconciled, the `oc adm top pod` output should become uniform and very close to 1.

When this is enabled we ingest an extra 2 metrics with time series for every pod. On cluster with large workloads, i.e. many pods and/or high pod turnover, this will cause additional resource overhead.

pkg/tasks/controlplane.go
```diff
@@ -55,6 +56,23 @@ func (t *ControlPlaneTask) Run(ctx context.Context) error {
...
+
+	if *t.config.ClusterMonitoringConfiguration.K8sPrometheusAdapter.DedicatedServiceMonitors.Enabled {
+		err = t.client.CreateOrUpdateServiceMonitor(ctx, smkpa)
+		if err != nil {
+			return errors.Wrap(err, "reconcilin prometheus-adapater dedicatedg control-plane kubelet ServiceMonitor failed")
+		}
+	} else {
+		err = t.client.DeleteServiceMonitor(ctx, smkpa)
+		if err != nil {
+			return errors.Wrap(err, "deleting prometheus-adapater dedicated control-plane etcd ServiceMonitor failed")
```
JoaoBraveCoding, line 72: I'm not 100% sure, but aren't we going to land here every time the DedicatedServiceMonitors are not enabled? Shouldn't we exclude the IfNotFound error?

pkg/tasks/prometheusadapter.go
```diff
@@ -159,6 +150,38 @@ func (t *PrometheusAdapterTask) Run(ctx context.Context) error {
...
+		if err != nil {
+			return errors.Wrap(err, "initializing PrometheusAdapter ConfigMap failed")
+		}
+		if *t.config.ClusterMonitoringConfiguration.K8sPrometheusAdapter.DedicatedServiceMonitors.Enabled {
+			err = t.client.CreateOrUpdateConfigMap(ctx, cmD)
+			if err != nil {
+				return errors.Wrap(err, "reconciling PrometheusAdapter ConfigMap for dedicated ServiceMonitors failed")
+			}
+			err = t.client.DeleteConfigMap(ctx, cm)
+			if err != nil {
```
JoaoBraveCoding, line 167: Same as previous comment for the deletes

pkg/tasks/prometheusadapter.go
```diff
@@ -159,6 +150,38 @@ func (t *PrometheusAdapterTask) Run(ctx context.Context) error {
 		}
 	}
 	{
+		cmD, err := t.factory.PrometheusAdapterConfigMapDedicated()
```
JoaoBraveCoding, line 153: Also, random idea, what if both ConfigMaps had the same name? This way we would not have to go through the update/delete.
Note that in this approach we would have to add an annotation to the deployments with a hash of the config to trigger redeployments.

pkg/tasks/controlplane.go
jan--f: Yes and we do: https://github.com/openshift/cluster-monitoring-operator/blob/ba2f4472bc07989f459cbf5f299bac8100d193df/pkg/client/client.go#L742-L756

JoaoBraveCoding: Ahhh the check was there, disregard then

pkg/tasks/prometheusadapter.go
jan--f: Not sure I understand the advantage of this approach :thinking: 

Wouldn't we still have to update the ConfigMap and pass data to `t.factory.PrometheusAdapterDeployment`?
Part of the motivation to change the ConfigMap name in the prometheus-adapter Deployment was that this makes it immediately obvious which config is used.

JoaoBraveCoding: If I'm thinking correctly the advantage would be that instead of a Create&Delete we would only need to call CreateOrUpdate, which would make this part of the code consistent with other areas of the operator.

However, if one of the goals of the change was to make it obvious what configuration was being used then this would be the way to go.

assets/control-plane/service-monitor-kubelet-prometheus-adapter.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - action: keep
+      regex: container_cpu_usage_seconds_total|container_memory_working_set_bytes
+      sourceLabels:
+      - __name__
+    - action: replace
+      replacement: pa_$1
+      sourceLabels:
+      - __name__
+      targetLabel: __name__
+    path: /metrics/cadvisor
```
simonpasquier, line 41: Targeting `/metrics/resources` should be more efficient on Prometheus side because we won't have to drop any metrics. It might also be less demanding on the kubelet side.

https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/

simonpasquier: the other advantage is that we can scope the prometheus-adapter's queries using the `metrics_path` label (e.g. `container_cpu_usage_seconds_total{metrics_path="/metrics/resources",...}`)

simonpasquier: From what I understand reading the kubelet code base, the `/metrics/resource` handler should be more lightweight than cAdvisor in terms of resource usage.

simonpasquier: yes there's a risk but we should be able to find which recording rules need to be fixed. At the same time, having a different metric name is indeed more obvious.

jan--f: I added the `/metrics/resource` endpoint but kept the metrics prefix. wdyt?

jsonnet/components/control-plane.libsonnet
```diff
@@ -141,6 +141,61 @@ function(params)
       },
     },
 
+    // This adds a kubelet ServiceMonitor for special use with
+    // prometheus-adapter
+    serviceMonitorKubeletPrometheusAdapter: super.serviceMonitorKubelet + {
```
simonpasquier, line 146: Using `self` doesn't require to override other fields like `labels`, `selector`, etc.
```suggestion
    serviceMonitorKubeletPrometheusAdapter: self.serviceMonitorKubelet + {
```

pkg/tasks/prometheusadapter.go
simonpasquier: I agree that having 2 different configmaps is more obvious.
Semi-related question: why not have the 2 configmaps always present and only deal with selecting the correct one in the adapter deployment? The advantage would be less code branches.

jsonnet/components/prometheus-adapter.libsonnet
```diff
@@ -19,7 +19,89 @@ function(params)
...
+
   pa {
+    configMapDedicatedServiceMonitors: {
+      apiVersion: 'v1',
+      kind: 'ConfigMap',
+      metadata: pa._metadata {
+        name: 'adapter-config-dedicated-sm',
+      },
+      data: { 'config.yaml': std.manifestYamlDoc(config) },
+    },
```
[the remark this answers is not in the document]
jan--f: Oh yeah good point. The current version somewhat anticipated getting https://github.com/prometheus-operator/kube-prometheus/pull/1844 merged to make this go away. But lets use the patching approach until this maybe merges.

pkg/manifests/config.go
```diff
@@ -305,6 +307,10 @@ type Audit struct {
 	Profile auditv1.Level `json:"profile"`
 }
 
+type DedicatedServiceMonitors struct {
+	Enabled *bool `json:"enabled"`
```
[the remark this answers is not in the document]
jan--f: Agreed, I actually specifically implemented this as a pointer since this seemed to be the convention in the config. But `EtcdConfig` was a poor example to go by :sweat_smile:

jan--f:
> I agree that having 2 different configmaps is more obvious.
Semi-related question: why not have the 2 configmaps always present and only deal with selecting the correct one in the adapter deployment? The advantage would be less code branches.

My thinking was that having a config map that is inert might be confusing. If anyone went looking for prometheus-adapter configs it might be easy to get them confused, so I though cleaning up what is not used would be better even if that means a slight increase in code complexity.

simonpasquier:
> My thinking was that having a config map that is inert might be confusing. If anyone went looking for prometheus-adapter configs it might be easy to get them confused, so I though cleaning up what is not used would be better even if that means a slight increase in code complexity.

Sounds good to me.

jsonnet/components/control-plane.libsonnet
```diff
@@ -141,6 +141,52 @@ function(params)
...
+                // Increase the scrape timeout to match the scrape interval
+                // because the kubelet metric endpoints might take more than the default
+                // 10 seconds to reply.
+                scrapeTimeout: '30s',
+                metricRelabelings: [
+                  {
+                    sourceLabels: ['__name__'],
+                    action: 'keep',
+                    regex: 'container_cpu_usage_seconds_total|container_memory_working_set_bytes',
+                  },
```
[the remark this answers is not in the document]
jan--f: Based on https://github.com/kubernetes/enhancements/pull/726/ I was under the impression the resource endpoint also exposes `node_` metrics, so I left this step in.

jan--f:
> I'm not sure that prometheus-operator/kube-prometheus#1844 is enough because we want to generate both configmaps anyway.

I think my assumption was, that I would simply create a second config, something along the lines of
``` jsonnet
  local pa2 = prometheusAdapter(cfg + {containerMetricsPrefix: params.prometheusAdapterMetricPrefix});

  pa2 {
    configMapDedicatedServiceMonitors: pa2.configMap {
      metadata+: {
        name: adapter-config-dedicated-sm',
      },
    },
  }
```
I haven't tested that yet and on second thought this is certainly not very efficient.

jsonnet/components/control-plane.libsonnet
```diff
@@ -141,6 +141,52 @@ function(params)
...
+            ,
+            function(e)
+              e {
+                path: '/metrics/resource',
+                tlsConfig+: {
+                  caFile: '/etc/prometheus/configmaps/kubelet-serving-ca-bundle/ca-bundle.crt',
+                  insecureSkipVerify: false,
+                  certFile: '/etc/prometheus/secrets/metrics-client-certs/tls.crt',
+                  keyFile: '/etc/prometheus/secrets/metrics-client-certs/tls.key',
+                },
```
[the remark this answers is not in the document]
jan--f: Nope that is indeed unnecessary.

pkg/manifests/config.go
```diff
@@ -291,6 +291,8 @@ type K8sPrometheusAdapter struct {
 
 	// Prometheus Adapter audit logging related configuration
 	Audit *Audit `json:"audit"`
+
+	DedicatedServiceMonitors *DedicatedServiceMonitors `json:"dedicatedServiceMonitors"`
```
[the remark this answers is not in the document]
jan--f: Yes, thanks for the reminder!

jsonnet/components/control-plane.libsonnet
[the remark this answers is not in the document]
simonpasquier: ah right!

jsonnet/components/control-plane.libsonnet
```diff
@@ -141,6 +141,42 @@ function(params)
       },
     },
 
+    // This adds a kubelet ServiceMonitor for special use with
+    // prometheus-adapter
```
simonpasquier, line 145:
```suggestion
    // prometheus-adapter if enabled by the configuration of the cluster monitoring operator.
```

simonpasquier (commented):
I guess that we want an end-to-end test to validate the feature?

jan--f:
> I guess that we want an end-to-end test to validate the feature?

Good point, thanks for the reminder. I extended the existing tests to cover both config variants. Do you think we need additional testing?

jsonnet/components/control-plane.libsonnet
```diff
@@ -141,6 +141,43 @@ function(params)
...
+          std.filterMap(
+            function(e)
+              'path' in e && e.path == '/metrics/cadvisor'
+            ,
+            function(e)
+              e {
+                path: '/metrics/resource',
+                honorTimestamps: true,
+                metricRelabelings: [
+                  {
```
simonpasquier, line 161:
```suggestion
                  // Keep container_cpu_usage_seconds_total and container_memory_working_set_bytes metrics, this is all that the Prometheus adapter needs. Node metrics are provided by node_exporter (Linux) or Windows exporter.
                  // The scrape_error metric can be useful for troubleshooting.
                  {
```

simonpasquier (commented):
Formal lgtm from my side. Just asking for a small comment on the metric relabel config for future ourselves :)

jsonnet/components/control-plane.libsonnet
jan--f: :+1: Lets be friendly to our future selves.

jsonnet/components/control-plane.libsonnet
```diff
@@ -141,6 +141,42 @@ function(params)
...
+                    sourceLabels: ['__name__'],
+                    action: 'keep',
+                    regex: 'container_cpu_usage_seconds_total|container_memory_working_set_bytes',
```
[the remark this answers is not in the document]
jan--f: Hmm after thinking about this a little more: isn't this the same `scrape_error` metric we already get from the control-plane ServiceMonitor? I.e. it will just duplicate the "Failed to scrape container metrics" signal, since it scrapes the same kubelet instance?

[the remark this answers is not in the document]
simonpasquier: I don't think so, my understanding is that kubelet will return `scrape_error 1` if it fails to gather container metrics for some reason while it will still return 200 OK to Prometheus meaning `up 1`.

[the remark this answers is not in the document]
simonpasquier: sorry I misread your comment but I think it's still useful to collect scrape_error for both service monitors. Also it's one series per kubelet so not a cardinality concern.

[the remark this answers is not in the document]
jan--f: I suppose I'm not quite clear what this metric means :thinking: 
I was under the impression this signals that kubelet is unable to get container metrics, so this would just add the same information since its the same kubelet we are scraping.
But its actually related to the service monitor?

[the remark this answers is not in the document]
simonpasquier: no it's indeed related to the kubelet collecting container metrics but I wouldn't necessarily assume that both endpoints /metrics/cadvisor and /metrics/resource return the same value (it will probably but who knows :shrug:).

jan--f:
/retitle OCPBUGS-1364: Dedicated kubelet ServiceMonitor for prometheus-adapter

simonpasquier (commented):
/lgtm

PR #1755, opened 2022-08-29
Bug 2111345: go.mod: update openshift-api to current release-4.11 tip

jan--f:


juzhao:
tested with PR, value for actions in AlertRelabelConfig CRD is changed to PascalCase
/label qe-approved

PR #1760, opened 2022-09-05
OCPBUGS-924: increase alertmanager's startupProbe failure threshold

sthaha:
On some clusters, during the initial boot, DNS resolution of services
may fail and this can cause Alertmanager pods to fail, which in turn
causes CMO to be in degraded state.

This patch fixes the issue by increasing the startupProbe's tolerance.

NOTE: The actual issue in the Bug[1] seems to be DNS related where
alertmanager fails to establish a quorum since its peers can't be
reached.

[1]: https://issues.redhat.com/browse/OCPBUGSM-47741

---

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -505,9 +505,13 @@ func (f *Factory) AlertmanagerMain(trustedCABundleCM *v1.ConfigMap) (*monv1.Aler
 		// returns (20s is twice the time that Alertmanager waits before
 		// declaring that it can start sending notfications).
 		//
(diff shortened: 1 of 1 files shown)


simonpasquier:
/label backport-risk-assessed

According to https://docs.google.com/document/d/1qbKUl4K9PPW4tCbdDgWSXd5klosX77ifkxQLqzjztGE/edit#heading=h.mml0rz88o0nj you need to recreate the parent bug in the OCPBUGS Jira project and clone up to the 4.10 release.

PR #1762, opened 2022-09-06
[release-4.9] OCPBUGS-937: Give precedence to CMO config map proxy config

juanrh:
Problem: Customers cannot configure a proxy for telemeter
client without setting up the cluster wide proxy.

Solution: Give precedence to CMO config map proxy config
over cluster proxy config for the proxy configuration,
otherwise cluster proxy config is always used, because
openshift.io/v1/Proxy/cluster is always present.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2116382

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -522,6 +522,22 @@ func (o *Operator) enqueue(obj interface{}) {
 	o.queue.Add(key)
 }
 
+type proxyConfigSupplier func(context.Context) (*ProxyConfig, error)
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
/lgtm

juzhao:
tested with the PR, set http parameter in cluster-monitoring-config configmap, telemeter client would use the setting to forward metrics to telemeter server. passed testing
/label qe-approved

PR #1763, opened 2022-09-07
Add odf_system_objects_total and odf_system_bucket_count

weirdwiz:
This commit adds two new metrics for object and bucket count, this
further deprecates the Noobaa metrics for the same.

Jira Issue: https://issues.redhat.com/browse/RHSTOR-3471


weirdwiz:
PTAL @simonpasquier

jan--f:
I'm afraid the label sets of these metrics are unsuitable to ship to telemetry. Given these definitions:
```
    - expr: |
        sum by (namespace, managedBy, job, service) (ocs_objectbucket_objects_total)
      labels:
        system_type: OCS
        system_vendor: Red Hat
      record: odf_system_objects_total
    - expr: |
        ocs_objectbucket_count_total
      labels:
        system_type: OCS
        system_vendor: Red Hat
      record: odf_system_bucket_count
```
this would ship two unneeded (I assume) labels to telemetry (`system_type` and `system_vendor`).
Also the label set of `odf_system_objects_total` seems to have the potential to create an unbounded value set.
You'll need convincing arguments why you need all these labels in telemetry.

weirdwiz:
@jan--f the other `odf_system` metrics contain the same labels `system_type` and `system_vendor`.

The labels for `odf_system_objects_total`, only have a few possible values as the operator is only installed in one namespace, and the value of managedBy, job, and service will only have fixed values per operator. But we can reduce the label set to just `job`.

wdyt @mulbc

simonpasquier:
If the system_type and system_vendor values are fixed, we could even add them as additional selectors to the telemeter config which would avoid any risk of shipping unwanted metrics:

```
- '{__name__="odf_system_objects_total",system_type="OCS",system_vendor="Red Hat"}'
- '{__name__="odf_system_bucket_count",system_type="OCS",system_vendor="Red Hat"}'
``` 

> The labels for odf_system_objects_total, only have a few possible values as the operator is only installed in one namespace, and the value of managedBy, job, and service will only have fixed values per operator. But we can reduce the label set to just job.

What if the odf_system_objects_total and odf_system_bucket_count recording rules selected exactly that namespace? E.g. assuming that the namespace is openshift-storage

```
sum (ocs_objectbucket_objects_total{namespace="openshift-storage"})
```

mulbc:
@weirdwiz - please implement what @simonpasquier suggested, it's fine by me

in the ocs-operator and noobaa code you use `sum by (namespace, job, service)` or `sum by (namespace, managedBy, job, service)` - please just remove the by-clause so that is reads 

```
sum (ocs_objectbucket_objects_total)
      labels:
        system_type: OCS
        system_vendor: Red Hat
      record: odf_system_objects_total
```

for example
As discussed with @jan--f this will allow them to accept this PR

Documentation/data-collection.md
```diff
@@ -361,15 +361,27 @@ data:
...
     # owners: (@openshift/team-ocs-committers)
     #
     # job:noobaa_total_object_count:sum is the total number of noobaa objects.
+    # This metric is deprecated in ODF 4.12, refer to odf_system_objects_total instead.
     - '{__name__="job:noobaa_total_object_count:sum"}'
     #
     # owners: (@openshift/team-ocs-committers)
     #
+    # odf_system_bucket_count is the total number of buckets in ODF system
+    - '{__name__="odf_system_bucket_count"}'
```
jan--f, line 376:
```suggestion
    - '{__name__="odf_system_bucket_count",system_type="OCS",system_vendor="Red Hat"}'
```

Documentation/data-collection.md
```diff
@@ -361,15 +361,27 @@ data:
...
+    #
+    # owners: (@openshift/team-ocs-committers)
+    #
+    # odf_system_objects_total is the total number of objects in ODF system
+    - '{__name__="odf_system_objects_total"}'
```
jan--f, line 381:
```suggestion
    - '{__name__="odf_system_objects_total",system_type="OCS",system_vendor="Red Hat"}'
```

jan--f (changes requested):
Lets also add the additional guarding for the static label values that Simon suggested.

weirdwiz:
PR up for OCS and Noobaa Mixins. 

https://github.com/red-hat-storage/ocs-operator/pull/1802
https://github.com/noobaa/noobaa-mixins/pull/30

jan--f:
/label docs-approved
/label px-approved
/label qe-approved
Adding labels for a telemetry addition.

weirdwiz:
@jan--f all the downstream PRs are merged now.

PR #1764, opened 2022-09-08
update to release.openshift.io/feature-set to match OCP 4.12

deads2k:
This updates the operator openshift/api master to pick up the updated feature-set string.  In so doing, it must bring in kube 1.25.0, which I've done. Minor fitting was required in a mock in order to build.

This was added in https://github.com/openshift/cluster-version-operator/pull/821 to allow more featuresets and allow for a future migration to include actual gates. Because usage of this manifest would prevent upgrades, there are not upgrade concerns for this change.

/cc @wking


deads2k:
> /retest

I think I got the generator problem,  but I don't see how to fix the vendor job.

jan--f:
> > /retest
> 
> I think I got the generator problem, but I don't see how to fix the vendor job.

Looks like we're running into pruning issues as outlined in https://go.dev/doc/go1.17#go-command.

We need to either change our Makefile to run `go mod tidy -compat=1.17` or update our module to go1.18.
Updating seems reasonable to me, @simonpasquier mind chiming in here?

jan--f:
@deads2k after a rebase and `make generate && make vendor` this should be good to go.
/label docs-approved
/label px-approved
/label qe-approved
Adding labels for a version bump of api dependencies.

PR #1765, opened 2022-09-09
[release-4.8] OCPBUGS-1098: Give precedence to CMO config map proxy config

juanrh:
Problem: Customers cannot configure a proxy for telemeter client without setting up the cluster wide proxy.

Solution: Give precedence to CMO config map proxy config over cluster proxy config for the proxy configuration, otherwise cluster proxy config is always used, because openshift.io/v1/Proxy/cluster is always present.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2112381

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -421,6 +421,22 @@ func (o *Operator) enqueue(obj interface{}) {
 	o.queue.Add(key)
 }
 
+type proxyConfigSupplier func() (*ProxyConfig, error)
+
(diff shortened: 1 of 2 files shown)


juanrh:
/retitle OCPBUGS-1098: Give precedence to CMO config map proxy config

juanrh:
/retitle [release-4.8] OCPBUGS-1098: Give precedence to CMO config map proxy config

PR #1767, opened 2022-09-13
OSD-13070: Fixed TargetDown expression to join on the proper label.

thavlice:
The node label does not exist for the 'up' metric: switched to the 'pod' label to join the proper kube-state-metrics metric.


Nikokolas3270:
Hello @simonpasquier, thanks for your feedback in your previous message.
We (me & @thavlice) changed `jsonnet/rules.libsonnet` file as you advised and re-generated file `assets/cluster-monitoring-operator/prometheus-rule.yaml` by running `make generate`.
Could you please take a look at the PR again?
Also we have 2 questions:
* Why is tracking file `assets/cluster-monitoring-operator/prometheus-rule.yaml` (in Git repo) needed if it can be generated at will from `jsonnet/rules.libsonnet`?
* I opened an OCPBUGS as requested: [OCPBUGS-1453](https://issues.redhat.com/browse/OCPBUGS-1453). Should we edit PR title & commit description to reference it (if "yes", @thavlice: can you please take care of it)
Thanks & best regards, Nicolas

simonpasquier:
> * Why is tracking file `assets/cluster-monitoring-operator/prometheus-rule.yaml` (in Git repo) needed if it can be generated at will from `jsonnet/rules.libsonnet`?

It's easier to review changes by looking at the YAML manifests that the jsonnet code + it is a requirement for the downstream builds (downloading the jsonnet dependencies isn't possible in the context of OCP builds).

> 
> * I opened an OCPBUGS as requested: [OCPBUGS-1453](https://issues.redhat.com/browse/OCPBUGS-1453). Should we edit PR title & commit description to reference it (if "yes", @thavlice: can you please take care of it)

/retitle OCPBUGS-1453: Fixed TargetDown expression to join on the proper label

Nikokolas3270:
> /hold
> 
> I think we need to explore possible edge cases. Like if kube-state-metrics is down, we'll have no `kube_pod_info` metric at all and the alert would never fire.

Indeed, that's an interesting point. Maybe we should "join" with `unless` clause instead of `on` clause.

Nikokolas3270:
@simonpasquier, @JoaoBraveCoding & @thavlice 
I propose the following prometheus expression to leverage the fact that `kube-state-metrics` may eventually be done:
```
100 *(count(up == 0 and on(pod) kube_pod_info) BY (job, namespace, service) /
      count(up      and on(pod) kube_pod_info) BY (job, namespace, service)) > 10  
or
100 *(count(up == 0 and on() absent(up{service='kube-state-metrics'})) BY (job, namespace, service) /
      count(up      and on() absent(up{service='kube-state-metrics'})) BY (job, namespace, service)) > 10
```
I will provide some argumentation in [OHSS-14889](https://issues.redhat.com/browse/OHSS-14889) to demonstrate the validity on the new proposed expression.

Nikokolas3270:
Hello! @simonpasquier & @JoaoBraveCoding, some test is failing but this does not seem linked to this PR (connection refused), could you please have a look? Thanks!

simonpasquier (commented):
sorry for the very late review :-/
I believe that the new query will omit targets for which the `pod` label isn't defined. 

image

The following version might address this edge case but it needs to be vetted. Ideally we want additional unit tests in https://github.com/openshift/cluster-monitoring-operator/tree/master/test/rules.

```
((
  count by (job, namespace, service) (up{pod!=""} == 0 and on(pod, namespace) topk by(pod, namespace) (1, kube_pod_info))
  /
  count by (job, namespace, service) (up{pod!=""} and on(pod, namespace) topk by(pod, namespace) (1, kube_pod_info)) 
)
or
# for jobs that don't have a pod label like kubelet, crio and apiserver.
(
  count by (job, namespace, service) (up{pod=""} == 0)
  /
  count by (job, namespace, service) (up{pod=""}) 
)
or
# in case kube-state-metrics is down.
(
  count by (job, namespace, service) (up == 0 and on() absent(up{service='kube-state-metrics'}))
  /
  count by (job, namespace, service) (up and on() absent(up{service='kube-state-metrics'})) 
)) > 10
```

Nikokolas3270:
Hi @simonpasquier,

Thanks for taking up a look, this might indeed be an issue.
We (me, @thavlice) will try to see if the alert is ever firing for the incriminated job... but even if it doesn't it is better to be on the safe side I guess.

JoaoBraveCoding (commented):
Alert seems correct, but as @simonpasquier mentioned it would be great if we could get some tests for it.

> Ideally we want additional unit tests in https://github.com/openshift/cluster-monitoring-operator/tree/master/test/rules

I would also suggest changing the Target Version on Jira to `4.13.0 `so that if everything is good with the PR we can merge it right away.

Nikokolas3270:
> Alert seems correct, but as @simonpasquier mentioned it would be great if we could get some tests for it.
> 
> > Ideally we want additional unit tests in https://github.com/openshift/cluster-monitoring-operator/tree/master/test/rules
> 
> I would also suggest changing the Target Version on Jira to `4.13.0 `so that if everything is good with the PR we can merge it right away.

@thavlice , @simonpasquier & @JoaoBraveCoding :
I took the liberty of writing some tests: https://github.com/Nikokolas3270/cluster-monitoring-operator/commit/e85fd4420c65df99ae36469354d8e722ce95d2ab
Could that be integrated in your PR Tomas?

Also as you can see I modified the expression:
- I am summing the `up` metrics. This avoids the first part of the expression to be undefined if all pods are up and running.
- Thanks to that, I factorised the "corner cases" parts in just one part.
- As the `kube_pod_info` metric never has a `pod` label set to the empty string, I am no longer performing test on this label. If the label is empty in the `up` metric, the expression will default to the second part.
- Also the `and` keyword is not used to combine vectors. It just retain elements of the left vector for which one (or several) elements have been found in the right vector. So there is no need to use `topk` or `max`.

JoaoBraveCoding:
From my POV the changes look really nice, from my side it would be ok to merge them with this PR and add the lgtm, but leave the final decision to @simonpasquier as he might see/know something that I missed.

One question though in the tests isn't the 2nd to last and last test similar? Will there be a case where a service in the up metric will not have the pod label? Sorry if this is a stupid question

Nikokolas3270:
Hello!

Changes requested by @simonpasquier in https://github.com/Nikokolas3270/cluster-monitoring-operator/commit/e85fd4420c65df99ae36469354d8e722ce95d2ab have been implemented.
As discussed, there is no need to get the `and` operator operates on a one-to-one basis; for this reason we are not using `max` to satisfy that requirement.

@JoaoBraveCoding: the 2 last tests are a bit different, one tests the absence of `pod` label while the other tests the absence of the `kube_pod_info` metric.

JoaoBraveCoding (approved):
Great job 👍 very happy to see this one ready to get merged

simonpasquier (commented):
/lgtm

PR #1772, opened 2022-09-19
Makefile: add docgen tool dependency to the docs target

jan--f:
`make generate` has failed for me. I suppose when embedmd is already present the `docs` target doesn't pull in the tooling install target.


jan--f:
/label docs-approved
/label px-approved
/label qe-approved
Adding labels for an internal build change.

simonpasquier (commented):
/lgtm

PR #1775, opened 2022-09-21
[WIP] *: remove merge patch of Thanos sidecar

simonpasquier:
Using strategic merge patch was necessary because the Prometheus CRD had no way to provide the TLS materials to the Thanos sidecar but with latest versions of the CRD, we can simplify.


simonpasquier:
/close

It needs some work on the prometheus operator side: we want to configure the thanos sidecar to listen on the pod's IP for gRPC traffic and localhost for HTTP but there's no possibility in the Prometheus CRD without using strategic merge patch.

PR #1780, opened 2022-09-27
OCPBUGS-1321: Node Exporter - ignore virtual NICs from OVNK cluster

raptorsun:

These NICs are to ignore:

- tun[0-9]*
- br[0-9]*
- ovn-k8s-mp[0-9]*
- br-ex
- br-int
- br-ext


raptorsun:
Hello @dcbw , could you please verify the patterns to ignore are correct?

simonpasquier:
Any reason why the change doesn't make sense in prometheus-operator/kube-prometheus directly? If anything, the upstream jsonnet should at least provide a way to tune the list of ignored interfaces (rather than manipulating the args directly).

PR #1785, opened 2022-09-30
MON-2693: Scrape profiles

JoaoBraveCoding:


pkg/manifests/manifests.go
```diff
@@ -1856,6 +1860,36 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
...
+	if scrapeProfile == "" {
+		return nil
+	}
+
+	if scrapeProfile != "full" && scrapeProfile != "operational" && scrapeProfile != "uponly" {
+		return errors.Wrap(ErrConfigValidation, `scrape profile provided is unknown, supported scrape profiles are: full, operational, uponly`)
+	}
+
+	labelSelector := &metav1.LabelSelector{
+		MatchExpressions: []metav1.LabelSelectorRequirement{
```
JoaoBraveCoding, line 1873: @simonpasquier I was hoping this approach would work however the selectors are ANDed, so in this case, when we configure a scrape profile, Prometheus ends up not scraping anything... The complex problem here is joining two different sets of Generators (ServiceMonitor, PodMonitor and Probe), one with no profile label configured and one with a profile label configured. After thinking about it, I came up with 2 options:
1. Create another Prometheus CRD to select Generators based on the `profile` label. By Default we would start running 2 Prometheus CRDs one without selectors and another with `profile: full`. The problem with this solution is that it goes against the goal of the issue which is to lower resource consumption.
2. Introduce a change to Prometheus operator that would allow the Prometheus CRD to be configured with an array of `labelSelectors` that would be ORed. The problem with this solution is that we would be changing the Prometheus CRD API.

JoaoBraveCoding: **UPDATE** After some more testing, if we use the key selector `NotIn`  we are able to get the behaviour we want, so please disregard my comment

pkg/manifests/manifests.go
```diff
@@ -1856,6 +1860,40 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
...
+				Key:      "monitoring.openshift.io/scrape-profile",
+				Operator: metav1.LabelSelectorOpNotIn,
+				Values:   profiles,
+			},
+		},
+	}
+
+	p.Spec.ServiceMonitorSelector = labelSelector
+	p.Spec.PodMonitorSelector = labelSelector
+	p.Spec.ProbeSelector = labelSelector
```
simonpasquier, line 1892: I think we need to leave ProbeSelector untouched as it's always nil.

simonpasquier (commented):
CMO would need to deploy all new service monitors so they can be picked up by Prometheus?

simonpasquier:
@JoaoBraveCoding it's ok to undraft it, isn't it?

pkg/manifests/types.go
```diff
@@ -178,6 +178,8 @@ type PrometheusK8sConfig struct {
 	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
 	// Defines the pod's topology spread constraints.
 	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+	// Possible values are full operational uponly
```
simonpasquier, line 181: the descriptioin would need a bit more details.

jsonnet/components/control-plane.libsonnet
```diff
@@ -46,12 +47,74 @@ function(params)
       },
     },
 
+    operationalServiceMonitorEtcd: self.serviceMonitorEtcd {
```
simonpasquier, line 50: we can probably factor out the generation of a new service monitor from the "full" version?

```
operationalServiceMonitorEtcd: serviceMonitorForScrapeProfile(
  self.serviceMonitorEtcd, // the original service monitor definition
  "operational", // the resulting scrape profile
  ['etcd_disk_backend_commit_duration_seconds_bucket', ...], // list of metrics being kept.
)
```

pkg/manifests/manifests.go
```diff
@@ -788,6 +798,32 @@ func (f *Factory) KubeStateMetricsServiceMonitor() (*monv1.ServiceMonitor, error
...
 }
 
+func (f *Factory) KubeStateMetricsOperationalServiceMonitor() (*monv1.ServiceMonitor, error) {
+	sm, err := f.NewServiceMonitor(f.assets.MustNewAssetReader(KubeStateMetricsOperationalServiceMonitor))
+	if err != nil {
+		return nil, err
+	}
+
+	sm.Spec.Endpoints[0].TLSConfig.ServerName = fmt.Sprintf("kube-state-metrics.%s.svc", f.namespace)
+	sm.Spec.Endpoints[1].TLSConfig.ServerName = fmt.Sprintf("kube-state-metrics.%s.svc", f.namespace)
```
simonpasquier, line 808: we should be able to remove this Go code and just generate the correct manifests from jsonnet. The reason why it's done this way for now is because it was envisioned at some point that the monitoring namespaces could be customized which isn't something that will happen any time soon (if it ever happens). I'd be ok to have another PR removing all the `...ServerName = fmt.Sprintf("...", ...)` lines in pkg/manifests/manifests.go wherever it isn't required. The same goes for the `sm.Namespace = f.namespace` line below.

pkg/tasks/nodeexporter.go
```diff
@@ -117,6 +118,26 @@ func (t *NodeExporterTask) Run(ctx context.Context) error {
 		return errors.Wrap(err, "initializing node-exporter ServiceMonitor failed")
 	}
 
+	err = t.client.CreateOrUpdateServiceMonitor(ctx, smn)
+	if err != nil {
+		return errors.Wrap(err, "reconciling node-exporter ServiceMonitor failed")
+	}
+
+	smn, err = t.factory.NodeExporterOperationalServiceMonitor()
```
simonpasquier, line 126: maybe have a `t.factory.NodeExporterServiceMonitors()` method which would return a `map[string]*monv1.ServiceMonitor` to reduce the boiler-plate code? 
The same comment applies to the other tasks.

pkg/manifests/manifests.go
```diff
@@ -1898,6 +1962,40 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
...
 }
 
+func (f *Factory) setupScrapeProfiles(p *monv1.Prometheus, scrapeProfile string) error {
+	if scrapeProfile == "" {
+		scrapeProfile = "full"
+	}
+
+	if scrapeProfile != "full" && scrapeProfile != "operational" && scrapeProfile != "uponly" {
+		return errors.Wrap(ErrConfigValidation, `scrape profile provided is unknown, supported scrape profiles are: full, operational, uponly`)
+	}
```
simonpasquier, line 1972: I'd do the verification and defaulting beforehand, e.g. in https://github.com/openshift/cluster-monitoring-operator/blob/274ceee6174dfa776db0a7f20c6e3a4100ca0f18/pkg/manifests/config.go#L189-L201

pkg/manifests/types.go
```diff
@@ -178,6 +178,8 @@ type PrometheusK8sConfig struct {
...
+	ScrapeProfile string `json:"scrapeProfile,omitempty"`
```
simonpasquier, line 182:
```suggestion
	ScrapeProfile ScrapeProfile `json:"scrapeProfile,omitempty"`
```

```golang
type ScrapeProfile string

const (
    FullScrapeProfile = "full"
    OperationalScrapeProfile = "operational"
)
```

simonpasquier (commented):
Good start! Should we first implement full & operational profiles only?

jsonnet/components/control-plane.libsonnet
JoaoBraveCoding: Do you mean abstracting the logic to create a jsonnet function? I'm just not sure because some ServiceMonitors have multiple endpoints, that export different metrics, see kubelet for an example. So what would the interface of that function look like? 

While working on https://issues.redhat.com/browse/MON-2694 I thought we could also just simply add the list of metrics to all the endpoints, this would allow for something like what you suggested.

However, the direction I was taking, [MON-2694](https://issues.redhat.com/browse/MON-2694) was more aligned with the goal of adding for each endpoint only the relevant subset of metrics. The tool would either fetch these metrics from a pre-compiled list or from a file that Openshift developers could provide as a parameter, both the list and the file would have a structure like so:
- namespace/serviceMonitorName/port/path:  []metrics

WDYT? @simonpasquier

simonpasquier: > Do you mean abstracting the logic to create a jsonnet function?

Yes

> I'm just not sure because some ServiceMonitors have multiple endpoints, that export different metrics, see kubelet for an example. So what would the interface of that function look like?

Good point. Thinking more about it, the "operational" profile is all about keeping a set of "known" metrics so the same list could be applied to all endpoints?

Also the current kubelet monitor endpoints have drop configurations in place that aren't "useful" for the operational profile since it knows which metrics to keep and dropping the drop  (sic) directives would be a bit more efficient.

E.g. the existing kubelet service monitor:

```
spec:
  endpoints:
  - bearerTokenFile: ""
    honorLabels: true
    interval: 30s
    metricRelabelings:
    - action: drop
      regex: kubelet_(pod_worker_latency_microseconds|pod_start_latency_microseconds|cgroup_manager_latency_microseconds|pod_worker_start_latency_microseconds|pleg_relist_latency_microseconds|pleg_relist_interval_microseconds|runtime_operations|runtime_operations_latency_microseconds|runtime_operations_errors|eviction_stats_age_microseconds|device_plugin_registration_count|device_plugin_alloc_latency_microseconds|network_plugin_operations_latency_microseconds)
      sourceLabels:
      - __name__
  ...
  - bearerTokenFile: ""
    honorLabels: true
    honorTimestamps: false
    interval: 30s
    metricRelabelings:
    - action: drop
      regex: container_(network_tcp_usage_total|network_udp_usage_total|tasks_state|cpu_load_average_10s)
      sourceLabels:
      - __name__
...
```

could be transformed into an operational service monitor looking like:

```
spec:
  endpoints:
  - bearerTokenFile: ""
    honorLabels: true
    interval: 30s
    metricRelabelings:
    - action: keep
      regex: "<list of operational metrics>"
      sourceLabels:
      - __name__
    port: https-metrics
...
    path: /metrics
  - bearerTokenFile: ""
    honorLabels: true
    honorTimestamps: false
    interval: 30s
    metricRelabelings:
    - action: keep
      regex: "<list of operational metrics>"
      sourceLabels:
      - __name__
    path: /metrics/cadvisor
...
```

pkg/manifests/manifests.go
JoaoBraveCoding: I agree with the defaulting however, regarding validation we usually do it on `manifests/manifests.go` so I'm not so sure we would want to move it to `manifests/config.go`. But do let me know if you prefer it that way!

pkg/manifests/manifests.go
JoaoBraveCoding: I'll open a PR refactoring this and then I'll rebase the PR that lands sooner

jsonnet/components/control-plane.libsonnet
JoaoBraveCoding: Got it, that makes sense. I just have one more question: 
Should we generate our ServiceMonitors with jsonnet and have that in this PR or should we redact that part from this PR and include it in the PR where we will provide [MON-2694](https://issues.redhat.com/browse/MON-2694)? I like more this idea because that way we would be using our own solution for generating the ServiceMonitors instead of using jsonnet? The tool would just consume the assets folder and output yaml files to the same folder. But let me know if I'm missing something, maybe instead of doing things with a dedicated tool we could just create jsonnet package that other developers could use 

WDYT @simonpasquier, if you want we can also quickly discuss it on a call

pkg/manifests/manifests.go
simonpasquier: I'm more in favor of bailing out early. It also means that ` NewConfig()` returns a semantically correct value and we don't have to worry about it later.

pkg/manifests/manifests.go
JoaoBraveCoding: https://github.com/openshift/cluster-monitoring-operator/pull/1873

pkg/manifests/manifests.go
```diff
@@ -1481,6 +1508,31 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
 	return nil
 }
 
+func (f *Factory) setupScrapeProfiles(p *monv1.Prometheus, scrapeProfile ScrapeProfile) error {
+	profiles := make([]string, 0, len(ScrapeProfiles)-1)
```
jan--f, line 1512: To me it wasn't immediately clear what this function does. Maybe this could be improved by better naming?
E.g. something like `profilesToIgnore` or `ignoredByPrometheus` might make the interaction between the label selector further below a bit clearer.

pkg/manifests/manifests.go
```diff
@@ -1481,6 +1508,31 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
...
+	for _, profile := range ScrapeProfiles {
+		if profile == scrapeProfile {
+			continue
+		}
+		profiles = append(profiles, string(profile))
+	}
+
+	labelSelector := &metav1.LabelSelector{
```
jan--f, line 1520: Maybe add a comment why this selector operator is chosen?

jan--f (commented):
This looks great, though I added a few minor sugestions inline.

jsonnet/utils/configure-authentication-for-monitors.libsonnet
```diff
@@ -14,7 +14,23 @@
...
+                              serverName: std.format('%s.%s.svc',
+                                                     [
+                                                       if o.metadata.name != 'thanos-sidecar' then
+                                                         if !std.objectHas(o.metadata.labels, 'monitoring.openshift.io/scrape-profile') || o.metadata.labels['monitoring.openshift.io/scrape-profile'] == 'full' then
+                                                           o.metadata.name
+                                                         else
+                                                           std.rstripChars(o.metadata.name, '-' + o.metadata.labels['monitoring.openshift.io/scrape-profile'])
+                                                       else
+                                                         'prometheus-' + o.metadata.labels['app.kubernetes.io/instance'] + '-' + o.metadata.name,
+                                                       o.metadata.namespace,
```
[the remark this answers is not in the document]
JoaoBraveCoding: I  think this came out great! Let me know what you think 👌

pkg/manifests/manifests.go
JoaoBraveCoding: I changed the name and added a method comment, let me know!

pkg/manifests/manifests.go
JoaoBraveCoding: Indeed I added an implementation comment, hopefully, it makes it clear!

JoaoBraveCoding:
Adding
/label docs-approved as https://github.com/openshift/openshift-docs/pull/56073
ping @juzhao and @Senthamilarasu-STA for remaining labels

pkg/manifests/manifests.go
```diff
@@ -1496,6 +1523,40 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
 	return nil
 }
 
+// setupProfilesToIgnore configures the label selectors of the Prometheus ("p")
+// to select any ServiceMonitor's or PodMonitor's that doesn't have the scrape
+// profile label or that matches the Scrape Profile ("sp").
+func (f *Factory) setupProfilesToIgnore(p *monv1.Prometheus, sp ScrapeProfile) error {
+	// Our goal is to configure Prometheus select both the resources
```
jan--f, line 1530:
```suggestion
	// Our goal is to configure Prometheus to select both the resources
```

jan--f (commented):
One thing that is still missing is to actually check the TechPreview feature gate. Not sure if that was discussed already. The lowest effort way to solve this is to check [`client.TechPreviewEnabled()](https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/client/client.go#L507) in `NewConfig` and error out if a profile other then `full` is specified.
That means that the service monitors still get created but the `minimal` flavor can't be used unless TP is set.
The alternative would be to gate the creation of the `minimal` ServiceMonitors, but I'm not sure how useful that is other then less confusion for users who look at the SMs.
@simonpasquier wdyt?

JoaoBraveCoding:
> One thing that is still missing is to actually check the TechPreview feature gate. Not sure if that was discussed already. The lowest effort way to solve this is to check [`client.TechPreviewEnabled()`](https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/client/client.go#L507) in `NewConfig`and error out if a profile other then`full`is specified. That means that the service monitors still get created but the`minimal`flavor can't be used unless TP is set. The alternative would be to gate the creation of the`minimal` ServiceMonitors, but I'm not sure how useful that is other then less confusion for users who look at the SMs. @simonpasquier wdyt?

I've added the check, it will error out if as you suggested another profile other than `full` is used. I've also made it so the minimal `ServiceMonitor's` are only created if `TechPreview` is enabled. If we prefer them to always be created it will also be easy to revert the change.

jsonnet/components/node-exporter.libsonnet
```diff
@@ -23,6 +24,65 @@ function(params)
...
+    // This changes node-exporter to be scraped with validating TLS.
+    serviceMonitor+: {
+      metadata+: {
+        labels+: {
+          'monitoring.openshift.io/scrape-profile': 'full',
+        },
+      },
+    },
+
+    minimalServiceMonitor: generateServiceMonitor.minimal(
```
juzhao, line 36: @JoaoBraveCoding Is this the metrics list exposed by node-exporter for `scrapeProfile: minimal` setting, same for other components

JoaoBraveCoding: This metric list only concerns metrics exported by `node-exporter` that are then used in alerts, rules, telemetry or console.
If you look at the metric list of another component you will see completely different sent of metrics.
Let me know if that answers the question or if I misunderstood it.

juzhao: yes, that's my question, thanks for the explanation

pkg/manifests/manifests.go
```diff
@@ -3348,7 +3348,7 @@ func (f *Factory) HashSecret(secret *v1.Secret, data ...string) (*v1.Secret, err
 	}, nil
 }
 
-func serviceMonitors(fullServiceMonitor, minimalServiceMonitor func() (*monv1.ServiceMonitor, error)) ([]*monv1.ServiceMonitor, error) {
+func (f *Factory) serviceMonitors(fullServiceMonitor, minimalServiceMonitor func() (*monv1.ServiceMonitor, error)) ([]*monv1.ServiceMonitor, error) {
```
jan--f, line 3351:
```suggestion
func serviceMonitors(appendMinimal bool, fullServiceMonitor, minimalServiceMonitor func() (*monv1.ServiceMonitor, error)) ([]*monv1.ServiceMonitor, error) {
```
I'd probably argue to keep this is a pure function. `serviceMonitors` doesn't really need to know about the TP flag and we don't need ths recevier arg here for any purpose.

JoaoBraveCoding:
> the PR is only for platform Prometheus, we don't need to care about user workload monitoring Prometheus, right?

Correct

pkg/manifests/manifests.go
```diff
@@ -1496,6 +1523,40 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
 	return nil
 }
 
+// setupProfilesToIgnore configures the label selectors of the Prometheus ("p")
+// to select any ServiceMonitor's or PodMonitor's that doesn't have the scrape
+// profile label or that matches the Scrape Profile ("sp").
+func (f *Factory) setupProfilesToIgnore(p *monv1.Prometheus, sp ScrapeProfile) error {
```
jan--f, line 1529:
```suggestion
func setupProfilesToIgnore(p *monv1.Prometheus, sp ScrapeProfile) error {
```
Iiuc `setupProfilesToIgnore` doesn't actually use the receiver argument?

pkg/operator/operator.go
```diff
@@ -857,6 +857,9 @@ func (o *Operator) loadConfig(key string) (*manifests.Config, error) {
 	}
 
 	cParsed, err := manifests.NewConfigFromString(configContent)
+	if tp {
+		cParsed, err = manifests.NewConfigFromStringTp(configContent)
+	}
```
jan--f, line 862: Nit: This code looks a little awkward to me. The first `err` might not actually be looked at. This not an issue now, since it runs the same code right now (lets ignore the inefficiency), but I can imagine future changes might make this less obvious. It might be better here to either separate the two calls with `if...else...` or change the signature of the new function to `func NewConfigFromStringTp(content string, tp bool) (*Config, error)` and expose the tech preview flag.

With the second option we could also add a comment to `NewConfigFromString` that this is deprecated and one should use `NewConfigFromStringTp(content, false)` as a drop in replacement. This gets rid of this little awkwardness.

jan--f (commented):
Two more nits, but otherwise this is shaping up nicely.

pkg/operator/operator.go
JoaoBraveCoding: Yeah, I agree, I was not too fond of it either. I think I'll follow the second approach as I believe that `NewConfigFromString` should always receive the `tp` argument since in the future there will be other features that are tech preview and will face a similar issue

jan--f:
/lgtm
Thanks :metal:

pkg/manifests/config.go
```diff
@@ -197,6 +199,24 @@ func NewConfig(content io.Reader) (*Config, error) {
...
+	// Validate ScrapeProfile field
+	foundProfile := false
+	for _, profile := range ScrapeProfiles {
+		if profile == c.ClusterMonitoringConfiguration.PrometheusK8sConfig.ScrapeProfile {
+			foundProfile = true
+			continue
+		}
+	}
+	if !foundProfile {
+		return nil, errors.Wrap(ErrConfigValidation, fmt.Sprintf(`scrape profile provided is unknown, supported scrape profiles are: %v`, ScrapeProfiles))
```
simonpasquier, line 217: nit: print the config-provided value
```suggestion
		return nil, errors.Wrap(ErrConfigValidation, fmt.Sprintf(`scrape profile %q is not supported, supported scrape profiles are: %q`, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.ScrapeProfile, strings.Join(ScrapeProfiles, ", ")))
```

pkg/manifests/types.go
```diff
@@ -186,6 +195,13 @@ type PrometheusK8sConfig struct {
 	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
 	// Defines the pod's topology spread constraints.
 	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+	// Defines the scraping profile which Prometheus will use to scrape metrics
+	// from the platform components. Supported values are `full` or `minimal`.
+	// In the `full` profile (default) Prometheus will scrape all the metrics
```
simonpasquier, line 200: strictly speaking, not all metrics are collected with the full profile as we've already made choices to drop some (e.g. kubelet, kube-state-metrics). Not sure how to phrase it differently though...

pkg/manifests/types.go
```diff
@@ -186,6 +195,13 @@ type PrometheusK8sConfig struct {
...
+	// that are exposed by the different platform components. In the `minimal`
+	// profile Prometheus will only scrape metrics necessary for platform
+	// alerts, recording rules, telemetry and console dashboards.
+	ScrapeProfile ScrapeProfile `json:"scrapeProfile,omitempty"`
```
simonpasquier, line 204: Should it say collectionProfile rather?

pkg/manifests/manifests.go
```diff
@@ -1496,6 +1523,40 @@ func (f *Factory) setupQueryLogFile(p *monv1.Prometheus, queryLogFile string) er
...
+			{
+				Key:      scrapeProfileLabel,
+				Operator: metav1.LabelSelectorOpNotIn,
+				Values:   profiles,
+			},
+		},
+	}
+
+	p.Spec.ServiceMonitorSelector = labelSelector
+	p.Spec.PodMonitorSelector = labelSelector
```
simonpasquier, line 1555: I wonder if it shouldn't update p.Spec.ProbeSelector too? This would be mostly for consistency as we're not supporting probes for cluster monitoring anyway.

pkg/tasks/controlplane.go
```diff
@@ -48,14 +48,16 @@ func (t *ControlPlaneTask) Run(ctx context.Context) error {
...
+		return errors.Wrap(err, "initializing control-plane kubelet ServiceMonitors failed")
 	}
 
-	err = t.client.CreateOrUpdateServiceMonitor(ctx, smk)
-	if err != nil {
-		return errors.Wrap(err, "reconciling control-plane kubelet ServiceMonitor failed")
+	for _, sm := range sms {
+		err = t.client.CreateOrUpdateServiceMonitor(ctx, sm)
+		if err != nil {
+			return errors.Wrap(err, "reconciling control-plane kubelet ServiceMonitor failed")
```
simonpasquier, line 59:
```suggestion
			return errors.Wrap(err, "reconciling %s/%s ServiceMonitor failed", sm.Namespace, sm.Name)
```

pkg/tasks/controlplane.go
```diff
@@ -75,16 +77,19 @@ func (t *ControlPlaneTask) Run(ctx context.Context) error {
...
 	}
 
 	if t.config.ClusterMonitoringConfiguration.EtcdConfig.IsEnabled() {
-		err = t.client.CreateOrUpdateServiceMonitor(ctx, sme)
-		if err != nil {
-			return errors.Wrap(err, "reconciling control-plane etcd ServiceMonitor failed")
+		for _, sm := range sms {
+			err = t.client.CreateOrUpdateServiceMonitor(ctx, sm)
+			if err != nil {
+				return errors.Wrap(err, "reconciling control-plane etcd ServiceMonitor failed")
```
simonpasquier, line 89: same here

simonpasquier (commented):
just a few minor suggestions from my side.

pkg/manifests/types.go
JoaoBraveCoding: I see, I changed it to `desired metrics` but I also have a hard time phrasing it differently without expending it too much

pkg/manifests/types.go
JoaoBraveCoding: Right, because of the documentation, yes we should be consistent but then I'll change the name everywhere to collectionProfile to make it consistent all around. Let me know if you feel the opposite and we should only change the user-facing name

pkg/manifests/manifests.go
JoaoBraveCoding: Initially I had it but I removed it because of this comment https://github.com/openshift/cluster-monitoring-operator/pull/1785#discussion_r986555681

pkg/manifests/config.go
```diff
@@ -188,6 +189,17 @@ func (cfg *TelemeterClientConfig) IsEnabled() bool {
 	return true
 }
 
+// collectionProfilesAsString the list of collection profiles cannot easily be
+// formated using Sprintf since they are not of type string, this function
+// returns a string with all CollectionProfiles separated by comma
+func collectionProfilesAsString() string {
```
simonpasquier, line 195: suggestion: a more idiomatic way is to create a custom type and implement the Stringer interface for it:

```golang
type CollectionProfiles []CollectionProfile

func (cps CollectionProfiles) String() string {
        collectionProfiles := make([]string, 0, len(cps))
        for _, collectionProfile := range cps {
                collectionProfiles = append(collectionProfiles, string(collectionProfile))
        }
        return strings.Join(collectionProfiles, ", ")
}

...
// in NewConfig()
                return nil, errors.Wrap(ErrConfigValidation, fmt.Sprintf(`%q is not supported, supported collection profiles are: %q`, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile, SupportedCollectionProfiles.String()))

...

var SupportedCollectionProfiles = CollectionProfiles{FullCollectionProfile, MinimalCollectionProfile}

```

simonpasquier (commented):
i think that there are a couple of other places where we want to log the service monitor as namespace/name (e.g. `return errors.Wrapf(err, "reconciling %s/%s ServiceMonitor failed", sm.Namespace, sm.Name)`).

pkg/manifests/manifests.go
simonpasquier: ah interesting, i'll look into why this has changed (probably no good reason)...

pkg/manifests/types.go
simonpasquier: :+1:

pkg/manifests/manifests.go
simonpasquier: hmm my previous comment was completely wrong: I can't find anywhere that probeSpec had been set to nil...

pkg/manifests/config.go
JoaoBraveCoding: Yes, this looks better! I implemented this

pkg/manifests/config.go
```diff
@@ -186,7 +189,15 @@ func (cfg *TelemeterClientConfig) IsEnabled() bool {
 	return true
 }
 
-func NewConfig(content io.Reader) (*Config, error) {
+func (cps CollectionProfiles) String() string {
+	jointProfiles := make([]string, 0, len(cps))
+	for _, cp := range cps {
+		jointProfiles = append(jointProfiles, string(cp))
+	}
+	return strings.Join(jointProfiles, ", ")
```
sthaha, line 197: nit: `string.Builder` is be a better choice than creating a slice and joining.

pkg/manifests/config.go
```diff
@@ -197,6 +208,24 @@ func NewConfig(content io.Reader) (*Config, error) {
...
+	}
+	// Validate CollectionProfile field
+	foundProfile := false
+	for _, profile := range SupportedCollectionProfiles {
+		if profile == c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile {
+			foundProfile = true
+			continue
+		}
+	}
+	if !foundProfile {
```
sthaha, line 225: may be replaced with `slices.Contains(SupportedCollectionProfiles, profile)` 
see: https://pkg.go.dev/golang.org/x/exp/slices#Contains

pkg/manifests/config.go
```diff
@@ -197,6 +208,24 @@ func NewConfig(content io.Reader) (*Config, error) {
...
+
+	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile != FullCollectionProfile && !tp {
+		return nil, errors.Wrap(ErrConfigValidation, "collectionProfiles is a TechPreview feature, to be able to use a profile different from the default (\"full\") please enable TechPreview")
+	}
+	// Validate CollectionProfile field
+	foundProfile := false
+	for _, profile := range SupportedCollectionProfiles {
+		if profile == c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile {
+			foundProfile = true
+			continue
```
sthaha, line 222: shouldn't this be a `break` instead?

assets/kube-state-metrics/minimal-service-monitor.yaml
```diff
@@ -0,0 +1,57 @@
...
...tus_capacity|kube_node_status_condition|kube_persistentvolume_info|kube_persistentvolume_status_phase|kube_persistentvolumeclaim_access_mode|kube_persistentvolumeclaim_info|kube_persistentvolumeclaim_labels|kube_persistentvolumeclaim_resource_requests_storage_bytes|kube_pod_container_resource_limits|kube_pod_container_resource_requests|kube_pod_container_status_last_terminated_reason|kube_pod_container_status_restarts_total|kube_pod_container_status_waiting_reason|kube_pod_info|kube_pod_owner|kube_pod_status_phase|kube_pod_status_ready|kube_pod_status_unschedulable|kube_poddisruptionbudget_status_current_healthy|kube_poddisruptionbudget_status_desired_healthy|kube_poddisruptionbudget_status_expected_pods|kube_replicaset_owner|kube_replicationcontroller_owner|kube_resourcequota|kube_state_metrics_list_total|kube_state_metrics_watch_total|kube_statefulset_metadata_generation|kube_statefulset_replicas|kube_statefulset_status_current_revision|kube_statefulset_status_observed_generation|kube_statefulset_status_replicas|kube_statefulset_status_replicas_ready|kube_statefulset_status_replicas_updated|kube_statefulset_status_update_revision|kube_storageclass_info|process_start_time_seconds)
```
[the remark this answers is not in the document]
JoaoBraveCoding: I added these but I'll have to investigate why the tool I developed didn't pick up these, I'll open an issue on the project and add a task to the GA epic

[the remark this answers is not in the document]
JoaoBraveCoding: Okay looking quickly at the bugs these seem to be metrics that are not used in our default alerting so it's normal that they got excluded from the list. The minimal profile is very restrictive and should only contain metrics that are essential to default alerts, default rules, console and telemetry

pkg/manifests/config.go
JoaoBraveCoding: I love these nit's they help me write better code! Thank you!

assets/kube-state-metrics/minimal-service-monitor.yaml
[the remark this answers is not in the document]
juzhao: thanks, it makes sense

assets/control-plane/minimal-service-monitor-kubelet.yaml
```diff
@@ -0,0 +1,106 @@
...
...d_periods_total|container_cpu_usage_seconds_total|container_fs_reads_bytes_total|container_fs_reads_total|container_fs_usage_bytes|container_fs_writes_bytes_total|container_fs_writes_total|container_memory_cache|container_memory_rss|container_memory_swap|container_memory_usage_bytes|container_memory_working_set_bytes|container_network_receive_bytes_total|container_network_receive_packets_dropped_total|container_network_receive_packets_total|container_network_transmit_bytes_total|container_network_transmit_packets_dropped_total|container_network_transmit_packets_total|container_spec_cpu_shares|kubelet_certificate_manager_client_expiration_renew_errors|kubelet_containers_per_pod_count_sum|kubelet_node_name|kubelet_pleg_relist_duration_seconds_bucket|kubelet_pod_worker_duration_seconds_bucket|kubelet_server_expiration_renew_errors|kubelet_volume_stats_available_bytes|kubelet_volume_stats_capacity_bytes|kubelet_volume_stats_inodes|kubelet_volume_stats_inodes_free|kubelet_volume_stats_inodes_used|kubelet_volume_stats_used_bytes|machine_cpu_cores|machine_memory_bytes|process_start_time_seconds|process_start_time_seconds|rest_client_requests_total|storage_operation_duration_seconds_count)
```
juzhao, line 18: two `process_start_time_seconds`, same in other files

Documentation/api.md
```diff
@@ -295,6 +295,7 @@ The `PrometheusK8sConfig` resource defines settings for the Prometheus component
 | retentionSize | string | Defines the maximum amount of disk space used by data blocks plus the write-ahead log (WAL). Supported values are `B`, `KB`, `KiB`, `MB`, `MiB`, `GB`, `GiB`, `TB`, `TiB`, `PB`, `PiB`, `EB`, and `EiB`. By default, no limit is defined. |
 | tolerations | [][v1.Toleration](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#toleration-v1-core) | Defines tolerations for the pods. |
 | topologySpreadConstraints | []v1.TopologySpreadConstraint | Defines the pod's topology spread constraints. |
+| collectionProfile | CollectionProfile | Defines the collection profile which Prometheus will use to collect metrics from the platform components. Supported values are `full` or `minimal`. In the `full` profile (default) Prometheus will collect all the metrics that are exposed by the platform components. In the `minimal` profile Prometheus will only collect metrics necessary for the default platform alerts, recording rules, telemetry and console dashboards. |
```
bburt-rh, line 298:
```suggestion
| collectionProfile | CollectionProfile | Defines the metrics collection profile that Prometheus uses to collect metrics from the platform components. Supported values are `full` or `minimal`. In the `full` profile (default), Prometheus collects all metrics that are exposed by the platform components. In the `minimal` profile Prometheus only collects metrics necessary for the default platform alerts, recording rules, telemetry and console dashboards. |
```

Documentation/openshiftdocs/modules/prometheusk8sconfig.adoc
```diff
@@ -42,6 +42,8 @@ Appears in: link:clustermonitoringconfiguration.adoc[ClusterMonitoringConfigurat
 
 |topologySpreadConstraints|[]v1.TopologySpreadConstraint|Defines the pod's topology spread constraints.
 
+|collectionProfile|CollectionProfile|Defines the collection profile which Prometheus will use to collect metrics from the platform components. Supported values are `full` or `minimal`. In the `full` profile (default) Prometheus will collect all the metrics that are exposed by the platform components. In the `minimal` profile Prometheus will only collect metrics necessary for the default platform alerts, recording rules, telemetry and console dashboards.
```
bburt-rh, line 45:
```suggestion
|collectionProfile|CollectionProfile| Defines the metrics collection profile that Prometheus uses to collect metrics from the platform components. Supported values are `full` or `minimal`. In the `full` profile (default), Prometheus collects all metrics that are exposed by the platform components. In the `minimal` profile Prometheus only collects metrics necessary for the default platform alerts, recording rules, telemetry and console dashboards.
```

pkg/manifests/types.go
```diff
@@ -186,6 +196,13 @@ type PrometheusK8sConfig struct {
 	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
 	// Defines the pod's topology spread constraints.
 	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+	// Defines the collection profile which Prometheus will use to collect
+	// metrics from the platform components. Supported values are `full` or
+	// `minimal`. In the `full` profile (default) Prometheus will collect all
```
bburt-rh, line 201:
```suggestion
	// `minimal`. In the `full` profile (default), Prometheus collects all
```

pkg/manifests/types.go
```diff
@@ -186,6 +196,13 @@ type PrometheusK8sConfig struct {
...
+	// the metrics that are exposed by the platform components. In the `minimal`
```
bburt-rh, line 202:
```suggestion
	// metrics that are exposed by the platform components. In the `minimal`
```

pkg/manifests/types.go
```diff
@@ -186,6 +196,13 @@ type PrometheusK8sConfig struct {
...
+	// profile Prometheus will only collect metrics necessary for the default
```
bburt-rh, line 203:
```suggestion
	// profile, Prometheus only collects metrics necessary for the default
```

PR #1788, opened 2022-10-03
Revert "[bot] Update jsonnet dependencies"

dgoodwin:
Reverts openshift/cluster-monitoring-operator#1770

Test revert, we think this may have caused regression https://issues.redhat.com/browse/TRT-589.

@simonpasquier apologies here, the PR is great and found a real problem, unfortunately it's long standing and we're scrambling to get the end of sprint payload, this is the last of two known blockers. We need to revert per TRT policy.

Bringing it back in will be delicate, we need to get the problem addressed and it will be challenging, there will be a card filed for that separate issue soon. We will assist however we can.

### DIFF
--- a/assets/control-plane/prometheus-rule.yaml
+++ b/assets/control-plane/prometheus-rule.yaml
@@ -31,7 +31,7 @@ spec:
       expr: |
         sum by (namespace, pod, cluster) (
           max by(namespace, pod, cluster) (
-            kube_pod_status_phase{namespace=~"(openshift-.*|kube-.*|default)",job="kube-state-metrics", phase=~"Pending|Unknown|Failed"}
+            kube_pod_status_phase{namespace=~"(openshift-.*|kube-.*|default)",job="kube-state-metrics", phase=~"Pending|Unknown"}
           ) * on(namespace, pod, cluster) group_left(owner_kind) topk by(namespace, pod, cluster) (
             1, max by(namespace, pod, owner_kind, cluster) (kube_pod_owner{owner_kind!="Job"})
           )
--- a/jsonnet/jsonnetfile.lock.json
+++ b/jsonnet/jsonnetfile.lock.json
@@ -18,7 +18,7 @@
           "subdir": "contrib/mixin"
         }
       },
-      "version": "a932fb58f26ec862be858c884380c53a4975fb3b",
+      "version": "b886bbc89f31d7ac59bdfbb3d14a4263dae85a1b",
       "sum": "IkDHlaE0gvvcPjSNurFT+jQ2aCOAbqHF1WVmXbAgkds="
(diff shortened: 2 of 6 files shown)


dgoodwin:
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.12-e2e-aws-sdn-upgrade 10

dgoodwin:
Consultation with David and Forrest, we are confident this PR needs to get backed out despite the fact it's totally correct and exposing a real problem we will pursue separately.

/override ci/prow/e2e-agnostic
/override ci/prow/e2e-agnostic-operator
/override ci/prow/e2e-agnostic-upgrade
/override ci/prow/e2e-aws-single-node

deads2k:
clean revert of the last commit.  This is almost certainly the problem given that we have Failed installer pods showing up as NotReady and this is one of the things added here.  We need the installer pods to not-fail before bringing this back in.

deads2k:
also, you can see the failure in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/1770/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-upgrade/1575111106949550080 during presubmits

simonpasquier:
Thanks for the quick reaction! No problem with reverting the PR of course :)
Let us know the link to the ticket for the failed installer pods.

> Bringing it back in will be delicate, we need to get the problem addressed and it will be challenging

There are several options that we take on our side (e.g. we don't want to be indefinitely blocked from syncing our alerts from upstream):
1. Override this specific alert in CMO and retain the previous expression (not including Failed pods). The downside is that we potentially miss pods that go failed.
2. Override this specific alert in CMO to exclude installer pods in failed phase (something like `... (kube_pod_status_phase{pod!~"installer.+",namespace=~"(openshift-.*|kube-.*|default)",job="kube-state-metrics", phase=~"Pending|Unknown|Failed"} or kube_pod_status_phase{pod=~"installer.+",namespace=~"(openshift-.*|kube-.*|default)",job="kube-state-metrics", phase=~"Pending|Unknown"}) ...`).
3. Update the e2e tests to skip KubePodNotReady alerts from installer pods.

Option 2 is probably the "best" solution since it makes the alert also resilient for OCP users (unless we want them to be alerted when an installer pod is failing?).

dgoodwin:
From what David said yesterday I think (2) would be good temporarily, but ultimately we would like to expose this to customers and ourselves when it happens. (once the installer pod problem is fixed)

raptorsun:
The https://github.com/openshift/cluster-monitoring-operator/pull/1790 is created to pin down the alert `KubePodNotReady` to the expression that ignores installer pods.

PR #1796, opened 2022-10-12
test/e2e: improve getActiveTarget()

simonpasquier:
It should fix the CMO e2e tests for
https://github.com/openshift/prometheus/pull/142.

getActiveTarget() can use the `scrapePool` key instead of iterating over the discovered labels.


simonpasquier:
/label qe-approved
/label px-approved
/label docs-approved

It only update the end-to-end tests, no functional change.

PR #1800, opened 2022-10-17
OCPBUGS-2439: set the argument path.udev.data in node exporter

raptorsun:

The diskstats collector in node exporter looks for the directory `/run/udev/data`. The correct path should be `/host/root/run/udev/data` in node exporter daemon set.

### DIFF
--- a/assets/node-exporter/daemonset.yaml
+++ b/assets/node-exporter/daemonset.yaml
@@ -40,6 +40,7 @@ spec:
         - --collector.textfile.directory=/var/node_exporter/textfile
         - --no-collector.cpufreq
         - --collector.netclass.ignored-devices=^(veth.*|[a-f0-9]{15}|tun[0-9]*|br[0-9]*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext)$
+        - --path.udev.data=/host/root/run/udev/data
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
don't we need the same change upstream?

simonpasquier:
/hold
I'd prefer the fix to land upstream first (less work overall since we wouldn't have to undo the patch in CMO).

raptorsun:
upstream PR created: https://github.com/prometheus-operator/kube-prometheus/pull/1913

raptorsun:
This bug has been [fixed in upstream ](https://github.com/prometheus-operator/kube-prometheus/pull/1913)and brought to master branch by PR #1818 which upgrades all jsonnet dependencies including kube-prometheus. 

The upstream does not have a 0.12 release branch yet. Shall we merge this PR or update the `kube-prometheus` version to [the fix commit](https://github.com/prometheus-operator/kube-prometheus/commit/fd63d9ea9ffa8fa0ba7264cee7e4243adb7e9da8)?

jan--f:
@raptorsun apologies for the late reply. I think updating `kube-prometheus` would actually be preferable. Carrying downstream fixes is usually more work in the long run.

simonpasquier (commented):
/lgtm

PR #1801, opened 2022-10-18
Documentation: fix TelemeterClientConfig

simonpasquier:
Remove fields which don't need to be documented.


simonpasquier:
/label px-approved

/assign @juzhao 
for adding the `qe-approved` label

/assign @bburt-rh 
for adding the `docs-approved` label

PR #1806, opened 2022-10-24
 *: don't expose Prometheus operator's port 

simonpasquier:
The Prometheus operator pod exposes the prometheus-operator container's
web service running on port 8080 by default. It isn't necessary because:
1) metrics are collected via the kube-rbac-proxy sidecar.
2) the admission webhook service is outsourced to the
   prometheus-operator-admission-webhook deployment.

This change configures the prometheus-operator container to listen on
127.0.0.1:8080 and removes 8080 from the list of ports exposed by the
prometheus-operator service. The same changes are applied to the
user-workload's Prometheus Operator.

### DIFF
--- a/assets/prometheus-operator/operator-certs-ca-bundle.yaml
+++ b/assets/prometheus-operator/operator-certs-ca-bundle.yaml
@@ -1,8 +1,5 @@
 apiVersion: v1
-data: {}
 kind: ConfigMap
 metadata:
-  annotations:
-    service.beta.openshift.io/inject-cabundle: "true"
   name: operator-certs-ca-bundle
   namespace: openshift-monitoring
--- a/jsonnet/components/prometheus-operator-user-workload.libsonnet
+++ b/jsonnet/components/prometheus-operator-user-workload.libsonnet
@@ -32,7 +32,9 @@ function(params)
         name: 'prometheus-user-workload-operator',
       },
     },
+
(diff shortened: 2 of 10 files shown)


assets/prometheus-operator/operator-certs-ca-bundle.yaml
```diff
@@ -1,8 +1,5 @@
 apiVersion: v1
-data: {}
 kind: ConfigMap
 metadata:
-  annotations:
-    service.beta.openshift.io/inject-cabundle: "true"
   name: operator-certs-ca-bundle
   namespace: openshift-monitoring
```
sthaha, line 5: Can this file be removed as well?

jsonnet/components/prometheus-operator-user-workload.libsonnet
```diff
@@ -80,13 +82,9 @@ function(params)
...
                         '--config-reloader-cpu-request=1m',
                         '--config-reloader-memory-request=10Mi',
+                        '--web.listen-address=127.0.0.1:8080',
                       ],
-                      securityContext: {
-                        allowPrivilegeEscalation: false,
-                        capabilities: {
-                          drop: ['ALL'],
-                        },
-                      },
```
sthaha, line 89: Isn't this needed?

assets/prometheus-operator/operator-certs-ca-bundle.yaml
sthaha: Nevermind, I see the TODO below

sthaha (commented):
/lgtm
/approve

sthaha:
Might want to reconsider the `*:` in  commit message `*: don't expose ...`

PR #1809, opened 2022-11-02
Switch ksm registry to registry.k8s.io

olivierlemasle:
Kubernetes has switched registry from` k8s.gcr.io` to `registry.k8s.io` in kubernetes/kubernetes#109938.

kube-state-metrics also did the switch in kubernetes/kube-state-metrics#1750.


JoaoBraveCoding:
For QE approval
/assign @juzhao

PR #1812, opened 2022-11-04
OCPBUGS-1998: pkg/client: Update daemonset degrade condition

slashpai:
Currently the CMO degrades when it is not able to
rollout daemonset pod on a node that is not ready. This blocks the upgrade when a node is unavailable. This commit is to fix that by comparing number of
unavailable daemonset pods and maximum unavailable daemonset pods allowed.

Fixes #OCPBUGS-1998

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -40,6 +40,7 @@ import (
 	"github.com/prometheus-operator/prometheus-operator/pkg/thanos"
 	thanosoperator "github.com/prometheus-operator/prometheus-operator/pkg/thanos"
 	admissionv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


JoaoBraveCoding (commented):
Wouldn't this be a good opportunity to scrap the `UpdateDaemonSet` code and use instead https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/apps.go#L181? It would free us from having to maintain this part of the code

juzhao:
tested with PR, issue is fixed, details please see from jira bug
/label qe-approved

slashpai:
> Wouldn't this be a good opportunity to scrap the `UpdateDaemonSet` code and use instead https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/apps.go#L181? It would free us from having to maintain this part of the code

I am not sure if we should try that for bugfix. I think it would be best to do that separately for cmo code base.

pkg/client/client.go
```diff
@@ -1298,27 +1299,47 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
...
+		if int(numberUnavailable) > maxUnavailable {
+			lastErr = errors.Errorf("Too many daemonset pods are unavailable (%d > %d max unavailable).", numberUnavailable, maxUnavailable)
+			return false, nil
+		}
+
 		if d.Generation > d.Status.ObservedGeneration {
 			lastErr = errors.Errorf("current generation %d, observed generation: %d",
 				d.Generation, d.Status.ObservedGeneration)
 			return false, nil
 		}
```
simonpasquier, line 1326: this should always the first check: it guarantees that the reported status is up-to-date.

simonpasquier (commented):
I wonder if the code added by https://github.com/openshift/cluster-monitoring-operator/pull/1279 doesn't become obsolete? IIUC with the new code, we compare against the number of nodes being considered ready.

slashpai:
> comparing number of
> unavailable daemonset pods and maximum unavailable daemonset pods allowed

The current fix compares number of unavailable daemonset pods and maximum unavailable daemonset pods allowed. https://github.com/openshift/cluster-monitoring-operator/pull/1279 added the case daemonset pods should be running on all  Ready nodes. So I think its safe to remove that code

slashpai:
> I wonder if the code added by #1279 doesn't become obsolete? IIUC with the new code, we compare against the number of nodes being considered ready.

The current fix compares number of unavailable daemonset pods and maximum unavailable daemonset pods allowed. https://github.com/openshift/cluster-monitoring-operator/pull/1279 added the case daemonset pods should be running on all  Ready nodes. So I think its safe to remove code added by #1279

pkg/client/client.go
slashpai: moved it as first case

simonpasquier:
/skip

This is hard to ensure that it's doing the right thing with end-to-end tests... Have you verified with a test cluster that it works as expected? If not, you'd probably have to manually increase the max % of unavailable pods to check that CMO doesn't go degraded when the number of unavailable pods is below the threshold.

Something else that the PR doesn't fix is that any error triggered by the node_exporter task sets CMO to (degraded=true,available=false) while one could argue that CMO should report (degraded=true,available=true). But this is a broader discussion to have (cc @jan--f).

simonpasquier (commented):
/lgtm

PR #1813, opened 2022-11-15
OCPBUGS-3665: adding check for node exporter daemon set

jan--f:
removing operator log

code review comments

correcting log message

changing to use context instead of todo

(cherry picked from commit 57df962638780d3e30d64e197121aa92489369e0)

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -947,9 +947,20 @@ func (c *Client) WaitForDaemonSetRollout(ds *appsv1.DaemonSet) error {
 				d.Status.DesiredNumberScheduled, d.Status.UpdatedNumberScheduled)
 			return false, nil
 		}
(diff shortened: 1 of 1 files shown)


jan--f:
/rename OCPBUGS-3665: adding check for node exporter daemon set

jan--f:
> The code assumes daemon runs on every node.
A daemon set can also run on a subset of nodes, as stated in this document.
Shall we further enhance the master branch by checking the .spec.template.spec.nodeSelector field before calculating the number of daemons to run? In case we add other daemon sets running on specific nodes in the future?

Good idea, but lets implement that if and when we need it. :)

raptorsun (approved):
/lgtm

juzhao:
tested with PR, issue is fixed, details please see from jira bug
/label cherry-pick-approved

PR #1814, opened 2022-11-15
OCPBUGS-3331: Pin Jsonnet Deps + Update go.mod for 4.12

raptorsun:
* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

This is the first PR of bug record [OCPBUGS-3331](https://issues.redhat.com/browse/OCPBUGS-3331) to pin down dependencies in master branch and then backport it to the branch `release-4.12`. 
The next 2 PRs will be:
1. restoring the jsonnet versions after this is merged.
2. backporting this modification to branch `release-4.12`.

Better merge this after [the patch PR on alert PodNotReady](https://github.com/openshift/cluster-monitoring-operator/pull/1792) is merged.

PR #1792, opened 2022-10-05
OCPBUGS-1994: Patch Alert KubePodNotReady

raptorsun:
* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

This PR contains Jsonnet depencies update on 5/10/2022, while the alert KubePodNotReady will keep excluding installer pods.
This works as a walkaround to [TRT-589](https://issues.redhat.com//browse/TRT-589). The alert will be no longer patched once the issue [TRT-589](https://issues.redhat.com//browse/TRT-589) is resolved.
The alert will stay as patched during future updates from Jsonnet components. (ref https://github.com/openshift/cluster-monitoring-operator/pull/1788).

PR #1815, opened 2022-11-18
Revert "OCPBUGS-3331: Pin Jsonnet Deps + Update go.mod for 4.12"

simonpasquier:
Reverts openshift/cluster-monitoring-operator#1814

The updated definition of the KubePodNotReady alerting rule breaks the CI as explained in https://issues.redhat.com//browse/OCPBUGS-1994.
To unrevert this revert, we'd need to incorporate the change in https://github.com/openshift/cluster-monitoring-operator/pull/1792 that keeps the previous alert definition.


raptorsun:
Shall we merge [this PR patching the alert KubePodNotReady](https://github.com/openshift/cluster-monitoring-operator/pull/1792) instead? 
It restores the alert to the old definition.

dgoodwin:
> Shall we merge [this PR patching the alert KubePodNotReady](https://github.com/openshift/cluster-monitoring-operator/pull/1792) instead? It restores the alert to the old definition.

A common and fair question, but with the goal of keeping nightlies stable, we've established widely agreed precedent throughout the org to favor reverts as they take us back to a known good state, whereas attempted fixes have a higher risk and we know from painful experience often don't work first try and then burn more time with payloads blocked. While the suggested PR is almost surely fine, we can be all but certain this revert is. 

We should merge this one and then for a followup PR you can unrevert this revert, layer in the new commit with the fix, and we can easily prove that works with /payload command and merge in when ready, without the pressure of a rush to get payloads back online.

raptorsun:
> > Shall we merge [this PR patching the alert KubePodNotReady](https://github.com/openshift/cluster-monitoring-operator/pull/1792) instead? It restores the alert to the old definition.
> 
> A common and fair question, but with the goal of keeping nightlies stable, we've established widely agreed precedent throughout the org to favor reverts as they take us back to a known good state, whereas attempted fixes have a higher risk and we know from painful experience often don't work first try and then burn more time with payloads blocked. While the suggested PR is almost surely fine, we can be all but certain this revert is.
> 
> We should merge this one and then for a followup PR you can unrevert this revert, layer in the new commit with the fix, and we can easily prove that works with /payload command and merge in when ready, without the pressure of a rush to get payloads back online.

OK. After merging this revert, we will merge the patch to the alert KubePodNotReady and then re-apply the content of PR#1792.

PR #1824, opened 2022-11-28
OCPBUGS-4168: Increase startupProbe for prometheus

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/OCPBUGS-4168

Problem: recently we have seen a couple of customers complaining that some startup probes are timing out due to the WAL replay that Prometheus has to perform once rebooted.

Solution: to avoid this we are doubling the startup probe, with this commit Prometheus should now have up to 30 minutes to boot

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1710,6 +1710,16 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 
 	for i, container := range p.Spec.Containers {
 		switch container.Name {
+		case "prometheus":
+			// Increase the startup probe timeout to 1h from 15m to avoid restart failures when the WAL replay
+			// takes a long time. See https://issues.redhat.com/browse/OCPBUGS-4168 for details.
+			// TODO (JoaoBraveCoding): Once prometheus-operator adds CRD support to configure startupProbe directly
+			// we should use that instead of using strategic merge patch
+			// See https://github.com/prometheus-operator/prometheus-operator/issues/4730
+			p.Spec.Containers[i].StartupProbe = &v1.Probe{
+				PeriodSeconds:    15,
+				FailureThreshold: 240,
+			}
 		case "prometheus-proxy":
 			p.Spec.Containers[i].Image = f.config.Images.OauthProxy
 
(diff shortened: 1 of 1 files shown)


pkg/manifests/manifests.go
```diff
@@ -1710,6 +1710,12 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 
 	for i, container := range p.Spec.Containers {
 		switch container.Name {
+		case "prometheus":
+			// TODO (JoaoBraveCoding): Once prometheus-operator adds CRD support to configure startupProbe directly
+			// we should use that instead of using strategic merge patch
+			p.Spec.Containers[i].StartupProbe = &v1.Probe{
+				PeriodSeconds: 30,
```
simonpasquier, line 1717: I'd rather keep the PeriodSeconds value unchanged and increase FailureThreshold. And for "safety", I'd set the timeout duration to 1h in total.

```suggestion
				PeriodSeconds: 15,
				FailureThreshold: 240,
```

pkg/manifests/manifests.go
```diff
@@ -1710,6 +1710,13 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 
 	for i, container := range p.Spec.Containers {
 		switch container.Name {
+		case "prometheus":
+			// TODO (JoaoBraveCoding): Once prometheus-operator adds CRD support to configure startupProbe directly
+			// we should use that instead of using strategic merge patch
```
simonpasquier, line 1715: I tend to link the bug in the comment as it's more obvious than going through the Git log :)

```suggestion
			// Increase the startup probe timeout to 1h from 15m to avoid restart failures when the WAL replay takes a long time.
			// See https://issues.redhat.com/browse/OCPBUGS-4168 for details.
			// TODO (JoaoBraveCoding): Once prometheus-operator adds CRD support to configure startupProbe directly
			// we should use that instead of using strategic merge patch
```

simonpasquier:
For consistency can you do the same for UWM Prometheus too?

simonpasquier (commented):
/lgtm

PR #1826, opened 2022-11-28
OCPBUGS-4181: Fixes externalURL field for Prometheus and Alertmanager

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/OCPBUGS-4181

Problem: currently when setting externalURL field CMO uses a package which removes one of the two "/" after "https" this causes that when users get alerts in their receivers that alerts have in the console URL a total of 3 "/"

Solution: instead of using the "path" package to join the "monitoring" path with the console URL use instead the net/url package. Add tests to manifest_test.go to test this behaviour.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -25,7 +25,6 @@ import (
 	"io"
 	"net"
 	"net/url"
-	"path"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
/lgtm

PR #1827, opened 2022-11-29
OCPBUGS-4184: use mTLS authentication for metrics scraping

simonpasquier:


JoaoBraveCoding (commented):
Only have one question, maybe it's obvious but since I'm not sure I'll ask. Why did we remove `serverName` in some resources but not in others? Appart from that LGTM 👍

simonpasquier:
> Why did we remove serverName in some resources but not in others? 

`serverName` is usually overriden by CMO before applying the manifest ([example](https://github.com/openshift/cluster-monitoring-operator/blob/3797f78018c90766b88b45372f9fd0a26247bee0/pkg/manifests/manifests.go#L431)). I've decided to avoid setting serverName to `server-name-replaced-at-runtime`  as I feel that it doesn't bring much information + there are e2e tests verifying that all monitoring targets are up. The remaining `serverName` fields come from the upstream mixins.

JoaoBraveCoding (approved):
/lgtm

PR #1832, opened 2022-11-30
Remove deprecated option from kube-state-metrics args

simonpasquier:
The `--metric-labels-allowlist` command-line argument contained the `poddisruptionbudget[*]` string which was introduced to support older versions of kube-state-metrics.

With https://github.com/kubernetes/kube-state-metrics/pull/1826, kube-state-metrics will now refuse to start if the allow-list contains an invalid resource name.

### DIFF
--- a/assets/kube-state-metrics/deployment.yaml
+++ b/assets/kube-state-metrics/deployment.yaml
@@ -39,7 +39,7 @@ spec:
           --metric-denylist=
           ^kube_secret_labels$,
           ^kube_.+_annotations$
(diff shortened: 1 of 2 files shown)


jan--f:
Nice, thanks!
/lgtm
/label docs-approved
/label px-approved
Ping @juzhao for qe approval.

PhilipGough (approved):
/lgtm

juzhao:
tested with PR
```
$ oc -n openshift-monitoring get deploy kube-state-metrics -oyaml | grep metric-labels-allowlist
        - --metric-labels-allowlist=pods=[*],nodes=[*],namespaces=[*],persistentvolumes=[*],persistentvolumeclaims=[*],poddisruptionbudgets=[*]
```
checked the prometheus API, the PDB label is in the search result

PR #1836, opened 2022-12-02
OCPBUGS-2141: compute doc link in PVC not configured message

sthaha:
Previously, when PVC is not configured for Prometheus, the CO monitoring status message for Degraded condition included a link to the documentation on configuring monitoring stack for version 4.8.

This commit fixes the link to point to doc for the current cluster version and in case of failure to compute, the cluster-version, points to the latest documentation.

JIRA: https://issues.redhat.com/browse/OCPBUGS-2141


juzhao:
link to `**latest**` is fine, since the docs usually not ready after GA
for example, if the current version is 4.13, due to there is not 
https://docs.openshift.com/container-platform/4.13/monitoring/configuring-the-monitoring-stack.html now
will get 404

sthaha:
@juzhao what this pr does is to see what the current cluster version is, say it is 4.12.1, then it will create a link to
https://docs.openshi..../4.12/monitoring/...
If the version can't be determined for whatever reason, then you will see the link to latest.

JoaoBraveCoding (approved):
lgtm will be happy to add the final lgtm once the CI is happy

PR #1839, opened 2022-12-05
go.mod: switch to go 1.19

sthaha:
Commit 70af87f25 makes use of url.JoinPath that available only in 1.19, hence switching to 1.19


jan--f:
/lgtm
/label docs-approved
/label px-approved
/label qe-approved
Adding labels for a simple go version bump.
/retest

PR #1843, opened 2022-12-07
OCPBUGS-2729: Node Exporter ignore virtual network device 'enP.*'.

raptorsun:
Network interfaces named "enP.*" are virtual network interfaces on Azure cluster. We will ignore them in Node Exporter.

### DIFF
--- a/jsonnet/main.jsonnet
+++ b/jsonnet/main.jsonnet
@@ -181,6 +181,12 @@ local inCluster =
             rateInterval: '1m',  // adjust the rate interval value to be 4 x the node_exporter's scrape interval (15s).
           },
         },
+        // NOTE: 3 patterns for virutal NICs will be ignored:
+        // 1. veth network interface associated with containers.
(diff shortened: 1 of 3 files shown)


jsonnet/main.jsonnet
```diff
@@ -181,6 +181,7 @@ local inCluster =
             rateInterval: '1m',  // adjust the rate interval value to be 4 x the node_exporter's scrape interval (15s).
           },
         },
+        ignoredNetworkDevices:: '^(veth.*|[a-f0-9]{15}|enP.*)$',
```
simonpasquier, line 184: Can you add a comment explaining the reasons for each exclusion (e.g. like we have in kube-prometheus)?

raptorsun: thanks, explanation added :)

simonpasquier: cool, thanks!

simonpasquier (commented):
/lgtm

PR #1848, opened 2022-12-09
OCPBUGS-4521: check that all targets are up after certificate recreation

simonpasquier:
This change adds an end-to-end test that triggers a rotation of the TLS certificate used by prometheus-k8s pods to authenticate against targets. The rotation is simulated by deleting the existing secret.


simonpasquier:
/hold

we need kubelet to load the correct CA => https://github.com/openshift/machine-config-operator/pull/3448

simonpasquier:
The new CMO PR is https://github.com/openshift/machine-config-operator/pull/3458

JoaoBraveCoding (approved):
/lgtm

PR #1856, opened 2022-12-22
WIP: [DONOTMERGE] jsonnet: Remove service.beta.openshift.io/inject-cabundle from webhook

slashpai:
This PR is to test if we needed `service.beta.openshift.io/inject-cabundle` on webhook


simonpasquier:
After more thought, I realize that we need to inject the CA bundle into the webhook configuration, otherwise the kube API server can't connect to the webhook service (e.g. it wouldn't trust the TLS certificate used by the webhook endpoint).

PR #1868, opened 2023-01-18
OCPBUGS-5353: unstack dashboards with limit markers.

raptorsun:

Dashboards showing CPU / memory usage used to stack the limit markers such as limit/request/quota over other metrics create confusion for users reading the diagrams. So we unstack all metrics in the charts containing these limit markers. 

This is the result of unstacking metrics:
image

This is what it looks like when stacking metrics:
image


simonpasquier:
can we make the change upstream?

simonpasquier:
Ok! can you add the explanation next to the added jsonnet code (with a link to the JIRA bug)?
Also did we discuss this issue with the console folks? Is there something to improve on this side (e.g. to support stacked and unstacked series on the same graph)?

raptorsun:
> Ok! can you add the explanation next to the added jsonnet code (with a link to the JIRA bug)? Also did we discuss this issue with the console folks? Is there something to improve on this side (e.g. to support stacked and unstacked series on the same graph)?

Yup. An explanation has been added to the file `dashboards.libsonnet` as well as on the JIRA bug ticket. I will check with the console team on what shall we do to the charts in UI.

PR #1870, opened 2023-01-23
*: remove kube-rbac-proxy sidecar container

simonpasquier:
This commit removes the need of running a kube-rbac-proxy sidecar
container next to the CMO container by implementing TLS +
authentication/authorization directly in CMO.

Supported authentication methods are client TLS certificates and bearer
tokens. Client TLS certificates are verified when issued by the
Kubernetes cluster's CA (read from the 'client-ca-file' key in the
kube-system/extension-apiserver-authentication ConfigMap).

There's a built-in authorization rule for the
'system:serviceaccount:openshift-monitoring:prometheus-k8s' user
(service account) on the '/metrics' endpoint. Other users will be
allowed based on their RBAC permissions.

The following endpoints are also available without authn/authz:
* '/healthz'
* '/livez'
* '/readyz'

The implementation mostly builds upon the
github.com/openshift/library-go and k8s.io/apiserver modules. This is
why many of the Go dependencies have been updated.


jan--f (changes requested):
Very cool, thank you!
This looks great, I only have one nit and a request to drop the `WIP` from the commit message. While you're add it maybe also mention that this updates a whole bunh of dependencies?

JoaoBraveCoding (approved):
Awesome work! I would agree with Jan on improving the commit message!
So a TL;DR for this PR is:
- Replacing kube-rbac-proxy with a dedicated server to serve CMO metrics
- Adhering to the k8s lib for metrics, which allows us to set metric stabilityLevel

simonpasquier:
> This looks great, I only have one nit and a request to drop the WIP from the commit message.

@jan--f thanks for noticing!

simonpasquier:
updated the commit message and the PR description :)

assets/cluster-monitoring-operator/service-monitor.yaml
```diff
@@ -7,11 +7,19 @@ metadata:
   namespace: openshift-monitoring
 spec:
   endpoints:
-  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+  - bearerTokenFile: ""
+    metricRelabelings:
+    - action: drop
```
raptorsun, line 12: why these metrics are dropped?

jan--f: The [jsonnet code](https://github.com/openshift/cluster-monitoring-operator/pull/1870/files#diff-a89d64ee6e92f3dd1e7faf2c0e4bdf29ea537f3a82d7dd8db62490a0e8799efc) has the answer :)

jan--f (approved):
/lgtm
:+1: :ship:

simonpasquier:
/label px-approved
/label docs-approved

No user facing changes so adding the PX and docs labels myself. I'll let @juzhao and/or @lihongyan1 verify that we have no regression and add the `qe-approved` label.

juzhao:
tested with PR, cluster-monitoring-operator endpoint exposes more metrics than nightly payload now, example: (apiserver|go_sched|workqueue)_.+ metrics, and the metrics are dropped from prometheus, according to the PR, this is expected:
PR metrics: http://pastebin.test.redhat.com/1089772
nightly payload metrics: http://pastebin.test.redhat.com/1089770

/lgtm

PR #1874, opened 2023-01-31
Fix 'make run-local' target

simonpasquier:
When running CMO locally, the KUBECONFIG file needs to be provided to the HTTP metrics server otherwise the initialization fails.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -199,7 +199,7 @@ func Main() int {
 
 	wg.Go(func() error { return o.Run(ctx) })
 
-	srv, err := metrics.NewServer("cluster-monitoring-operator", config, *certFile, *keyFile)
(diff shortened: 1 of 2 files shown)


jan--f (approved):
Thanks, TIL. I would have assumed that `BuildConfigFromFlags()` creates a stand-alone config with all needed information.
/lgtm
/retest

jan--f:
/label docs-approved
/label px-approved
/label qe-approved
Adding labels for a change to internal testing functionality.

simonpasquier:
This is because in local mode, there's no in-cluster config:

https://github.com/openshift/cluster-monitoring-operator/blob/b1b2cceb9086d896dae48f47302eff0db3accd12/vendor/k8s.io/apiserver/pkg/server/options/authentication.go#L421-L429

PR #1879, opened 2023-02-03
MON-2932: jsonnet/dashboards: add role template variable to node related dashbo…

jan--f:
…ards

This adds a new `role` template variable to the two node related dashboards. This feature allows for node role based filtering of the `node` and `instance` template variables.
I proposed this feature upstream first, but since not all k8s distributions assign node roles by default this was rejected.

For the node-rsrc-use dashboard we switch to the same template variable queries as in the k8s-resources-node dashboard (i.e. use `kube_node_info` instead of `node_exporter_build_info`. The label values that matter are the same and this avoids more jsonnet complexity, since we'd otherwise have to rewrite label names in order to match `instance` and `node` label values.

2023-02-07-101231_1172x585
2023-02-07-101159_1057x605

### DIFF
--- a/jsonnet/components/dashboards.libsonnet
+++ b/jsonnet/components/dashboards.libsonnet
@@ -1,4 +1,5 @@
 local kubernetesGrafana = import 'github.com/brancz/kubernetes-grafana/grafana/grafana.libsonnet';
+local grafana = import 'github.com/grafana/grafonnet-lib/grafonnet/grafana.libsonnet';
 
 function(params)
   local cfg = params;
@@ -49,7 +50,12 @@ function(params)
       false
   ;
 
-  local glib = kubernetesGrafana(cfg) {};
+  local nodeDashboards = [
+    'k8s-resources-node.json',
+    'node-rsrc-use.json',
+  ];
+
+  local glib = kubernetesGrafana(cfg);
 
   local unstackDashboards = function(dashboards)
     std.map(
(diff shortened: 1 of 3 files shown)


jan--f:
/label docs-approved
/label px-approved
Adding doc approval since this doesn't need doc changes and px approval as per agreement.
Ping @juzhao for qe approval.
cc @kyoto

jsonnet/components/dashboards.libsonnet
```diff
@@ -49,7 +50,16 @@ function(params)
...
+  local nodeDashboards = [
+    'k8s-resources-node.json',
+    'node-rsrc-use.json',
+  ];
+
+  local glib = kubernetesGrafana(cfg) + {
+    _config+: {
+      clusterLabel: 'cluster',
+    },
+  };
```
jan--f, line 62: This is the default cluster label that are added in the dashboards. I should be able to pull this from the cfg structure of on of the mixin imports. So far I have failed to find the right config though.

JoaoBraveCoding: Sorry if I'm missing something but in `github.com/brancz/kubernetes-grafana/grafana/grafana.libsonnet` I don't see this `clusterLabel` defined so I'm not sure if I'm missing something or you meant to pull it from somewhere else 🤔

juzhao:
a small issue with the PR, https://issues.redhat.com/browse/OCPBUGS-7184, since it's a low priority bug, we could add
/label qe-approved

jsonnet/components/dashboards.libsonnet
jan--f: The default config in question lives here: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/07ee0700006ecbc8704ff0b2790fccac18c36c6d/config.libsonnet#L85

Through a convoluted path, we pull that in and I _think_ this should be available in our `cfg` object.

JoaoBraveCoding: Thanks for pointing out where is the value that we want. But AFAICT that mixin is not being included in `dashboards.libsonnet` nor is it being passed in the function pram in `main.libsonnet` so AFAIK it seems that here we would either have to include that jsonnet file that you linked or pass the config in the params, if I understand things correctly

simonpasquier:
/lgtm
/hold
my suggestion can be implemented later so feel free to remove my hold @jan--f

jsonnet/components/dashboards.libsonnet
jan--f: yes, that looks perfect. Thank you! :heart:

jan--f:
/unhold
This is good to go after we get an lgtm stamp.

PR #1882, opened 2023-02-07
MON-2959: Allow configuring secrets in alertmanager (platform)

slashpai:
Related-to #[MON-2798](https://issues.redhat.com//browse/MON-2798)

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1235,6 +1235,16 @@ func (c *Client) WaitForSecret(ctx context.Context, s *v1.Secret) (*v1.Secret, e
 	return result, nil
 }
 
+func (c *Client) WaitForSecretByNsName(ctx context.Context, obj types.NamespacedName) (*v1.Secret, error) {
+	secret := v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      obj.Name,
+			Namespace: obj.Namespace,
+		},
+	}
+	return c.WaitForSecret(ctx, &secret)
+}
+
 func (c *Client) WaitForRouteReady(ctx context.Context, r *routev1.Route) (string, error) {
 	host := ""
 	var lastErr error
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -582,6 +582,10 @@ func (f *Factory) AlertmanagerMain(trustedCABundleCM *v1.ConfigMap) (*monv1.Aler
 		a.Spec.Resources = *f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Resources
 	}
 
+	if f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Secrets != nil {
+		a.Spec.Secrets = append(a.Spec.Secrets, f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Secrets...)
+	}
+
 	if f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.EnableUserAlertManagerConfig &&
 		!f.config.UserWorkloadConfiguration.Alertmanager.Enabled {
 		a.Spec.AlertmanagerConfigSelector = &metav1.LabelSelector{}
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -35,6 +35,7 @@ import (
 	policyv1 "k8s.io/api/policy/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/strings/slices"
 )
 
 type fakeInfrastructureReader struct {
@@ -2596,6 +2597,9 @@ func TestAlertmanagerMainConfiguration(t *testing.T) {
     requests:
       cpu: 10m
       memory: 75Mi
+  secrets:
+  - test-secret
+  - slack-api-token
   volumeClaimTemplate:
     spec:
       resources:
@@ -2646,6 +2650,14 @@ ingress:
 	}
 	if memoryRequestPtr.String() != "75Mi" {
 		t.Fatal("Alertmanager memory request is not configured correctly:", memoryRequestPtr.String())
+	}
+
+	if !slices.Contains(a.Spec.Secrets, "test-secret") {
+		t.Fatal("Alertmanager secret `test-secret` is not configured correctly")
+	}
+
+	if !slices.Contains(a.Spec.Secrets, "slack-api-token") {
+		t.Fatal("Alertmanager secret `slack-api-token` is not configured correctly")
 	}
 
 	if a.Spec.NodeSelector["type"] != "worker" {
--- a/pkg/manifests/types.go
+++ b/pkg/manifests/types.go
@@ -73,6 +73,12 @@ type AlertmanagerMainConfig struct {
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 	// Defines resource requests and limits for the Alertmanager container.
 	Resources *v1.ResourceRequirements `json:"resources,omitempty"`
+	// Defines a list of secrets that need to be mounted into the Alertmanager.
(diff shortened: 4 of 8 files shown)


pkg/manifests/types.go
```diff
@@ -73,6 +73,10 @@ type AlertmanagerMainConfig struct {
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 	// Defines resource requests and limits for the Alertmanager container.
 	Resources *v1.ResourceRequirements `json:"resources,omitempty"`
+	// Defines a list of Secrets in the same namespace as the Alertmanager object
+	// which shall be mounted into the Alertmanager Pods as volumes named secret-<secret-name>.
+	// and are mounted into /etc/alertmanager/secrets/<secret-name> in the ‘alertmanager’ container.
```
sthaha, line 78: How about we rephrase this a bit .. 

```suggestion
	// Defines a list of secrets that need to be mounted into the Alertmanager. 
	// The secrets must reside within the same namespace as the Alertmanager object. 
	// They will be added as volumes named secret-<secret-name> and mounted at
	// /etc/alertmanager/secrets/<secret-name> within the 'alertmanager' container of
	// the Alertmanager Pods.

```

pkg/client/client.go
```diff
@@ -1423,6 +1423,12 @@ func (c *Client) CreateIfNotExistSecret(ctx context.Context, s *v1.Secret) error
 	return errors.Wrap(err, "retrieving Secret object failed")
 }
 
+func (c *Client) CheckSecretExists(ctx context.Context, namespace, secretName string) error {
+	sClient := c.kclient.CoreV1().Secrets(namespace)
+	_, err := sClient.Get(ctx, secretName, metav1.GetOptions{})
+	return err
+}
+
```
sthaha, line 1431: I think we should replace the call to `CheckSecretExists` with just the content of the function below since IMHO, it don't simplify the code. 
It may even be better to call the function `GetSecret(..) (*corev1.Secret, error)` and the caller can ignore the the secret returned.

pkg/client/client.go
```diff
@@ -1423,6 +1423,12 @@ func (c *Client) CreateIfNotExistSecret(ctx context.Context, s *v1.Secret) error
 	return errors.Wrap(err, "retrieving Secret object failed")
 }
 
+func (c *Client) CheckSecretExists(ctx context.Context, namespace, secretName string) error {
```
sthaha, line 1426:
```suggestion
func (c *Client) GetSecret(ctx context.Context, namespace, secretName string) (*corev1.Secret,  error) {
```

pkg/tasks/alertmanager.go
```diff
@@ -106,6 +106,14 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 		return errors.Wrap(err, "creating Alertmanager RBAC proxy metric Secret failed")
 	}
 
+	if t.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Secrets != nil {
+		for _, secret := range t.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Secrets {
+			if err = t.client.CheckSecretExists(ctx, "openshift-monitoring", secret); err != nil {
```
sthaha, line 111:
```suggestion
			if _, err := t.client.GetSecret(ctx, "openshift-monitoring", secret); err != nil {
```

sthaha: It might also be worth replacing "openshift-monitoring" with a const `MonitoringNamespace`

pkg/manifests/manifests_test.go
```diff
@@ -2648,6 +2651,17 @@ ingress:
 		t.Fatal("Alertmanager memory request is not configured correctly:", memoryRequestPtr.String())
 	}
 
+	foundSecrets := 0
```
sthaha, line 2654:
```suggestion
	secretsFound := 0
```

pkg/manifests/manifests_test.go
```diff
@@ -2648,6 +2651,17 @@ ingress:
...
+	for _, s := range a.Spec.Secrets {
+		if s == "test-secret" || s == "slack-api-token" {
+			foundSecrets++
+		}
+	}
+
+	if foundSecrets != 2 {
+		t.Fatal("Alertmanager secrets not configured correctly")
```
sthaha, line 2662: it might be worth logging what secrets were found and what was missing?

pkg/client/client.go
slashpai: Thanks for that. I saw wee already have GetSecret() method so I have re-used that :)

pkg/tasks/alertmanager.go
slashpai: Done

pkg/manifests/manifests_test.go
slashpai: I updated the test to use `slices` package to compare secrets, does that look good now? :)

slashpai:
cc: @juzhao Please help with QE review

pkg/manifests/manifests.go
```diff
@@ -582,6 +582,10 @@ func (f *Factory) AlertmanagerMain(trustedCABundleCM *v1.ConfigMap) (*monv1.Aler
 		a.Spec.Resources = *f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Resources
 	}
 
+	if f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Secrets != nil {
+		a.Spec.Secrets = append(a.Spec.Secrets, f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Secrets...)
```
sthaha, line 586: Would this append by any case add duplicate entries to AlertManager when the reconcile loop start again?

pkg/tasks/alertmanager.go
```diff
@@ -20,8 +20,11 @@ import (
 	"github.com/openshift/cluster-monitoring-operator/pkg/client"
 	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/types"
 )
 
+const MonitoringNamespace = "openshift-monitoring"
```
sthaha, line 26: is there a pkg where we keep all global constant?

pkg/tasks/alertmanager.go
```diff
@@ -106,6 +109,18 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
...
 	}
 
+	if t.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Secrets != nil {
+		for _, secret := range t.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.Secrets {
+			obj := types.NamespacedName{
+				Name:      secret,
+				Namespace: MonitoringNamespace,
+			}
+			if _, err = t.client.WaitForSecretByNsName(ctx, obj); err != nil {
+				return errors.Wrap(err, "retrieving Alertmanager main config secret failed")
```
sthaha, line 119: How about .. 

```suggestion
				return errors.Wrap(err, "failed to find Alertmanager secret %q", secret)
```

pkg/manifests/manifests.go
slashpai: I didnt see issue like that

pkg/tasks/alertmanager.go
slashpai: I realized we don't need to use const here since we can use `t.client.Namespace()` so I have removed the const

sthaha (commented):
/lgtm

slashpai:
/label docs-approved
Since https://issues.redhat.com/browse/RHDEVDOCS-4842 is already created for docs

slashpai:
hold was not needed :)

pkg/manifests/manifests.go
simonpasquier: The Alertmanager object is read from the assets directory (it's not the "live" object retrieved from the Kube API).

sthaha: Aah! thanks for the explanation @simonpasquier 🙇

PR #1886, opened 2023-02-10
STOR-1154: Add vSphere topology to telemetry

jsafrane:

Added aggregated vSphere topology metrics to telemetry. Approved by @jan--f in https://issues.redhat.com/browse/MON-2966.

When at it, I moved all vSphere prometheus rules to a separate group (they IMO don't belong to `openshift-kubernetes.rules`) + added aggregation for the new metrics.


simonpasquier:
/lgtm
/hold
until https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/135 merges.

jsafrane:
/hold cancel
the metric have been merged to vmware-vsphere-csi-driver-operator

/retest-required

PR #1890, opened 2023-02-14
WIP: CMO deploys monitoring console-plugin

sthaha:
**DRAFT** submitted for early feedback


sthaha:
Testing it locally 
```
  go run ./cmd/operator/... \
    -images=prometheus-operator=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:1d704524e203614a92f903ed50247546dcf854a6e983edb1f2c7e9268a6b3247 \
    -images=prometheus-config-reloader=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:0b682f7e323cafc28bd09fa60903f3ae9e073aeb3c04b194f7401432e50a6c31 \
    -images=prometheus-operator-admission-webhook=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:cb86ba5d99371e93eced2f86bfb20d041668cabb7d09501d24fd8ea0617402e4 \
    -images=configmap-reloader=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:7ded11abcea8ed136acf36d6a5bacd218dffadac564ef0f5bce150620c992879 \
    -images=prometheus=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:b10ab06cd40456afe5f928c2642832ea92a8ad05b0636b9286b82d39a66eb2d4 \
    -images=alertmanager=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:aaa094ab3806e1688b381459841ace8bf528b4965f8791cb0353546a518ba543 \
    -images=oauth-proxy=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:5925f28a7682ef5cb91d4a3e3c9f6877ccc339ddb3533ad6376771314a79cb5a \
    -images=node-exporter=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:3576c4bdb825087fc187266bfc13ac84c644ce18779fe864adc03303fab67063 \
    -images=kube-state-metrics=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:7d236b37ff6ef642a63a15ecce4951705e434c340ee6914f432eddc551ca66b8 \
    -images=openshift-state-metrics=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:8a254827cbaf147d18fdd55e381dcec8abb4ea622a1d5a52b23b3cd71f16715d \
    -images=kube-rbac-proxy=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ec61024e2c37e5aba93593220b12ba91a3ba4f547b2562a36539ff1d9df74a23 \
    -images=telemeter-client=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:570c1c5b42268585f572478e5d22b85c7c23e42d96f71c0af95c31d54104cd67 \
    -images=prom-label-proxy=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:9a2f06206b2a3ff6fdd4d4d3a9b4b6ad43ec595fa8e765c47f5565633ba27ecf \
    -images=k8s-prometheus-adapter=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:a61b6f8fab1abbe03db8cd221179dd01c4eb622e981b93b76ca58a8fb77158c0 \
    -images=thanos=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:2327c5c575e389003c7b3d3ffb0a95a7ecaccac6403cf82d50c35eb57716fba1 \
    -images=console-plugin=quay.io/anpicker/monitoring-plugin:4-13-test  # 👈 this is new \
    -assets assets/ \
    -telemetry-config /tmp/telemetry-config.yaml \
    -kubeconfig <path>/to/kubeconfig # 👈 replace this\ 
    -namespace=openshift-monitoring \
    -configmap=cluster-monitoring-config \
    -logtostderr=true \
    -v=5 2>&1 | tee tmp/$(git branch --show-current)/cbot.log
```

pkg/client/client.go
```diff
@@ -1761,6 +1793,54 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
 	return int(podCapacityTotal), nil
 }
 
+func (c *Client) CreateOrUpdateConsolePlugin(ctx context.Context, plg *consolev1.ConsolePlugin) error {
+
+	conClient := c.osconclient.ConsoleV1().ConsolePlugins()
+	existing, err := conClient.Get(ctx, plg.GetName(), metav1.GetOptions{})
+	klog.Infof("Got error: %v", err)
+	if apierrors.IsNotFound(err) {
+		klog.Infof("creating console-plugin: %v", plg.Spec)
```
sthaha, line 1802:
```suggestion

```

pkg/client/client.go
```diff
@@ -1761,6 +1793,54 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
...
+		_, err := conClient.Create(ctx, plg, metav1.CreateOptions{})
+		klog.Infof("Got error Create: %v", err)
```
sthaha, line 1804:
```suggestion

```

pkg/client/client.go
```diff
@@ -1761,6 +1793,54 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
...
+	if err != nil {
+		return errors.Wrapf(err, "retrieving console %s failed", clusterConsole)
+	}
+
+	if slices.Contains(console.Spec.Plugins, name) {
+		klog.V(5).Info("console already contains plugin", name)
+		return nil
+	}
+
+	klog.Info("Enabling console plugin in Console")
```
sthaha, line 1832:
```suggestion

```

pkg/client/client.go
```diff
@@ -1761,6 +1793,54 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
...
+	console.Spec.Plugins = append(console.Spec.Plugins, name)
+	_, err = consoleClient.Update(ctx, console, metav1.UpdateOptions{})
+	if err != nil {
+		klog.Error(err, fmt.Sprintf("Could not update resource - APIVersion: %s, Kind: %s, Name: %s",
+			console.APIVersion, console.Kind, console.Name))
```
sthaha, line 1837:
```suggestion

```

pkg/client/client.go
```diff
@@ -1761,6 +1793,54 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
...
+		return err
```
sthaha, line 1838:
```suggestion
		return errors.Wrap(err, "failed to update console-plugin")
```

pkg/client/client.go
```diff
@@ -1761,6 +1793,54 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
 	return int(podCapacityTotal), nil
 }
 
+func (c *Client) CreateOrUpdateConsolePlugin(ctx context.Context, plg *consolev1.ConsolePlugin) error {
+
+	conClient := c.osconclient.ConsoleV1().ConsolePlugins()
+	existing, err := conClient.Get(ctx, plg.GetName(), metav1.GetOptions{})
+	klog.Infof("Got error: %v", err)
```
simonpasquier, line 1800: The error is handled below, no need to log it here?
```suggestion
```

pkg/client/client.go
```diff
@@ -1761,6 +1793,54 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
...
+	_, err = conClient.Update(ctx, required, metav1.UpdateOptions{})
+	return errors.Wrap(err, "updating ConsolePlugin object failed")
+}
+
+func (c *Client) RegisterConsolePlugin(ctx context.Context, name string) error {
+	consoleClient := c.osopclient.OperatorV1().Consoles()
+
+	console, err := consoleClient.Get(ctx, clusterConsole, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "retrieving console %s failed", clusterConsole)
```
simonpasquier, line 1824: nit
```suggestion
		return errors.Wrapf(err, "retrieving console %q failed", clusterConsole)
```

pkg/manifests/manifests.go
```diff
@@ -31,8 +31,10 @@ import (
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
+	consolev1 "github.com/openshift/api/console/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	securityv1 "github.com/openshift/api/security/v1"
+
```
simonpasquier, line 37: nit
```suggestion
```

pkg/tasks/console_plugin.go
```diff
@@ -0,0 +1,104 @@
+// Copyright 2019 The Cluster Monitoring Operator Authors
```
simonpasquier, line 1: nit
```suggestion
// Copyright 2023 The Cluster Monitoring Operator Authors
```

assets/console-plugin/console-plugin.yaml
```diff
@@ -0,0 +1,21 @@
+---
+# Source: openshift-console-plugin/templates/consoleplugin.yaml
+apiVersion: console.openshift.io/v1
+kind: ConsolePlugin
+metadata:
+  name: monitoring-plugin
```
simonpasquier, line 6: do we need to specify a namespace?

assets/console-plugin/configmap.yaml
```diff
@@ -0,0 +1,29 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: monitoring-plugin
+  namespace: openshift-monitoring
+  labels:
+    app: monitoring-plugin
+    app.kubernetes.io/name: monitoring-plugin
+    app.kubernetes.io/instance: monitoring-plugin
+    app.kubernetes.io/part-of: monitoring-plugin
```
simonpasquier, line 12:
```suggestion
    app.kubernetes.io/name: monitoring-plugin
    app.kubernetes.io/instance: monitoring-plugin
    app.kubernetes.io/part-of: openshift-monitoring
```

assets/console-plugin/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+kind: Deployment
+metadata:
+  name: monitoring-plugin
+  namespace: openshift-monitoring
+  labels:
+    app: monitoring-plugin
+    app.kubernetes.io/name: monitoring-plugin
+    app.kubernetes.io/instance: monitoring-plugin
+    app.kubernetes.io/part-of: monitoring-plugin
+    app.openshift.io/runtime-namespace: monitoring-plugin
```
simonpasquier, line 13: out of curiosity, what's the purpose of this label? I suppose that it's required by the console server?

assets/console-plugin/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+spec:
```
simonpasquier, line 14: We lack anti-affinity rules + pod disruption budget to prevent service disruption in case of node upgrades.

simonpasquier (commented):
2 questions/remarks from me:
* for consistency we should aim at generating the assets from jsonnet.
* is there any authn/authz for the console plugin deployment?

assets/console-plugin/console-plugin.yaml
sthaha: No, ConsolePlugins are cluster scoped.

assets/console-plugin/deployment.yaml
sthaha: @kyoto, could you please help answer this?

assets/console-plugin/deployment.yaml
sthaha: Good point! . I will add those.

sthaha: Done, could you please check again?

manifests/image-references
```diff
@@ -66,3 +66,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-thanos:latest
+  - name: console-plugin
+    from:
+      kind: DockerImage
+      name: quay.io/anpicker/monitoring-plugin:4-13-test
```
sthaha, line 72: @simonpasquier do you know if these must always refer to quay.io/openshift? Since we are missing this image, I wonder if this would work to `launch <pr>` ?

assets/console-plugin/deployment.yaml
kyoto: `app.openshift.io/runtime-namespace` does not appear to be used, so let's remove it.

manifests/image-references
sthaha: Resolving this. This can't be done until the console-plugin image is published to `quay.io/openshift` repo. `ci/images` also fails for the same reason. OpenShift CI  only loads know images.

cc: @kyoto

jsonnet/versions.yaml
```diff
@@ -22,3 +22,4 @@ versions:
   prometheusAdapter: 0.10.0
   prometheusOperator: 0.63.0
   thanos: 0.30.2
+  consolePlugin: 4-13
```
jan--f, line 25:
```suggestion
  consolePlugin: 4.13
```
Or is `4-13` intentional for testing with Andy's image build?

jan--f (commented):
I think this looks good.

@simonpasquier raised a good point in a chat earlier:
We have several options that can apply to most CMO components. It might be worth discussing which we should add intitally or which could be added in a follow-up PR.

1. nodeSelector
Most components can be [moved to specific nodes](https://docs.openshift.com/container-platform/4.12/monitoring/configuring-the-monitoring-stack.html#moving-monitoring-components-to-different-nodes_configuring-the-monitoring-stack). This should probably apply to the console plugin deployment as well?
2. Same for [tolerations](https://docs.openshift.com/container-platform/4.12/monitoring/configuring-the-monitoring-stack.html#assigning-tolerations-to-monitoring-components_configuring-the-monitoring-stack).
3. [TopologySpreadConstraints](https://docs.openshift.com/container-platform/4.12/monitoring/configuring-the-monitoring-stack.html#configuring_pod_topology_spread_constraintsfor_monitoring_configuring-the-monitoring-stack) are worth considering for high availability purposes.
4. [Setting logging levels](https://docs.openshift.com/container-platform/4.12/monitoring/configuring-the-monitoring-stack.html#setting-log-levels-for-monitoring-components_configuring-the-monitoring-stack) doesn't apply iiuc.

I mostly just want to bring this up so we can discuss. I'd prefer to merge this PR and add the above in follow-ups. If some of those make it into 4.13, great! If not I could live with a later addition.

simonpasquier:
> 4\. 
> 
From https://github.com/openshift/cluster-monitoring-operator/pull/1890#pullrequestreview-1305766195

> I mostly just want to bring this up so we can discuss. I'd prefer to merge this PR and add the above in follow-ups. If some of those make it into 4.13, great! If not I could live with a later addition.

I'd consider tolerations & node selector to be a must because some users have a requirement to separate infrastructure workloads from user workloads.

jsonnet/components/console-plugin.libsonnet
```diff
@@ -0,0 +1,16 @@
+local cm = (import 'console-plugin/config-map.json');
+local deploy = (import 'console-plugin/deployment.json');
+local cp = (import 'console-plugin/console-plugin.json');
+local pdb = (import 'console-plugin/pod-disruption-budget.json');
+local sa = (import 'console-plugin/service-account.json');
+local svc = (import 'console-plugin/service.json');
```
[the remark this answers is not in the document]
sthaha: These files are generated just like manifests are to `tmp/<random/assets` (instead of `tmp/random/manifests`) and included by the `build-jsonnet` script.

jsonnet/versions.yaml
sthaha: yes, that is intentional (for now) to use Andy's image.

pkg/manifests/manifests.go
```diff
@@ -2442,6 +2450,46 @@ func (f *Factory) NewValidatingWebhook(manifest io.Reader) (*admissionv1.Validat
...
 
+func (f *Factory) NewConsolePlugin(manifest io.Reader) (*consolev1.ConsolePlugin, error) {
+	return NewConsolePlugin(manifest)
+}
+
+func (f *Factory) ConsolePlugin() (*consolev1.ConsolePlugin, error) {
+	return f.NewConsolePlugin(f.assets.MustNewAssetReader(ConsolePlugin))
+}
+
+func (f *Factory) ConsolePluginDeployment() (*appsv1.Deployment, error) {
```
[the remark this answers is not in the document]
sthaha: Good idea! I will work on it but if it doesn't make it in time, can we do that as a follow up commit?

sthaha:
@kyoto could you please help answer the concern that Simon raised? 

> reiterating my question from earlier: is there any authn/authz for the console plugin service? If not, does it mean that any user who's able to run a pod in the cluster can get access to the monitoring APIs?

kyoto:
> reiterating my question from earlier: is there any authn/authz for the console plugin service? If not, does it mean that any user who's able to run a pod in the cluster can get access to the monitoring APIs?

@simonpasquier The Service we're adding is just for serving the plugin's resources (JavaScript, CSS) as described in https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md#delivering-plugins

Once the new plugin frontend code is installed and running, it accesses Thanos and Alertmanager in the same way as the old Console code.

Does that answer your question?

pkg/manifests/manifests.go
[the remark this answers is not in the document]
sthaha: I have added the nodeSelector, topologySpreadContraints, Tolerations, and Resources as config. Does that look good to you? @simonpasquier

jsonnet/versions.yaml
sthaha: @jan--f just noting that before this patch goes in, we will have to change this to match the image tag that @kyoto is publishing.

sthaha:
> I think this looks good.
> 
> @simonpasquier raised a good point in a chat earlier: We have several options that can apply to most CMO components. It might be worth discussing which we should add intitally or which could be added in a follow-up PR.
> 
> 1. nodeSelector
>    Most components can be [moved to specific nodes](https://docs.openshift.com/container-platform/4.12/monitoring/configuring-the-monitoring-stack.html#moving-monitoring-components-to-different-nodes_configuring-the-monitoring-stack). This should probably apply to the console plugin deployment as well?
> 2. Same for [tolerations](https://docs.openshift.com/container-platform/4.12/monitoring/configuring-the-monitoring-stack.html#assigning-tolerations-to-monitoring-components_configuring-the-monitoring-stack).
> 3. [TopologySpreadConstraints](https://docs.openshift.com/container-platform/4.12/monitoring/configuring-the-monitoring-stack.html#configuring_pod_topology_spread_constraintsfor_monitoring_configuring-the-monitoring-stack) are worth considering for high availability purposes.
> 4. [Setting logging levels](https://docs.openshift.com/container-platform/4.12/monitoring/configuring-the-monitoring-stack.html#setting-log-levels-for-monitoring-components_configuring-the-monitoring-stack) doesn't apply iiuc.
> 
> I mostly just want to bring this up so we can discuss. I'd prefer to merge this PR and add the above in follow-ups. If some of those make it into 4.13, great! If not I could live with a later addition.

I didn't see a log-level that we can set since the plugin iiuc, is only an nginx that serves static content (js+css)
I have added all configs mentioned above + resource limits, could you please take a look again?

jsonnet/components/console-plugin.libsonnet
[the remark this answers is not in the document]
sthaha: So the way this works is that  `make build-jsonnet` will first convert `assets/console-plugin/*.yaml` to `tmp/<rand>/console-plugin/*.json` and then this directory is included in the jsonnet path (-J) by build-jsonnet.sh script. These files are imported by the `console-plugin.libsonnet` which is then used to generate the yaml files. So essentially, we just need to maintain the yaml files until this is completely replaced by proper jsonnet in a follow up PR.

jan--f:
/retitle [WIP] MON-2967: CMO deploys monitoring console-plugin

jan--f:
This will not make 4.13 but rather target 4.14.

jsonnet/components/console-plugin.libsonnet
```diff
@@ -0,0 +1,210 @@
+local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
+
+function(params)
```
sthaha, line 3: @raptorsun A little bird to me that you are one of the jsonnet experts. It would be great if you   could please help review this jsonnet magic :)

sthaha: I didn't realise ksonnet was deprecated 🤦 .. I will convert everything to not use it.

sthaha:
Just noting that we need https://github.com/openshift/release/pull/36531/files to be merged before tests start passing.

assets/console-plugin/config-map.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
+data:
+  nginx.conf: |
+    error_log /dev/stdout info;
+    events {}
+    http {
+      access_log         /dev/stdout;
```
raptorsun, line 7: Is there a risk of excessive logging? Maybe we do not need every access to be logged, we just need to log exceptional access such as errors.
Here is conditional logging config: https://docs.nginx.com/nginx/admin-guide/monitoring/logging/#enabling-conditional-logging

sthaha: I am not sure what is the right amount of logging that the new plugin should have.  This is based on the helm chart for plugin. 

@kyoto any thoughts?

sthaha: I have removed the access logs.

bparees:
the monitor-plugin image promotion has merged, so this should be able to successfully reference the image now.

/retest

manifests/image-references
```diff
@@ -66,3 +66,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-thanos:latest
+  - name: console-plugin
```
bparees, line 69: sorry I think this needs to be `monitoring-plugin` since that is what the image is being promoted as.

sthaha: I see! thank you!

pkg/manifests/manifests.go
```diff
@@ -2558,6 +2569,72 @@ func (f *Factory) NewValidatingWebhook(manifest io.Reader) (*admissionv1.Validat
...
+	}
+
+	// ensure console-plugin container is present even if config isn't defined so that,
+	// we validate that deployment has the expected container name. Thereby avoiding
+	// any suprises should user add config later.
+	podSpec := d.Spec.Template.Spec
+	containers := podSpec.Containers
+	idx := slices.IndexFunc(containers, containerNameEquals(MonitoringPluginDeploymentContainer))
+	if idx < 0 {
+		return nil, errors.Errorf("failed to find console-plugin container: %s", MonitoringPluginDeploymentContainer)
```
jan--f, line 2593: Maybe log where we failed to find the container too, i.e. the deployement?

jan--f (commented):
:+1:  I think this looks good, only one small nit.
Holding of on adding lgtm, more eyes on this would be good.

pkg/manifests/manifests.go
sthaha: I think the convention we have in CMO is to either log or return error since errors are logged. @simonpasquier can confirm

jan--f: Sorry my mistake, I merely meant let's include where the container couldn't be found in the message. Didn't mean to request extra logging...apologies!

pkg/client/client.go
```diff
@@ -1833,6 +1864,43 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
...
+		return errors.Wrapf(err, "retrieving console %q failed", clusterConsole)
+	}
+
+	if slices.Contains(console.Spec.Plugins, name) {
+		klog.V(5).Info("console already contains plugin", name)
+		return nil
+	}
+
+	console.Spec.Plugins = append(console.Spec.Plugins, name)
+	_, err = consoleClient.Update(ctx, console, metav1.UpdateOptions{})
```
simonpasquier, line 1900: I wonder if we could use `Patch()` instead to be more specific about what we update in the resource? Especially we could use JSON patch to append the item to the plugins array (https://datatracker.ietf.org/doc/html/rfc6902?_gl=1*1xvkbi9*_ga*NjMxMDY2NzQ4LjE2ODA3NzYyMTU.*_ga_4RQQZ3WGE9*MTY4MjUxMDY4My4yLjEuMTY4MjUxMDcwMC40My4wLjA.#appendix-A.2).

pkg/manifests/manifests.go
simonpasquier: yes I agree with @sthaha: you want to either return the error or log it but not both. And it's not to this repository but more a general Go idiom.

pkg/manifests/manifests.go
```diff
@@ -2558,6 +2569,72 @@ func (f *Factory) NewValidatingWebhook(manifest io.Reader) (*admissionv1.Validat
...
+
+func (f *Factory) MonitoringPluginDeployment() (*appsv1.Deployment, error) {
+	d, err := f.NewDeployment(f.assets.MustNewAssetReader(MonitoringPluginDeployment))
+	if err != nil {
+		return nil, err
+	}
+
+	// ensure console-plugin container is present even if config isn't defined so that,
+	// we validate that deployment has the expected container name. Thereby avoiding
+	// any suprises should user add config later.
```
simonpasquier, line 2588: not sure what you mean by "any surprises should user add config later"?

pkg/client/client.go
sthaha: Would it be okay to address this in separate patch so that the feature itself can be tested?

pkg/manifests/manifests.go
sthaha: The initial code I wrote was to return if the config for console-plugin is nil. 
Now .. lets say the `console-plugin` container is absent (for whatever reason .. may be a typo, jsonnet issue). User then adds the config to set resource limits. They would then start experiencing error which may be confusing (like it confused me).
By ensuring that the container is present, errors will be caught earlier.

pkg/manifests/manifests.go
sthaha: The error would be something like 
```
UpdateConsolePluginFailed: failed to find console-plugin container: console-plugin
```
Doesn't that suffice?

pkg/client/client.go
```diff
@@ -194,6 +219,12 @@ func OpenshiftRouteClient(osrclient openshiftrouteclientset.Interface) Option {
 	}
 }
 
+func OpenshiftOperatorClient(osopclient openshiftoperatorclientset.Interface) Option {
```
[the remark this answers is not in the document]
sthaha: @simonpasquier, you are right! I went with the convention I saw in the file to create these Client functions. I am happy to remove this in the follow up patch to use `Patch`

pkg/manifests/manifests.go
sthaha: @jan--f, I will add the deployment info in the subsequent patch, since I have some modifications anyway

pkg/client/client.go
sthaha: @simonpasquier , I tried the following (hoping this is what you meant)

```go
	jsonPatch := []jsonpatch.Operation{{
		Operation: "add",
		Path:      "/spec/plugins/-",
		Value:     name,
	}}

	patch, err := json.Marshal(jsonPatch)
	if err != nil {
		panic(err)
	}

	_, err = consoleClient.Patch(ctx, clusterConsole, types.JSONPatchType, patch, metav1.PatchOptions{})
```

But this fails with the default `console` spec (below) which doesn't have the `plugins` array defined. It also doesn't really save us from having to `get` and check if the plugin is already added. 

Do you have a better way of doing this? 

```
apiVersion: operator.openshift.io/v1
kind: Console
metadata:
  name: cluster
spec:
  logLevel: Normal
  managementState: Managed
  operatorLogLevel: Normal
```

pkg/manifests/manifests.go
```diff
@@ -2558,6 +2569,74 @@ func (f *Factory) NewValidatingWebhook(manifest io.Reader) (*admissionv1.Validat
...
+	podSpec := d.Spec.Template.Spec
+	containers := podSpec.Containers
+	idx := slices.IndexFunc(containers, containerNameEquals(MonitoringPluginDeploymentContainer))
+	if idx < 0 {
+		return nil, errors.Errorf(
+			"failed to find console-plugin container %q in deployment %q",
+			MonitoringPluginDeploymentContainer, MonitoringPluginDeployment)
```
sthaha, line 2595: @jan--f looks good enough?

pkg/client/client.go
simonpasquier: I haven't tested fully but I believe that we'd have to deal with 2 different patches:
* When `.spec.plugins` isn't defined
```json
[{
		"op": "add",
		"path":      "/spec/plugins/",
		"value":     ["monitoring-plugin"],
}]
```

* When `.spec.plugins` is defined
```json
[{
		"op": "add",
		"path":      "/spec/plugins/-",
		"value":     "monitoring-plugin",
}]
```

> It also doesn't really save us from having to get and check if the plugin is already added.

Sure but it should avoid the risk of overwriting the full object while we only care about inserting one item in the list.

pkg/manifests/manifests.go
simonpasquier: got it! the other option would be to fail hard (e.g. panic) since it would mean that the asset and the Go code are out of sync.

pkg/client/client.go
sthaha: > Sure but it should avoid the risk of overwriting the full object while we only care about inserting one item in the list.

I have implemented this but I am not sure if this is less error prone that the method used earlier because we don't risk changing the whole object as k8s api-server already prevent us from doing this (occ) ?

 Isn't `get`, `modify`, `update` proven to be working without issues? If the object changes, after `get` IIUC, k8s will reject the change we made based on the `metadata.resourceVersion` and return a `conflict` error asking us to retry

simonpasquier: > Isn't get, modify, update proven to be working without issues? If the object changes, after get IIUC, k8s will reject the change we made based on the metadata.resourceVersion and return a conflict error asking us to retry

Sure but it's still less elegant and efficient than a patch operation, especially since the resource may be managed by several controllers.

Tai-RedHat:
keep failing on build cluster with PR:
```
error: unable to create a release: operator "cluster-monitoring-operator" contained an invalid image-references file: no input image tag named "monitoring-plugin"
{"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"k8s.io/test-infra/prow/entrypoint/run.go:79","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"2023-05-17T00:22:19Z"}
---}
```
[logs](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws-modern/1658627206542266368), [logs2](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws-modern/1658636066040582144)

sthaha:
@bparees could you please help shed some light into why `monitoring-plugin` image is missing when using  clusterbot to launch the cluster ( with `launch openshift/cluster-monitoring-operator#1890`  )

I see that the monitoring-plugin image is missing in the [image-stream](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws-modern/1658636066040582144/artifacts/build-resources/imagestreams.json) that @Tai-RedHat tried launching where as the [`e2e-aws-ovn-upgrade` image stream](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/1890/pull-ci-openshift-cluster-monitoring-operator-master-e2e-aws-ovn/1653988023500541952/artifacts/build-resources/imagestreams.json) contains it.

bparees:
> @bparees could you please help shed some light into why monitoring-plugin image is missing when using clusterbot to launch the cluster ( with launch openshift/cluster-monitoring-operator#1890 )

sorry i'm pretty swamped at the moment, can the test platform team help you?

Tai-RedHat:
will do **Post-merge test** since the PR still can not build with cluster-bot.

jan--f:
/label docs-approved
/label px-approved
/label qe-approved
Adding qe approval based on https://github.com/openshift/cluster-monitoring-operator/pull/1890#issuecomment-1567711656

PR #1900, opened 2023-02-22
OCPBUGS-7391: wait for service CA secrets

simonpasquier:
The Prometheus adapter and control plane tasks would exit immediately if they couldn't find the TLS secrets generated by the service CA operator.

With this change, the tasks are waiting up to 5 minutes for the secrets to appear which should speed up a bit the reconciliation (no need to run another attempt).

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1204,22 +1204,21 @@ func (c *Client) WaitForSecret(ctx context.Context, s *v1.Secret) (*v1.Secret, e
 	var lastErr error
 	if err := wait.Poll(1*time.Second, 5*time.Minute, func() (bool, error) {
 		var err error
+
 		result, err = c.kclient.CoreV1().Secrets(s.Namespace).Get(ctx, s.Name, metav1.GetOptions{})
-
-		if apierrors.IsNotFound(err) {
-			lastErr = err
-			return false, nil
-		}
-
 		if err != nil {
 			lastErr = err
-			klog.V(4).ErrorS(err, "WaitForSecret: failed to get Secret")
 			return false, nil
 		}
 
-		for _, v := range result.Data {
+		if len(result.Data) == 0 {
+			lastErr = errors.New("secret contains no data")
+			return false, nil
+		}
+
+		for k, v := range result.Data {
 			if len(v) == 0 {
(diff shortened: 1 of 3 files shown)


pkg/client/client.go
```diff
@@ -1245,6 +1244,52 @@ func (c *Client) WaitForSecretByNsName(ctx context.Context, obj types.Namespaced
...
+		var err error
+
+		result, err = c.kclient.CoreV1().ConfigMaps(cm.Namespace).Get(ctx, cm.Name, metav1.GetOptions{})
+		if err != nil {
+			lastErr = err
+			return false, nil
+		}
+
+		if len(result.Data) == 0 {
+			lastErr = errors.New("configmap contains no data")
```
JoaoBraveCoding, line 1260: nit: either the error message should say `configmap` or `ConfigMap`, I'll let you decide, I suggested this since online 1276 you had written `ConfigMap`. Also shouldn't we say which ConfigMap the error refers to?
```suggestion
			lastErr = errors.New("ConfigMap contains no data")
```

pkg/client/client.go
```diff
@@ -1245,6 +1244,52 @@ func (c *Client) WaitForSecretByNsName(ctx context.Context, obj types.Namespaced
 	return c.WaitForSecret(ctx, &secret)
 }
 
+func (c *Client) WaitForConfigMap(ctx context.Context, cm *v1.ConfigMap) (*v1.ConfigMap, error) {
```
JoaoBraveCoding, line 1247: nit: Is this code not used? Or am I missing something 🤔 Do we have plans to start using it?

JoaoBraveCoding (commented):
just two nit's but it lgtm!

pkg/client/client.go
simonpasquier: I used the lower-case version here because the convention for error messages is that they shouldn't be capitalized.

pkg/client/client.go
simonpasquier: good catch! I forgot to modify the Prometheus adapter and control plane tasks.

JoaoBraveCoding (approved):
lovely! great job!
/lgtm

PR #1904, opened 2023-02-27
MON-3229: Remove the dependency on the apiserver auth

marioferh:
## Description

When the control plane nodes are under pressure or the apiserver is just not available, no telemetry data is emitted by the monitoring stack 

## Solution 

Remove the dependency on the apiserver would be to use mTLS communication between telemeter-client and the Prometheus pods.

Add /federate endpoint to rbac proxy and allow telemeter-client to authenticate via mTLS to reach Prometheus metrics.

Add mTLS auth to telemeter-client https://github.com/openshift/telemeter/pull/455

## Type of change

- [X] Remove the dependency on the apiserver auth
- [X] kube-rbac-proxy exposing the /federate endpoint
- [X] Add tests
- [X] No user facing changes, so no entry in CHANGELOG was needed.

### DIFF
--- a/jsonnet/components/prometheus.libsonnet
+++ b/jsonnet/components/prometheus.libsonnet
@@ -203,7 +203,25 @@ function(params)
       data: {},
     },
 
-    kubeRbacProxySecret: generateSecret.staticAuthSecret(cfg.namespace, cfg.commonLabels, 'kube-rbac-proxy'),
+    kubeRbacProxySecret: generateSecret.staticAuthSecret(
+      cfg.namespace,
+      cfg.commonLabels,
+      'kube-rbac-proxy',
+      {
+        authorization+: {
+          static+: [
+            {
+              user: {
+                name: 'system:serviceaccount:openshift-monitoring:prometheus-k8s',
+              },
+              verb: 'get',
+              path: '/federate',
+              resourceRequest: false,
+            },
+          ],
+        },
+      },
+    ),
 
     // Secret holding the token to authenticate against the Telemetry server when using native remote-write.
     telemetrySecret: {
@@ -397,7 +415,7 @@ function(params)
             args: [
               '--secure-listen-address=0.0.0.0:9092',
               '--upstream=http://127.0.0.1:9090',
-              '--allow-paths=/metrics',
+              '--allow-paths=/metrics,/federate',
               '--config-file=/etc/kube-rbac-proxy/config.yaml',
               '--tls-cert-file=/etc/tls/private/tls.crt',
               '--tls-private-key-file=/etc/tls/private/tls.key',
--- a/jsonnet/components/prometheus.libsonnet
+++ b/jsonnet/components/prometheus.libsonnet
@@ -210,19 +210,9 @@ function(params)
       {
         authorization+: {
           static+: [
-            // prometheus-k8s is needed for mTLS due to cert is signed with this SA
-            // instead for token auth telemeter-client SA is required
             {
               user: {
                 name: 'system:serviceaccount:openshift-monitoring:prometheus-k8s',
-              },
-              verb: 'get',
-              path: '/federate',
-              resourceRequest: false,
-            },
-            {
-              user: {
-                name: 'system:serviceaccount:openshift-monitoring:telemeter-client',
               },
               verb: 'get',
               path: '/federate',
@@ -432,7 +422,6 @@ function(params)
               '--client-ca-file=/etc/tls/client/client-ca.crt',
               '--tls-cipher-suites=' + cfg.tlsCipherSuites,
               '--logtostderr=true',
-              '--v=10',
             ],
             terminationMessagePolicy: 'FallbackToLogsOnError',
             volumeMounts: [
--- a/jsonnet/components/prometheus.libsonnet
+++ b/jsonnet/components/prometheus.libsonnet
@@ -210,9 +210,19 @@ function(params)
       {
         authorization+: {
           static+: [
+            // prometheus-k8s is needed for mTLS due to cert is signed with this SA
+            // instead for token auth telemeter-client SA is required
             {
               user: {
                 name: 'system:serviceaccount:openshift-monitoring:prometheus-k8s',
+              },
+              verb: 'get',
+              path: '/federate',
+              resourceRequest: false,
+            },
+            {
+              user: {
+                name: 'system:serviceaccount:openshift-monitoring:telemeter-client',
               },
               verb: 'get',
               path: '/federate',
@@ -422,6 +432,7 @@ function(params)
               '--client-ca-file=/etc/tls/client/client-ca.crt',
               '--tls-cipher-suites=' + cfg.tlsCipherSuites,
               '--logtostderr=true',
+              '--v=10',
             ],
             terminationMessagePolicy: 'FallbackToLogsOnError',
             volumeMounts: [
--- a/jsonnet/utils/generate-secret.libsonnet
+++ b/jsonnet/utils/generate-secret.libsonnet
@@ -1,5 +1,5 @@
 {
-  staticAuthSecret(cfgNamespace, cfgCommonLabels, cfgName):: {
+  staticAuthSecret(cfgNamespace, cfgCommonLabels, cfgName, additionalConfig={}):: {
     apiVersion: 'v1',
     kind: 'Secret',
     metadata: {
@@ -23,7 +23,7 @@
             },
           ],
         },
-      },),
+      } + additionalConfig),
     },
   },
 }
--- a/test/e2e/telemeter_test.go
+++ b/test/e2e/telemeter_test.go
@@ -121,7 +121,7 @@ func TestTelemeterClient(t *testing.T) {
 			`federate_samples{job="telemeter-client"}`,
 			func(v float64) error {
 				if v < 10 {
-					return fmt.Errorf("expecting federate samples from telemeter client more than 10 but got %f", v)
+					return fmt.Errorf("expecting federate samples from telemeter client greater than or equal to 10 but got %f", v)
 				}
 				return nil
 			},
--- a/test/e2e/telemeter_test.go
+++ b/test/e2e/telemeter_test.go
@@ -17,6 +17,7 @@ package e2e
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 	"time"
 
@@ -97,4 +98,33 @@ func TestTelemeterRemoteWrite(t *testing.T) {
 			return nil
 		},
 	)
+}
+
+// TestTelemeterClient verifies that the telemeter client can collect metrics from the monitoring stack and forward them to the telemeter server.
+func TestTelemeterClient(t *testing.T) {
+	{
+		f.PrometheusK8sClient.WaitForQueryReturn(
+			t,
+			5*time.Minute,
+			`metricsclient_request_send{client="federate_to",job="telemeter-client",status_code="200"}`,
+			func(v float64) error {
+				if v == 0 {
+					return fmt.Errorf("expecting metricsclient request send more than 0 but got none")
+				}
+				return nil
+			},
+		)
+
+		f.PrometheusK8sClient.WaitForQueryReturn(
+			t,
+			5*time.Minute,
+			`federate_samples{job="telemeter-client"}`,
+			func(v float64) error {
+				if v < 10 {
+					return fmt.Errorf("expecting federate samples from telemeter client more than 10 but got %f", v)
+				}
+				return nil
+			},
+		)
+	}
 }
--- a/assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
+++ b/assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
@@ -20,9 +20,4 @@ stringData:
         "user":
           "name": "system:serviceaccount:openshift-monitoring:prometheus-k8s"
         "verb": "get"
-      - "path": "/federate"
-        "resourceRequest": false
-        "user":
-          "name": "system:serviceaccount:openshift-monitoring:telemeter-client"
-        "verb": "get"
 type: Opaque
--- a/assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
+++ b/assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
@@ -15,4 +15,9 @@ stringData:
         "user":
           "name": "system:serviceaccount:openshift-monitoring:prometheus-k8s"
         "verb": "get"
+      - "path": "/federate"
+        "resourceRequest": false
+        "user":
+          "name": "system:serviceaccount:openshift-monitoring:prometheus-k8s"
+        "verb": "get"
 type: Opaque
--- a/assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
+++ b/assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
@@ -20,4 +20,9 @@ stringData:
         "user":
           "name": "system:serviceaccount:openshift-monitoring:prometheus-k8s"
         "verb": "get"
+      - "path": "/federate"
+        "resourceRequest": false
+        "user":
+          "name": "system:serviceaccount:openshift-monitoring:telemeter-client"
+        "verb": "get"
 type: Opaque
--- a/assets/prometheus-k8s/prometheus.yaml
+++ b/assets/prometheus-k8s/prometheus.yaml
@@ -82,7 +82,7 @@ spec:
   - args:
     - --secure-listen-address=0.0.0.0:9092
     - --upstream=http://127.0.0.1:9090
-    - --allow-paths=/metrics
+    - --allow-paths=/metrics,/federate
     - --config-file=/etc/kube-rbac-proxy/config.yaml
     - --tls-cert-file=/etc/tls/private/tls.crt
     - --tls-private-key-file=/etc/tls/private/tls.key


simonpasquier:
I think that we also need to update the telemeter client so it can authenticate against the Prometheus server using a client TLS certificate. Right now, the forwarder package (which reads the metrics from the /federate endpoint) only supports bearer token authentication.

https://github.com/openshift/telemeter/blob/ee1ba4699b82ecb2033f886af12b9e2e451d2296/pkg/forwarder/forwarder.go#L149-L182

marioferh:
> 
Testing TLS changes
https://github.com/openshift/telemeter/pull/455

jsonnet/utils/generate-secret.libsonnet
```diff
@@ -21,6 +21,14 @@
               path: '/metrics',
               resourceRequest: false,
             },
+            {
```
raptorsun, line 24: We only need prometheus to expose `/federate` endpoint to telemeter client. 
I think it better to limit the access to `/federate` endpoint to the `kube-rbac-proxy` before `prometheus-k8s` .
We can make a dedicated function to `prometheus-k8s` deployment or make a wrap function to add this config about `/federate` to the result of `staticAuthSecret`

jsonnet/components/prometheus.libsonnet
```diff
@@ -397,13 +397,14 @@ function(params)
...
               '--upstream=http://127.0.0.1:9090',
-              '--allow-paths=/metrics',
+              '--allow-paths=/metrics,/federate',
               '--config-file=/etc/kube-rbac-proxy/config.yaml',
               '--tls-cert-file=/etc/tls/private/tls.crt',
               '--tls-private-key-file=/etc/tls/private/tls.key',
               '--client-ca-file=/etc/tls/client/client-ca.crt',
               '--tls-cipher-suites=' + cfg.tlsCipherSuites,
               '--logtostderr=true',
+              '--v=10',
```
raptorsun, line 407: when we are ready to merge the PR, don't forget to remove this  verbose output

pkg/manifests/manifests.go
```diff
@@ -1105,6 +1106,10 @@ func (f *Factory) PrometheusRBACProxySecret() (*v1.Secret, error) {
 	return f.NewSecret(f.assets.MustNewAssetReader(PrometheusRBACProxySecret))
 }
 
+func (f *Factory) PrometheusRBACProxySecretFederate() (*v1.Secret, error) {
```
raptorsun, line 1109: it has to add call to this function here https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/tasks/prometheus.go#L133-L136
So that the secret will be created by the operator.

marioferh: done

assets/prometheus-k8s/kube-rbac-proxy-secret-federate.yaml
```diff
@@ -0,0 +1,23 @@
...
+      "static":
+      - "path": "/metrics"
+        "resourceRequest": false
+        "user":
+          "name": "system:serviceaccount:openshift-monitoring:prometheus-k8s"
+        "verb": "get"
+      - "path": "/federate"
+        "resourceRequest": false
+        "user":
+          "name": "system:serviceaccount:openshift-monitoring:telemeter-client"
```
simonpasquier, line 21: I like that we use a different username but CMO would need to request a new certificate from the Kube API (similar to https://github.com/openshift/cluster-monitoring-operator/blob/d1179dcd3090bc31e3de08259c0c792a1f3a0061/pkg/operator/operator.go#L380).

simonpasquier (commented):
we need to pass the TLS certificate and key to the telemeter deployment.

test/e2e/prometheus_test.go
```diff
@@ -346,3 +349,73 @@ func TestBodySizeLimit(t *testing.T) {
 	)
 
 }
+
+func TestFederateEndpointProxy(t *testing.T) {
```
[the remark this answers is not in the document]
simonpasquier: see https://github.com/openshift/origin/pull/27915

assets/telemeter-client/kube-rbac-proxy-secret.yaml
```diff
@@ -15,4 +15,9 @@ stringData:
         "user":
           "name": "system:serviceaccount:openshift-monitoring:prometheus-k8s"
         "verb": "get"
+      - "path": "/federate"
```
simonpasquier, line 18: hmm this should go to `assets/prometheus-k8s/kube-rbac-proxy-secret.yaml` rather as this is the configuration of the prometheus-k8s's kube-rbac-proxy sidecar that needs to allow /federate?
Conversely the telemeter-client service account needs the GET permission on /federate.

pkg/operator/operator.go
```diff
@@ -405,6 +405,34 @@ func New(
 	}
 	o.controllersToRunFunc = append(o.controllersToRunFunc, csrController.Run, o.ruleController.Run, o.relabelController.Run)
 
+	csrFederateController, err := csr.NewClientCertificateController(
```
simonpasquier, line 408: It might be worth commenting about the purpose of this controller? e.g.

```suggestion
	// csrFederateController runs a controller that requests a client TLS certificate for the telemeter client. This certificate is used to authenticate against the Prometheus /federate API endpoint.
	csrFederateController, err := csr.NewClientCertificateController(
```

pkg/operator/operator.go
```diff
@@ -405,6 +405,34 @@ func New(
...
+		csr.ClientCertOption{
+			SecretNamespace: "openshift-monitoring",
+			SecretName:      "metrics-client-certs",
```
simonpasquier, line 411: you want a different secret name I believe.

pkg/operator/operator.go
```diff
@@ -405,6 +405,34 @@ func New(
...
+		},
+		csr.CSROption{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "system:openshift:openshift-monitoring-",
+				Labels: map[string]string{
+					"metrics.openshift.io/csr.subject": "prometheus",
```
simonpasquier, line 417: wonder if we need a different value here too?

pkg/operator/operator.go
marioferh: Right, sorry copy paste

test/e2e/telemeter_client_test.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2019 The Cluster Monitoring Operator Authors
```
sthaha, line 1: nit: 2023 :)

marioferh:
| /test e2e-aws-ovn-single-node
-- | --

test/e2e/telemeter_client_test.go
marioferh: Right, also the comment explaining the teste is wrong. Thanks

assets/admission-webhook/deployment.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
     app.kubernetes.io/managed-by: cluster-monitoring-operator
     app.kubernetes.io/name: prometheus-operator-admission-webhook
     app.kubernetes.io/part-of: openshift-monitoring
-    app.kubernetes.io/version: 0.63.0
+    app.kubernetes.io/version: 0.65.1
```
sthaha, line 8: @marioferh ,  suggestion for next time. Could you please keep the unrelated changes in a separate commit?

marioferh: sure, sorry it was make versions but now I know it is not necessary :)

simonpasquier: @marioferh can you revert the changes to jsonnet/versions.yaml and regenerate the assets? it would indeed help reviewing the PR.

test/e2e/telemeter_client_test.go
```diff
@@ -0,0 +1,97 @@
...
+		"overrides": [{
+			"group": "apps",
+			"kind": "Deployment",
+			"name": "cluster-monitoring-operator",
+			"namespace": "openshift-monitoring",
+			"unmanaged": false
+		}]
+	}
+}`)
+		_, err := f.OpenShiftConfigClient.ConfigV1().ClusterVersions().Patch(ctx, "version", types.MergePatchType, patch, metav1.PatchOptions{})
```
simonpasquier, line 63: You shouldn't have to put CMO into unmanaged state? IIUC the test only needs to check that telemeter client collects metrics from Prometheus.

test/e2e/telemeter_client_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"time"
+
+	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// TestTelemeterClient verifies that the monitoring stack can get data on
+// telemeter client from /federate endpoint
```
simonpasquier, line 30:
```suggestion
// TestTelemeterClient verifies that the telemeter client can collect metrics from the monitoring stack and forward them to the telemeter server.
```

jsonnet/components/prometheus.libsonnet
```diff
@@ -167,6 +167,10 @@ function(params)
           resourceNames: ['nonroot'],
           verbs: ['use'],
         },
+        {
+          nonResourceURLs: ['/federate'],
+          verbs: ['get'],
+        },
```
simonpasquier, line 173: is this needed?

test/e2e/telemeter_client_test.go
marioferh: its unmanaged: false to diferenciate the other test that is unmanaged: true
But I can delete de line if its clearer

jsonnet/components/prometheus.libsonnet
marioferh: Yep, without this change in the role auth is rejected

test/e2e/telemeter_client_test.go
marioferh: its unmanaged: false to diferenciate the other test that is unmanaged: true
But I can delete de line if its clearer

marioferh: done

test/e2e/telemeter_client_test.go
marioferh: done

marioferh:
https://github.com/openshift/cluster-monitoring-operator/pull/1904#discussion_r1213055609

About this one, I think is better to wait for the fixed one https://github.com/openshift/cluster-monitoring-operator/pull/1924 and do a rebase. I prefer that to check every file and revert it

test/e2e/telemeter_client_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"time"
+
+	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// TestTelemeterClient verifies that the telemeter client can collect metrics from the monitoring stack and forward them to the telemeter server.
+
```
simonpasquier, line 30: (nit)
```suggestion
```

test/e2e/telemeter_client_test.go
simonpasquier: yes please, I think that you only need to keep the code starting at L71. Also the test case could go in test/e2e/test_telemeter.go

pkg/operator/operator.go
```diff
@@ -405,6 +405,37 @@ func New(
...
+			SecretNamespace: "openshift-monitoring",
+			SecretName:      "federate-client-certs",
+		},
+		csr.CSROption{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "system:openshift:openshift-monitoring-",
+				Labels: map[string]string{
+					"metrics.openshift.io/csr.subject": "prometheus",
+				},
+			},
```
simonpasquier, line 420: I'd try with an empty CSROption, I don't think that it is used anywhere by the controller.

```suggestion
		csr.CSROption{},
```

jsonnet/components/prometheus.libsonnet
simonpasquier: hmm I'm wondering why it is the case?

marioferh:
https://github.com/openshift/cluster-monitoring-operator/pull/1904#discussion_r1213185167

If /federate is not added there then this error is shown:
`I0529 12:28:05.416685       1 request.go:1154] Response Body: {"kind":"SubjectAccessReview","apiVersion":"authorization.k8s.io/v1","metadata":{"creationTimestamp":null,"managedFields":[{"manager":"kube-rbac-proxy","operation":"Update","apiVersion":"authorization.k8s.io/v1","time":"2023-05-29T12:28:05Z","fieldsType":"FieldsV1","fieldsV1":{"f:spec":{"f:groups":{},"f:nonResourceAttributes":{".":{},"f:path":{},"f:verb":{}},"f:user":{}}}}]},"spec":{"nonResourceAttributes":{"path":"/federate","verb":"get"},"user":"system:serviceaccount:openshift-monitoring:prometheus-k8s","groups":["system:authenticated"]},"status":{"allowed":false}}`

jsonnet/components/prometheus.libsonnet
marioferh: If /federate is not added there then this error is shown:
I0529 12:28:05.416685       1 request.go:1154] Response Body: {"kind":"SubjectAccessReview","apiVersion":"authorization.k8s.io/v1","metadata":{"creationTimestamp":null,"managedFields":[{"manager":"kube-rbac-proxy","operation":"Update","apiVersion":"authorization.k8s.io/v1","time":"2023-05-29T12:28:05Z","fieldsType":"FieldsV1","fieldsV1":{"f:spec":{"f:groups":{},"f:nonResourceAttributes":{".":{},"f:path":{},"f:verb":{}},"f:user":{}}}}]},"spec":{"nonResourceAttributes":{"path":"/federate","verb":"get"},"user":"system:serviceaccount:openshift-monitoring:prometheus-k8s","groups":["system:authenticated"]},"status":{"allowed":false}}

test/e2e/telemeter_client_test.go
marioferh: The idea is to decouple tests as this one is testing telemeter-client and the other one prometheus remoteWrite. In the future when telemeter client is not needed it will clearer delete this test.
In any case, if needed I can move it to test_telemeter.go

test/e2e/telemeter_client_test.go
marioferh: done

jsonnet/components/prometheus.libsonnet
simonpasquier: `system:serviceaccount:openshift-monitoring:prometheus-k8s` <- it isn't using the proper certificate?

pkg/operator/operator.go
marioferh: It returns an error.:
```
E0602 11:43:35.715902  131677 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory addr
ess or nil pointer dereference)            
```

simonpasquier: hmm do you have the full stack trace?

jsonnet/components/prometheus.libsonnet
marioferh: It is using this one https://github.com/openshift/cluster-monitoring-operator/pull/1990
but still rejected

marioferh: https://github.com/brancz/kube-rbac-proxy/tree/master/examples/non-resource-url

marioferh: You are right, just the proxy doesn't show the allowed true :S
trying with a curl:
> GET /federate? HTTP/2
> Host: prometheus-k8s.openshift-monitoring.svc:9092
> User-Agent: curl/7.61.1
> Accept: */*

marioferh:
it seems as everything is working with last patch of https://github.com/openshift/telemeter/pull/455
As example:
with cmd: `curl --connect-timeout 5 -v -s --fail --cert /etc/tls/private/tls.crt --key /etc/tls/private/tls.key --cacert ./etc/serving-certs-ca-bundle/service-ca.crt  "https://prometheus-k8s.openshift-monitoring.svc:9092/federate?"   -G --data-urlencode 'match[]={job="telemeter-client"}'`

```
# TYPE metricsclient_request_send untyped                                              
metricsclient_request_send{client="federate_to",container="kube-rbac-proxy",endpoint="https",instance="10.131.0.29:8443",job="telemeter-client",namespace="openshift-monitoring",pod="telemeter-client-858d8d9c5b-hwxq2",service="telemeter-client",status_code="200",prometheus="openshift-monitoring/k8s",prometheus_replica="prometheus-k8s-1"} 2 1686586
998902          
```

After this PR is merged I can add some test in telemeter side. Also this tests in tests origin is checking always that cmo and telemeter federate scraping is working right: https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus.go#L276-L305

raptorsun:
/retitle MON-3229: Remove the dependency on the apiserver auth

jsonnet/components/prometheus.libsonnet
```diff
@@ -203,7 +203,25 @@ function(params)
...
+    kubeRbacProxySecret: generateSecret.staticAuthSecret(
+      cfg.namespace,
+      cfg.commonLabels,
+      'kube-rbac-proxy',
+      {
+        authorization+: {
+          static+: [
+            {
+              user: {
+                name: 'system:serviceaccount:openshift-monitoring:prometheus-k8s',
```
sthaha, line 215: nit: This seems to repeat at various place. Perhaps we should store this in a variable / config.

test/e2e/telemeter_test.go
```diff
@@ -98,3 +99,32 @@ func TestTelemeterRemoteWrite(t *testing.T) {
...
+				return nil
+			},
+		)
+
+		f.PrometheusK8sClient.WaitForQueryReturn(
+			t,
+			5*time.Minute,
+			`federate_samples{job="telemeter-client"}`,
+			func(v float64) error {
+				if v < 10 {
```
sthaha, line 123: I think we should either improve the error message to read `greater than or equal to 10`  or change `<` to  `<=` ? E.g . if v = 10, should this result in error? In the case it won't but the message says "v should be more than 10

marioferh: done

marioferh:
/hold
wait for more reviews

jsonnet/components/prometheus.libsonnet
raptorsun: the other reference to this user is here https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/utils/generate-secret.libsonnet#L18

I think it not so repetitive as all the `*rbac-secret.yaml` files are generated from that function.

marioferh: if its ok for @sthaha we can moving forward with this PR and maybe do a refactor in the future with other variables configs or SA.

sthaha: @marioferh , `nit` can always be ignored 🤗

PR #1906, opened 2023-02-28
Adds startupProbe to prometheus-adapter

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/OCPBUGS-7694

Problem: in clusters with a large nb of CRDs deployed prom-adapter takes too long to discover all of them which makes it fail the livenessProbe

Solution: introduce a startupProbe that gives 3 minutes and 30 seconds for prom-adapter to initialize

### DIFF
--- a/assets/prometheus-adapter/deployment.yaml
+++ b/assets/prometheus-adapter/deployment.yaml
@@ -82,6 +82,14 @@ spec:
             drop:
             - ALL
           readOnlyRootFilesystem: true
+        startupProbe:
+          failureThreshold: 18
+          httpGet:
+            path: /livez
+            port: https
+            scheme: HTTPS
+          initialDelaySeconds: 30
+          periodSeconds: 10
         terminationMessagePolicy: FallbackToLogsOnError
         volumeMounts:
         - mountPath: /tmp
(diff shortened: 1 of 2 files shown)


assets/prometheus-adapter/deployment.yaml
```diff
@@ -82,6 +82,14 @@ spec:
             drop:
             - ALL
           readOnlyRootFilesystem: true
+        startupProbe:
+          failureThreshold: 18
+          httpGet:
+            path: /livez
```
raptorsun, line 88: Is it better to use `/readyz`? 
If a pod is fully started, we may consider it ready to serve traffic. In most cases `/livez` and `/readyz` return 200 OK within the same second though.

JoaoBraveCoding: I used `/livez` because according to the k8s docs once a pod passes the startupProbe then the livenessProbe takes over, hence I used the same endpoint as the livenessProbe.

JoaoBraveCoding:
Closing in favor of https://github.com/prometheus-operator/kube-prometheus/pull/2029

PR #1905, opened 2023-02-28
OCPBUGS-7282: Node Exporter ignores network interface under name "cali[a-f0-9]*"

raptorsun:
* [x] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

This PR resolves https://issues.redhat.com/browse/OCPBUGS-7282 .
The issue is similar to https://issues.redhat.com/browse/OCPBUGS-2729 .

PR #1909, opened 2023-03-02
OCPBUGS-8215: bugfix in Node Exporter argument setting

raptorsun:

This is the fix to the bug found by @juzhao  in https://github.com/openshift/cluster-monitoring-operator/pull/1905#issuecomment-1451186414

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -2209,7 +2209,8 @@ func (f *Factory) setTLSSecurityConfiguration(args []string, tlsCipherSuitesArg 
 func setArg(args []string, argName string, argValue string) []string {
 	found := false
 	for i, arg := range args {
-		if strings.HasPrefix(arg, argName) {
+		if arg == argName ||
+			(argName[len(argName)-1] == '=' && strings.HasPrefix(arg, argName)) {
 			args[i] = argName + argValue
 			found = true
 		}
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -4151,6 +4151,79 @@ enableUserWorkload: true
 	}
 }
 
+func TestSetArg(t *testing.T) {
+	tests := []struct {
+		name         string
+		args         []string
+		expectedArgs []string
+		argName      string
+		argValue     string
+	}{
+		{
+			name: "shared prefix",
+			args: []string{
+				"--collector.netclass",
+				"--collector.netclass.ignored-devices=^(veth.*|[a-f0-9]{15}|enP.*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext|br[0-9]*|tun[0-9]*|cali[a-f0-9]*)$",
+			},
+			expectedArgs: []string{
+				"--collector.netclass",
+				"--collector.netclass.ignored-devices=^(veth.*|[a-f0-9]{15}|enP.*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext|br[0-9]*|tun[0-9]*|cali[a-f0-9]*)$",
+			},
+			argName:  "--collector.netclass",
+			argValue: "",
+		},
+		{
+			name: "modify value",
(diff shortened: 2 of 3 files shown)


simonpasquier (commented):
/lgtm

simonpasquier (commented):
This breaks the unit tests.

raptorsun:
> This breaks the unit tests.

Unit test are fixed now :)

juzhao:
issue is fixed, details see from bug comment
/label qe-approved

raptorsun:
> /retest-required

I find out why the test `e2e-agnostic-operator` fails, it misses the metric `node_network_speed_bytes` because the test node does not have the `ethtool` netlink interface. I have created a new PR to fix this by disabling netlink mode of netclass collector by default. I suggest we use the new PR to fix this bug.
https://github.com/openshift/cluster-monitoring-operator/pull/1911

pkg/manifests/manifests.go
```diff
@@ -2140,7 +2140,8 @@ func (f *Factory) setTLSSecurityConfiguration(args []string, tlsCipherSuitesArg
 func setArg(args []string, argName string, argValue string) []string {
 	found := false
 	for i, arg := range args {
-		if strings.HasPrefix(arg, argName) {
+		if arg == argName ||
+			(len(argName) > 0 && argName[len(argName)-1] == '=' && strings.HasPrefix(arg, argName)) {
```
simonpasquier, line 2144: Is there a case where `argName` can be empty or is it just by precaution?

raptorsun: this is by precaution, preventing a potential out of index error :)

raptorsun:
The PR to fix the failling test `e2e-agnostic-operator` is https://github.com/openshift/cluster-monitoring-operator/pull/1912 . 
That PR#1912 should merge  before this one.

raptorsun:
Node exporter cannot collect the metric `node_network_speed_bytes` in the CI cluster, but in a cluster created by cluster bot the metric `node_network_speed_bytes` can be scraped. Still checking where comes such discrepancy  🤔

raptorsun:
The cluster created by cluster bot on GCP is able to emit `speed` metric in `/host/root/sys/class/net/*/`, 
while the `speed` file of a NIC is not readable in AWS CI clusters. 

Here is what happens when reading the  `speed` file on a node in an AWS CI cluster.
```
drwxr-xr-x. 10 root root    0 Mar 14 11:22 queues
-r--r--r--.  1 root root 4096 Mar 14 11:20 speed
drwxr-xr-x.  2 root root    0 Mar 14 11:22 statistics
lrwxrwxrwx.  1 root root    0 Mar 14 11:20 subsystem -> ../../../../../class/net
-r--r--r--.  1 root root 4096 Mar 14 11:23 testing
-rw-r--r--.  1 root root 4096 Mar 14 11:23 threaded
-rw-r--r--.  1 root root 4096 Mar 14 11:23 tx_queue_len
-r--r--r--.  1 root root 4096 Mar 14 11:20 type
-rw-r--r--.  1 root root 4096 Mar 14 11:20 uevent
lrwxrwxrwx.  1 root root    0 Mar 14 11:22 upper_ovs-system -> ../../../../virtual/net/ovs-system
sh-4.4$ pwd
/host/root/sys/class/net/ens5
sh-4.4$ cat speed
cat: speed: Invalid argument
```

simonpasquier:
@raptorsun this needs a backport on `release-4.13`

pkg/manifests/manifests.go
simonpasquier: IIUC an empty `argName` can't be triggered by user input (e.g. via CMO configmap) so I think that it's better to fail. If we hit this situation, it means that we have introduced a bug in the CMO logic.

```suggestion
			(argName[len(argName)-1] == '=' && strings.HasPrefix(arg, argName)) {
```

simonpasquier (commented):
/lgtm

PR #1910, opened 2023-03-02
OCPBUGS-12903: Add new web console usage metrics

jerolimov:

Fixes: [ODC-7258](https://issues.redhat.com/browse/ODC-7258)

This PR adds new console metrics that are collected in [ODC-7232](https://issues.redhat.com/browse/ODC-7232) and should make them later available in Superset DataHat or Tableau.

See also PR https://github.com/openshift/console/pull/12527 for detailed information about the collected metrics.

**Update:** PR https://github.com/openshift/console/pull/12684 reduced the metrics cardinality based on the feedback in this PR here.

A quick overview:

0. GaugeVec `cluster_version_capability` with label `name` is already part of telemetry, and could be used in internal Superset or Tableau, or :question:
1. Sum Counter `console_auth_login_requests_total`
2. Sum CounterVec `console_auth_login_successes_total` by label `role`
3. Sum CounterVec `console_auth_login_failures_total` by label `reason`
4. Sum CounterVec `console_auth_logout_requests_total` by label `reason`
5. ~~Sum CounterVec `console_usage_total` by labels `event` and `perspective`~~
   :warning: **Update:** Removed this metric completely from this PR.
7. Max GaugeVec `console_usage_users` by label `role`
   **Update:** with PR https://github.com/openshift/console/pull/12684 only used roles was reported.
9. Max GaugeVec `console_plugins_info` by labels `name` and `state`
   :warning: **Update:** with PR https://github.com/openshift/console/pull/12684 the plugins are grouped by their name to reduce cardinality.
   Name could be just `"redhat"`, `"demo"` and `"other"` at the moment
10. Max GaugeVec `console_customization_perspectives_info` by labels `name` and `state`
   :warning: **Update:** with PR https://github.com/openshift/console/pull/12684 the perspectives are grouped as well.
   Name could be just `"admin"`, `"dev"`, `"acm"` and `"other"`

An example of what the console `/metrics` endpoint provides:

```
# 1
# HELP console_auth_login_requests_total Total number of login requests from the frontend.
# TYPE console_auth_login_requests_total counter
console_auth_login_requests_total 27

# 2
# HELP console_auth_login_successes_total Total number of successful logins. Role label is based on RBAC can list namespaces check.
# TYPE console_auth_login_successes_total counter
console_auth_login_successes_total{role="cluster-admin"} 3
console_auth_login_successes_total{role="developer"} 23
console_auth_login_successes_total{role="kubeadmin"} 1

# 3
# HELP console_auth_login_failures_total Total number of login failures.
# TYPE console_auth_login_failures_total counter
console_auth_login_failures_total{reason="unknown"} 0

# 4
# HELP console_auth_logout_requests_total Total number of logout requests from the frontend.
# TYPE console_auth_logout_requests_total counter
console_auth_logout_requests_total{reason="unknown"} 0

# 5 --- NOT PART OF THIS PR ANYMORE --- keep that here to give the code review comments a context
#
# HELP console_usage_total Total number of events like "page_views" (loading index.html without history.push) and "page_impressions".
# TYPE console_usage_total counter
# console_usage_total{event="page_impression",perspective="admin"} 13
# console_usage_total{event="page_impression",perspective="dev"} 10
# console_usage_total{event="page_view",perspective="admin"} 212
# console_usage_total{event="page_view",perspective="dev"} 432

# 6
# HELP console_usage_users The number of console users splitten into roles (cluster-admin, developer, and unknown if a RBAC check fails)
# TYPE console_usage_users gauge
console_usage_users{role="cluster-admin"} 3
console_usage_users{role="developer"} 32
console_usage_users{role="kubeadmin"} 1

# 7
# HELP console_plugins_info List all plugins with their name and state as label. State is currently always enabled. Reports 1 for each plugin (per console pod instance).
# TYPE console_plugins_info gauge
#
# OLD:
# console_plugins_info{name="acm",state="enabled"} 1
# console_plugins_info{name="crane-ui-plugin",state="enabled"} 1
# console_plugins_info{name="logging-view-plugin",state="enabled"} 1
# console_plugins_info{name="mce",state="enabled"} 1
# console_plugins_info{name="my-plugin",state="enabled"} 1
#
# NEW:
console_plugins_info{name="redhat",state="notfound"} 4
console_plugins_info{name="demo",state="notfound"} 1

# 8
# HELP console_customization_perspectives_info List of customized perspectives, for example perspective=dev with state=disabled and 1 as metric.
# TYPE console_customization_perspectives_info gauge
#
# OLD:
# console_customization_perspectives_info{name="dev",state="only-for-developers"} 1
# console_customization_perspectives_info{name="dev1",state="only-for-developers"} 1
# console_customization_perspectives_info{name="dev2",state="only-for-developers"} 1
# console_customization_perspectives_info{name="dev3",state="only-for-developers"} 1
#
# NEW:
console_customization_perspectives_info{name="dev",state="only-for-developers"} 1
console_customization_perspectives_info{name="other",state="only-for-developers"} 3
```


jsonnet/rules.libsonnet
```diff
@@ -611,5 +611,42 @@ function(params) {
...
+          expr: 'max(console_usage_users) by (role)',
+          record: 'cluster:console_usage_users:max',
+        },
+        {
+          expr: 'sum(console_plugins_info) by (name, state)',
+          record: 'cluster:console_plugins_info:sum',
+        },
+        {
+          expr: 'sum(console_customization_perspectives_info) by (name, state)',
+          record: 'cluster:console_customization_perspectives_info:sum',
```
jan--f, line 647: I understand that this is not something for this PR but maybe its possible to change without too much hassle. I would argue to rename the label `name` to `perspective`. Its more descriptive and used in other places as well.

jan--f (commented):
Hey @jerolimov thanks for the comprehensive context and the drive-by typo fixes :smile: 

Can you please add some more information on possible label value sets? We are usually trying to only ingest bounded value sets.

For example `perspective`: I suppose this will be either `admin` or `dev`? `state` will likely one of `enabled`, `disabled` and maybe `failed` or so?
`role`, `reason`, `event`, `name` (plugin name) seem like they have the potential to have unbounded value sets. Is that the case? If so we need a strong argument, why that is needed or we work on the recording rules to make those value sets finite.

jerolimov:
Hey @jan--f, thanks for the review. :smile:

> Can you please add some more information on possible label value sets? We are usually trying to only ingest bounded value sets.

Sure, FYI, I tried to explain it in this [google docs](https://docs.google.com/document/d/1PqbKv_-q2PW8mK3lwGEjpLwO5jdf9TojOchnjUY9YMU/edit#heading=h.ted8499p8t4l) as well.

> For example `perspective`: I suppose this will be either `admin` or `dev`? `state` will likely one of `enabled`, `disabled` and maybe `failed` or so?

**For console_customization_perspectives_info**

`name`s are in theory unbound, that's correct, but we don't expect much more than `admin`, `dev`, `acm` (or multicluster or something similar), and maybe one or two more in the future. For now, I wouldn't expect more than 1-3 here (the reality can always surprise you...) and for PM it's important to know which perspectives are configured.

Whenever this escalates we are fine to reduce this to `admin`, `dev`, `other`  or so. Just to understand this: Would you notice this and would you inform us to change this if needed?

In the `state`, we report if a customer has "customized" perspective, which means the customer disabled a perspective them completely, or disabled/enabled it only for a specific user group. The user groups are already reduced to these 5 cases:

* `enabled`
* `disabled`
* `only-for-cluster-admins`
* `only-for-developers`
* `custom-permissions` (rules are RbAC rules, and we use this as fallback for everything we don't know)

> `role`, `reason`, `event`, `name` (plugin name) seem like they have the potential to have unbounded value sets. Is that the case? If so we need a strong argument, why that is needed or we work on the recording rules to make those value sets finite.

**`role`s:**

* in the login case we use exactly these 3:
   * `kubeadmin`
   * `cluster-admin`
   * `developer`
* in the users metric we use also `unknown` if an internal error occurs and we can't assign one of the 3 roles.

**`reason`s**

The reason for failures is currently always `unknown`.

Internally we have some more errors which we might want to use. But we (I) decided to just use `unknown`  to have such a metric bound here.

If needed we could add reasons like login `canceled` or logout `timeout or `manually` in the future. But we have no plan for this at the moment.

**`event`s** we use currently exactly three events that tracks the usage of the console:

* `page_view`
* `page_impression`
* `perspective_changed`

There aren't other events planned at the moment.

**`console_plugins_info`**:

This is maybe more unbound than perspectives, as the console supports "dynamic plugins" and other teams are invited to extend the console.

Currently, I would expect values like `demo`, `acm`, `gitops`, `pipelines`, and maybe a few more. But this could evolve over time without our control. I believe different PMs are really interested to see the adoption rate of the plugin infrastructure and the plugins themself.

If this escalates I would expect we need to filter them or track just "redhat" vs "marketplace" vs "customer" plugins. But I don't think we have that distinction yet.

CHANGELOG.md
```diff
@@ -11,11 +11,12 @@
 - [#1882](https://github.com/openshift/cluster-monitoring-operator/issues/1882) Allow configuring secrets in alertmanager component (platform)
 - [#1876](https://github.com/openshift/cluster-monitoring-operator/pull/1876) Add nodeExporter.collectors.tcpstat settings.
 - [#1888](https://github.com/openshift/cluster-monitoring-operator/pull/1888) Add nodeExporter.collectors.netdev settings.
-- [#1884](https://github.com/openshift/cluster-monitoring-operator/issues/1882) Allow configuring secrets in alertmanager component (UWM)
+- [#1884](https://github.com/openshift/cluster-monitoring-operator/issues/1884) Allow configuring secrets in alertmanager component (UWM)
```
raptorsun, line 14: This should be in the 4.14 version now.

jerolimov: I moved my PR to 4.14, but I kept this specific line as it was just a drive-by fix. Just the link wasn't correct.

jerolimov:
> Oh and in the PR description you have a small error:
> ..
> That second label is probably called `state`?

Oh yeah, copy and paste error. Fixed that. :smirk:

jerolimov:
> Ok so the anticipated label value sets have me a bit worried or rather the open-endedness. We don't currently check for cardinality increases in telemeter. Even if we did, at that stage it would be too late :)
> 
> So for this PR, let's change the recording rule in a way so that we really only collect the label values that we need to have in telemetry. For example: For rule `cluster:console_customization_perspectives_info:sum` you currently expect possible values to be one of `admin, dev,acm`. So lets reflect that in the recording rule like so:
> 
> ```
>  expr: 'sum(console_customization_perspectives_info{name=~"admin|dev|acm") by (name, state)',
>  record: 'cluster:console_customization_perspectives_info:sum',
> ```

Thanks for that recommendation.

Just FYI: I checked our anonymized analytics where we have an event when a user switches perspective. We have currently **_only_** users switching between "admin" and "dev".

So this shows that other perspectives aren't really used today (on the clusters we monitoring with frontend analytics). So I'm not worried about too many time series, but I understand that you are. :smile:

Is there maybe to group all other perspectives' as "other" or something instead of just dropping them? Does this make sense to you? When there is no easy way I can apply the filter to `admin|dev` for now.

> This allows us to protect against future label value proliferation, to reason about the number of time-series this PR adds and it forces you to actually reason about what data you need.
> 
> Also fair warning: This looks like it'll add quite a few time-series, so we'll like have to get @eparis's thumbs up. No big deal, just another admin step. grinning

Sure, that's fine.

eparis:
From a strategic PoV I approve this PR given 2 things.
1. I do not support `console_usage_total`   I am willing to have a separate conversation about that metric.
2. I would like to see the cardinality discussion closed: https://github.com/openshift/cluster-monitoring-operator/pull/1910#issuecomment-1472353514

jerolimov:
Hey @jan--f, hey @eparis,

### On perspectives / plugins names

I opened a PR on our side (https://github.com/openshift/console/pull/12684) that reduced the cardinality already in our backend.

1. For `console_customization_perspectives_info` it just returns `"admin"`, `"dev"`, `"acm"`. All other perspectives would be grouped as `"other"`.
2. Same for `console_plugins_info`, I followed the recommendation and grouped them by "vendor" (also if we don't have a marketplace or something yet). For now, I grouped the well-known plugin names as `"redhat"`, `"demo"`, and any unknown plugin will be grouped as `"other"`.

In theory, with the console PR, we don't need a filter here. But to get this PR into 4.14 (and 4.13) I added the recommended filters to this PR here anyway. So that it can get merged independently from our change!

See `rules.libsonnet`:

https://github.com/jerolimov/cluster-monitoring-operator/blob/b35870dd9949d76ab23f45f3c62b5dafdc620f37/jsonnet/rules.libsonnet#L637-L644

### Usage metric (page views)

On our side, we still have the metric `console_usage_total` that contains the number of page views and page impressions.

I agree that this information should be also accessible via other frontend analytics/telemetry tools like Segment/Amplitude. But its also true that we don't have frontend analytics on all clusters available.

Anyway, to get this in let us continue that in a separate thread or PR, I removed this metric completely from this PR.

Can you take a new look if this PR is fine now? Thanks

jerolimov:
/cc @raptorsun @jan--f @simonpasquier 

Should I include the changes in `jsonnet/versions.yaml` that I can generate with `./hack/generate-versions.sh` to fix the versions job or should you or me create another PR just for that change??

The other CI job looks flaky, is it fine to just rerun them?

/retest

jsonnet/rules.libsonnet
```diff
@@ -615,5 +615,38 @@ function(params) {
         },
       ],
     },
+    {
+      name: 'openshift-console.rules',
```
simonpasquier, line 619: I'd rather see the rules live in openshift/console-operator if this is possible.

jerolimov: @simonpasquier does this mean that I should just change the label or move the metrics code from openshift/console to openshift/console-operator?

simonpasquier: I meant to say that the console-operator should deploy a PrometheusRule containing the recording rule. Since the metric is owned/managed by the console team, it's more convenient if you also own the rule.
Does it make sense?

jerolimov: Yep, that makes sense to me. I opened https://github.com/openshift/console-operator/pull/755 for this. PTAL if this PR is then fine. Thanks.

Documentation/data-collection.md
```diff
@@ -417,6 +417,48 @@ data:
...
+    # cluster:console_plugins_info:max list the web console plugins grouped by a vendor name and their installation state.
+    # The plugin name is currently mapped into the groups redhat, demo and other.
+    #
+    - '{__name__="cluster:console_plugins_info:max"}'
+    #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_customization_perspectives_info:max list customized web console perspectives.
+    # The perspective id is mapped into the groups admin, dev, acm and other.
+    # The state is a reduced value based on the RBAC rules that are required to have access to perspective:
+    # For example: enabled, disabled, only-for-cluster-admins, only-for-developers, custom-permissions
```
simonpasquier, line 458:
```suggestion
    # cluster:console_customization_perspectives_info:max reports information about customized web console perspectives.
    # Labels:
    # * `name` , one of `admin`, `dev`, `acm` or `other`.
    # `* state` , one of ???
```

Documentation/data-collection.md
```diff
@@ -417,6 +417,48 @@ data:
...
     - '{__name__="console_url"}'
     #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_auth_login_requests_total:sum gives the total number of login requests initiated from the web console.
+    #
+    - '{__name__="cluster:console_auth_login_requests_total:sum"}'
+    #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_auth_login_successes_total:sum gives the total number of successful logins initiated from the web console.
+    # Role is kubeadmin, cluster-admin or developer based on a permission check if ther user can list all namespaces.
```
simonpasquier, line 427:
```suggestion
    # Labels:
    # * `role`, one of `kubeadmin`, `cluster-admin` or `developer` . The value is based whether or not the logged-in user can all namespaces.
```

Documentation/data-collection.md
```diff
@@ -417,6 +417,48 @@ data:
...
+    #
+    - '{__name__="cluster:console_auth_login_successes_total:sum"}'
+    #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_auth_login_failures_total:sum gives the total number of login failures initiated from the web console.
+    # This might include canceled OAuth logins depenending on the user OAuth provider/configuration.
```
simonpasquier, line 433:
```suggestion
    # This might include canceled OAuth logins depenending on the user OAuth provider/configuration.
    # Labels:
    # * `reason`, one of ...
```

Documentation/data-collection.md
```diff
@@ -417,6 +417,48 @@ data:
...
+    #
+    - '{__name__="cluster:console_auth_login_failures_total:sum"}'
+    #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_auth_logout_requests_total:sum gives the total number of logout requests send from the web console.
```
simonpasquier, line 438:
```suggestion
    # cluster:console_auth_logout_requests_total:sum gives the total number of logout requests send from the web console.
    # Labels:
    # * `reason`, one of ...
```

Documentation/data-collection.md
```diff
@@ -417,6 +417,48 @@ data:
...
+    #
+    - '{__name__="cluster:console_auth_logout_requests_total:sum"}'
+    #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_usage_users:max contains the number of web console users splitten into the roles kubeadmin,
+    # cluster-admin and developer depending if the user can list all namespaces.
```
simonpasquier, line 444: same here

Documentation/data-collection.md
```diff
@@ -417,6 +417,48 @@ data:
...
+    #
+    - '{__name__="cluster:console_usage_users:max"}'
+    #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_plugins_info:max list the web console plugins grouped by a vendor name and their installation state.
+    # The plugin name is currently mapped into the groups redhat, demo and other.
```
simonpasquier, line 450: same here

simonpasquier (commented):
Sorry for the late reply :-(

For future readers and maintainers, it would be good to list explicitly which labels are present for each recorded metric and what are the expected values for each. This will avoid going back to the console and console-operator code base every time.

jerolimov:
I addressed your comments by [adding this label change](https://github.com/openshift/cluster-monitoring-operator/compare/51bf44f728f895d32c3331efce63f6f82c827faf..7340ded2ef5d1313bee3eb5d64ddc960ad1b3070).

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -409,6 +409,58 @@ data:
...
+    #
+    - '{__name__="cluster:console_plugins_info:max"}'
+    #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_customization_perspectives_info:max reports information about customized web console perspectives.
+    # Labels:
+    # * `name`, one of `admin`, `dev`, `acm` or `other`
+    # * `state`, one of `enabled`, `disabled`, `only-for-cluster-admins`, `only-for-developers` or `custom-permissions`
+    # 
+    # For example: enabled, disabled, only-for-cluster-admins, only-for-developers, custom-permissions
```
simonpasquier, line 460: I guess this sentence can be removed.
```suggestion
```

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -409,6 +409,58 @@ data:
...
+    #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_usage_users:max contains the number of web console users splitten into the roles.
+    # Labels:
+    # * `role`: `kubeadmin`, `cluster-admin` or `developer`. The value is based whether or not the user can list all namespaces.
+    #
+    - '{__name__="cluster:console_usage_users:max"}'
+    #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_plugins_info:max reports information about the web console plugins and their state.
```
simonpasquier, line 447: to be sure: we can have all label combinations for the same cluster? E.g. we can have

```
cluster:console_plugins_info:max{name="redhat",state="enabled"} 1
cluster:console_plugins_info:max{name="redhat",state="disabled"} 2
cluster:console_plugins_info:max{name="redhat",state="notfound"} 3
cluster:console_plugins_info:max{name="demo",state="enabled"} 4
cluster:console_plugins_info:max{name="demo",state="disabled"} 5
cluster:console_plugins_info:max{name="demo",state="notfound"} 6
cluster:console_plugins_info:max{name="other",state="enabled"} 7
cluster:console_plugins_info:max{name="other",state="disabled"} 8
cluster:console_plugins_info:max{name="other",state="notfound"} 9
```

jerolimov: Just in theory yes

But this looks like a misconfiguration and is nothing we expect. Most clusters will have no plugin, or some from `redhat` or `other`, that are `enabled` or `disabled`.

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -409,6 +409,58 @@ data:
...
+    # Labels:
+    # * `name`: `redhat`, `demo` or `other`.
+    # * `state`: `enabled`, `disabled` or `notfound`
+    #
+    - '{__name__="cluster:console_plugins_info:max"}'
+    #
+    # owners: (@openshift/hybrid-application-console-maintainers)
+    # cluster:console_customization_perspectives_info:max reports information about customized web console perspectives.
```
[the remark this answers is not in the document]
jerolimov: In the absolute worst case, yes.

It would surprise me to have more than a few clusters using this at all.

The most common case (on these few clusters) will be `admin` and `dev` will be `disabled`, `only-for-cluster-admins` or `only-for-developers`.

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
jerolimov: Removed.

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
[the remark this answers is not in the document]
simonpasquier: So you're betting on a cardinality of 6?

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
simonpasquier: Ok so cardinality of 9 is the worst case but very unlikely. For most clusters, it should something between 0 and 4.

jerolimov: Correct, I expect that most clusters will have `redhat` or `other` plugins in state `enabled` or `disabled`. So 0-4 is a reasonable assumption here.

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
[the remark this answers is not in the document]
jerolimov: Correct, I think 4 to 6 is a reasonable assumption here for the clusters that use this feature.

The console returns just one state at each time, but the cluster admin can switch (change the config) this from `disabled` to `only-for-cluster-admins` or `only-for-developers` and back.

simonpasquier (commented):
|Metric|Cardinality|
|---|---|
|cluster:console_auth_login_requests_total:sum|1|
|cluster:console_auth_login_successes_total:sum	|3|
|cluster:console_auth_login_failures_total:sum |1|
|cluster:console_auth_logout_requests_total:sum |1|
|cluster:console_usage_users:max| 3 |
|cluster:console_plugins_info:max| worst case is 9 but it should be between 0 and 4 in practice |
|cluster:console_customization_perspectives_info:max | 4 at most but can change over time |
|Total|Minimum is 9, Maximum is between 17 and 22|

/lgtm
/hold

@jan--f @eparis see the recap above about how many Telemetry series are added there. We're a bit above the usual numbers but according to @jerolimov customer clusters should send no more than 17 series and in practice we should rather around 10-ish (the worst scenario of 22 series assumes many installed plugins in all possible states).

simonpasquier:
/hold cancel

The metrics have been approved by @eparis

jerolimov:
@jan--f @simonpasquier  Looks like the [e2e-agnostic-operator](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-operator) job is failing because of this assertion error:

https://search.ci.openshift.org/?search=prometheus_remote_storage_samples_pending&maxAge=168h&context=1&type=build-log&name=pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-operator&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

Do you take a look? :smirk:

simonpasquier (commented):
/lgtm

PR #1894, opened 2023-02-15
MON-2895: toggle netlink implementation of netclass collector

raptorsun:
* [x] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

This PR follows https://github.com/openshift/cluster-monitoring-operator/pull/1893 . It should be merged after that one.

PR #1911, opened 2023-03-03
OCPBUGS-8215: bugfix in Node Exporter argument setting + turn off netlink mode of netclass collector

raptorsun:

This PR has one additional commit than https://github.com/openshift/cluster-monitoring-operator/pull/1909
The test from that PR fails because the metric `node_network_speed_bytes` is missing.
This is due to lack of `ethtool` netlink interface on the host. 
To be safe, we disable the netlink mode of netclass collector by default, so that we can be sure to have the metric  `node_network_speed_bytes` on less equipped nodes.


pkg/manifests/config.go
```diff
@@ -209,7 +209,7 @@ func defaultClusterMonitoringConfiguration() ClusterMonitoringConfiguration {
 				},
 				NetClass: NodeExporterCollectorNetClassConfig{
 					Enabled:    true,
-					UseNetlink: true,
+					UseNetlink: false,
```
simonpasquier, line 212: hmm I missed that we enabled the netlink driver by default in #1894. To avoid any regression/change, I agree that it's better to keep the default as it was before (e.g. using sysfs).

simonpasquier (commented):
IIUC the issue was missed in the initial because the e2e tests didn't check for the metrics?

raptorsun:
> IIUC the issue was missed in the initial because the e2e tests didn't check for the metrics?

e2e test checked the metrics, but the setArg function has a problem to overwrite arguments with the same prefix, so `--collector.netclass` overwrites `--collector.netclass.netlink` at runtime, it still uses the old sysfs implementation, which brings the metric `node_network_speed_bytes`.

raptorsun:
This PR is closed.
We seperate the 2 fixes in 2 PRs:
- https://github.com/openshift/cluster-monitoring-operator/pull/1909
- https://github.com/openshift/cluster-monitoring-operator/pull/1912

The order of merge is 
1. https://github.com/openshift/cluster-monitoring-operator/pull/1912
2. https://github.com/openshift/cluster-monitoring-operator/pull/1909

PR #1912, opened 2023-03-03
OCPBUGS-8282: turn off netlink mode of netclass collector in Node Exporter.

raptorsun:

The test from [the PR fixing setArg](https://github.com/openshift/cluster-monitoring-operator/pull/1909) fails because the metric node_network_speed_bytes is missing.
This is due to lack of `ethtool` netlink interface in the base image.
To be safe, we disable the netlink mode of netclass collector by default, so that we can be sure to have the metric node_network_speed_bytes with default settings.


pkg/manifests/types.go
```diff
@@ -368,6 +368,9 @@ type NodeExporterCollectorNetClassConfig struct {
 	// `node_network_address_assign_type`,
 	// `node_network_name_assign_type`,
 	// `node_network_device_id`.
+	// Due to lack of `ethtool` interface in the base image, these metrics are missing, too:
```
simonpasquier, line 371: I wouldn't go into the details of why the metric is missing.

pkg/manifests/types.go
```diff
@@ -368,6 +368,9 @@ type NodeExporterCollectorNetClassConfig struct {
...
+	// `node_network_speed_bytes`,
+	// the `duplex` label in the metric `node_network_info`.
```
simonpasquier, line 373: same here

pkg/manifests/types.go
```diff
@@ -367,7 +367,9 @@ type NodeExporterCollectorNetClassConfig struct {
 	// This implementation improves the performance of the `netclass` collector by omitting these metrics:
 	// `node_network_address_assign_type`,
 	// `node_network_name_assign_type`,
-	// `node_network_device_id`.
+	// `node_network_device_id`,
+	// `node_network_speed_bytes`,
+	// the `duplex` label in the metric `node_network_info`.
```
simonpasquier, line 372:
```suggestion
	// `node_network_speed_bytes`.
	// In addition, The `node_network_info` metric lacks the `duplex` label.
```

simonpasquier (commented):
/lgtm

PR #1917, opened 2023-03-15
OCPBUGS-7694: add startup probe for prometheus-adapter

simonpasquier:
The startup probe prevents that kubelet kills the prometheus-adapter pod on clusters with many CRDs and/or slow API because it might take some time for prometheus-adapter to discover the available API resources and the /livez endpoint only becomes available once this discovery phase is completed.

The startup probe will wait 3m30s before failing which should be enough even in most unfavorable cases.


simonpasquier:
/retitle OCPBUGS-7694: add startup probe for prometheus-adapter

PR #1937, opened 2023-04-05
OCPBUGS-11434: node-exporter: disable btrfs collector

vrutkovs:
[Profiling](https://pprof.me/9802ad7/) shows that node-exporter is using ~16% of CPU time on updating btrfs details. This FS is not included in RHEL kernels, so it can be safely disabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Note: This CHANGELOG is only for the monitoring team to track all monitoring related changes. Please see OpenShift release notes for official changes.
 
+## 4.14
+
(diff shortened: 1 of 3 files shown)


simonpasquier:
Thanks for the PR! I think that we want to address it upstream first https://github.com/prometheus-operator/kube-prometheus/issues/2063

vrutkovs:
I think its orthogonal to upstream's "btrfs collector takes too much CPU" - node-exporter running openshift will *never* report any btrfs metrics, so its safe to disable it now (and backport to other releases)

simonpasquier (commented):
/lgtm

PR #1926, opened 2023-03-28
OCPBUGS-10387: label for infra nodes for metric cluster:capacity_cpu_cores:sum

raptorsun:
As stated in [OCPBUGS-10387](https://issues.redhat.com/browse/OCPBUGS-10387), certain clusters do not label their infra nodes in the metric `cluster:capacity_cpu_cores:sum`, due to lack of label `label_node_role_kubernetes_io="infra"` in the metric `kube_node_labels`.
This PR add logic to automatically add a label  `label_node_role_kubernetes_io="infra"` to infra nodes when generating the metric `metric cluster:capacity_cpu_cores:sum`.

This PR is based on the PR https://github.com/openshift/cluster-monitoring-operator/pull/1926 that adds unit tests for the current recording rule of  `cluster:capacity_cpu_cores:sum`.

* [x] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #1939, opened 2023-04-06
add unit test for metric "cluster:capacity_cpu_cores:sum"

raptorsun:

### DIFF
--- a/test/rules/cluster_capacity_cpu.yaml
+++ b/test/rules/cluster_capacity_cpu.yaml
@@ -1,0 +1,165 @@
+# Tests for cluster:capacity_cpu_cores:sum
+# Verifying basic logic and infra nodes are correctly labeled.
+
+rule_files:
+  - rules.yaml
+
+evaluation_interval: 30s
+
+tests:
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
master nodes also have the `control-plane` role, it would be good to test that the output is correct in this case.

raptorsun:
> master nodes also have the `control-plane` role, it would be good to test that the output is correct in this case.

Thanks for the review :)

The tests about `cluster:master_nodes` and `cluster:infra_nodes` are moved into a separate file. 
`kube_node_role(role="control-plane")` is added to master nodes in the test cases.

simonpasquier:
IIUC this is now obsoleted by #1926? If yes, feel free to close this PR.

raptorsun:
> IIUC this is now obsoleted by #1926? If yes, feel free to close this PR.

Yes. These tests will be obsolete when we merge #1926 This is just for comparing the behavior before and after we modify the recording rule of `cluster:capacity_cpu_cores:sum`.
/close

PR #1942, opened 2023-04-10
[release-4.10]  OCPBUGS-11579: jsonnet: Add prometheus container in UWM prometheus asset file

slashpai:
openshift#1824 increased the
startupProbe timeout from 15m to 1h for both platform and uwm proemetheus. But it didn't take effect on UWM prometheus because in assets file it didn't have a container named `prometheus` on UWM prometheus pod so CMO never entered the code path that was added in openshift#1824
since it uses strategic merge patch.

Hence this commit adds a `prometheus` container in uwm prometheus jsonnet file so it gets added
in assets file.

Fixes #OCPBUGS-10690

### DIFF
--- a/jsonnet/components/prometheus-user-workload.libsonnet
+++ b/jsonnet/components/prometheus-user-workload.libsonnet
@@ -464,6 +464,9 @@ function(params)
               },
             },
           },
+          {
+            name: 'prometheus',
+          },
         ],
       },
     },
--- a/assets/prometheus-user-workload/prometheus.yaml
+++ b/assets/prometheus-user-workload/prometheus.yaml
@@ -147,6 +147,7 @@ spec:
       requests:
         cpu: 1m
         memory: 10Mi
+  - name: prometheus
   enableFeatures: []
   enforcedNamespaceLabel: namespace
   externalLabels: {}


slashpai:
@juzhao @jan--f

jsonnet/components/prometheus-user-workload.libsonnet
```diff
@@ -464,6 +464,9 @@ function(params)
               },
             },
           },
+          {
+            name: 'prometheus',
+          },
```
sthaha, line 469: I think we should add a comment that explains why an empty container with name `prometheus` is needed.

slashpai: We missed the comment on 4.14 branch . Should we add it from there?

sthaha: I think a comment would be nice since an empty container with `prometheus` doesn't tell the reader why this is important. I.E. why this is required for the startup probes to be setup properly.

slashpai: Added comment on source file https://github.com/openshift/cluster-monitoring-operator/pull/1943

PR #1943, opened 2023-04-11
jsonnet: add comment why empty prometheus container needed

slashpai:


slashpai:
This is just adding a comment on code and no code changes

/label docs-approved
/label px-approved
/label qe-approved

PR #1944, opened 2023-04-11
[release-4.12] OCPBUGS-11623: node-exporter: disable btrfs collector

vrutkovs:
[Profiling](https://pprof.me/9802ad7/) shows that node-exporter is using ~16% of CPU time on updating btrfs details. This FS is not included in RHEL kernels, so it can be safely disabled

Cherry-pick of https://github.com/openshift/cluster-monitoring-operator/pull/1941 on release-4.12


jsonnet/components/node-exporter.libsonnet
```diff
@@ -181,6 +181,8 @@ function(params)
                               // https://issues.redhat.com/browse/OCPBUGS-1321
                               '--collector.netclass.ignored-devices=^(veth.*|[a-f0-9]{15}|tun[0-9]*|br[0-9]*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext)$',
                               '--path.udev.data=/host/root/run/udev/data',
+                              // Disable btrfs collector as btrfs is not included in RHEL kernels
+                              '--no-collector.btrfs'
```
jan--f, line 185:
```suggestion
                              '--no-collector.btrfs',
```
This should fix the `fmt-jsonnet` check.

PR #1946, opened 2023-04-11
MON-3146: Addd build number to vsphere vcenter information

gnufied:
Add build number to vsphere vcenter information

xref https://issues.redhat.com/browse/MON-3146?filter=-2


sdodson:
/retitle OCPBUGS-13066: Add build number to vsphere vcenter information
This is a bug because without this build detail we cannot decide if it's safe to upgrade to 4.13 where vSphere CSI migration is available.

sdodson:
/retitle OCPBUGS-13006: Add build number to vsphere vcenter information

PR #1950, opened 2023-04-20
OCPBUGS-11889: disable CORS headers on Thanos querier

machine424:

* [x] Change commit message + changelog entry
* [ ] Backport to 4.12/4.13?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - [#1937](https://github.com/openshift/cluster-monitoring-operator/pull/1937) Disables btrfs collector
 - [#1910](https://github.com/openshift/cluster-monitoring-operator/pull/1910) Add new web console usage metrics
+- [#1950](https://github.com/openshift/cluster-monitoring-operator/pull/1950) Disable CORS headers on Thanos querier by defaut and add a flag to enable them back.
 
 ## 4.13
 
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -3277,7 +3277,8 @@ func TestThanosQuerierConfiguration(t *testing.T) {
       cpu: 3m
       memory: 4Mi
   logLevel: debug
-  enableRequestLogging: true`, false)
+  enableRequestLogging: true
+  enableCORS: true`, false)
 
 	if err != nil {
 		t.Fatal(err)
@@ -3397,6 +3398,19 @@ grpc:
 				}
 				if got.value != expectResult {
 					t.Fatalf("unexpected flag value for Thanos query, wanted %s but got %s", expectResult, got.value)
+				}
+			}
+
+			{
+				// test CORS headers flag
+				const (
+					disableCORSFlag = "--web.disable-cors"
+				)
+
+				_, ok := getArgValue(c, disableCORSFlag)
+				if ok {
+					// For now this "index out of range" prior to this if the flag is set
+					t.Fatalf("CORS headers should be enabled on Thanos query")
 				}
 			}
 
--- a/pkg/manifests/types.go
+++ b/pkg/manifests/types.go
@@ -258,6 +258,10 @@ type ThanosQuerierConfig struct {
 	// The possible values are `error`, `warn`, `info`, and `debug`.
 	// The default value is `info`.
 	LogLevel string `json:"logLevel,omitempty"`
+	// A Boolean flag that enables setting CORS headers.
+	// The headers would allow access from any origin.
+	// The default value is `false`.
+	EnableCORS bool `json:"enableCORS,omitempty"`
 	// Defines the nodes on which the pods are scheduled.
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 	// Defines resource requests and limits for the Thanos Querier container.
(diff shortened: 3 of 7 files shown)


pkg/manifests/manifests_test.go
```diff
@@ -3400,6 +3401,19 @@ grpc:
...
 
+			{
+				// test CORS headers flag
+				const (
+					disableCORSFlag = "--web.disable-cors"
+				)
+
+				_, ok := getArgValue(c, disableCORSFlag)
+				if ok {
+					// For now this "index out of range" prior to this if the flag is set
```
machine424, line 3412: I'll remove this

pkg/manifests/types.go
```diff
@@ -258,6 +258,9 @@ type ThanosQuerierConfig struct {
 	// The possible values are `error`, `warn`, `info`, and `debug`.
 	// The default value is `info`.
 	LogLevel string `json:"logLevel,omitempty"`
+	// A Boolean flag that enables setting CORS headers.
+	// The default value is `false`.
+	EnableCORS bool `json:"enableCORS,omitempty"`
```
sthaha, line 263: Just a thought: 
should we use `*bool` so that we can determine (in future) without changing the API if the value has been explicitly set by user?

sthaha: Do note it is only a `suggestion` , feel free to ignore :) 

> Is there a convention? 

I don't think there is one. In this specific case, I think we may want to change the default behaviour later (i.e. if user hasn't explicitly set this field) and making it a pointer (like k8s api types do) helps with that.

> Do we need to apply the default/get rid of nil like in here 

I would says so ...  k8s pointer package provides utils for this. 

```
import "k8s.io/utils/pointer"

pointer.BoolPtr(false)
```

machine424: > I don't think there is one. In this specific case, I think we may want to change the default behaviour later (i.e. if user hasn't explicitly set this field) and making it a pointer (like k8s api types do) helps with that.

My bad, I meant a convention for "applying the default/getting rid of nil like in the example". ;)

Thanks, I'll take a look at the package.

pkg/manifests/manifests_test.go
machine424: I'd keep it for guidance.

pkg/manifests/types.go
sthaha: I haven't tested it, so it would be nice to know if it is possible to convert this to pointer later and detect if the field was set 🤔 . Have you tried it already?

machine424: It looks straightforward to me, converting it to a pointer will still allow the user to provide the same values, we'd only need to deal with nil (here we'd know the user didn't provide a value and we could set the value it points to to false to avoid breaking the API), I just think it's not worth all the extra code for now as detecting missing values is not needed for now.

jan--f: +1 If we really need to know if a user set this or not, a pointer is good. But since this effectively passes through a boolean cli flag `*bool` seems unnecessary here.
Or is there a specific use case you have in mind Sunil?

sthaha: > since this effectively passes through a boolean cli flag *bool seems unnecessary here

Based on the test added in this PR, I thought that isn't the case. I.E. this config will  be stored in the configmap.

machine424: Sorry, is one of the added implies that the code detects/should detect missing values?

sthaha: For the record, we resolved this. My concern was with mashalling and unmarshalling which is never the case since we only unmarshall.

pkg/manifests/types.go
```diff
@@ -258,6 +258,9 @@ type ThanosQuerierConfig struct {
 	// The possible values are `error`, `warn`, `info`, and `debug`.
 	// The default value is `info`.
 	LogLevel string `json:"logLevel,omitempty"`
+	// A Boolean flag that enables setting CORS headers.
+	// The default value is `false`.
```
jan--f, line 262:
```suggestion
	// The default value is `false`.
	// In enabled, Thanos sets CORS headers to be allowed by all.
```

jan--f (changes requested):
One last nit, other this looks good.

pkg/manifests/types.go
machine424: I assume you're talking about `Access-Control-Allow-Origin: *`

How about

```
// A Boolean flag that enables setting CORS headers.
// The headers would allow access from any origin.
// The default value is `false`.
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - [#1937](https://github.com/openshift/cluster-monitoring-operator/pull/1937) Disables btrfs collector
 - [#1910](https://github.com/openshift/cluster-monitoring-operator/pull/1910) Add new web console usage metrics
+- [#1950](https://github.com/openshift/cluster-monitoring-operator/pull/1950) Disable CORS headers on Thanos querier by defaut and add a flag to enable them back.
```
simonpasquier, line 7: note for the future: you don't have to update the CHANGELOG.md especially for bug fixes as they are tracked/documented in JIRA.

pkg/manifests/types.go
machine424: I pushed a this version.

CHANGELOG.md
jan--f: The line is blurry here as this stems from an RFE :)

simonpasquier: well this is linked to an OCPBUGS-XXX ticket :)

PR #1957, opened 2023-05-02
OCPBUGS-12488: go.mod: update golang.org/x/net to v0.7.0

machine424:


machine424:
It's normal that `ci/prow/versions` fails as we don't systematically upgrade components on z versions, I'll create a ticket to improve it.

simonpasquier:
> I'll create a ticket to improve it.

I've just closed a ticket yesterday about it that I filed a while ago. The `versions` job is only informing and I'm not sure that we have time to "fix" the glitch...

danielmellado:
Should we just disable it once we cut a branch then? Either that or just ignoring it for z releases.

marioferh (commented):
/lgtm

PR #1960, opened 2023-05-09
OCPBUGS-13095: Uncomment cluster:vsphere_infrastructure_failure_domains:max

jsafrane:
We do want this metric in telemetry. This looks like an oversight / typo. I will make sure that it's enabled in rhobs/configuration on the server side.


simonpasquier (commented):
/lgtm

PR #1963, opened 2023-05-11
MON-3222: Make netdev/netclass interfaces configurable

jotak:
- Introduce new setting for netdev/netclass: ignoredNetworkDevices
- Keep using the current exclusion list as a default if this isn't explicitely configured
- Add a removeArg function to allow setting an explicit empty list (so that it isn't overriden by the upstream jsonnet value)
- Add tests


jotak:
Opening this PR to validate / discuss the approach, cc @simonpasquier :
- Currently I created 2 different settings for netdev and netclass - we mentioned maybe having just one, I can definitely change that, but 2 settings is more flexible and would avoid the pain of having to deal with a breaking change if it becomes required someday.
- As discussed, I opted for a simple config of just a exclusion list to override or not the default - as opposed to a list + a flag "add/replace"
- Using now the [NETOBSERV-1021](https://issues.redhat.com/browse/NETOBSERV-1021) jira but let me know if you create a MON* jira that needs to be tracked instead.

Documentation/api.md
```diff
@@ -243,6 +243,7 @@ The `NodeExporterCollectorNetClassConfig` resource works as an on/off switch for
 | -------- | ---- | ----------- |
 | enabled | bool | A Boolean flag that enables or disables the `netclass` collector. |
 | useNetlink | bool | A Boolean flag that activates the `netlink` implementation of the `netclass` collector. This implementation improves the performance of the `netclass` collector by omitting these metrics: `node_network_address_assign_type`, `node_network_name_assign_type`, `node_network_device_id`, `node_network_speed_bytes`. In addition, The `node_network_info` metric lacks the `duplex` label. |
+| ignoredNetworkDevices | *[]string | A list of network devices to be excluded from collection, using regular expressions. By default, the following list is used: `veth.*`, `[a-f0-9]{15}`, `enP.*`, `ovn-k8s-mp[0-9]*`, `br-ex`, `br-int`, `br-ext`, `br[0-9]*`, `tun[0-9]*`, `cali[a-f0-9]*` |
```
jotak, line 246: Should I add some kind of warning here, that removing filters from the default list will increase cardinality?

jotak: Especially for filters such as `[a-f0-9]{15}` or `cali[a-f0-9]*` , removal must be considered with care

simonpasquier, line 246: Does it make sense to have 1 exclude list for each collector? I feel that in most (all?) cases, you want to apply the same list to both collectors.

simonpasquier, line 246: I wouldn't document the exact list as it would quickly become out of date. Maybe include some general phrasing that by default, node_exporter will try to collect metrics only from physical network interfaces.

assets/node-exporter/daemonset.yaml
```diff
@@ -36,8 +36,8 @@ spec:
         - --path.udev.data=/host/root/run/udev/data
         - --no-collector.wifi
         - --collector.filesystem.mount-points-exclude=^/(dev|proc|sys|run/k3s/containerd/.+|var/lib/docker/.+|var/lib/kubelet/pods/.+)($|/)
-        - --collector.netclass.ignored-devices=^(veth.*|[a-f0-9]{15}|enP.*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext|br[0-9]*|tun[0-9]*|cali[a-f0-9]*)$
-        - --collector.netdev.device-exclude=^(veth.*|[a-f0-9]{15}|enP.*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext|br[0-9]*|tun[0-9]*|cali[a-f0-9]*)$
+        - --collector.netclass.ignored-devices=^(veth.*|[a-f0-9]{15})$
+        - --collector.netdev.device-exclude=^(veth.*|[a-f0-9]{15})$
```
simonpasquier, line 40: to make it obvious that CMO is going to override the parameters, I'd use a recognizable value like `.*`

```suggestion
        - --collector.netdev.device-exclude=^.*$
```

pkg/manifests/manifests.go
```diff
@@ -847,6 +854,13 @@ func (f *Factory) updateNodeExporterArgs(args []string) []string {
 		if f.config.ClusterMonitoringConfiguration.NodeExporterConfig.Collectors.NetClass.UseNetlink {
 			args = setArg(args, "--collector.netclass.netlink", "")
 		}
+		excludedDev := *f.config.ClusterMonitoringConfiguration.NodeExporterConfig.Collectors.NetClass.IgnoredNetworkDevices
+		if len(excludedDev) > 0 {
+			args = setArg(args, "--collector.netclass.ignored-devices=", regexListToArg(excludedDev))
+		} else {
+			// make sure upstream default isn't used
```
simonpasquier, line 861: To keep it simple, I'd configure node_exporter with `--collector.netclass.ignored-devices=""` if the list is empty (it removes 1 if branch).

Documentation/api.md
jotak: my intent was that users who override this parameter can easily know what's the setting baseline, so they have a known working starting point when they want to update this setting - but yes this is assuming this doc is always kept up to date. Let me know if you still prefer that I remove it.

jotak: For this particular RFE it's fine to have only one, but in case later a customer wants to tune one and not the other, splitting a single setting into two would be a breaking change, or would need to keep both settings versions thus making it more complex, that's why I thought it would be better to have 2 right from the beginning. I'm fine with both options.

If we make a single option, where should it go? In `NodeExporterConfig`, in `NodeExporterCollectorConfig` or in a new struct for common network settings?

jan--f: I agree with @simonpasquier here, seems most likely user will want to ignore devices across collectors.
I'd propose to add this as a field to `NodeExporterConfig`. If we later need per collector ignore lists, they could be added and would override the list set in `NodeExporterConfig`.

jsonnet/main.jsonnet
```diff
@@ -188,21 +188,6 @@ local inCluster =
             clusterLabel: $.values.common.dashboardClusterLabel,
           },
         },
```
simonpasquier, line 190:
```suggestion
        },
        // The list of ignored devices is replaced by CMO at runtime depending on its configuration.
        ignoredNetworkDevices:: '^.*$',
```

pkg/manifests/types.go
```diff
@@ -343,6 +343,19 @@ type NodeExporterCollectorTcpStatConfig struct {
...
+	// `[a-f0-9]{15}`,
+	// `enP.*`,
+	// `ovn-k8s-mp[0-9]*`,
+	// `br-ex`,
+	// `br-int`,
+	// `br-ext`,
+	// `br[0-9]*`,
+	// `tun[0-9]*`,
+	// `cali[a-f0-9]*`
+	IgnoredNetworkDevices *[]string `json:"ignoredNetworkDevices,omitempty"`
```
simonpasquier, line 358: (nit) we can use a slice instead of a pointer. An empty slice means "use the default list".
 
```suggestion
	IgnoredNetworkDevices []string `json:"ignoredNetworkDevices,omitempty"`
```

jotak: (and btw that's why I needed the ability to remove the  `--collector.netclass.ignored-devices` argument in manifest.go)

jotak:
thanks @simonpasquier @jan--f for the review, I think I addressed all the points

pkg/manifests/types.go
```diff
@@ -281,6 +281,10 @@ type NodeExporterConfig struct {
 	// For nodes with a high CPU count, you can set the limit to a low number, which  saves resources by preventing Go routines from being scheduled to run on all CPUs.
 	// However, I/O performance degrades if the `maxProcs` value is set too low, and there are many metrics to collect.
 	MaxProcs uint32 `json:"maxProcs,omitempty"`
+	// A list of network devices, as regular expressions, to be excluded from the relevant collectors configuration such as `netdev` and `netclass`.
+	// Some devices are excluded by default to minimize the impact on memory usage.
```
simonpasquier, line 285:
```suggestion
	// When not set, the Cluster Monitoring Operator uses a predefined list of devices to be excluded to minimize the impact on memory usage.
```

pkg/manifests/manifests.go
```diff
@@ -836,8 +837,22 @@ func (f *Factory) updateNodeExporterArgs(args []string) []string {
 		args = setArg(args, "--no-collector.tcpstat", "")
 	}
 
+	var excludedDevices string
+	if f.config.ClusterMonitoringConfiguration.NodeExporterConfig.Collectors.NetDev.Enabled ||
+		f.config.ClusterMonitoringConfiguration.NodeExporterConfig.Collectors.NetClass.Enabled {
+		devs := *f.config.ClusterMonitoringConfiguration.NodeExporterConfig.IgnoredNetworkDevices
+		if len(devs) > 0 {
```
simonpasquier, line 844: Do we need to a special case for zero-length list?

pkg/manifests/manifests.go
```diff
@@ -874,11 +896,14 @@ func (f *Factory) NodeExporterDaemonSet() (*appsv1.DaemonSet, error) {
 		switch container.Name {
 		case "node-exporter":
 			ds.Spec.Template.Spec.Containers[i].Image = f.config.Images.NodeExporter
-			ds.Spec.Template.Spec.Containers[i].Args = f.updateNodeExporterArgs(ds.Spec.Template.Spec.Containers[i].Args)
+			ds.Spec.Template.Spec.Containers[i].Args, err = f.updateNodeExporterArgs(ds.Spec.Template.Spec.Containers[i].Args)
 		case "kube-rbac-proxy":
 			ds.Spec.Template.Spec.Containers[i].Image = f.config.Images.KubeRbacProxy
 			ds.Spec.Template.Spec.Containers[i].Args = f.setTLSSecurityConfiguration(container.Args, KubeRbacProxyTLSCipherSuitesFlag, KubeRbacProxyMinTLSVersionFlag)
 		}
+		if err != nil {
```
simonpasquier, line 904: (nit) I would move it next to L899

Documentation/openshiftdocs/modules/nodeexporterconfig.adoc
```diff
@@ -22,6 +22,8 @@ Appears in: link:clustermonitoringconfiguration.adoc[ClusterMonitoringConfigurat
 
 |maxProcs|uint32|The target number of CPUs on which the Node Exporter's process will run. The default value here is `0`, which means that Node Exporter runs on all CPUs. If a kernel deadlock occurs or if performance degrades when reading from `sysfs` concurrently, you can change this value to `1`, which limits Node Exporter to running on one CPU. For nodes with a high CPU count, you can set the limit to a low number, which  saves resources by preventing Go routines from being scheduled to run on all CPUs. However, I/O performance degrades if the `maxProcs` value is set too low, and there are many metrics to collect.
 
+|ignoredNetworkDevices|*[]string|A list of network devices, as regular expressions, to be excluded from the relevant collectors configuration such as `netdev` and `netclass`. Some devices are excluded by default to minimize the impact on memory usage. If you modify this setting, watch the prometheus-k8s deployment closely for excessive memory usage.
```
simonpasquier, line 25: cc @bburt-rh for a review of the documentation.

pkg/manifests/manifests.go
jotak: well, an empty list would generate this: `--collector.netclass.ignored-devices=^()$` ... which technically could work... that would only match empty strings, which I guess is not supposed to happen for a network device label? I'm not 100% sure

Anyway it seems to me that removing this command line argument would be cleaner and more clear, than passing a regex that matches nothing?

pkg/manifests/manifests.go
jotak: done

pkg/manifests/manifests.go
simonpasquier: I don't feel strongly on this one but at least a comment would help for the future :)

pkg/manifests/types.go
```diff
@@ -283,6 +283,10 @@ type NodeExporterConfig struct {
 	// For nodes with a high CPU count, you can set the limit to a low number, which  saves resources by preventing Go routines from being scheduled to run on all CPUs.
 	// However, I/O performance degrades if the `maxProcs` value is set too low, and there are many metrics to collect.
 	MaxProcs uint32 `json:"maxProcs,omitempty"`
+	// A list of network devices, as regular expressions, to be excluded from the relevant collectors configuration such as `netdev` and `netclass`.
+	// When not set, the Cluster Monitoring Operator uses a predefined list of devices to be excluded to minimize the impact on memory usage.
```
simonpasquier, line 287: should we detail that passing an empty list = exclude no interface?

simonpasquier (commented):
I'm not too sure how we can validate with the [e2e tests](https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/node_exporter_test.go)... Maybe adding a wildcard exclude (e.g. '.*') and verify that we get no network metrics?

jotak:
@simonpasquier on e2e tests, I'm ok to try adding some but would appreciate some help bc having some troubles to run them. Basically ; my understanding is that they require a running cluster, however it seems they don't seem to prepare / cleanup everything before (or after) running? I keep getting errors related to objects already existing (e.g. `serviceaccounts "cluster-monitoring-operator-e2e" already exists`)

pkg/manifests/manifests.go
jotak: comment done

pkg/manifests/types.go
jotak: done

test/e2e/node_exporter_test.go
```diff
@@ -254,3 +254,60 @@ nodeExporter:
...
+			predicate: func(f float64) bool { return f > 0 },
+		},
+		{
+			name: "excluding 'lo'",
+			config: `
+nodeExporter:
+  ignoredNetworkDevices:
+  - lo`,
+			filter:    `device="lo"`,
+			predicate: func(f float64) bool { return f == 0 },
```
simonpasquier, line 285: when excluding devices, the query needs to use absent() otherwise it could see metrics from the past.

simonpasquier:
/skip

LGTM in general! We'd need feedback from @bburt-rh about the configuration documentation.

test/e2e/node_exporter_test.go
jotak: yep I figured that out when I was finally able to test locally :-)

jsonnet/main.jsonnet
```diff
@@ -189,21 +189,8 @@ local inCluster =
...
-        // 4. OVN virtual interfaces ovn-k8s-mp[0-9]*
-        // 5. virtual tunnels and bridges: tun[0-9]*|br[0-9]*|br-ex|br-int|br-ext
-        // 6. Calico Virtual NICs cali[a-f0-9]*
-        // Refer to:
-        // https://issues.redhat.com/browse/OCPBUGS-1321
-        // https://issues.redhat.com/browse/OCPBUGS-2729
-        // https://issues.redhat.com/browse/OCPBUGS-7282
-        ignoredNetworkDevices:: '^(veth.*|[a-f0-9]{15}|enP.*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext|br[0-9]*|tun[0-9]*|cali[a-f0-9]*)$',
+        // The list of ignored devices is replaced by CMO at runtime depending on its configuration.
+        ignoredNetworkDevices:: '^.*$',
```
raptorsun, line 193: It is safe to delete `ignoredNetworkDevices`, the operator will generate the arguments for node-exporter.

pkg/manifests/types.go
```diff
@@ -283,6 +283,11 @@ type NodeExporterConfig struct {
...
```
raptorsun, line 287: It is better to tell the user the default pattern is `^(veth.*|[a-f0-9]{15}|enP.*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext|br[0-9]*|tun[0-9]*|cali[a-f0-9]*)$`. 
This comment will appear in the generated documents.

raptorsun:
Overrall looks good, just some non-functional suggestions. thank you, @jotak  :D

jsonnet/main.jsonnet
jotak: That's what I did at first, but  @simonpasquier suggested to reintroduce this in order to make it more explicit that it's going to be overiden

pkg/manifests/types.go
jotak: Thanks for the review! This point was actually discussed here: https://github.com/openshift/cluster-monitoring-operator/pull/1963#discussion_r1211644529 (here also I initially documented the default list)

Documentation/api.md
```diff
@@ -343,6 +343,7 @@ The `NodeExporterConfig` resource defines settings for the `node-exporter` agent
...
 | maxProcs | uint32 | The target number of CPUs on which the Node Exporter's process will run. The default value here is `0`, which means that Node Exporter runs on all CPUs. If a kernel deadlock occurs or if performance degrades when reading from `sysfs` concurrently, you can change this value to `1`, which limits Node Exporter to running on one CPU. For nodes with a high CPU count, you can set the limit to a low number, which  saves resources by preventing Go routines from being scheduled to run on all CPUs. However, I/O performance degrades if the `maxProcs` value is set too low, and there are many metrics to collect. |
+| ignoredNetworkDevices | *[]string | A list of network devices, as regular expressions, to be excluded from the relevant collectors configuration such as `netdev` and `netclass`. When not set, the Cluster Monitoring Operator uses a predefined list of devices to be excluded to minimize the impact on memory usage. When set as an empty list, no devices are excluded. If you modify this setting, watch the prometheus-k8s deployment closely for excessive memory usage. |
```
bburt-rh, line 346:
```suggestion
| ignoredNetworkDevices | *[]string | A list of network devices, as regular expressions, to be excluded from the relevant collector configuration such as `netdev` and `netclass`. When not set, the Cluster Monitoring Operator uses a predefined list of devices to be excluded to minimize the impact on memory usage. When set as an empty list, no devices are excluded. If you modify this setting, monitor the prometheus-k8s deployment closely for excessive memory usage. |
```

Documentation/openshiftdocs/modules/nodeexporterconfig.adoc
```diff
@@ -22,6 +22,8 @@ Appears in: link:clustermonitoringconfiguration.adoc[ClusterMonitoringConfigurat
 
 |maxProcs|uint32|The target number of CPUs on which the Node Exporter's process will run. The default value here is `0`, which means that Node Exporter runs on all CPUs. If a kernel deadlock occurs or if performance degrades when reading from `sysfs` concurrently, you can change this value to `1`, which limits Node Exporter to running on one CPU. For nodes with a high CPU count, you can set the limit to a low number, which  saves resources by preventing Go routines from being scheduled to run on all CPUs. However, I/O performance degrades if the `maxProcs` value is set too low, and there are many metrics to collect.
 
+|ignoredNetworkDevices|*[]string|A list of network devices, as regular expressions, to be excluded from the relevant collectors configuration such as `netdev` and `netclass`. When not set, the Cluster Monitoring Operator uses a predefined list of devices to be excluded to minimize the impact on memory usage. When set as an empty list, no devices are excluded. If you modify this setting, watch the prometheus-k8s deployment closely for excessive memory usage.
```
bburt-rh, line 25:
```suggestion
|ignoredNetworkDevices|*[]string|A list of network devices, as regular expressions, to be excluded from the relevant collector configuration such as `netdev` and `netclass`. When not set, the Cluster Monitoring Operator uses a predefined list of devices to be excluded to minimize the impact on memory usage. When set as an empty list, no devices are excluded. If you modify this setting, monitor the `prometheus-k8s` deployment closely for excessive memory usage.
```

jsonnet/main.jsonnet
```diff
@@ -189,21 +189,8 @@ local inCluster =
...
-        // 3. enP.* virtual NICs on Azure cluster.
-        // 4. OVN virtual interfaces ovn-k8s-mp[0-9]*
-        // 5. virtual tunnels and bridges: tun[0-9]*|br[0-9]*|br-ex|br-int|br-ext
-        // 6. Calico Virtual NICs cali[a-f0-9]*
-        // Refer to:
-        // https://issues.redhat.com/browse/OCPBUGS-1321
-        // https://issues.redhat.com/browse/OCPBUGS-2729
-        // https://issues.redhat.com/browse/OCPBUGS-7282
-        ignoredNetworkDevices:: '^(veth.*|[a-f0-9]{15}|enP.*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext|br[0-9]*|tun[0-9]*|cali[a-f0-9]*)$',
+        // The list of ignored devices is replaced by CMO at runtime depending on its configuration.
```
bburt-rh, line 192:
```suggestion
        // The list of ignored devices is replaced by the Cluster Monitoring Operator (CMO) at runtime depending on the configuration of the CMO.
```
The phrase `its configuration` has an unclear antecedent. What is `its`? the CMO? or the list?

pkg/manifests/config.go
```diff
@@ -303,6 +303,35 @@ func (c *Config) applyDefaults() {
 	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile == "" {
 		c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile = FullCollectionProfile
 	}
+
+	// defaultExcludedDevices is the default for 2 arguments "--collector.netclass.ignored-devices" and "--collector.netdev.device-exclude".
```
bburt-rh, line 307:
```suggestion
	// `defaultExcludedDevices` is the default for 2 arguments `--collector.netclass.ignored-devices` and `--collector.netdev.device-exclude`.
```

pkg/manifests/config.go
```diff
@@ -303,6 +303,35 @@ func (c *Config) applyDefaults() {
...
+	// 6 kinds of virtual NICs are ignored by default:
```
bburt-rh, line 308:
```suggestion
	// The following virtual NICs are ignored by default:
```
Better not to specify a specific number in case the total number ever changes.

pkg/manifests/config.go
```diff
@@ -303,6 +303,35 @@ func (c *Config) applyDefaults() {
...
+	// 1. veth network interface associated with containers.
```
bburt-rh, line 309:
```suggestion
	// * `veth` network interface associated with containers.
```

pkg/manifests/config.go
```diff
@@ -303,6 +303,35 @@ func (c *Config) applyDefaults() {
...
+	// 2. OVN renames veth.* to <rand-hex>@if<X> where X is /sys/class/net/<if>/ifindex
```
bburt-rh, line 310:
```suggestion
	// * OVN renames `veth.*` to `<rand-hex>@if<X>` where `X` is `/sys/class/net/<if>/ifindex`
```

pkg/manifests/config.go
```diff
@@ -303,6 +303,35 @@ func (c *Config) applyDefaults() {
...
+	// thus [a-f0-9]{15}
```
bburt-rh, line 311:
```suggestion
	// thus `[a-f0-9]{15}`
```

pkg/manifests/config.go
```diff
@@ -303,6 +303,35 @@ func (c *Config) applyDefaults() {
...
+	// 3. enP.* virtual NICs on Azure cluster.
+	// 4. OVN virtual interfaces ovn-k8s-mp[0-9]*
+	// 5. virtual tunnels and bridges: tun[0-9]*|br[0-9]*|br-ex|br-int|br-ext
+	// 6. Calico Virtual NICs cali[a-f0-9]*
```
bburt-rh, line 315:
```suggestion
	// * `enP.*` virtual NICs on Azure cluster
	// * OVN virtual interfaces `ovn-k8s-mp[0-9]*`
	// * virtual tunnels and bridges: `tun[0-9]*|br[0-9]*|br-ex|br-int|br-ext`
	// * Calico Virtual NICs `cali[a-f0-9]*`
```

pkg/manifests/manifests.go
```diff
@@ -850,8 +851,24 @@ func (f *Factory) updateNodeExporterArgs(args []string) []string {
 		args = setArg(args, "--no-collector.tcpstat", "")
 	}
 
+	var excludedDevices string
+	if f.config.ClusterMonitoringConfiguration.NodeExporterConfig.Collectors.NetDev.Enabled ||
+		f.config.ClusterMonitoringConfiguration.NodeExporterConfig.Collectors.NetClass.Enabled {
+		devs := *f.config.ClusterMonitoringConfiguration.NodeExporterConfig.IgnoredNetworkDevices
+		// An empty list would generate a regular expression matching empty strings: ^()$
+		// It's more safe to avoid that, and not set the exclusion regex at all
```
bburt-rh, line 859:
```suggestion
		// An empty list generatea a regular expression matching empty strings: `^()$`
		// It is therefore preferable not to set the exclusion regex at all.
```

pkg/manifests/types.go
bburt-rh, line 287:
```suggestion
	// A list of network devices, as regular expressions, to be excluded from the relevant collector configuration such as `netdev` and `netclass`.
	// When not set, the Cluster Monitoring Operator uses a predefined list of devices to be excluded to minimize the impact on memory usage.
```

jsonnet/main.jsonnet
jotak: you're right that's CMO config

jotak:
thanks @bburt-rh - all changes accepted

jan--f:
@jotak The changes to `Documentation/api.md` need to be worked into `pkg/manifests/config.go`. Then run `make generate` to propagate those changes to the api documentation.

jotak:
@jan--f : done

jotak:
sorry, hold on, I did that the wrong way around

jotak:
@jan--f this time should be good. `make generate` did not trigger changes in doc, as expected

jan--f:
/label px-approved
Adding px approval as per agreement.
Ping @juzhao for qe approval.
/lgtm

Tai-RedHat:
I build this PR with cluster-bot
`launch openshift/cluster-monitoring-operator#1963`

check the ds, they did not show as `^.*$`
```
% oc -n openshift-monitoring get ds node-exporter -oyaml | grep -A10 containers: 
      containers:
      - args:
        - --web.listen-address=127.0.0.1:9100
        - --path.sysfs=/host/sys
        - --path.rootfs=/host/root
        - --path.udev.data=/host/root/run/udev/data
        - --no-collector.wifi
        - --collector.filesystem.mount-points-exclude=^/(dev|proc|sys|run/k3s/containerd/.+|var/lib/docker/.+|var/lib/kubelet/pods/.+)($|/)
        - --collector.netclass.ignored-devices=^(veth.*|[a-f0-9]{15}|enP.*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext|br[0-9]*|tun[0-9]*|cali[a-f0-9]*)$
        - --collector.netdev.device-exclude=^(veth.*|[a-f0-9]{15}|enP.*|ovn-k8s-mp[0-9]*|br-ex|br-int|br-ext|br[0-9]*|tun[0-9]*|cali[a-f0-9]*)$
        - --collector.cpu.info
```
is this expected?

jotak:
> I build this PR with cluster-bot `launch openshift/cluster-monitoring-operator#1963`
> 
> check the ds, they did not show as `^.*$`
> 
> is this expected?

Yes it is expected, the ds is always overriden with injected config at runtime. And since we don't want to introduce any breaking or disruptive change, the default config is set to the same exclusion list as it is currently, which corresponds to the regex that you're seeing.

simonpasquier (commented):
/hold cancel

PR #1969, opened 2023-05-22
OCPBUGS-14561: Prevent ci/prow/versions from failing on PR against release-xxx

machine424:
…implify.

Make hack/generate-versions.sh do nothing on PR that are not against the main branch as the components versions are only updated on the main branch (on release-xx branches, components kepp their versions when fixed)

Add a 'make check-versions' target to run the versions check. CI jobs will be updated to use this.

### DIFF
--- a/hack/generate-versions.sh
+++ b/hack/generate-versions.sh
@@ -15,20 +15,20 @@ TMP_BIN="$(pwd)/tmp/bin"
 # Ensure that we use the binaries from the versions defined in hack/tools/go.mod.
 PATH="${TMP_BIN}:${PATH}"
 
+MAIN_BRANCH="master"
+
 # Set default variable values
 : "${VERSION_FILE:=jsonnet/versions.yaml}"
-: "${INTERACTIVE:=true}"
+# PULL_BASE_REF will also be set by CI
+: "${PULL_BASE_REF:=master}"
+
+if [ "$PULL_BASE_REF" != "$MAIN_BRANCH" ]; then
+	echo >&2 "Components versions are only updated on '${MAIN_BRANCH}' branch for now. Nothing to do against '${PULL_BASE_REF}'."
+	exit 0
+fi
 
 version_from_remote() {
-	curl --retry 5 --silent --fail "https://raw.githubusercontent.com/${1}/master/VERSION"
-}
-
-# Fallback mechanism when VERSION file is empty or not found
-version_from_user() {
-	ver=""
-	echo >&2 -n "Cannot determine version of ${1}. Please provide version manually (without alphabetical prefixes) and press ENTER: "
-	read -r ver
-	echo "$ver"
+	curl --retry 5 --silent --fail "https://raw.githubusercontent.com/${1}/${MAIN_BRANCH}/VERSION"
 }
 
 CONTENT="$(gojsontoyaml -yamltojson <"${VERSION_FILE}")"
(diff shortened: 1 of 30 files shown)


hack/generate-versions.sh
```diff
@@ -15,20 +15,20 @@ TMP_BIN="$(pwd)/tmp/bin"
...
+fi
 
-# Fallback mechanism when VERSION file is empty or not found
-version_from_user() {
-	ver=""
-	echo >&2 -n "Cannot determine version of ${1}. Please provide version manually (without alphabetical prefixes) and press ENTER: "
-	read -r ver
-	echo "$ver"
+version_from_remote() {
+	curl --retry 5 --silent --fail "https://raw.githubusercontent.com/${1}/${MAIN_BRANCH}/VERSION"
```
machine424, line 31: (we can still use this if we change (I think we should do that for clarity) components versions on release-xxx branches when fixing them)

hack/generate-versions.sh
```diff
@@ -42,14 +42,6 @@ for c in $COMPONENTS; do
 	REMOTE="$(version_from_remote "$SLUG")"
 	REMOTE="${REMOTE#v}"
 
-	if [ "$REMOTE" = "" ] && [ "$INTERACTIVE" != "true" ]; then
```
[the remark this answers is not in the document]
jan--f: I have never used `INTERACTIVE` afair.

jan--f (approved):
/lgtm

machine424:
I added a commit to synchronize the versions as the bot that does it for thanos is disabled.

this being said, I cannot find the config of the bot that syncs kubeRbacProxy.

jsonnet/versions.yaml
```diff
@@ -14,11 +14,11 @@ repos:
...
-  kubeRbacProxy: 0.14.0
+  kubeRbacProxy: 0.14.1
   kubeStateMetrics: 2.8.2
   nodeExporter: 1.5.0
   promLabelProxy: 0.6.0
   prometheus: 2.43.0
   prometheusAdapter: 0.10.0
   prometheusOperator: 0.63.0
-  thanos: 0.31.0
+  thanos: 0.30.2
```
sthaha, line 24: Could you please help me understand why addressing https://issues.redhat.com//browse/MON-3173 require versions to be changed? 
If it isn't strictly needed, we shouldn't introduce changes to the product in a PR that addresses a development workflow tweak.

machine424: I explained this in the separate commit where I make the change: the versions job was failing because these versions were not aligned with the ones that are already used by the product.

jan--f: While I agree a separate PR would be ideal, I'd be happy to merge this since its in its own commit and clearly states why it was changed.

machine424:
I opened https://github.com/openshift/release/pull/39773 to use the new target in CI.

jsonnet/versions.yaml
machine424: If https://github.com/openshift/cluster-monitoring-operator/pull/1924 gets merged before this one, I'd discard my commit.

sthaha: > While I agree a separate PR would be ideal, I'd be happy to merge this since its in its own commit and clearly states why it was changed.

I am not against merging this but .. I wanted to understand why the versions are being downgraded. You might have more context than I do since the commit doesn't provide much information to me. It seems evident from the code that the versions are being downgraded but the message could be clearer about the reason behind this change.

E.g. Is it because the synbot disabled? 

```
   Synchronize version of thanos as it was downgraded and the sync bot is disabled

    Synchronize version of kubeRbacProxy

```


> the versions job was failing because these versions were not aligned with the ones that are already used in the product

If it isn't too much to ask for could you please incorporate that the message in the commit to indicate this. E.g. 

```
Fix versions ci job

`versions` ci job is failing because of versions of thanos and kube-rbac-proxy 
are different to the ones that are already used in the product.  This patch fixes
that by setting the correct version of thanos and kube-rbac-proxy. 
```

jan--f: :+1: for better docs, I have something in the pipeline for rhobs/handbook.

i'd argue for merging this. We need a bug to move this anyway and #1924 would need yet another one. So lets combine them and skip the additional Jira interaction.

machine424: I can help with rhobs/handbook, it falls within https://issues.redhat.com/browse/MON-3217

sthaha: +1 on merging this. The question wasn't intended to block the merge.

PR #1971, opened 2023-05-23
OCPBUGS-13939: Extend remote write test timeout

jan--f:

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -166,14 +166,14 @@ func TestPrometheusRemoteWrite(t *testing.T) {
           key: ca.crt`,
 			expected: []remoteWriteTest{
 				{
-					query:       fmt.Sprintf(`ceil(delta(prometheus_remote_storage_samples_pending{pod="%[1]s",prometheus_replica="%[1]s"}[1m]))`, "prometheus-k8s-0"),
-					expected:    func(v float64) bool { return v != 0 },
-					description: "prometheus_remote_storage_samples_pending indicates no remote write progress, expected a continuously changing delta",
+					query:       `sum (prometheus_build_info{cluster_id="",prometheus_replica="prometheus-k8s-0"})`,
+					expected:    func(v float64) bool { return v == 2 },
+					description: "expected 2 prometheus_build_info metrics for prometheus-k8s-0, found %[1]s",
 				},
 				{
-					query:       fmt.Sprintf(`ceil(delta(prometheus_remote_storage_samples_pending{pod="%[1]s",prometheus_replica="%[1]s"}[1m]))`, "prometheus-k8s-1"),
-					expected:    func(v float64) bool { return v != 0 },
-					description: "prometheus_remote_storage_samples_pending indicates no remote write progress, expected a continuously changing delta",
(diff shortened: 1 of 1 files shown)


test/e2e/prometheus_test.go
```diff
@@ -219,9 +219,10 @@ func TestPrometheusRemoteWrite(t *testing.T) {
 `,
 			expected: []remoteWriteTest{
 				{
-					query:       `absent(prometheus_remote_storage_samples_pending{__tmp_openshift_cluster_id__=~".+"})`,
+					query: `absent(prometheus_build_info{__tmp_openshift_cluster_id__=~".+"})`,
+					// absent returns 1 if query is empty
 					expected:    func(v float64) bool { return v == 1 },
-					description: "Expected to find 0 time series of metric prometheus_remote_storage_samples_pending with the temporary cluster_id label",
+					description: "Expected to find 0 time series of metric prometheus_build_info with the temporary cluster_id label, found %[1]s",
```
simonpasquier, line 225: if the test fails, there will be no result hence no `%[1]s`. In fact the description will never be invoked.

jan--f: Right this will only be invoked if the `validate` function fails. In that case there should be a query result already and the descriptions gets wrapped in [WaitForqueryReturn](https://github.com/openshift/cluster-monitoring-operator/blob/d0dad7cb7ce69c7ddd5c1b8dbe9263dd3a644302/test/e2e/framework/client.go#L375). Or am I missing something :thinking:

test/e2e/prometheus_test.go
```diff
@@ -166,14 +166,14 @@ func TestPrometheusRemoteWrite(t *testing.T) {
           key: ca.crt`,
 			expected: []remoteWriteTest{
 				{
-					query:       fmt.Sprintf(`ceil(delta(prometheus_remote_storage_samples_pending{pod="%[1]s",prometheus_replica="%[1]s"}[1m]))`, "prometheus-k8s-0"),
-					expected:    func(v float64) bool { return v != 0 },
-					description: "prometheus_remote_storage_samples_pending indicates no remote write progress, expected a continuously changing delta",
+					query:       `sum (prometheus_build_info{cluster_id="",,prometheus_replica="prometheus-k8s-0"})`,
```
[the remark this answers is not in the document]
jan--f: :facepalm:

test/e2e/prometheus_test.go
simonpasquier: If absent(...) returns something, it will always be 1. Otherwise WaitForQueryReturn will return an error here: https://github.com/openshift/cluster-monitoring-operator/blob/d0dad7cb7ce69c7ddd5c1b8dbe9263dd3a644302/test/e2e/framework/client.go#L371

sthaha (commented):
/lgtm

I will defer to @simonpasquier to take a final look for approval

slashpai:
https://github.com/openshift/cluster-monitoring-operator/pull/1973 to increase poll wait time for alertmanager. Apparently it takes longer for UWM alertmanager to become ready and `TestAlertmanagerUWMSecrets` is also one of the tests that fails often too.

PR #1972, opened 2023-05-24
OCPBUGS-14007: test/e2e: don't fail on telemeter remote write failed samples

jan--f:
This test adds a lot of flakiness. Assuming 0 failed samples is too strict for determining that remote write is working. This commit removes the check for failed samples so that we consider telemetry via remote write successful as long as some samples are send.

### DIFF
--- a/test/e2e/telemeter_test.go
+++ b/test/e2e/telemeter_test.go
@@ -17,7 +17,6 @@ package e2e
 import (
 	"context"
 	"errors"
-	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jan--f:
cc @marioferh

test/e2e/telemeter_test.go
```diff
@@ -98,17 +97,4 @@ func TestTelemeterRemoteWrite(t *testing.T) {
 			return nil
 		},
 	)
-
-	// Check that the Telemeter server returns no error.
```
simonpasquier, line 102: maybe we can comment out the check and note that it's disabled because flaky?

marioferh: Right maybe add a TODO for the future in order to explain why we don have this test. If a newcomer comes to the repo maybe its difficult to understand why we dont check failed samples.

marioferh (commented):
it needs make versions

test/e2e/telemeter_test.go
jan--f: I suppose the question is whether we'll want re-enable it in the future. I would propose that expecting 0 failed samples/s over 5 minutes couples our test too tightly to telemeter and the network conditions in between.

I'm not against bringing a check of this flavor back in the future, just maybe not quite so strict. In that light I'd opt to remove the code.

simonpasquier: fine by me

PR #1980, opened 2023-05-30
OCPBUGS-14376: Fix golangci-lint ineffassign violations

danielmellado:
This PR fixes golangci-lint ineffassign violations as per the new
Makefile option.


pkg/tasks/thanos_ruler_user_workload.go
```diff
@@ -233,6 +233,9 @@ func (t *ThanosRulerUserWorkloadTask) create(ctx context.Context) error {
 				return errors.Wrap(err, "initializing Thanos Querier Route failed")
 			}
 			queryURL, err = t.client.GetRouteURL(ctx, querierRoute)
+			if err !nil {
```
jan--f, line 236:
```suggestion
			if err != nil {
```

PR #1981, opened 2023-05-30
OCPBUGS-14377: Fix golangci-int wastedassign violations

danielmellado:
This commit fixes all wastedassign violations as per the new
golangci-int makefile target.


sthaha (commented):
/lgtm

PR #1979, opened 2023-05-30
OCPBUGS-14375: Fix golangci-lint gosimple violations

danielmellado:
This PR fixes gosimple violations as per the new makefile target for
golangci-lint.

Signed-off-by: Daniel Mellado <dmellado@redhat.com>

PR #1983, opened 2023-05-30
Skip some errcheck golangci-lint violations

danielmellado:
This commit skips a few errcheck golangci-lint violations, specifically
on the informers or goroutines.


pkg/operator/operator.go
```diff
@@ -1187,17 +1188,11 @@ func newRunReportForError(reason string, err error) runReport {
 	}
 }
 func (r runReport) isDegraded() bool {
-	if isNilOrAsExpected(r.degraded) {
-		return false
-	}
-	return true
```
jan--f, line 1193: This and the change to `isUnavailable` is already part of #1979. Do we need this here?

danielmellado: Probably rebased at some point, I'll check but it won't do any harm :D

pkg/alert/relabel_controller.go
```diff
@@ -61,6 +61,8 @@ type RelabelConfigController struct {
 }
 
 // NewRelabelConfigController returns a new RelabelConfigController instance.
+//
+//nolint:errcheck
```
simonpasquier, line 65: I'd prefer `nolint:errcheck` to be inserted where we explicitly agree to skip err checking.

pkg/client/client.go
```diff
@@ -1834,6 +1834,8 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
 // over `existing` keys and values. Additionally function performs filtering of labels and annotations from `exiting` map
 // where keys starting from string defined in `metadataPrefix` are deleted. This prevents issues with preserving stale
 // metadata defined by the operator
+//
+//nolint:errcheck
```
simonpasquier, line 1838: I'd move the marker to the specific code lines (or go with `_ = mergo.Merge(...)`) otherwise it might mask future violations.

pkg/operator/operator.go
```diff
@@ -177,6 +177,7 @@ type Operator struct {
 	relabelController *alert.RelabelConfigController
 }
 
+//nolint:errcheck
```
simonpasquier, line 180: same remark here

pkg/alert/relabel_controller.go
```diff
@@ -116,12 +116,14 @@ func NewRelabelConfigController(ctx context.Context, client *client.Client) (*Re
 		secretInformer:        secretInformer,
 	}
 
+	//nolint:errcheck
```
simonpasquier, line 119: what would it take to handle the error and return if not nil?

pkg/alert/relabel_controller.go
```diff
@@ -116,12 +116,14 @@ func NewRelabelConfigController(ctx context.Context, client *client.Client) (*Re
...
 	relabelConfigInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    controller.handleAlertRelabelConfigAdd,
 		UpdateFunc: controller.handleAlertRelabelConfigUpdate,
 		DeleteFunc: controller.handleAlertRelabelConfigDelete,
 	})
 
+	//nolint:errcheck
```
simonpasquier, line 126: same here

pkg/alert/rule_controller.go
```diff
@@ -58,6 +58,8 @@ type RuleController struct {
 }
 
 // NewRuleController returns a new AlertingRule controller instance.
+//
+//nolint:errcheck
```
simonpasquier, line 62: missed this one in my previous review but here also I'd prefer to check for errors if possible

danielmellado: missed it, done!

simonpasquier (commented):
/lgtm

PR #1984, opened 2023-05-30
OCPBUGS-14379: Skip specific govet violation on operator main.go

danielmellado:
This commit skips the warning of unbuffered os.Signal channel due to its explicitly wanted use
as per to do graceful shutdown.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -82,6 +82,7 @@ type telemetryConfig struct {
 	Matches []string `json:"matches"`
 }
 
+//nolint:govet
 func Main() int {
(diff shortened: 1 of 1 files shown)


jan--f:
/lgtm
/label docs-approved
/label px-approved
/label qe-approved
Adding labels for an internal refactoring.

cmd/operator/main.go
```diff
@@ -82,6 +82,7 @@ type telemetryConfig struct {
 	Matches []string `json:"matches"`
 }
 
+//nolint:govet
```
machine424, line 85: why not using a buffered channel?

PR #1985, opened 2023-05-30
OCPBUGS-14381: Skip golang-ci lint unused false positives

danielmellado:
This commit skips unused violations for specific parts of the code where
it'll be either be used indirectly or unsure about whether it may impact
later on. It also moves some part of the code to the tests packages to avoid
such issues.


pkg/operator/operator.go
```diff
@@ -1186,13 +1186,16 @@ func newRunReportForError(reason string, err error) runReport {
 		},
 	}
 }
+
+//nolint:unused
 func (r runReport) isDegraded() bool {
```
jan--f, line 1191: This and `isUnavailable` are only used in `operator_test.go` which is why golint misses this I suppose. Should those functions live in the test file?

pkg/tasks/telemeter.go
```diff
@@ -278,6 +278,7 @@ func (t *TelemeterClientTask) destroy(ctx context.Context) error {
 	return errors.Wrap(err, "creating Telemeter Client serving certs CA Bundle ConfigMap failed")
 }
 
+//nolint:unused
```
machine424, line 281: I think we can create a "technical debt" issue to get rid of this, IIUC it's no longer using after https://github.com/openshift/cluster-monitoring-operator/commit/23356b2df280acacfb8132616f3abe3e98ef2dbc and I don't think the function can/is being used by any other package.

pkg/operator/operator.go
sthaha: > Should those functions live in the test file?

Yes, that is good idea! I would covert this to a utility function in test since it makes the test readable.

pkg/tasks/telemeter.go
jan--f: why not remove it in this PR then?

pkg/operator/operator_test.go
```diff
@@ -446,6 +446,20 @@ func TestUpgradeableStatus(t *testing.T) {
 	}
 }
 
+func isDegraded(r runReport) bool {
```
simonpasquier, line 449: (nit) this could be turned into a test helper

```suggestion
func assertDegraded(t *testing.T, r runReport) {
    t.Helper()
    ...
}
```

pkg/manifests/image.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//nolint:unused
```
simonpasquier, line 15: why not remove the file?

pkg/client/condition.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//nolint:unused
```
simonpasquier, line 15: which function(s) is(are) unused?

pkg/operator/operator_test.go
danielmellado: +1, I'll refactor the tests overall in a separate PR. There's a few more functions I'd like to move. Thanks for the tip!

pkg/client/condition.go
simonpasquier: yes this would be my option: remove dead code and see if the build/tests fail.

danielmellado: ok, found all their usage in the end, basically belonged to the tests only

PR #1986, opened 2023-06-01
MON-3216: Add ownership labels to kube resources

machine424:
Generalize the use of recommended "app.kubernetes.io/managed-by" and "app.kubernetes.io/part-of" labels on Kube resources created and managed by CMO and CVO.

See https://issues.redhat.com/browse/MON-3216 for the whys.

We chose this "inject at the end" approach as it requires less changes, but
we're aware that:
- We need to label resources managed by CVO individually (see utils/add-labels.libsonnet)
- Items of a XXXList (RoleList in role-specific-namespaces.yaml for example) are missed
- Resources created by some CR are missed (as we don't tell their operators to add the labels)

A draft of the other approach (Add the labels explicitly to each resource) can be found at
https://github.com/machine424/cluster-monitoring-operator/commit/53867351d1eb6bb3bfbd69a8df222993712d9fd2#diff-3f0ac462e0855b3c5693e93fd73febeec9422783681a4a15b2a82f9f40efec89, with this approcah we can get better coverage, we have better control
but it's harder to maintain and is more intrusive.


jsonnet/components/node-exporter.libsonnet
```diff
@@ -120,18 +121,8 @@ function(params)
...
-      metadata+: {
-        labels+: {
-          'app.kubernetes.io/managed-by': 'cluster-monitoring-operator',
-        },
-      },
       spec+: {
         template+: {
-          metadata+: {
-            labels+: {
-              'app.kubernetes.io/managed-by': 'cluster-monitoring-operator',
```
machine424, line 132: we make upstream inject it now (even though CMO doesn't manage the Pods directly))

jsonnet/components/telemeter-client.libsonnet
```diff
@@ -19,15 +19,23 @@ function(params) {
...
+  },
 
   // Remapping everything as this is the only way I could think of without refactoring imported library
-  prometheusRule: tc.telemeterClient.prometheusRule,
-  clusterRoleBindingView: tc.telemeterClient.clusterRoleBindingView,
-  clusterRoleBinding: tc.telemeterClient.clusterRoleBinding,
-  clusterRole: tc.telemeterClient.clusterRole,
-  serviceAccount: tc.telemeterClient.serviceAccount,
-  service: tc.telemeterClient.service,
+  prometheusRule: tc.telemeterClient.prometheusRule + commonLabels,
```
machine424, line 29: I chose this over a function that will implicitly inject the labels.

(better solution is to make the upstream take care of it)

machine424:
/ retitle MON-3216: Add ownership labels to kube resources

machine424:
I'll be simulating an upgrade on a cluster, this shouldn't break controllers as I made sure to only ADD labels.

simonpasquier (commented):
I think that we can accomplish the same with less modifications to the jsonnet files and a 100% coverage:

https://github.com/openshift/cluster-monitoring-operator/commit/8954acf50c86d4de75c264052af27eacafa36f00

jsonnet/components/telemeter-client.libsonnet
machine424: we chose to inject at the end

simonpasquier:
> We need to label resources managed by CVO individually (see utils/add-labels.libsonnet)

sgtm

> Items of a XXXList (RoleList in role-specific-namespaces.yaml for example) are missed

is it possible to extend jsonnet/utils/add-labels.libsonnet and support *List objects?

> Resources created by some CR are missed (as we don't tell their operators to add the labels)

Are you talking about the resources managed by the Prometheus operator (e.g. Prometheus, Alertmanager & ThanosRuler)? We could pass the --labels argument to inject labels in the child resources.

machine424:
> is it possible to extend jsonnet/utils/add-labels.libsonnet and support *List objects?

Yep, we do sth like that in add-annotations.libsonnet, but as I see it it'll make the util complicated, so maybe in another PR (or the next time someone asks for those resources)

> Are you talking about the resources managed by the Prometheus operator (e.g. Prometheus, Alertmanager & ThanosRuler)? We could pass the --labels argument to inject labels in the child resources.

Yep, (and maybe others managed by other operators), it'd be great to have `app.kubernetes.io/part-of: openshift-monitoring` on them (normally they're passed via `podMetadata.metadata`) I didn't know about --labels, the labels setting will no longer be centralized in jsonnet + we'll have another "implicit" injection, but I'll add it.

simonpasquier:
> (or the next time someone asks for those resources)

I'm afraid it will be never then :stuck_out_tongue_closed_eyes:

machine424:
or `maybe in another PR` ;) I'll be thinking about a simpler solution.
(I added --labels to prom-op)

jsonnet/utils/add-labels.libsonnet
```diff
@@ -0,0 +1,19 @@
...
+  local managedBy(o) =
+    local name = o.metadata.name;
+    // Some of the resources are genrated via jsonnet and moved under manifects/ to be managed by CVO
+    if o.kind == 'CustomResourceDefinition' || (o.kind == 'Role' && name == 'cluster-monitoring-operator-alert-customization') || (o.kind == 'Deployment' && name == 'cluster-monitoring-operator') || (o.kind == 'ClusterRole' && std.setMember(name, ['cluster-monitoring-operator-namespaced', 'cluster-monitoring-operator'])) then
+      'cluster-version-operator'
+    else
+      'cluster-monitoring-operator',
+
+  addLabels(o, labels): {
+    // ignore ConfigMapList as this breaks some conversions in hack/build-jsonnet.sh
```
simonpasquier, line 11: can you make it a TODO?

machine424: done

jsonnet/utils/add-labels.libsonnet
```diff
@@ -0,0 +1,19 @@
+{
+  local managedBy(o) =
+    local name = o.metadata.name;
+    // Some of the resources which are generated by jsonnet are moved under manifests/ and managed by CVO.
+    if o.kind == 'CustomResourceDefinition' || (o.kind == 'Role' && name == 'cluster-monitoring-operator-alert-customization') || (o.kind == 'Deployment' && name == 'cluster-monitoring-operator') || (o.kind == 'ClusterRole' && std.setMember(name, ['cluster-monitoring-operator-namespaced', 'cluster-monitoring-operator'])) then
+      'cluster-version-operator'
+    else
+      'cluster-monitoring-operator',
+
+  addLabels(o, labels): {
```
jan--f, line 10:
```suggestion
  addLabels(labels,o ): {
```
Nit but what do you think of swapping the arguments here. I think that would improve readability in the main file. Other wise we have
```
addLabels(
    ...32 lines...,
    $labelsToAdd)

jan--f (commented):
One nit, otherwise this looks good.

jsonnet/utils/add-labels.libsonnet
machine424: done

jan--f (approved):
/lgtm

PR #1990, opened 2023-06-06
Add federate-client-certs

marioferh:


pkg/tasks/prometheus.go
```diff
@@ -280,6 +280,17 @@ func (t *PrometheusTask) create(ctx context.Context) error {
 		return errors.Wrap(err, "waiting for Metrics Client Certs secret failed")
 	}
 
+	federateCerts, err := t.factory.FederateClientCerts()
+	if err != nil {
+		return errors.Wrap(err, "initializing Federate Client Certs secret failed")
+	}
+
+	//nolint:ineffassign
+	federateCerts, err = t.client.WaitForSecret(ctx, federateCerts)
```
simonpasquier, line 289: (nit)
```suggestion
	//nolint:ineffassign
	_, err = t.client.WaitForSecret(ctx, federateCerts)
```

assets/prometheus-k8s/prometheus.yaml
```diff
@@ -195,6 +195,7 @@ spec:
   - prometheus-k8s-thanos-sidecar-tls
   - kube-rbac-proxy
   - metrics-client-certs
+  - federate-client-certs
```
simonpasquier, line 198: hmm I expect the federate-client-certs secret to be mounted into the telemeter client deployment?

marioferh: right

jsonnet/components/cluster-monitoring-operator.libsonnet
```diff
@@ -67,6 +67,17 @@ function(params) {
     data: {},
   },
 
+  federateClientCerts: {
```
sthaha, line 70: suggestion: 
It would we really nice if we started documenting what these secrets are getting used for. 

E.g.
```
// federate-client-certs secret is used by <X> to <achieve-what> 
```

sthaha (commented):
/lgtm

I do have `suggestion` that you may want to consider. Ideally we would want to provide as much information as possible to future maintainers about obscure parts of the system.

jsonnet/components/cluster-monitoring-operator.libsonnet
marioferh: Sure.

PR #1991, opened 2023-06-07
OCPBUGS-14123: make TestBodySizeLimit less flaky

raptorsun:
TestBodySizeLimit checks Prometheus config file instead of the metric "prometheus_target_scrapes_exceeded_body_size_limit_total".

The package "github.com/prometheus/prometheus/discovery/kubernetes" is added to `vendor/` to parse the field `kubernetes_sd_configs` in the config file.


simonpasquier (commented):
/lgtm

PR #1996, opened 2023-06-12
OCPBUGS-13153: Limit the value of GOMAXPROCS on node-exporter.

machine424:
xxxxx

* [x] Add to jsonnet
* [x] Add tests
* [x] Adjust commit message

### DIFF
--- a/jsonnet/components/node-exporter.libsonnet
+++ b/jsonnet/components/node-exporter.libsonnet
@@ -250,6 +250,21 @@ function(params)
                               '--collector.textfile.directory=' + textfileDir,
                               '--no-collector.btrfs',
                             ],
+                      command: [
+                        '/bin/sh',
+                        '-c',
+                        |||
+                          export GOMAXPROCS=4
+                          # We don't take CPU affinity into account as the container doesn't have integer CPU requests.
+                          # In case of error, fallback to the default value.
+                          NUM_CPUS=$(grep -c '^processor' "/proc/cpuinfo" 2>/dev/null || echo "0")
+                          if [ "$NUM_CPUS" -lt "$GOMAXPROCS" ]; then
+                            export GOMAXPROCS="$NUM_CPUS"
+                          fi
+                          echo "ts=$(date --iso-8601=seconds) num_cpus=$NUM_CPUS gomaxprocs=$GOMAXPROCS"
+                          exec /bin/node_exporter "$0" "$@"
+                        |||,
+                      ],
                       terminationMessagePolicy: 'FallbackToLogsOnError',
                       volumeMounts+: [{
                         mountPath: textfileDir,
--- a/assets/node-exporter/daemonset.yaml
+++ b/assets/node-exporter/daemonset.yaml
@@ -41,6 +41,19 @@ spec:
         - --collector.cpu.info
         - --collector.textfile.directory=/var/node_exporter/textfile
         - --no-collector.btrfs
+        command:
+        - /bin/sh
+        - -c
+        - |
+          export GOMAXPROCS=4
+          # We don't take CPU affinity into account as the container doesn't have integer CPU requests.
+          # In case of error, fallback to the default value.
+          NUM_CPUS=$(grep -c '^processor' "/proc/cpuinfo" 2>/dev/null || echo "0")
+          if [ "$NUM_CPUS" -lt "$GOMAXPROCS" ]; then
+            export GOMAXPROCS="$NUM_CPUS"
+          fi
+          echo "ts=$(date --iso-8601=seconds) num_cpus=$NUM_CPUS gomaxprocs=$GOMAXPROCS"
+          exec /bin/node_exporter "$0" "$@"
         image: quay.io/prometheus/node-exporter:v1.6.0
         name: node-exporter
         resources:


test/e2e/node_exporter_test.go
```diff
@@ -257,3 +257,17 @@ nodeExporter:
 	}
 
 }
+
+func TestNodeExporterGoMaxProcs(t *testing.T) {
+	t.Run("limited GOMAXPROCS", func(st *testing.T) {
+		f.PrometheusK8sClient.WaitForQueryReturn(
+			t, 5*time.Minute, `max(go_sched_gomaxprocs_threads{job="node-exporter"})`,
```
machine424, line 264: we could add a metric with the number of CPU into the equation but it'll add complexity...

test/e2e/node_exporter_test.go
```diff
@@ -257,3 +257,17 @@ nodeExporter:
...
+			func(v float64) error {
+				if v > 4 {
```
machine424, line 266: don't know if tests are likely to run on nodes with more than 4 cpu...

test/e2e/node_exporter_test.go
```diff
@@ -257,3 +257,17 @@ nodeExporter:
...
+					return fmt.Errorf(`expecting max(go_sched_gomaxprocs_threads{job="node-exporter"}) <= 4 but got %v.`, v)
```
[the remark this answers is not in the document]
machine424: it was removed by default in recent versions of prom go client, and node-exporter does not seem to add them (they didn't add them back since), and we'll not ask for them just for a test.

I remove the test, we can add one that check the "Set GOMAXPROCS=N" in logs, I cannot think of another that wouldn't make the patch more complex.

jan--f:
This seems to be a recurring issue for [golang](https://github.com/golang/go/issues/33803) based [containers](https://github.com/prometheus-operator/prometheus-operator/issues/501) in k8s. While this fixes the customer issue reported I wonder if we should think about a different solution, one that is easier to test and that fixes this once and for all.

Reading about this one often finds `go.uber.org/automaxproc` mentioned and [prometheus itself ](https://github.com/prometheus/prometheus/pull/10498)has made use of that. Maybe node-exporter could benefit from a similar solution?

machine424:
> This seems to be a recurring issue for https://github.com/golang/go/issues/33803 based https://github.com/prometheus-operator/prometheus-operator/issues/501 in k8s. While this fixes the customer issue reported I wonder if we should think about a different solution, one that is easier to test and that fixes this once and for all.

> Reading about this one often finds go.uber.org/automaxproc mentioned and https://github.com/prometheus/prometheus/pull/10498has made use of that. Maybe node-exporter could benefit from a similar solution?

Yes, I still don't understand why `go.uber.org/automaxproc` hasn't been integrated into Go yet, I've always found limiting `GOMAXPROCS` handy to calm down some programs when running them on big machines (it helped me once calm down Prometheus before they add automaxproc).

We think about implementing the same for node-exporter, but it wouldn't help in this case as we don't set CPU limits, maybe we should think about setting them...

(Actually, the CU runs a single node cluster (with 4 CPU reserved to platform components) and normally, the default value of GOMAXPROCS takes CPU affinity into account https://github.com/golang/go/blob/bce7aec3cdca8580585095007e9b7cea11a8812f/src/runtime/os_linux.go#L107, but the affinity doesn't seem to be set for node-exporter as its CPU requests is not an integer I think)

jan--f:
> We think about implementing the same for node-exporter, but it wouldn't help in this case as we don't set CPU limits, maybe we should think about setting them...

We are thinking about that already ;)
https://issues.redhat.com/browse/MON-3168

> but the affinity doesn't seem to be set for node-exporter as its CPU requests is not an integer I think)

Not sure I understand this part. I'm a bit torn, I feel like we can improve our understanding of what's going on here but at the same time I don't have anything constructive and will be out for the rest of the week. Maybe @openshift/openshift-team-monitoring  can chime in here?

jsonnet/components/node-exporter.libsonnet
```diff
@@ -214,6 +214,21 @@ function(params)
                               '--collector.textfile.directory=' + textfileDir,
                               '--no-collector.btrfs',
                             ],
+                      command: [
+                        '/bin/sh',
+                        '-c',
+                        |||
+                          export GOMAXPROCS=4
+                          # We don't take CPU affinity into account as the container dones't have integer CPU requests.
```
simonpasquier, line 222:
```suggestion
                          # We don't take CPU affinity into account as the container doesn't have integer CPU requests.
```

jsonnet/components/node-exporter.libsonnet
```diff
@@ -214,6 +214,21 @@ function(params)
...
+                          # In case of error, fallback to the default value.
+                          NUM_CPUS=$(grep -c '^processor' "/proc/cpuinfo" 2>/dev/null || echo "0")
+                          if [ "$NUM_CPUS" -lt "$GOMAXPROCS" ]; then
+                            export GOMAXPROCS="$NUM_CPUS"
+                          fi
+                          echo "Set GOMAXPROCS=$GOMAXPROCS"
```
simonpasquier, line 228: (suggestion) we can make the logged message look like a regular log message...
```suggestion
                          echo "ts=$(date --iso-8601=seconds) num_cpus=$NUM_CPUS gomaxprocs=$GOMAXPROCS"

```

simonpasquier (commented):
I understand the desire to fix it once for all in node_exporter and the `automaxproc` library + ability to configure CPU limits in CMO could be the right solution (btw we don't even need the automaxprocs lib as we could inject the CPU limits via the downward API). At the same time, this patch provides a tactical solution that should be relatively easy to backport...

raptorsun:
We need to check the env var `GOMAXPROCS` is not overwriting the settings `nodeExporter.maxProcs` just added by https://issues.redhat.com//browse/MON-2894

jsonnet/components/node-exporter.libsonnet
machine424: Done.

num_cpus will also ease debugging.

(We get `XXX+00:00` instead of `XXXMSZ`, I don't think this is a problem, it's even better to not make the user thinks it's node_exporter...)

(node_exporter can also `ts=2023-06-19T23:25:13.022Z caller=node_exporter.go:186 level=debug msg="Go MAXPROCS" procs=5` when debug, I don't think this is confusing as there is `Starting node_exporter` in between)

machine424:
> We need to check the env var GOMAXPROCS is not overwriting the settings nodeExporter.maxProcs just added by https://issues.redhat.com//browse/MON-2894

> According to[ the comment in the runtime code,](https://cs.opensource.google/go/go/+/refs/heads/master:src/runtime/proc.go;drc=261e26761805e03c126bf3934a8f39302e8d85fb;bpv=1;bpt=1;l=694) the main function of node exporter should be called after setting the process count by the env var GOMAXPROCS. It should not overwrite the setting of GOMAXPROCS in node exporter.

> But I suggest to give it a test to be sure :)


Yes, you're right, it's clear that the flag has precedence https://github.com/prometheus/node_exporter/commit/fb26b9fc72c3b18d6720303db747101907ed6ad1, but I had done some manual tests and the flag does have precedence in >= 4.14

I even wanted to add the check to the flag e2e test https://github.com/openshift/cluster-monitoring-operator/pull/1998/files#diff-28d0ddad61ad014c06899788cff47f72864db5df662c49235c59cee301e74b57R243 but removed it as the `go_sched_gomaxprocs_threads` metric is not available in latest node exporter versions https://github.com/openshift/cluster-monitoring-operator/pull/1996#discussion_r1227704236 (maybe we can see with if they will expose the metric again or want to expose gomaxprocs and other params on a /path as prometheus does and …we can add the check back in the future)

PR #1999, opened 2023-06-13
Fix link metrics

marioferh:

### DIFF
--- a/Documentation/sample-metrics.md
+++ b/Documentation/sample-metrics.md
@@ -1,7 +1,7 @@
 # Sample Metrics
 
 To understand what metrics are collected by the Telemeter service, we can replicate the request that the Telemeter client makes against a running OpenShift cluster's Prometheus service.
-The Telemeter client makes an HTTP GET request to Prometheus' `/federate` endpoint with the [metrics match rules](https://github.com/openshift/telemeter/blob/master/jsonnet/telemeter/metrics.jsonnet) URL encoded as query parameters.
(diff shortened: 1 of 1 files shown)


Documentation/sample-metrics.md
```diff
@@ -1,7 +1,7 @@
 # Sample Metrics
 
 To understand what metrics are collected by the Telemeter service, we can replicate the request that the Telemeter client makes against a running OpenShift cluster's Prometheus service.
-The Telemeter client makes an HTTP GET request to Prometheus' `/federate` endpoint with the [metrics match rules](https://github.com/openshift/telemeter/blob/master/jsonnet/telemeter/metrics.jsonnet) URL encoded as query parameters.
+The Telemeter client makes an HTTP GET request to Prometheus' `/federate` endpoint with the [metrics match rules](https://github.com/openshift/cluster-monitoring-operator/blob/master/Documentation/data-collection.md) URL encoded as query parameters.
```
sthaha, line 4: isn't https://github.com/openshift/cluster-monitoring-operator/blob/master/Documentation/telemetry/telemeter_query the right one?

sthaha: We should may be create a header in https://github.com/openshift/cluster-monitoring-operator/blob/master/Documentation/sample-metrics.md and point to that

marioferh: In fact the query is there, just remove the link will be fine

PR #2002, opened 2023-06-15
Log as cmo service account in make run-local script

marioferh:
After some issues with cert creation:
 {[{Denied True CSRDenied CSR "system:openshift:openshift-monitoring-dvpqj" was created by an unexpected user: "system:admin"

This change allows correct cmo sa to test changes in local: make run-local

Add: 
make run-local-as-cmo

### DIFF
--- a/hack/local-cmo.sh
+++ b/hack/local-cmo.sh
@@ -1,11 +1,35 @@
 #!/usr/bin/env bash
 set -eu -o pipefail
 
+cmo_login='true'
+
 # Script will:
 # 1. Patch CVO to exclude CMO from being managed
 # 2. Scale in-cluster CMO to 0
 # 3. Try to run CMO locally (CMO needs to be built before starting)
 
+
+echo "script starting"
+
+while test $# -gt 0; do
+  case "$1" in
+    -h|--help)
+      echo "local_cmo [options]"
+      echo " "
+      echo "options:"
+      echo "-h, --help                show brief help"
+      echo "-n, --no-cmo-login        run script as kube:admin"
+      exit 0
+      ;;
+    -n|--no-cmo-login)
+      cmo_login='false'
+      shift
+      ;;
+    *)
+      break
+      ;;
+  esac
+done
 
 validate() {
   local ret=0
@@ -78,6 +102,29 @@ images_from_deployment() {
   kc get deployment cluster-monitoring-operator -o json | jq -r '.spec.template.spec.containers[] | select(.name=="cluster-monitoring-operator") | .args[] | select(.|test("\\-images.*"))'
 }
 
+#log as cmo sa in order to be the correct user
+login_as_cmo() {
+	local cmo_sa="system:serviceaccount:openshift-monitoring:cluster-monitoring-operator"
+  local tmp_dir="tmp/cmo_token"
+	local token="$tmp_dir/cmo-token"
+
+	# use while to break out
+	while [[ -f "$token" ]]; do
+		oc login -u "$cmo_sa" || {
+			rm -f "$token"
+			break
+		}
+		oc get prometheus -n openshift-monitoring && return 0
+	done
+
+	echo "create a new cmo token -> $token"
+	mkdir -p "$tmp_dir"
+	oc create -n openshift-monitoring token cluster-monitoring-operator >"$token"
+  oc login --token="$(cat "$token")"
+	oc project openshift-monitoring
+	oc whoami
+}
+
 run() {
   echo "Running: $*"
   "$@"
@@ -88,6 +135,7 @@ main(){
   cd "$(git rev-parse --show-cdup)"
 
   validate || exit 1
+  oc login -u system:admin
   disable_managed_cmo
 
   local operator_config=manifests/0000_50_cluster-monitoring-operator_04-config.yaml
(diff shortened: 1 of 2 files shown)


hack/local-cmo.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 set -eu -o pipefail
+set -x
+set -v
```
sthaha, line 4: -x is too verbose IMHO and hinders readability.

assets/prometheus-k8s/prometheus.yaml
```diff
@@ -82,7 +82,7 @@ spec:
   - args:
     - --secure-listen-address=0.0.0.0:9092
     - --upstream=http://127.0.0.1:9090
-    - --allow-paths=/metrics
+    - --allow-paths=/metrics,/federate
```
sthaha, line 85: Unrelated change?

hack/local-cmo.sh
```diff
@@ -78,12 +80,40 @@ images_from_deployment() {
   kc get deployment cluster-monitoring-operator -o json | jq -r '.spec.template.spec.containers[] | select(.name=="cluster-monitoring-operator") | .args[] | select(.|test("\\-images.*"))'
 }
 
+#log as cmo sa in order to be the correct user
+kc_cmo_login() {
+	local cmo_sa="system:serviceaccount:openshift-monitoring:cluster-monitoring-operator"
+	local tmp_dir
+	tmp_dir="$(git rev-parse --show-toplevel)/tmp/$(git branch --show-current)"
```
sthaha, line 87: I think we can use "PROJECT_DIR/tmp" itself for this.

assets/prometheus-k8s/prometheus.yaml
marioferh: yep sorry I was working above another change :S

hack/local-cmo.sh
marioferh: Yep, it was for test purposes only

jsonnet/components/prometheus.libsonnet
```diff
@@ -203,7 +203,25 @@ function(params)
       data: {},
     },
 
-    kubeRbacProxySecret: generateSecret.staticAuthSecret(cfg.namespace, cfg.commonLabels, 'kube-rbac-proxy'),
+    kubeRbacProxySecret: generateSecret.staticAuthSecret(
```
sthaha, line 206: Unrelated patch?

marioferh: Yep

hack/local-cmo.sh
```diff
@@ -78,12 +80,40 @@ images_from_deployment() {
...
 run() {
   echo "Running: $*"
   "$@"
 }
 
 main(){
+  echo "log as cmo sa?"
+  echo "1.- Yes, remember: oc login -u system:admin when finished"
+  echo "2.- continue as admin"
+  read select
```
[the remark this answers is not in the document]
marioferh: You are right, not so clear but is here: https://github.com/openshift/cluster-monitoring-operator/blob/43ffa9e515659e02a313ff504e308fb5ac0c17a5/hack/local-cmo.sh#L133
I will improve it

hack/local-cmo.sh
```diff
@@ -1,12 +1,36 @@
...
+    -n|--no-cmo-login)
+      cmo_login='false'
+      shift
+      ;;
+    *)
+      break
+      ;;
+  esac
+done
+
```
sthaha, line 33: Could we please move this to a parse_args function? The rule of thumb for maintainable scripts is to keep as much as possible in a function so that there aren't any adhoc execution and the flow is predictable.

hack/local-cmo.sh
```diff
@@ -100,6 +148,12 @@ main(){
   local -a images
   while read -r img; do images+=( "$img" ); done < <(images_from_deployment)
 
+  if [ $cmo_login == 'true' ]; then
+    echo "loggin as cmo sa"
+    echo "run: oc login -u system:admin when finished in order to continue as admin"
+    login_as_cmo
+  fi
```
sthaha, line 155:
```suggestion
  if  $cmo_login; then
    echo "logging as cmo sa"
    login_as_cmo
    echo "run: oc login -u system:admin when finished in order to continue as admin"
  fi
```

hack/local-cmo.sh
```diff
@@ -1,12 +1,36 @@
...
+    -h|--help)
+      echo "local_cmo [options]"
+      echo " "
+      echo "options:"
+      echo "-h, --help                show brief help"
+      echo "-n, --no-cmo-login        run script as kube:admin"
+      exit 0
+      ;;
+    -n|--no-cmo-login)
+      cmo_login='false'
```
sthaha, line 25:
```suggestion
      CMO_LOGIN=false
```

hack/local-cmo.sh
```diff
@@ -1,12 +1,36 @@
 #!/usr/bin/env bash
 set -eu -o pipefail
 
+cmo_login='true'
```
sthaha, line 4:
```suggestion
declare CMO_LOGIN=true
```

hack/local-cmo.sh
sthaha: I think we can improve the script by switching back to the old use when the script exits `trap restore_login EXIT`

marioferh:
closing this to continue with: https://github.com/openshift/cluster-monitoring-operator/pull/2010

PR #2010, opened 2023-06-22
MON-3253: refactor local-cmo.sh to use CMO SA by default

sthaha:
This PR refines the local-cmo.sh script by implementing the
following improvements:

- By default, the script now authenticates as the CMO service account
  (see tmp/cmo-kubeconfig) before launching the operator. To maintain
  the previous behavior, a --no-cmo-login option is provided.

- Instead of utilizing the compiled binary, the script now uses "go
  run cmd/operator" to execute the operator. This change enables
  automatic integration of source code modifications upon script reruns.

- Introduces a --dry-run flag, allowing the script to simulate
  operations without applying any modifications to the cluster.

These enhancements optimize the execution and flexibility of the
local-cmo.sh script for smoother operator management.

signed-by: sthaha@redhat.com

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 /operator
 
 jsonnet/vendor/
-tmp/
+/tmp
 
 # These are empty target files, created on every docker build. Their sole
 # purpose is to track the last target execution time to evaluate, whether the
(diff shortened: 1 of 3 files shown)


marioferh (approved):
Tested
/lgtm

PR #2011, opened 2023-06-23
OCPBUGS-14922: skip console-plugin installation if console CO is absent

sthaha:
This PR fixes the failure to upgrade from OpenShift version 4.13 to 4.14 in environments (such as telco) where `console` ClusterOperator is disabled. The upgrade fails happens when CMO executes the console-plugin task which fails due to missing ConsolePlugin.

The fix checks for the presence of `console` ClusterOperator and bails out (with a log message) if it is absent.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -377,6 +377,25 @@ func (c *Client) ConsoleListWatch(ctx context.Context) *cache.ListWatch {
 	}
 }
 
+func (c *Client) ClusterVersionListWatch(ctx context.Context, name string) *cache.ListWatch {
+	clusterVersionInterface := c.oscclient.ConfigV1().ClusterVersions()
+
+	return &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			return clusterVersionInterface.List(ctx,
+				metav1.ListOptions{
+					FieldSelector: fields.OneTermEqualSelector("metadata.name", name).String(),
+				})
+		},
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			return clusterVersionInterface.Watch(ctx,
+				metav1.ListOptions{
+					FieldSelector: fields.OneTermEqualSelector("metadata.name", name).String(),
+				})
+		},
+	}
+}
+
 func (c *Client) ClusterOperatorListWatch(ctx context.Context, name string) *cache.ListWatch {
 	ClusterOperatorInterface := c.oscclient.ConfigV1().ClusterOperators()
 
@@ -1860,6 +1879,19 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
 	}
 
 	return int(podCapacityTotal), nil
+}
+
+func (c *Client) HasClusterCapability(ctx context.Context, capability configv1.ClusterVersionCapability) (bool, error) {
+	version, err := c.oscclient.ConfigV1().ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	return slices.Contains(version.Status.Capabilities.EnabledCapabilities, capability), nil
+}
+
+func (c *Client) HasConsoleCapability(ctx context.Context) (bool, error) {
+	return c.HasClusterCapability(ctx, configv1.ClusterVersionCapabilityConsole)
 }
 
 func (c *Client) CreateOrUpdateConsolePlugin(ctx context.Context, plg *consolev1.ConsolePlugin) error {
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/component-base/metrics/legacyregistry"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	apiutilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/informers"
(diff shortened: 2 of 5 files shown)


pkg/client/client.go
```diff
@@ -1862,6 +1862,14 @@ func (c *Client) PodCapacity(ctx context.Context) (int, error) {
 	return int(podCapacityTotal), nil
 }
 
+func (c *Client) HasConsoleOperatorEnabled(ctx context.Context) (bool, error) {
+	_, err := c.oscclient.ConfigV1().ClusterOperators().Get(ctx, "console", metav1.GetOptions{})
```
simonpasquier, line 1866: (suggesetion) I'd rather check the status of the clusterversion/cluster resource which should tell us if the `Console` capability is enabled or not.

https://github.com/openshift/api/blob/ea40115b9fa67188f89075ec4e34a9ea346ea123/config/v1/types_cluster_version.go#L441-L454

https://github.com/openshift/api/blob/ea40115b9fa67188f89075ec4e34a9ea346ea123/config/v1/types_cluster_version.go#L271-L273

sthaha: Thanks @simonpasquier, Nice find. I will modify the code rely on this instead.

pkg/tasks/monitoring_plugin.go
```diff
@@ -37,6 +38,19 @@ func NewMonitoringPluginTask(client *client.Client, factory *manifests.Factory,
 }
 
 func (t *MonitoringPluginTask) Run(ctx context.Context) error {
+	// ensure console is enabled and bail out if it isn't
+	{
+		enabled, err := t.client.HasConsoleCapability(ctx)
+		if err != nil {
+			return errors.Wrap(err, "failed to determine if console cluster-operator is enabled")
```
simonpasquier, line 45: (nit)
```suggestion
			return errors.Wrap(err, "failed to determine if the console capability is enabled")
```

pkg/operator/operator.go
```diff
@@ -406,6 +406,20 @@ func New(
 	}
 	o.informers = append(o.informers, informer)
 
+	// console cluster-operator is an optional component and can be enabled
+	// after installation.
+	informer = cache.NewSharedIndexInformer(
+		o.client.ClusterOperatorListWatch(ctx, "console"),
```
simonpasquier, line 412: To be consistent with the way CMO infers the availability of the console operator, should it list/watch the "cluster" ClusterVersion resource instead?
btw CMO already watches the console CO but only for updates.

https://github.com/openshift/cluster-monitoring-operator/blob/70dd5c9a414448ace1d07090730510d5922b6b18/pkg/operator/operator.go#L374-L383

pkg/tasks/monitoring_plugin.go
```diff
@@ -37,6 +38,19 @@ func NewMonitoringPluginTask(client *client.Client, factory *manifests.Factory,
...
+		}
+
+		if !enabled {
+			klog.V(4).Infof("Skipping installation of Console Plugin as ClusterOperator console is not enabled")
```
simonpasquier, line 49:
```suggestion
			klog.V(4).Infof("Skipping installation of Console Plugin as the console capability is not enabled")
```

simonpasquier (commented):
The handleEvent() function needs to react on changes to the clusterversion object.

https://github.com/openshift/cluster-monitoring-operator/blob/8d675913996ba7a66deb89bb828e0ac283620876/pkg/operator/operator.go#L568-L596

I'd also replace the existing if blocks by type switches (e.g. https://go.dev/tour/methods/16).

pkg/operator/operator.go
```diff
@@ -406,6 +406,20 @@ func New(
 	}
 	o.informers = append(o.informers, informer)
 
+	// Many of the cluster capabilities such as Console can be enabled after
+	// installation. So this watches for any updates to the ClusterVersion - version
```
simonpasquier, line 410: Maybe worth adding a comment that the console capability (like other capabilities) can only go from disabled -> enabled and not the other way around, meaning that CMO doesn't have to deal with removal of the console plugin resources.

pkg/tasks/monitoring_plugin.go
```diff
@@ -37,6 +38,19 @@ func NewMonitoringPluginTask(client *client.Client, factory *manifests.Factory,
 }
 
 func (t *MonitoringPluginTask) Run(ctx context.Context) error {
+	// ensure console is enabled and bail out if it isn't
```
simonpasquier, line 41: Maybe worth adding a comment that the console capability (like other capabilities) can only go from disabled -> enabled and not the other way around, meaning that CMO doesn't have to deal with removal of the console plugin resources.

sthaha: Are we okay with doing this in a follow-up patch, so that this blocker is resolved?

sthaha: Scratch, that, I have added the change in this PR since I had to address a comment.

pkg/operator/operator.go
```diff
@@ -565,32 +580,27 @@ func (o *Operator) keyFunc(obj interface{}) (string, bool) {
...
+		gk := rtObj.GetObjectKind().GroupVersionKind().GroupKind()
+		metaObj := obj.(metav1.Object)
+		name := metaObj.GetName()
+		if ns := metaObj.GetNamespace(); ns != "" {
+			name = ns + "/" + name
+		}
+		// NOTE: use %T to print the type if the gv information is absent
+		objKind := gk.String()
+		if objKind == "" {
+			objKind = fmt.Sprintf("%T", obj)
```
sthaha, line 601: This is helpful as follows
```
I0628 16:14:51.525173 1595999 operator.go:603] Triggering an update due to a change in ClusterOperator.config.openshift.io - ingress

I0628 16:14:52.194462 1595999 operator.go:603] Triggering an update due to a change in *v1.Infrastructure - cluster
I0628 16:14:52.199262 1595999 operator.go:603] Triggering an update due to a change in *v1.APIServer - cluster
I0628 16:14:52.211466 1595999 operator.go:603] Triggering an update due to a change in *v1.ClusterVersion - version
```

PR #2012, opened 2023-06-23
OCPBUGS-9835: Add Content-Security-Policy headers to prometheus and alertmanager

marioferh:
Add http Content-Security-Policy headers
Add e2e test


pkg/manifests/manifests_test.go
```diff
@@ -1419,6 +1419,30 @@ func TestRemoteWriteAuthorizationConfig(t *testing.T) {
 	}
 }
 
+func TestPrometheusK8sHttpConfig(t *testing.T) {
```
marioferh, line 1422: Tested here and with curl. Not sure if it is needed to add some e2e test.
`Allow: GET, OPTIONS, POST
Content-Length: 19
Content-Security-Policy: frame-ancestors 'none'
Content-Type: text/plain; charset=utf-8
Date: Fri, 23 Jun 2023 13:08:26 GMT
Gap-Auth: system:serviceaccount:openshift-monitoring:prometheus-k8s@cluster.local
Gap-Upstream-Address: localhost:9090
X-Content-Type-Options: nosniff`

pkg/manifests/manifests_test.go
```diff
@@ -1419,6 +1419,52 @@ func TestRemoteWriteAuthorizationConfig(t *testing.T) {
...
+	config :=
```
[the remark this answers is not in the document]
raptorsun: same for line 1447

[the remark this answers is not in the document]
marioferh: Done thanks!

pkg/manifests/manifests_test.go
simonpasquier: Given that there are several services that require the changes and that the configuration isn't uniform, an e2e test would be useful.

simonpasquier: If we implement e2e tests, I wouldn't require unit tests as they feel redundant.

marioferh: done

test/e2e/alertmanager_test.go
```diff
@@ -135,6 +135,130 @@ enableUserWorkload: true`,
 	}
 }
 
+func TestAlertmanagerPolicyHeaders(t *testing.T) {
```
simonpasquier, line 138: (suggestion) I would group the checks into one test with sub-tests (one per component). IMHO it will be easier to maintain and share code.

test/e2e/alertmanager_test.go
```diff
@@ -135,6 +135,130 @@ enableUserWorkload: true`,
...
+		foreground := metav1.DeletePropagationForeground
+		if err := f.KubeClient.CoreV1().Namespaces().Delete(ctx, testNs, metav1.DeleteOptions{PropagationPolicy: &foreground}); err != nil {
+			t.Logf("err deleting namespace %s: %v", testNs, err)
+		}
+	})
+
+	sa := "editor"
+	cr := "monitoring-rules-edit"
+
+	_, err = f.CreateServiceAccount(testNs, sa)
```
simonpasquier, line 204: instead of creating a new service account + bindings, we could reuse the [e2e service account](https://github.com/openshift/cluster-monitoring-operator/blob/d4fa84c8be2e3e20f6d65829022d793339b0e491/test/e2e/framework/framework.go#L59) which should already have all RBACs.

test/e2e/prometheus_test.go
```diff
@@ -370,6 +371,95 @@ func TestBodySizeLimit(t *testing.T) {
...
+
+		if resp.StatusCode != http.StatusOK {
+			return fmt.Errorf("expecting 200 status code, got %d (%q)", resp.StatusCode, resp.Body)
+		}
+
+		if resp.Header == nil {
+			return fmt.Errorf("expecting policy headers but is empty")
+		}
+
+		if resp.Header["Content-Security-Policy"][0] != "frame-ancestors 'none'" {
```
simonpasquier, line 456: (suggestion) simpler
```suggestion
		if resp.Header.Get("Content-Security-Policy") != "frame-ancestors 'none'" {
```

test/e2e/alertmanager_test.go
marioferh: thanks, working on it

test/e2e/prometheus_test.go
```diff
@@ -370,6 +371,95 @@ func TestBodySizeLimit(t *testing.T) {
...
+	err = framework.Poll(2*time.Second, 10*time.Second, func() error {
+		_, err := f.CreateServiceAccount(testNs, testAccount)
+		return err
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = framework.Poll(2*time.Second, 10*time.Second, func() error {
+		_, err = f.CreateClusterRoleBinding(testNs, testAccount, "admin")
```
raptorsun, line 406: this role binding should be cleaned up. or reuse the e2e test account "cluster-monitoring-operator-e2e" 
created in https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/framework/framework.go#L184

test/e2e/prometheus_test.go
```diff
@@ -370,6 +371,95 @@ func TestBodySizeLimit(t *testing.T) {
...
+		if resp.Header == nil {
+			return fmt.Errorf("expecting policy headers but is empty")
+		}
+
+		if resp.Header["Content-Security-Policy"][0] != "frame-ancestors 'none'" {
+			return fmt.Errorf("expecting frame-ancestors 'none' policy headers, got %q", resp.Header["Content-Security-Policy"])
+		}
+		return nil
+	})
+}
```
raptorsun, line 461: missing err check by the end
```
	if err != nil {
		t.Fatal(err)
	}
```

test/e2e/alertmanager_test.go
marioferh: If I change token to:

`		token, err := f.GetServiceAccountToken(namespaceName, e2eServiceAccount)
		if err != nil {
			return err
		}`

I am getting: 
`mariofar respo &{403 Forbidden 403 HTTP/1.1 1 1 map[Content-Length:[142] Content-Type:[text/plain; charset=utf-8] Date:[Mon, 03 Jul 2023 16:56:11 GMT] X-Content-Type-Options:[nosniff]] 0xc000525d40 142 [] false false map[] 0xc0000da900 0xc0001454a0}
`

test/e2e/alertmanager_test.go
marioferh: Updated, pleas tell me if now looks better

test/e2e/alertmanager_test.go
marioferh: Fixed.

test/e2e/prometheus_test.go
marioferh: Done, Fixed.

test/e2e/prometheus_test.go
marioferh: Done

test/e2e/security_headers_test.go
```diff
@@ -0,0 +1,94 @@
...
+		client = framework.NewPrometheusClient(
+			host,
+			token,
+			&framework.QueryParameterInjector{
+				Name:  "namespace",
+				Value: f.Ns,
+			},
+		)
+		return nil
+	})
```
jan--f, line 67:
```suggestion
	})
	
	if err != nil {
		t.Fatal(err)
	}
```
No point in trying the request without a client is there?

simonpasquier (commented):
/lgtm

marioferh:
> Maybe add one more error bailout?

where?

PR #2015, opened 2023-06-26
OCPBUGS-12714: turn on netlink mode of netclass collector for node exporter

raptorsun:


simonpasquier (commented):
you'd need to clarify in the documentation (pkg/manifests/types.go) that the netlink driver is enabled by default now.

simonpasquier (commented):
Also we don't have any unit test for this option?

simonpasquier:
Finally I think that `UseNetlink` should be changed to a bool pointer so we can distinguish between the value isn't defined (==true), false or true.

simonpasquier:
> Thanks @simonpasquier for the timely review :) Documentation has been updated mentioning that netlink mode is activated by default. The unit test is already in `pkg/manifests/manifests_test.go`, added when developing the `UseNetlink` flag. This PR just adjust the argument settings of the node exporter container in the test. About the case of an empty value of `UseNetlink`, I suppose if we do not have other action than the default action for the empty value, we can stay with a true/false value setting instead of a true/fase/null value setting.

I still think that the current way we init and deserialize the configuration is error-prone and not very consistent. As a follow-up, I'd recommend that we clean up the code and find a simpler way.

PR #2017, opened 2023-06-27
MON-3180: Expose and propagate TopologySpreadConstraints for telemeter-client

marioferh:
Give users a TopologySpreadConstraints field in the TelemeterClientConfig field and propagate this to the pod that is created.


jan--f (approved):
/lgtm

PR #2018, opened 2023-06-27
OCPBUGS-15440: fix CMO to apply console-plugin pod.spec config

sthaha:
The console-plugin's deployment configs that need to applied to the deployment.spec.template.spec weren't getting applied because the changes were made to the copy of the object instead of the original object itself. The tests didn't catch the issue because the pod assertions didn't validate if any pods were found for the combination of namespace and label-selector

This commit fixes both by refering (pointer) to the original spec instead of the copy and by fixing the assertions fail if it fails to find any pods.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -2621,8 +2621,8 @@ func (f *Factory) MonitoringPluginDeployment() (*appsv1.Deployment, error) {
 
 	// ensure console-plugin container is present even if config isn't defined so that,
 	// we validate that deployment has the expected container name. Thereby avoiding
-	// any suprises should user add config later.
-	podSpec := d.Spec.Template.Spec
+	// any surprises should user add config later.
+	podSpec := &d.Spec.Template.Spec
 	containers := podSpec.Containers
 	idx := slices.IndexFunc(containers, containerNameEquals(MonitoringPluginDeploymentContainer))
 	if idx < 0 {
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -754,7 +754,7 @@ func TestClusterMonitorConsolePlugin(t *testing.T) {
 		deploymentName = "monitoring-plugin"
 		cpu            = "10m"
 		mem            = "13Mi"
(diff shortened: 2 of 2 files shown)


simonpasquier (commented):
/lgtm

note for a follow-up: we can also add a unit test in the manifests package for testing all CMO options related to the monitoring console plugin.

sthaha:
@simonpasquier @jan--f I have removed the assertion that  I added that to `AssertPodConfiguration` to check that at least one pod matches the selector passed. 
```
			if len(pods.Items) == 0 {
				return fmt.Errorf("%w - failed to find Pods in %s matching %s", err, namespace, labelSelector)
			}

```
due to the following unrelated test failures which may be another bug to fix

```
=== RUN   TestClusterMonitorOSMConfig/assert_pod_configuration_is_as_expected
    assertions.go:347: timed out waiting for the condition: %!w(<nil>) - failed to find Pods in openshift-monitoring matching k8s-app=openshift-state-metrics
--- FAIL: TestClusterMonitorOSMConfig (307.04s)

    --- PASS: TestClusterMonitorTelemeterClientConfig/test_the_telemeter-client_deployment_is_rolled_out (6.02s)
    --- FAIL: TestClusterMonitorTelemeterClientConfig/assert_pod_configuration_is_as_expected (300.01s)

=== RUN   TestClusterMonitorTelemeterClientConfig/assert_pod_configuration_is_as_expected
    assertions.go:347: timed out waiting for the condition: %!w(<nil>) - failed to find Pods in openshift-monitoring matching app.kubernetes.io/component=dashboard

```

sthaha:
Just a note - for `osm` below are the labels set 
```
app.kubernetes.io/component=exporter
app.kubernetes.io/managed-by=cluster-monitoring-operator
app.kubernetes.io/name=openshift-state-metrics
app.kubernetes.io/part-of=openshift-monitoring
pod-template-hash=6d87776c5f
```

sthaha:
https://github.com/openshift/cluster-monitoring-operator/pull/2028 should fix the issues found while testing this PR.

marioferh (commented):
/lgtm

PR #2021, opened 2023-06-27
OCPBUGS-15469: Limit the value of GOMAXPROCS on node-exporter to 4.

machine424:
On nodes with multiple CPU cores, this should help avoid lock contentions without having any side effects, see the ticket for more details.

node-exporter versions that have "--runtime.gomaxprocs" can still override this: the flag has precedence over this automatic setting.

Make the entrypoint set the env var before runnning the process. Get the CPU count from /proc/cpuinfo without taking CPU affinity (via cpuset) into account as the container's CPU requests is not an integer, thus no affinity is applied.

### DIFF
--- a/assets/node-exporter/daemonset.yaml
+++ b/assets/node-exporter/daemonset.yaml
@@ -41,6 +41,19 @@ spec:
         - --collector.cpu.info
         - --collector.textfile.directory=/var/node_exporter/textfile
         - --no-collector.btrfs
+        command:
+        - /bin/sh
+        - -c
+        - |
+          export GOMAXPROCS=4
+          # We don't take CPU affinity into account as the container doesn't have integer CPU requests.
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
/lgtm

sdodson:
/retitle OCPBUGS-15469: Limit the value of GOMAXPROCS on node-exporter to 4.

Tai-RedHat:
tested with PR

jan--f:
@Tai-RedHat can you add `cherry-picked-approved` or do we need @juzhao for that?

Tai-RedHat:
> @Tai-RedHat can you add `cherry-picked-approved` or do we need @juzhao for that?

I'm not in the allowlist yet, I can not add `cherry-picked-approved`. I will submit a pr for that later.
Anyway, @juzhao can you help add this time?

PR #2023, opened 2023-06-27
OCPBUGS-15473: Limit the value of GOMAXPROCS on node-exporter to 4

machine424:
On nodes with multiple CPU cores, this should help avoid lock contentions without having any side effects, see the ticket for more details.

node-exporter versions that have "--runtime.gomaxprocs" can still override this: the flag has precedence over this automatic setting.

Make the entrypoint set the env var before runnning the process. Get the CPU count from /proc/cpuinfo without taking CPU affinity (via cpuset) into account as the container's CPU requests is not an integer, thus no affinity is applied.


sdodson:
/retitle OCPBUGS-15473: Limit the value of GOMAXPROCS on node-exporter to 4

Tai-RedHat:
tested pr with cluster-bot

raptorsun:
/retitle OCPBUGS-15473: Limit the value of GOMAXPROCS on node-exporter to 4

PR #2025, opened 2023-06-28
MON-3176: Expose and propagate TopologySpreadConstraints for prometheus-adapter

marioferh:
Give users a TopologySpreadConstraints field in the K8sPrometheusAdapter field and propagate this to the pod that is created.


jan--f (approved):
/lgtm

PR #2027, opened 2023-06-28
Add new alert for kube-api degraded

danielmellado:
This PR adds a new alert for kube-api-server been in degraded state.


simonpasquier:
I think that the bug needs to move to the API team and the alerting rule owned by https://github.com/openshift/cluster-kube-apiserver-operator/.

danielmellado:
Closing in favor of the patch from the api folks

PR #2029, opened 2023-06-29
MON-2694: Write tooling to facilitate implementation of metric collection profiles for OCP components

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/MON-2694

Alternative https://github.com/openshift/cluster-monitoring-operator/pull/2053


hack/monitorgen/main.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		log.Fatalf("failed to read asset %v: %e", *path, err)
+	}
+
+	m, err := manifests.NewServiceMonitor(bytes.NewReader(f))
+	if err != nil {
+		log.Fatalf("failed to convert to service monitor: %e", err)
+	}
+
+	// The script is only inteded to be used with resources of kind ServiceMonitor
```
jan--f, line 69:
```suggestion
	// The script is only intended to be used with resources of kind ServiceMonitor
```

hack/monitorgen/main.go
```diff
@@ -0,0 +1,173 @@
...
+	if m.Kind != "ServiceMonitor" {
+		return
```
jan--f, line 71: Can this be reached after creating the `ServiceMonitor` object above? Is this is needed a log message might be good?

hack/monitorgen/main.go
```diff
@@ -0,0 +1,173 @@
...
+	if *pathMetricsFile != "" {
+		metricsToKeep, err = readMetricsFromFile(*pathMetricsFile)
+		if err != nil {
+			return
+		}
+	}
+
+	if len(metricsToKeep) == 0 {
+		return
+	}
```
jan--f, line 93: Shouldn't this part be in `generateCollectionProfileMinimal`? When we add another profile, we'd want to add a function `generateCollectionProfileNewProfile` to main no?

hack/monitorgen/main.go
```diff
@@ -0,0 +1,173 @@
...
+		if len(endpoint.MetricRelabelConfigs) > 0 {
+			endpoint.MetricRelabelConfigs = removeDropMetricRelableConfigs(endpoint.MetricRelabelConfigs)
+		}
+		// Add keep relabel configs
+		endpoint.MetricRelabelConfigs = append(endpoint.MetricRelabelConfigs, buildRelabelConfig(metricsToKeep))
+	}
+
+	yamlData, err := yaml.Marshal(mMinimal)
+	if err != nil {
+		log.Fatalf("failed marshling monitor: %e", err)
```
jan--f, line 118:
```suggestion
		log.Fatalf("failed marsahling monitor: %e", err)
```

hack/monitorgen/main.go
```diff
@@ -0,0 +1,173 @@
...
+		SourceLabels: []monitoringv1.LabelName{
+			"__name__",
+		},
+		Regex: fmt.Sprintf("(%s)", jointMetrics),
+	}
+}
+
+// builfFileName takes a file name like metrics.yaml and add "-minimal" between
+// "." and "yaml" resulting in "metrics-minimal.yaml"
+func builfFileName(fileName string) string {
```
jan--f, line 162:
```suggestion
func buildFileName(fileName string) string {
```

jan--f (commented):
Thanks, I like where this is going.

I think the `monitorgen` interface could maybe be improved. Right now we have a hard coded list of metrics, which can be overridden with a file passed on the cli.
This doesn't extend nicely when we think about adding additional profiles. The hard-coded part is clear, however `metrics-file` less so.

We could forego the hardcoded metrics part and require the user to pass a file with metrics. Keeping future profiles in mind however, this file would require some structure to have a list of metrics to keep per profile. we would have a set of files with the default lists, similar to the files in `metrics/`.

Another option would be to scope the `metrics-file` option per profile, so rename it `minimal-metrics-file`. That allows us to add addition profiles with the same cli arg set.

It might also be worth it to discuss how we'll update the keep lists in the future. That might inform how we generate the ServiceMonitors.

hack/monitorgen/main.go
JoaoBraveCoding: Reworked this part of the code

hack/monitorgen/main.go
JoaoBraveCoding: Yes, I noticed that indeed can, I can add the log message

PR #2030, opened 2023-07-02
MON-3231: add metric for the configured profile

rexagod:
- adds a metric with the label value set to the current collection profile.


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+	Help:           "Latest reconciliation state. Set to 1 if last reconciliation succeeded, else 0.",
+	StabilityLevel: metrics.ALPHA,
+})
+
+// ConfiguredCollectionProfile is a gauge that indicates the configured collection profile.
+var ConfiguredCollectionProfile = metrics.NewGaugeVec(&metrics.GaugeOpts{
+	Name:           "cluster_monitoring_operator_configured_collection_profile",
+	Help:           "The configured collection profile",
+	StabilityLevel: metrics.ALPHA,
+	// TODO: Use LabelValueAllowLists, once GA'd.
```
simonpasquier, line 27: Can you expand a bit? I'm not sure to know to which component "once GA'd" applies :)

simonpasquier (commented):
Since collection profiles are gated via TechPreview, I would only expose the metric if tech preview is explicitly enabled.

I'd also recommend exposing all collection profile values, e.g.

```
cluster_monitoring_operator_collection_profile{..., collection_profile="full"} 1
cluster_monitoring_operator_collection_profile{..., collection_profile="minimal"} 0
```

Having 0 for all series would mean that the collection profile is misconfigured.

pkg/manifests/config.go
```diff
@@ -215,14 +217,19 @@ func NewConfig(content io.Reader, tp bool) (*Config, error) {
 	// The operator should only create some manifests if techPreview is enabled
 	c.TechPreview = tp
 
+	cmometrics.ConfiguredCollectionProfile.WithLabelValues("").Set(0)
```
jan--f, line 220: Iiuc this is a no-op, since `CollectionProfile` will default to `full`: https://github.com/openshift/cluster-monitoring-operator/blob/ec2e0d3da7199f68c73f5a081ebf2a131d56f243/pkg/manifests/config.go#L303

jan--f (commented):
Generally this looks good to me. Can you please add separate commits for the lint fix, the metrics refactor and for adding the new metric? This will make future change archeology much easier.
Also this does not add the telemetry config yet, we could add the CMO side for this here too?

One thing I'd like to raise for discussion: Should we maybe add a `cluster_monitoring_operator_configuration` metric instead of single metrics for config options we're interested in? We're mostly interested in the label we set anyway and I think such a metadata metric can make it easier to add additional signals and send them to telemetry.

jan--f:
> One thing I'd like to raise for discussion: Should we maybe add a cluster_monitoring_operator_configuration metric instead of single metrics for config options we're interested in? We're mostly interested in the label we set anyway and I think such a metadata metric can make it easier to add additional signals and send them to telemetry.

Briefly discussed in [slack](https://redhat-internal.slack.com/archives/G79AW9Q7R/p1688373468751679), individual metrics have clear benefits.

pkg/metrics/metrics.go
rexagod: ACK, updated.

pkg/manifests/config.go
rexagod: Ah, right!

rexagod:
> Can you please add separate commits for the lint fix, the metrics refactor and for adding the new metric? This will make future change archeology much easier.

@jan--f Does it sound good if do this once the PR is LGTM'd and good to merge? That way I can keep adding any lint fixes or refactors during further reviews and split them all in the end.

Documentation/data-collection.md
```diff
@@ -584,6 +584,14 @@ data:
     # monitoring routes.
     - '{__name__="monitoring:haproxy_server_http_responses_total:sum"}'
     #
+    # owners: (@openshift/openshift-team-monitoring)
+    #
+    # cluster_monitoring_operator_configured_collection_profile contains information about the last configured
+    # collection profile.
+    # Possible label values are:
+    #   collection_profile= full|minimal (refer: cluster-monitoring-operator/pkg/manifests#SupportedCollectionProfiles)
+    - '{__name__="cluster_monitoring_operator_configured_collection_profile"}'
```
simonpasquier, line 593: You want to aggregate and record the metric to remove variable/unnecessary labels such as pod, job, service and instance. E.g. `max by(profile) (cluster_monitoring_operator_collection_profile)`

https://rhobs-handbook.netlify.app/products/openshiftmonitoring/telemetry.md/#configure-recording-rules

pkg/manifests/config.go
```diff
@@ -218,9 +220,23 @@ func NewConfig(content io.Reader, tp bool) (*Config, error) {
...
-		return nil, errors.Wrap(ErrConfigValidation, fmt.Sprintf(`%q is not supported, supported collection profiles are: %q`, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile, SupportedCollectionProfiles.String()))
+
+	// Validate the configured collection profile iff tech preview is enabled, even if the default profile is set.
+	if tp {
+		didMisconfigureCollectionProfile := false
+		if !slices.Contains(SupportedCollectionProfiles, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile) {
+			didMisconfigureCollectionProfile = true
+		}
+		for _, profile := range SupportedCollectionProfiles {
+			var v float64 = 0
```
simonpasquier, line 231: (nit)
```suggestion
			var v float64
```

pkg/manifests/config.go
```diff
@@ -218,9 +220,23 @@ func NewConfig(content io.Reader, tp bool) (*Config, error) {
...
+			if profile == c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile {
+				v = 1
+			}
+			cmometrics.ConfiguredCollectionProfile.WithLabelValues(string(profile)).Set(v)
+		}
+		if didMisconfigureCollectionProfile {
```
simonpasquier, line 237: (nit) we can get rid of the intermediate variable
```suggestion
		if !slices.Contains(SupportedCollectionProfiles, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile) {
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+// ConfiguredCollectionProfile is a gauge vector that indicates the configured collection profile.
+// All collection profiles, as stated in manifests.ConfiguredCollectionProfile, are present as label values at all times.
+// A metric value of 1 indicates expected behavior, whereas 0 indicates a misconfigured collection profile.
+var ConfiguredCollectionProfile = metrics.NewGaugeVec(&metrics.GaugeOpts{
+	Name:           "cluster_monitoring_operator_configured_collection_profile",
+	Help:           "The configured collection profile.",
+	StabilityLevel: metrics.ALPHA,
+	// TODO: Use LabelValueAllowLists, once GA'd: https://pkg.go.dev/k8s.io/component-base/metrics#MetricLabelAllowList.
+	// Refer (KEP-2305, currently beta): https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/2305-metrics-cardinality-enforcement#summary.
+}, []string{"last_configured_collection_profile"})
```
simonpasquier, line 31:
```suggestion
}, []string{"profile"})
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+
+// ConfiguredCollectionProfile is a gauge vector that indicates the configured collection profile.
+// All collection profiles, as stated in manifests.ConfiguredCollectionProfile, are present as label values at all times.
+// A metric value of 1 indicates expected behavior, whereas 0 indicates a misconfigured collection profile.
+var ConfiguredCollectionProfile = metrics.NewGaugeVec(&metrics.GaugeOpts{
+	Name:           "cluster_monitoring_operator_configured_collection_profile",
+	Help:           "The configured collection profile.",
+	StabilityLevel: metrics.ALPHA,
+	// TODO: Use LabelValueAllowLists, once GA'd: https://pkg.go.dev/k8s.io/component-base/metrics#MetricLabelAllowList.
+	// Refer (KEP-2305, currently beta): https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/2305-metrics-cardinality-enforcement#summary.
```
simonpasquier, line 30: I don't think that the dimension of the metric is unbounded.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	Name:           "cluster_monitoring_operator_last_reconciliation_successful",
+	Help:           "Latest reconciliation state. Set to 1 if last reconciliation succeeded, else 0.",
+	StabilityLevel: metrics.ALPHA,
+})
+
+// ConfiguredCollectionProfile is a gauge vector that indicates the configured collection profile.
+// All collection profiles, as stated in manifests.ConfiguredCollectionProfile, are present as label values at all times.
+// A metric value of 1 indicates expected behavior, whereas 0 indicates a misconfigured collection profile.
+var ConfiguredCollectionProfile = metrics.NewGaugeVec(&metrics.GaugeOpts{
+	Name:           "cluster_monitoring_operator_configured_collection_profile",
```
simonpasquier, line 26:
```suggestion
	Name:           "cluster_monitoring_operator_collection_profile",
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	Help:           "The configured collection profile.",
```
simonpasquier, line 27:
```suggestion
	Help:           "Information about collection profiles. Set to 1 for the configured collection profile, else 0.",
```

pkg/manifests/config.go
```diff
@@ -31,6 +31,8 @@ import (
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
 	"k8s.io/klog/v2"
+
+	cmometrics "github.com/openshift/cluster-monitoring-operator/pkg/metrics"
```
simonpasquier, line 35: (nit) do we need the import alias?

Documentation/data-collection.md
rexagod: Ah! Apologies, I totally missed that. Addressed.

pkg/metrics/metrics.go
rexagod: ACK, I thought about using this in a best practice sort of way, but it doesn't really help here, since it's guaranteed the cardinality will be bounded.

rexagod: I've removed the comments.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+// ReconcileStatus is a gauge that indicates the latest reconciliation state.
+var ReconcileStatus = metrics.NewGauge(&metrics.GaugeOpts{
+	Name:           "cluster_monitoring_operator_last_reconciliation_successful",
+	Help:           "Latest reconciliation state. Set to 1 if last reconciliation succeeded, else 0.",
+	StabilityLevel: metrics.ALPHA,
+})
+
+// ConfiguredCollectionProfile is a gauge vector that indicates the configured collection profile.
+// All collection profiles, as stated in manifests.ConfiguredCollectionProfile, are present as label values at all times.
+// A metric value of 1 indicates expected behavior, whereas 0 indicates a misconfigured collection profile.
```
jan--f, line 24: This comment needs an update to be consistent with the metric help.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+var ConfiguredCollectionProfile = metrics.NewGaugeVec(&metrics.GaugeOpts{
```
jan--f, line 25:
```suggestion
var CollectionProfile = metrics.NewGaugeVec(&metrics.GaugeOpts{
```
And update the comment above?

jan--f (changes requested):

jan--f:
/lgtm
/hold
for better commits :)

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -576,6 +576,14 @@ data:
     # monitoring routes.
     - '{__name__="monitoring:haproxy_server_http_responses_total:sum"}'
     #
+    # owners: (@openshift/openshift-team-monitoring)
+    #
+    # profile:cluster_monitoring_operator_collection_profile:max contains information about the configured
+    # collection profile. Set to 1 for the configured collection profile, else 0.
```
simonpasquier, line 582:
```suggestion
    # collection profile.

File: jsonnet/rules.libsonnet [hunk 14]
```diff
@@ -331,6 +331,10 @@ function(params) {
           // Returns 1 if all control plane nodes are ready 0 otherwise.
           // Should be used to suppress alerts during control plane upgrades or disruption.
         },
+        {
+          expr: 'max by (profile) (cluster_monitoring_operator_collection_profile)',
```
simonpasquier, line 335: We only need to report the selected profile.
```suggestion
          expr: 'max by (profile) (cluster_monitoring_operator_collection_profile == 1)',
```

simonpasquier (commented):
one last nit :)

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+})
+
+// CollectionProfile is a gauge vector that holds information about the collection profiles.
+// All collection profiles, as stated in manifests.SupportedCollectionProfiles, are present as label values.
+// Set to 1 for the configured collection profile, else 0.
+var CollectionProfile = metrics.NewGaugeVec(&metrics.GaugeOpts{
+	Name:           "cluster_monitoring_operator_collection_profile",
+	Help:           "Information about collection profiles. Set to 1 for the configured collection profile, else 0.",
+	StabilityLevel: metrics.ALPHA,
+}, []string{"profile"})
```
rexagod, line 29: This entire file came up as one single blob while [patching](https://github.com/tpope/vim-fugitive/blob/43f18ab9155c853a84ded560c6104e6300ad41da/doc/fugitive.txt#L313), and although this snippet comes under the metric addition, the file this is in, is itself a part of the refactor, so I included this here. PLMK if this doesn't make sense and I'll move this bit over to the latest commit.

rexagod: Let me break this from here and move it into the latest commit, it makes more sense there, I think.

PR #1931, opened 2023-03-31
MON-3127: add nodeExporter.collectors.ksmd settings.

raptorsun:
* [x] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #2032, opened 2023-07-04
MON-3269: add nodeExporter.collectors.processes settings.

raptorsun:

This PR is similar to #1931 , adding an on/off switch to `processes` collector.


jan--f:
/lgtm
Thanks Haoyu!

PR #2033, opened 2023-07-05
MON-3178: Expose and propagate TopologySpreadConstraints for prometheus-operator

marioferh:
Give users a TopologySpreadConstraints field in the prometheus-operator field and propagate this to the pod that is created.


jan--f (approved):
/lgtm

PR #2036, opened 2023-07-06
MON-3183:  Expose and propagate TopologySpreadConstraints for UWM prometheus

marioferh:
Give users a TopologySpreadConstraints field in the PrometheusRestrictedConfig field and propagate this to the pod that is created.


marioferh:
wrong branch

PR #2037, opened 2023-07-06
MON-3182: Expose and propagate TopologySpreadConstraints for UWM alertmanager 

marioferh:
Give users a TopologySpreadConstraints field in the AlertmanagerUserWorkloadConfig field and propagate this to the pod that is created.

### DIFF
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -2822,9 +2822,22 @@ ingress:
 
 }
 
-func TestAlertManagerUserWorkloadSecretsConfiguration(t *testing.T) {
+func TestAlertManagerUserWorkloadConfiguration(t *testing.T) {
 	c := NewDefaultConfig()
-	c.UserWorkloadConfiguration.Alertmanager.Secrets = []string{"test-secret", "slack-api-token"}
+	uwc, err := NewUserConfigFromString(`alertmanager:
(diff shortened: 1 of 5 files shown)


pkg/manifests/manifests_test.go
```diff
@@ -2822,9 +2822,22 @@ ingress:
 
 }
 
-func TestAlertManagerUserWorkloadSecretsConfiguration(t *testing.T) {
+func TestAlertManagerUserWorkloadConfiguration(t *testing.T) {
 	c := NewDefaultConfig()
-	c.UserWorkloadConfiguration.Alertmanager.Secrets = []string{"test-secret", "slack-api-token"}
+	uwc, err := NewUserConfigFromString(`alertmanager:
```
marioferh, line 2827: I think this is the correct way to check UWM.

pkg/manifests/manifests_test.go
```diff
@@ -2822,9 +2822,22 @@ ingress:
...
+  topologySpreadConstraints:
+  - maxSkew: 1
+    topologyKey: type
+    whenUnsatisfiable: DoNotSchedule
+    labelSelector:
+      matchLabels:
+        foo: bar
+  secrets:
+  - test-secret
```
marioferh, line 2836: Also added here previous test and change the test name in order to include all UWM test

PR #2039, opened 2023-07-11
MON-669: Remove etcd ServiceMonitors management code as they'll be no…

machine424:
…w managed by cluster-etcd-operator.

See https://github.com/openshift/cluster-etcd-operator/pull/1070


pkg/tasks/controlplane.go
```diff
@@ -77,45 +76,20 @@ func (t *ControlPlaneTask) Run(ctx context.Context) error {
 		}
 	}
 
-	sms, err = t.factory.ControlPlaneEtcdServiceMonitors()
+	// NOTE: This is temporary, to clean these resources that used to be managed by CMO, now
```
machine424, line 79: I'll create a task for this and assign it to me.

machine424: Maybe one also to move RBAC related resources as well https://github.com/openshift/cluster-monitoring-operator/blob/b886eb231ddf370b577ec4e572b1563d8889e28f/assets/prometheus-k8s/role-binding-specific-namespaces.yaml#L60-L78 (if possible)

machine424:
the migration is not atomic, there may be some transient phase where we have 0 etcd metrics (only this commit is taken into account) or duplicated etcd metrics (only the etcd commit is taken into account), but this should eventually work as expected (when the cluster updates the operators), of course I'm only talking about test clusters (that use master branches).
We'll make sure both commits are in 4.14.
(I don't think any code to ensure atomicity is worth it)

pkg/tasks/controlplane.go
jan--f: Does the etcd operator create this `RoleBinding` already? Our code does a noop if the RB already exists, so it should be safe to create them on either/both sides. Not really sure what the best practice is. CMO only creates RBs for SMs that it sets up.
So sounds like CEO would be the right place.

machine424: No, it doesn't create the `Role/RoleBinding` for the `openshift-etcd` namespace.
Yes, it makes more sense for me too to delegate this to CEO, I'll create the ticket then, maybe aim for 4.14 as well.

jan--f: sure, let's move it to CEO as soon as convenient.

jan--f:
Sorry...just a wee bit too late :(

PR #2043, opened 2023-07-12
Add telemeter client SA to rbac proxy

marioferh:
After change oauth proxy to rbac proxy https://github.com/openshift/cluster-monitoring-operator/pull/2005

When mTLS does not work and auth from telemeter-client and prometheus is done via token, telemter-client SA needs to be allowed.

### DIFF
--- a/assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
+++ b/assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
@@ -20,4 +20,9 @@ stringData:
         "user":
           "name": "system:serviceaccount:openshift-monitoring:prometheus-k8s"
         "verb": "get"
+      - "path": "/federate"
+        "resourceRequest": false
+        "user":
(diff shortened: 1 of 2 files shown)


assets/prometheus-k8s/kube-rbac-proxy-secret.yaml
```diff
@@ -20,4 +20,9 @@ stringData:
         "user":
           "name": "system:serviceaccount:openshift-monitoring:prometheus-k8s"
         "verb": "get"
+      - "path": "/federate"
+        "resourceRequest": false
+        "user":
+          "name": "system:serviceaccount:openshift-monitoring:telemeter-client"
```
marioferh, line 26: this error if telemter-client SA is not allowed:
`auth.go:97] Forbidden (user=system:serviceaccount:openshift-monitoring:telemeter-client, verb=get, resource=, subresource=). Reason: "".`

PR #2044, opened 2023-07-13
MON-3274: collect the number of LIST and WATCH requests to the apiserver from telemetry

p0lyn0mial:

adds `apiserver_list_watch_request_success_total:rate:sum` represents the rate of change for successful LIST and WATCH requests over a 5 minute period.

link to the issue https://issues.redhat.com/browse/MON-3274


jsonnet/rules.libsonnet
```diff
@@ -626,5 +626,14 @@ function(params) {
         },
       ],
     },
+    {
+      name: 'apiserver-list-watch.rules',
+      rules: [
+        {
+          expr: 'sum by(verb) (rate(apiserver_request_total{verb=~"LIST|WATCH",code=~"2.."}[5m]))',
+          record: 'apiserver_list_watch_request_total:rate:sum',
```
jan--f, line 634:
```suggestion
          record: 'apiserver_list_watch_request_success_total:rate:sum',
```
Since the new recording rule only covers `2xx` codes this name might be better?

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -994,6 +994,11 @@ data:
     #
     # cluster:vsphere_csi_migration:max shows if CSI migration for vSphere was enabled or not.
     - '{__name__="cluster:vsphere_csi_migration:max"}'
+    #
+    # owners: (@openshift/openshift-team-api, @polynomial)
+    #
+    # apiserver_list_watch_request_total:rate:sum represents the rate of change for the LIST and WATCH requests over a 5 minute period.
+    - '{__name__="apiserver_list_watch_request_total:rate:sum"}'
```
jan--f, line 1001:
```suggestion
    - '{__name__="apiserver_list_watch_request_success_total:rate:sum", verb=~"LIST|WATCH"}'
```
This seems redundant right now, but we want to keep adding these additional cardinality protections to our telemetry rules. This way the recording rule can changed in the future to include other verbs, without immediately affecting telemetry.

jan--f (changes requested):
Generally this looks good, I have only two small nits.
Please also note that for telemetry to be ingested the receiving end need a change as well: https://rhobs-handbook.netlify.app/products/openshiftmonitoring/telemetry.md/#synchronize-the-telemeter-servers-configuration
Its always worth linking the PRs.

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -994,6 +994,11 @@ data:
     #
     # cluster:vsphere_csi_migration:max shows if CSI migration for vSphere was enabled or not.
     - '{__name__="cluster:vsphere_csi_migration:max"}'
+    #
+    # owners: (@openshift/openshift-team-api, @polynomial)
+    #
+    # apiserver_list_watch_request_total:rate:sum represents the rate of change for the LIST and WATCH requests over a 5 minute period.
```
jan--f, line 1000:
```suggestion
    # apiserver_list_watch_request_success_total:rate:sum represents the rate of change for successful LIST and WATCH requests over a 5 minute period.
```

jan--f (approved):
/lgtm

PR #2022, opened 2023-06-27
MON-3211: Implement switching to metrics-server

slashpai:
As part of epic https://issues.redhat.com/browse/MON-3153

**Note:** Not all options of Prometheus Adapter are implemented in Metrics Server in this PR

**(I will add ChangeLog once everything finalized to avoid rebase issues)**

* [ ] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #2046, opened 2023-07-18
MON-3215: drop techpreview gate for collection profiles

rexagod:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Note: This CHANGELOG is only for the monitoring team to track all monitoring related changes. Please see OpenShift release notes for official changes.
 
+## 4.16
+
+- [#2046](https://github.com/openshift/cluster-monitoring-operator/pull/2046) Drop TechPreview gate for CollectionProfiles.
+
 ## 4.15
 
 - [#2022](https://github.com/openshift/cluster-monitoring-operator/pull/2022) Add support to switch to metrics server from prometheus-adapter when the `MetricsServer` feature gate is enabled.
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -428,22 +428,16 @@ func (c *Config) LoadEnforcedBodySizeLimit(pcr PodCapacityReader, ctx context.Co
 }
 
 func (c *Config) Precheck() error {
-	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile != FullCollectionProfile && !c.TechPreview {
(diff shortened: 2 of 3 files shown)


pkg/manifests/config.go
```diff
@@ -52,7 +52,6 @@ const (
 type Config struct {
 	Images      *Images `json:"-"`
 	RemoteWrite bool    `json:"-"`
-	TechPreview bool    `json:"-"`
```
slashpai, line 55: We have [metrics-server PR](https://github.com/openshift/cluster-monitoring-operator/pull/2022) also using same config so I think we can still keep this config and drop the check only for collection profile?

pkg/manifests/config.go
```diff
@@ -190,7 +189,7 @@ func (cps CollectionProfiles) String() string {
 	return sb.String()
 }
 
-func NewConfig(content io.Reader, tp bool) (*Config, error) {
+func NewConfig(content io.Reader) (*Config, error) {
```
slashpai, line 192: same comment above regarding metrics-server

pkg/manifests/config.go
```diff
@@ -458,14 +449,14 @@ func calculateBodySizeLimit(podCapacity int) string {
 // NewConfigFromString transforms a string containing configuration in the
 // openshift-monitoring/cluster-monitoring-configuration format into a data
 // structure that facilitates programmatical checks of that configuration. The
-// content of the data structure might change if TechPreview is enabled (tp), as
+// content of the data structure might change if TechPreview is enabled, as
 // some features are only meant for TechPreview.
-func NewConfigFromString(content string, tp bool) (*Config, error) {
```
slashpai, line 463: same comment above regarding metrics-server

simonpasquier:
I agree with @slashpai to keep the TP bool because it will be reusable in the future.

rexagod:
- [ ] Replace TP with FG.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Note: This CHANGELOG is only for the monitoring team to track all monitoring related changes. Please see OpenShift release notes for official changes.
 
+## 4.16
+
+- [#2046](https://github.com/openshift/cluster-monitoring-operator/pull/2046) Drop TechPreview gate for CollectionProfiles.
```
slashpai, line 5: IIUC we added this as a TechPreviewNoUpgrade FeatureGate so for this PR  there is no user change log?

rexagod: ACK, the TP was replaced by the newly introduced ways of using feature-gates in https://github.com/openshift/cluster-monitoring-operator/pull/2153. At the moment there is no changelog entry for this, but we might want to point out to users that this is available in TP behind a certain FG if they want to use it. WDYT?

PR #2047, opened 2023-07-20
OCPBUGS-17125: backport metrics collection profiles selector logic to prevent users from double scraping when they upgrade from 4.12 to 4.13

JoaoBraveCoding:
Part of https://github.com/openshift/enhancements/pull/1298

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -62,6 +62,8 @@ const (
 
 	platformAlertmanagerService     = "alertmanager-main"
 	userWorkloadAlertmanagerService = "alertmanager-user-workload"
+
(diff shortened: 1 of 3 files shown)


jan--f:
@JoaoBraveCoding Mind adding the[ test case](https://github.com/openshift/cluster-monitoring-operator/blob/91d735bd8662965037aae60c846c53baa79752ac/pkg/manifests/manifests_test.go#L1755-L1809) as well?

jan--f:
/retitle OCPBUGS-17125: backport metrics collection profiles selector logic to prevent users from double scraping when they upgrade from 4.12 to 4.13

PR #2048, opened 2023-07-25
MON-2645: remove unused rebalancer code

simonpasquier:
This commit removes the logic that allowed CMO to rebalance the stateful pods onto distinct nodes. It was required when upgrading to 4.8 which introduced hard anti-affinity rules. It is now useless and creates cognitive load for code maintainers.


marioferh (approved):
/lgtm

PR #2049, opened 2023-07-25
OCPBUGS-10699: remove Kube*QuotaOvercommit alerts

rexagod:
Quota alerts should not be defined for system namespaces.


simonpasquier (commented):
/lgtm

PR #2051, opened 2023-07-26
MON-2645: remove unnecessary RBAC permissions

simonpasquier:
This is a follow-up of #2048. This commit also removes the RBAC permissions that were needed by CMO to rebalance the pods.

It refactors the client code to remove the need for the list permission on pods.


machine424:
thanks
/lgtm

marioferh (approved):
/lgtm

PR #2055, opened 2023-07-31
MON-3291: Adjust node-exporter's MaxProcs documentation now that we s…

machine424:
…et a default value to GOMAXPROCS via https://issues.redhat.com//browse/OCPBUGS-13153


Documentation/api.md
```diff
@@ -380,7 +380,7 @@ The `NodeExporterConfig` resource defines settings for the `node-exporter` agent
...
+| maxProcs | uint32 | The target number of CPUs on which the Node Exporter's process will run. Use this to override the  default value which is set to the minimum of `4` and the number of CPU on the host. The default value is computed at runtime and set via the `GOMAXPROCS` environment variable before launching Node Exporter. If a kernel deadlock occurs or if performance degrades when reading from `sysfs` concurrently, you can change this value to `1`, which limits Node Exporter to running on one CPU. For nodes with a high CPU count, setting the limit to a low number saves resources by preventing Go routines from being scheduled to run on all CPUs. However, I/O performance degrades if the `maxProcs` value is set too low, and there are many metrics to collect. |
```
simonpasquier, line 383: > the minimum of `4` and the number of CPU on the host

The wording confused me a bit. Maybe `to 4 or the number of CPUs on the host, whichever is smaller.`

machine424: Done.
(Know that I thought about using min(4, xxx) at the beginning haha)

simonpasquier: haha, anyway let's wait for @bburt-rh inputs :)

Documentation/api.md
```diff
@@ -380,7 +380,7 @@ The `NodeExporterConfig` resource defines settings for the `node-exporter` agent
...
+| maxProcs | uint32 | The target number of CPUs on which the Node Exporter's process will run. Use this to override the default value which is set to 4 or the number of CPUs on the host, whichever is smaller. The default value is computed at runtime and set via the `GOMAXPROCS` environment variable before launching Node Exporter. If a kernel deadlock occurs or if performance degrades when reading from `sysfs` concurrently, you can change this value to `1`, which limits Node Exporter to running on one CPU. For nodes with a high CPU count, setting the limit to a low number saves resources by preventing Go routines from being scheduled to run on all CPUs. However, I/O performance degrades if the `maxProcs` value is set too low, and there are many metrics to collect. |
```
bburt-rh, line 383:
```suggestion
| maxProcs | uint32 | The target number of CPUs on which the Node Exporter's process will run. Use this setting to override the default value, which is set either to `4` or to the number of CPUs on the host, whichever is smaller. The default value is computed at runtime and set via the `GOMAXPROCS` environment variable before Node Exporter is launched. If a kernel deadlock occurs or if performance degrades when reading from `sysfs` concurrently, you can change this value to `1`, which limits Node Exporter to running on one CPU. For nodes with a high CPU count, setting the limit to a low number saves resources by preventing Go routines from being scheduled to run on all CPUs. However, I/O performance degrades if the `maxProcs` value is set too low, and there are many metrics to collect. |
```

Documentation/api.md
bburt-rh: @simonpasquier @machine424 - I added some suggested changes.

Documentation/api.md
machine424: done

PR #2057, opened 2023-08-01
OCPBUGS-17035: fix KRP permissions for Thanos Querier

rexagod:
fix KRP configuration for thanos-querier in order to enforce restricted
access only to entities allowed to [http-verb] for pods.metrics.k8s.io

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>

### DIFF
--- a/jsonnet/components/thanos-querier.libsonnet
+++ b/jsonnet/components/thanos-querier.libsonnet
@@ -144,7 +144,8 @@ function(params)
               },
             },
             resourceAttributes: {
-              apiVersion: 'metrics.k8s.io/v1beta1',
+              apiVersion: 'v1beta1',
+              apiGroup: 'metrics.k8s.io',
               resource: 'pods',
               namespace: '{{ .Value }}',
             },
--- a/test/e2e/user_workload_monitoring_test.go
+++ b/test/e2e/user_workload_monitoring_test.go
@@ -26,15 +26,17 @@ import (
 	"time"
 
 	"github.com/Jeffail/gabs"
-	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
 	"github.com/pkg/errors"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/cert"
+
+	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
 )
 
 type scenario struct {
@@ -525,16 +527,10 @@ func assertTenancyForMetrics(t *testing.T) {
 
 	err := framework.Poll(2*time.Second, 10*time.Second, func() error {
 		_, err := f.CreateServiceAccount(userWorkloadTestNs, testAccount)
-		return err
-	})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	// Grant enough permissions to the account so it can read metrics.
-	err = framework.Poll(2*time.Second, 10*time.Second, func() error {
-		_, err = f.CreateRoleBindingFromClusterRole(userWorkloadTestNs, testAccount, "admin")
-		return err
+		if !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+		return nil
 	})
 	if err != nil {
 		t.Fatal(err)
@@ -575,6 +571,32 @@ func assertTenancyForMetrics(t *testing.T) {
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Logf("Running query %q", tc.query)
+
+			var cleanupFn func() error
+			// Grant just-enough permissions to the account so it can read metrics.
+			err = framework.Poll(2*time.Second, 10*time.Second, func() error {
+				cleanupFn, err = f.CreateRoleBindingFromTypedRole(userWorkloadTestNs, testAccount, &rbacv1.Role{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "tenancy-test-metrics",
+					},
+					Rules: []rbacv1.PolicyRule{
+						{
+							APIGroups: []string{"metrics.k8s.io"},
+							Resources: []string{"pods"},
+							Verbs:     []string{"get"},
+						},
+					},
+				})
+				return err
+			})
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer func() {
+				if err := cleanupFn(); err != nil {
+					t.Fatal(err)
+				}
+			}()
 
 			err = framework.Poll(5*time.Second, time.Minute, func() error {
 				// The tenancy port (9092) is only exposed in-cluster so we need to use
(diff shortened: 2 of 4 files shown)


jsonnet/components/thanos-querier.libsonnet
```diff
@@ -144,7 +144,7 @@ function(params)
               },
             },
             resourceAttributes: {
-              apiVersion: 'metrics.k8s.io/v1beta1',
+              apiVersion: 'v1',
```
rexagod, line 147: Verified that `apiVersion` is a supported field: https://github.com/brancz/kube-rbac-proxy/blob/master/examples/resource-attributes/README.md?plain=1#L74.

jsonnet/components/thanos-querier.libsonnet
```diff
@@ -144,7 +149,7 @@ function(params)
...
```
simonpasquier, line 152: IIUC the API group needs to metrics.k8s.io and the API version is [v1beta1](https://github.com/kubernetes/metrics/tree/master/pkg/apis/metrics/v1beta1)
```suggestion
              apiVersion: 'v1beta1',
              apiGroup: 'metrics.k8s.io',
```

jsonnet/components/thanos-querier.libsonnet
```diff
@@ -70,6 +70,11 @@ function(params)
           resources: ['subjectaccessreviews'],
           verbs: ['create'],
         },
+        {
+          apiGroups: [''],
+          resources: ['pods'],
+          verbs: ['get', 'create'],
+        },
```
simonpasquier, line 77: I don't think that there's a need to add a new RBAC permission to the thanos-querier SA.

simonpasquier (commented):
We need to update the [assertTenancyForMetrics](https://github.com/openshift/cluster-monitoring-operator/blob/2662e6f961562d5eac447abd94d38b8735b90eaa/test/e2e/user_workload_monitoring_test.go#L523) e2e test.
Right now it creates a test service account to verify that a user can query metrics from their namespace but it [grants](https://github.com/openshift/cluster-monitoring-operator/blob/2662e6f961562d5eac447abd94d38b8735b90eaa/test/e2e/user_workload_monitoring_test.go#L536) the "admin" role which is too permissive. I'd advise to create a role with just-enough permissions allowing to get pod metrics.

jsonnet/components/thanos-querier.libsonnet
rexagod: I set the `apiVersion` to `v1` considering the `resource` (in the line just beneath this) is set to `pod`s, which is available in `core/v1`, am I missing something?

jsonnet/components/thanos-querier.libsonnet
rexagod: Ah! Right, makes sense!

jsonnet/components/thanos-querier.libsonnet
rexagod: I went through the KRP and `SubjectAccessReview` documentation (and types) figuring this out because for some reason I kept associating `pods` to `core/v1`. 🤦🏼

rexagod: Now that I've gained common sense, do you think it'd make sense to add a `sub resource: metrics` field in here as well?

test/e2e/user_workload_monitoring_test.go
```diff
@@ -533,7 +535,18 @@ func assertTenancyForMetrics(t *testing.T) {
 
 	// Grant enough permissions to the account so it can read metrics.
```
simonpasquier, line 536: (nit)
```suggestion
	// Grant just-enough permissions to the account so it can read metrics.
```

test/e2e/user_workload_monitoring_test.go
```diff
@@ -533,7 +535,18 @@ func assertTenancyForMetrics(t *testing.T) {
...
 	err = framework.Poll(2*time.Second, 10*time.Second, func() error {
-		_, err = f.CreateRoleBindingFromClusterRole(userWorkloadTestNs, testAccount, "admin")
+		_, err = f.CreateRoleBindingFromTypedClusterRole(userWorkloadTestNs, testAccount, &rbacv1.ClusterRole{
```
simonpasquier, line 538: can we create a Role rather than a ClusterRole?

test/e2e/user_workload_monitoring_test.go
```diff
@@ -531,9 +533,20 @@ func assertTenancyForMetrics(t *testing.T) {
...
-		_, err = f.CreateRoleBindingFromClusterRole(userWorkloadTestNs, testAccount, "admin")
+		_, err = f.CreateRoleBindingFromTypedRole(userWorkloadTestNs, testAccount, &rbacv1.Role{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "tenancy-test-metrics",
+			},
+			Rules: []rbacv1.PolicyRule{
+				{
+					APIGroups: []string{"metrics.k8s.io"},
+					Resources: []string{"pods"},
+					Verbs:     []string{"get"},
```
simonpasquier, line 546: we can also add new test scenarios like
* POST queries denied for roles without 'create' verb.
* POST queries allowed for roles with 'create' verb.

rexagod:
Cluster creations are failing for cluster-bot and openshift-installer for the past ~5 hours. I eventually pushed here to see if the CI can verify the newly added tests but that seems to be failing cluster creations as well.

test/e2e/user_workload_monitoring_test.go
```diff
@@ -576,6 +573,32 @@ func assertTenancyForMetrics(t *testing.T) {
...
+			var cleanupFn func() error
+			// Grant just-enough permissions to the account so it can read metrics.
+			err = framework.Poll(2*time.Second, 10*time.Second, func() error {
+				cleanupFn, err = f.CreateRoleBindingFromTypedRole(userWorkloadTestNs, testAccount, &rbacv1.Role{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "tenancy-test-metrics",
+					},
+					Rules: []rbacv1.PolicyRule{
+						{
+							APIGroups: []string{""},
```
simonpasquier, line 585: The APIGroup should be "metrics.k8s.io".

rexagod: I went with that initially, but it seems that the test fails when that's the case, I've stated my take at why this happens in https://github.com/openshift/cluster-monitoring-operator/pull/2057#issuecomment-1734984657. WDYT?

rexagod: Ah, my bad I thought you meant this for https://github.com/openshift/cluster-monitoring-operator/pull/2057/commits/42172986ae8bace69606e4a05ed47033c9c2aadc#diff-99cf99acf538c21d66511836754b4422d6f67d715ae4f47c6c05ebc78c0b1cbfR739.

jsonnet/components/thanos-querier.libsonnet
jan--f: Is this actually resolved? It's still `apiVersion: 'v1'` and missing the `apiGroup` value..?

rexagod:
Tested for `GET` and `POST` requests with a `ServiceAccount` linked to a `ClusterRole` having `get` and `create` verb permissions on `pods.metrics.k8s.io`. PTAL below for details.

<details>
<summary>Details</summary>

> _[From the ticket]_ [...] and thus I want to only add the minimal necessary permissions [...]

I used the same `ClusterRole` as we use for Thanos Querier's KRP container, hence the additional `*reviews` permissions, but only `pods.metrics.k8s.io` is needed to query the endpoint in question as shown below.

#### `-X GET`

Screenshot 2023-11-12 at 00 21 47

#### `-X POST`

Screenshot 2023-11-12 at 00 24 23

</details>

test/e2e/user_workload_monitoring_test.go
```diff
@@ -681,6 +703,123 @@ func assertTenancyForMetrics(t *testing.T) {
 	if err != nil {
 		t.Fatalf("the account has access to the rules endpoint of Thanos querier: %v", err)
 	}
+
+	for _, tc := range []struct {
+		// NOTE: role verbs need to factor in the permissions metrics-server requires for the respective use-case.
```
jan--f, line 708: Not sure I follow the meaning here? Is this still up-to-date or am I missing something? If the later may a bit more context might be good?

rexagod: Ah, this seems to be an artefact from earlier patches on this PR, not relevant now. Dropped.

jan--f:
Thanks! To me this looks good.
/lgtm
Just putting a 
/hold
for @simonpasquier to get a chance to look at this.

simonpasquier (commented):
/lgtm

PR #2059, opened 2023-08-02
Remove deprecated visual_web_terminal_sessions metric

mark-nc:
* [ x ] No user facing changes, so no entry in CHANGELOG was needed.


mark-nc:
The Visual Web Terminal component of Red Hat Advanced Cluster Management for Kubernetes was deprecated in RHACM 2.3 and removed from RHACM 2.4 which is now EOL.  Thus, this metric is no longer being reported and no longer is needed in the allowlist.

mark-nc:
All tests have passed.  @marioferh @sthaha it is ready to review.

marioferh (commented):
/lgtm

PR #2060, opened 2023-08-03
MON-3093: fix single node replica issue

marioferh:
While running the e2e test locally with Hypershift cluster from cluster-bot I noticed that it fails on step waiting for 2 prometheus instances.

“wait for prometheus-k8s: expected 2 Prometheus instances but got: 1: timed out waiting for the condition” 

Since Hypershift clusters from cluster-bot are single worker node

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


marioferh:
@simonpasquier thanks, much better :)

test/e2e/main_test.go
```diff
@@ -94,10 +95,16 @@ func testMain(m *testing.M) error {
 			return false, nil
 		}
 
-		if v != 2 {
-			loopErr = fmt.Errorf("expected 2 Prometheus instances but got: %v", v)
+		i, _ := f.OperatorClient.GetInfrastructure(ctx, "cluster")
```
simonpasquier, line 98: you still want to handle the error?

marioferh: yes, maybe v = 1 but its not single node then we want to catch the error

raptorsun: better check the error, or 0 == 0 may happen ;)

danielmellado: +1 with Haoyu's comments :D

marioferh: fixed

simonpasquier (commented):
/lgtm

PR #2062, opened 2023-08-06
MON-3302: add RHACS telemetry metrics

stehessel:

I didn't see any other Telemeter metrics in the changelog, let me know if you prefer an entry for it. Related Jira ticket: https://issues.redhat.com/browse/MON-3302

As mentioned in the ticket, if possible we would like to backport this to existing OpenShift versions so we don't have to wait until customers upgrade to OpenShift 4.14.


simonpasquier (commented):
Can rox_sensor_secured_nodes ever be different from the number of nodes in a cluster? If not, the nodes capacity is already reported in Telemetry via `cluster:node_instance_type_count:sum` . And the same goes for cpu capacity with `cluster:capacity_cpu_cores:sum`.

Instead of 2 metrics per sensor, you could have a single `rhacs:telemetry:rox_sensor_info` metric with the following labels
* `central_id`
* `hosting`
* `install_method`
* `build`
* `sensor_version`
* `sensor_id`

Reporting the number of nodes/vcpus per sensor would require a few PromQL joins but nothing too fancy IMHO.

The same remark goes for the central metrics, IIUC:
* `rox_central_secured_clusters` is equal to the sum of clusters running a sensor.
*  `rox_central_secured_nodes` is equal to the sum of the nodes reported by all the sensors.
* `rox_central_secured_vcpu` is equal to the sum of the vcpus reported by all the sensors.

If so, reporting 1 info metric from the sensor with (sensor_id, central_id) + 1 info metric from the central with (central_id) is enough to provide the same information.

stehessel:
So I talked with our release team and we can likely still squeeze in the change to the recording rules. If you have time, feel free to take a look at https://github.com/stackrox/stackrox/pull/7673.

simonpasquier:
> How long does the backporting process usually take?

For OCP, it's usually a matter of 1-2 weeks per minor release.

> How critical are above suggestions (replacing sensor node + vcpu metric with info metric; remove branding from recording rule) for you? Do they justify waiting another release cycle vs just going with what we have now?

From experience, a temporary solution lasts forever. I'd very much prefer to have a clean implementation from the start.

simonpasquier:
> For Central it's not the same, ...

Thanks, it clarifies a lot.

simonpasquier:
> as long as cluster:capacity_cpu_cores:sum is in K8s cpu units and not physical core in case of hyperthreading

`cluster:capacity_cpu_cores:sum` is computed from the `kube_node_status_capacity` metric which comes from the Kubernetes node status subresource. IIUC this is exactly the same source of information than RHACS:

https://github.com/stackrox/stackrox/blob/54e2a4a2204f55171ce334357a181dc52e85c961/sensor/kubernetes/clustermetrics/cluster_metrics.go#L134-L140

stehessel:
@simonpasquier I removed the sensor gauge metrics and the branding label. Can you take another look? Do these test failures need to be debugged? I'm not sure how they are related to my changes.

simonpasquier:
Will do. The e2e test failures are flakes.

Documentation/data-collection.md
```diff
@@ -991,6 +991,30 @@ data:
     #
     # apiserver_list_watch_request_success_total:rate:sum represents the rate of change for successful LIST and WATCH requests over a 5 minute period.
     - '{__name__="apiserver_list_watch_request_success_total:rate:sum", verb=~"LIST|WATCH"}'
+    #
+    # owners: (https://github.com/stackrox/stackrox, @stackrox/eng)
+    #
+    # rhacs:telemetry:rox_central_secured_clusters provides the number of clusters secured by a Central instance of a
+    # Red Hat Advanced Cluster Security installation.
+    - '{__name__="rhacs:telemetry:rox_central_secured_clusters"}'
```
simonpasquier, line 999: (suggestion) Looking at the `rhacs:telemetry:rox_central_secured_clusters` expression:

```
max by (build, central_id, central_version, hosting, install_method) (                                                                                                      
  rox_central_secured_clusters{branding=~"rhacs|RHACS"}                                                                                                                     
)
```

I'd recommend to normalize the branding value in the recorded metric, either when producing the metric (e.g. always upper-case the label value) or in the rule expression using label_replace().

Documentation/data-collection.md
```diff
@@ -991,6 +991,30 @@ data:
     #
     # apiserver_list_watch_request_success_total:rate:sum represents the rate of change for successful LIST and WATCH requests over a 5 minute period.
     - '{__name__="apiserver_list_watch_request_success_total:rate:sum", verb=~"LIST|WATCH"}'
+    #
+    # owners: (https://github.com/stackrox/stackrox, @stackrox/eng)
+    #
+    # rhacs:telemetry:rox_central_secured_clusters provides the number of clusters secured by a Central instance of a
+    # Red Hat Advanced Cluster Security installation.
```
simonpasquier, line 998: For future readers, I'd expand a bit on the labels of the metric.

IIUC
* `build`, "release" or "internal".
* `central_id` , unique ID identifying the central cluster.
* `central_version`, the product's full version.
* `hosting`, "cloud-service" or "self-managed".
* `install_method`, "operator", "manifest" or "helm".

Documentation/data-collection.md
```diff
@@ -991,6 +991,30 @@ data:
...
+    #
+    # rhacs:telemetry:rox_central_secured_nodes provides the number of nodes secured by a Central instance of a
+    # Red Hat Advanced Cluster Security installation.
+    - '{__name__="rhacs:telemetry:rox_central_secured_nodes"}'
+    #
+    # owners: (https://github.com/stackrox/stackrox, @stackrox/eng)
+    #
+    # rhacs:telemetry:rox_central_secured_vcpu provides the number of vCPUs secured by a Central instance of a
+    # Red Hat Advanced Cluster Security installation.
+    - '{__name__="rhacs:telemetry:rox_central_secured_vcpu"}'
```
simonpasquier, line 1011: (nit) `rhacs:telemetry:rox_central_secured_vcpus` would read better

Documentation/data-collection.md
```diff
@@ -991,6 +991,30 @@ data:
...
+    #
+    # owners: (https://github.com/stackrox/stackrox, @stackrox/eng)
+    #
+    # rhacs:telemetry:rox_sensor_info provides information about a Sensor instance of a Red Hat Advanced
+    # Cluster Security installation.
+    - '{__name__="rhacs:telemetry:rox_sensor_info"}'
```
simonpasquier, line 1017: Again looking at the recording expression:

```
max by (build, central_id, hosting, install_method, sensor_id, sensor_version) (
  rox_sensor_secured_nodes{branding=~"rhacs|RHACS"} > bool 0
)
```

Not sure why "> bool 0" is needed. The assumption for an info metric is that it always reports 1.

```
group by (build, central_id, hosting, install_method, sensor_id, sensor_version) (
  rox_sensor_secured_nodes{branding=~"rhacs|RHACS"}
)
```

simonpasquier (commented):
I'd recommend to add a `rhacs:telemetry:rox_central_info` metric similar to `rhacs:telemetry:rox_sensor_info` and remove the labels other than `central_id` from the `rhacs:telemetry:rox_central_secured_*` metrics.

stehessel:
Hi @simonpasquier , I followed your suggestions. Can you please take another look?

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -983,6 +983,55 @@ data:
     #
     # apiserver_list_watch_request_success_total:rate:sum represents the rate of change for successful LIST and WATCH requests over a 5 minute period.
     - '{__name__="apiserver_list_watch_request_success_total:rate:sum", verb=~"LIST|WATCH"}'
+    #
+    # owners: (https://github.com/stackrox/stackrox, @stackrox/eng)
+    #
+    # rhacs:telemetry:rox_central_info provides information about a Central instance of a Red Hat Advanced
+    # Red Hat Advanced Cluster Security installation.
```
simonpasquier, line 990: (nit)
```suggestion
    # rhacs:telemetry:rox_central_info provides information about a Central instance of a Red Hat Advanced
    # Cluster Security installation.
```

simonpasquier (commented):
Thanks for the updated version! I only found one nit in a comment otherwise I'll defer to @jan--f for a second check.

Having said that, I'm still not sure to understand/agree with the way monitoring is enabled and the fact that service monitors and prometheus rules are dropped in the openshift-monitoring namespaces. In particular for operator-based installations, you would benefit from using the `operatorframework.io/cluster-monitoring=true` annotation as described in https://github.com/openshift/enhancements/blob/master/enhancements/olm/olm-managed-operator-metrics.md#openshift-monitoring-prometheus-operator-support.

stehessel:
> Having said that, I'm still not sure to understand/agree with the way monitoring is enabled and the fact that service monitors and prometheus rules are dropped in the openshift-monitoring namespaces. In particular for operator-based installations, you would benefit from using the operatorframework.io/cluster-monitoring=true annotation as described in https://github.com/openshift/enhancements/blob/master/enhancements/olm/olm-managed-operator-metrics.md#openshift-monitoring-prometheus-operator-support.

The reason behind this implementation is that ACS supports multiple installation methods (manifest, Helm, OLM operator), and we wanted the monitoring to be consistent for all of them.

stehessel:
@jan--f I don't think this repo has the `/lgtm` automation, so you need to click `Approve` manually.

@simonpasquier Is there anything else holding up the merge now that @jan--f has approved?

jan--f:
@stehessel this is just prow doing its thing. The PR should merge once the bot is satisfied.

PR #2066, opened 2023-08-07
OCPBUGS-17346: Avoid recreating some resources, created by prometheus-operator, during 4.13->4.14 upgrade

machine424:
The option --labels was set to the prometheus-operator in https://github.com/openshift/cluster-monitoring-operator/pull/1986
to make it add the label "app.kubernetes.io/part-of: openshift-monitoring" to resources it creates.

prometheus-operator adds the labels to some controllers matchLabels as well (to the prometheus statefulset e.g.)
which makes the prometheus-operator recreate the statefulset as matchLabels is an immutable field.

Recreating the statefulset leads to downtime as the prometheus-operator run the deletion in a foreground mode, check
https://github.com/prometheus-operator/prometheus-operator/issues/3875 for the whys.

We consider that adding the ownership labels is not worth the downtime, and we'll be thinking of a way to avoid it.

Of course, this will lead to the recreation of these resources if https://github.com/openshift/cluster-monitoring-operator/pull/1986

### DIFF
--- a/assets/prometheus-operator-user-workload/deployment.yaml
+++ b/assets/prometheus-operator-user-workload/deployment.yaml
@@ -40,7 +40,6 @@ spec:
         - --config-reloader-cpu-request=1m
         - --config-reloader-memory-request=10Mi
         - --web.listen-address=127.0.0.1:8080
-        - --labels=app.kubernetes.io/part-of=openshift-monitoring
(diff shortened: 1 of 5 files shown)


simonpasquier (commented):
/lgtm

PR #2067, opened 2023-08-07
MON-3304: Add option to specify resource limits for all components

rexagod:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [#1950](https://github.com/openshift/cluster-monitoring-operator/pull/1950) Disable CORS headers on Thanos querier by default and add a flag to enable them back.
 - [#1963](https://github.com/openshift/cluster-monitoring-operator/pull/1963) Add nodeExporter settings for network devices list.
 - [#2049](https://github.com/openshift/cluster-monitoring-operator/pull/2049) Remove Kube*QuotaOvercommit alerts.
+- [#2067](https://github.com/openshift/cluster-monitoring-operator/pull/2067) Add options to specify resource requests and limits for all components.
 
 ## 4.13
 
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -39,6 +39,7 @@ Configuring Cluster Monitoring is optional. If the config does not exist or is e
 * [NodeExporterConfig](#nodeexporterconfig)
 * [OpenShiftStateMetricsConfig](#openshiftstatemetricsconfig)
 * [PrometheusK8sConfig](#prometheusk8sconfig)
+* [PrometheusOperatorAdmissionWebhookConfig](#prometheusoperatoradmissionwebhookconfig)
 * [PrometheusOperatorConfig](#prometheusoperatorconfig)
 * [PrometheusRestrictedConfig](#prometheusrestrictedconfig)
 * [RemoteWriteSpec](#remotewritespec)
@@ -131,6 +132,7 @@ The `ClusterMonitoringConfiguration` resource defines settings that customize th
 | kubeStateMetrics | *[KubeStateMetricsConfig](#kubestatemetricsconfig) | `KubeStateMetricsConfig` defines settings for the `kube-state-metrics` agent. |
 | prometheusK8s | *[PrometheusK8sConfig](#prometheusk8sconfig) | `PrometheusK8sConfig` defines settings for the Prometheus component. |
 | prometheusOperator | *[PrometheusOperatorConfig](#prometheusoperatorconfig) | `PrometheusOperatorConfig` defines settings for the Prometheus Operator component. |
+| prometheusOperatorAdmissionWebhook | *[PrometheusOperatorAdmissionWebhookConfig](#prometheusoperatoradmissionwebhookconfig) | `PrometheusOperatorAdmissionWebhookConfig` defines settings for the Prometheus Operator's admission webhook component. |
 | openshiftStateMetrics | *[OpenShiftStateMetricsConfig](#openshiftstatemetricsconfig) | `OpenShiftMetricsConfig` defines settings for the `openshift-state-metrics` agent. |
 | telemeterClient | *[TelemeterClientConfig](#telemeterclientconfig) | `TelemeterClientConfig` defines settings for the Telemeter Client component. |
 | thanosQuerier | *[ThanosQuerierConfig](#thanosquerierconfig) | `ThanosQuerierConfig` defines settings for the Thanos Querier component. |
@@ -167,6 +169,7 @@ The `K8sPrometheusAdapter` resource defines settings for the Prometheus Adapter 
 | -------- | ---- | ----------- |
 | audit | *Audit | Defines the audit configuration used by the Prometheus Adapter instance. Possible profile values are: `metadata`, `request`, `requestresponse`, and `none`. The default value is `metadata`. |
 | nodeSelector | map[string]string | Defines the nodes on which the pods are scheduled. |
+| resources | *[v1.ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#resourcerequirements-v1-core) | Defines resource requests and limits for the PrometheusAdapter container. |
 | tolerations | [][v1.Toleration](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#toleration-v1-core) | Defines tolerations for the pods. |
 | topologySpreadConstraints | []v1.TopologySpreadConstraint | Defines a pod's topology spread constraints. |
 | dedicatedServiceMonitors | *[DedicatedServiceMonitors](#dedicatedservicemonitors) | Defines dedicated service monitors. |
@@ -185,6 +188,7 @@ The `KubeStateMetricsConfig` resource defines settings for the `kube-state-metri
 | Property | Type | Description |
 | -------- | ---- | ----------- |
 | nodeSelector | map[string]string | Defines the nodes on which the pods are scheduled. |
+| resources | *[v1.ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#resourcerequirements-v1-core) | Defines resource requests and limits for the KubeStateMetrics container. |
 | tolerations | [][v1.Toleration](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#toleration-v1-core) | Defines tolerations for the pods. |
 | topologySpreadConstraints | []v1.TopologySpreadConstraint | Defines a pod's topology spread constraints. |
 
@@ -382,6 +386,7 @@ The `NodeExporterConfig` resource defines settings for the `node-exporter` agent
 | collectors | [NodeExporterCollectorConfig](#nodeexportercollectorconfig) | Defines which collectors are enabled and their additional configuration parameters. |
 | maxProcs | uint32 | The target number of CPUs on which the Node Exporter's process will run. Use this setting to override the default value, which is set either to `4` or to the number of CPUs on the host, whichever is smaller. The default value is computed at runtime and set via the `GOMAXPROCS` environment variable before Node Exporter is launched. If a kernel deadlock occurs or if performance degrades when reading from `sysfs` concurrently, you can change this value to `1`, which limits Node Exporter to running on one CPU. For nodes with a high CPU count, setting the limit to a low number saves resources by preventing Go routines from being scheduled to run on all CPUs. However, I/O performance degrades if the `maxProcs` value is set too low, and there are many metrics to collect. |
 | ignoredNetworkDevices | *[]string | A list of network devices, as regular expressions, to be excluded from the relevant collector configuration such as `netdev` and `netclass`. When not set, the Cluster Monitoring Operator uses a predefined list of devices to be excluded to minimize the impact on memory usage. When set as an empty list, no devices are excluded. If you modify this setting, monitor the `prometheus-k8s` deployment closely for excessive memory usage. |
+| resources | *[v1.ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#resourcerequirements-v1-core) | Defines resource requests and limits for the NodeExporter container. |
 
 [Back to TOC](#table-of-contents)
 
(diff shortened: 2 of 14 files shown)


pkg/manifests/types.go
```diff
@@ -49,6 +49,8 @@ type ClusterMonitoringConfiguration struct {
 	PrometheusK8sConfig *PrometheusK8sConfig `json:"prometheusK8s,omitempty"`
 	// `PrometheusOperatorConfig` defines settings for the Prometheus Operator component.
 	PrometheusOperatorConfig *PrometheusOperatorConfig `json:"prometheusOperator,omitempty"`
+	// `PrometheusOperatorConfig` defines settings for the Prometheus Operator's admission webhook component.
```
juzhao, line 52:
```suggestion
	// `PrometheusOperatorAdmissionWebhookConfig` defines settings for the Prometheus Operator's admission webhook component.
```

Documentation/api.md
```diff
@@ -131,6 +132,7 @@ The `ClusterMonitoringConfiguration` resource defines settings that customize th
 | kubeStateMetrics | *[KubeStateMetricsConfig](#kubestatemetricsconfig) | `KubeStateMetricsConfig` defines settings for the `kube-state-metrics` agent. |
 | prometheusK8s | *[PrometheusK8sConfig](#prometheusk8sconfig) | `PrometheusK8sConfig` defines settings for the Prometheus component. |
 | prometheusOperator | *[PrometheusOperatorConfig](#prometheusoperatorconfig) | `PrometheusOperatorConfig` defines settings for the Prometheus Operator component. |
+| prometheusOperatorAdmissionWebhook | *[PrometheusOperatorAdmissionWebhookConfig](#prometheusoperatoradmissionwebhookconfig) | `PrometheusOperatorConfig` defines settings for the Prometheus Operator's admission webhook component. |
```
juzhao, line 135:
```suggestion
| prometheusOperatorAdmissionWebhook | *[PrometheusOperatorAdmissionWebhookConfig](#prometheusoperatoradmissionwebhookconfig) | `PrometheusOperatorAdmissionWebhookConfig` defines settings for the Prometheus Operator's admission webhook component. |
```

Documentation/api.md
```diff
@@ -430,6 +436,21 @@ The `PrometheusK8sConfig` resource defines settings for the Prometheus component
 
 [Back to TOC](#table-of-contents)
 
+## PrometheusOperatorAdmissionWebhookConfig
+
+#### Description
+
+The `PrometheusOperatorConfig` resource defines settings for the Prometheus Operator's admission webhook workload.
```
juzhao, line 443:
```suggestion
The `PrometheusOperatorAdmissionWebhookConfig` resource defines settings for the Prometheus Operator's admission webhook workload.
```

Documentation/openshiftdocs/modules/prometheusoperatoradmissionwebhookconfig.adoc
```diff
@@ -0,0 +1,25 @@
...
+	// file will be overwritten when the content is re-generated. If you wish to 
+	// make edits, read the docgen utility instructions in the source code for the 
+	// CMO.
+	:_content-type: ASSEMBLY
+
+== PrometheusOperatorAdmissionWebhookConfig
+
+=== Description
+
+The `PrometheusOperatorConfig` resource defines settings for the Prometheus Operator's admission webhook workload.
```
juzhao, line 12:
```suggestion
The `PrometheusOperatorAdmissionWebhookConfig` resource defines settings for the Prometheus Operator's admission webhook workload.
```

Documentation/openshiftdocs/modules/clustermonitoringconfiguration.adoc
```diff
@@ -27,6 +27,8 @@ The `ClusterMonitoringConfiguration` resource defines settings that customize th
 
 |prometheusOperator|*link:prometheusoperatorconfig.adoc[PrometheusOperatorConfig]|`PrometheusOperatorConfig` defines settings for the Prometheus Operator component.
 
+|prometheusOperatorAdmissionWebhook|*link:prometheusoperatoradmissionwebhookconfig.adoc[PrometheusOperatorAdmissionWebhookConfig]|`PrometheusOperatorConfig` defines settings for the Prometheus Operator's admission webhook component.
```
juzhao, line 30:
```suggestion
|prometheusOperatorAdmissionWebhook|*link:prometheusoperatoradmissionwebhookconfig.adoc[PrometheusOperatorAdmissionWebhookConfig]|`PrometheusOperatorAdmissionWebhookConfig` defines settings for the Prometheus Operator's admission webhook component.
```

pkg/manifests/types.go
```diff
@@ -662,6 +676,13 @@ type ThanosRulerConfig struct {
 	VolumeClaimTemplate *monv1.EmbeddedPersistentVolumeClaim `json:"volumeClaimTemplate,omitempty"`
 }
 
+// The `PrometheusOperatorConfig` resource defines settings for the Prometheus
```
juzhao, line 679:
```suggestion
// The `PrometheusOperatorAdmissionWebhookConfig` resource defines settings for the Prometheus
```

Documentation/openshiftdocs/modules/telemeterclientconfig.adoc
```diff
@@ -23,6 +23,8 @@ Appears in: link:clustermonitoringconfiguration.adoc[ClusterMonitoringConfigurat
 | Property | Type | Description 
 |nodeSelector|map[string]string|Defines the nodes on which the pods are scheduled.
 
+|resources|*v1.ResourceRequirements|Defines resource requests and limits for the Telemeter container.
```
juzhao, line 26:
```suggestion
|resources|*v1.ResourceRequirements|Defines resource requests and limits for the Telemeter Client container.
```

pkg/manifests/types.go
```diff
@@ -662,6 +676,13 @@ type ThanosRulerConfig struct {
...
+// Operator's admission webhook workload.
+type PrometheusOperatorAdmissionWebhookConfig struct {
+	// Defines resource requests and limits for the console-plugin container.
```
juzhao, line 682:
```suggestion
	// Defines resource requests and limits for the prometheus-operator-admission-webhook container.
```

Documentation/openshiftdocs/modules/prometheusoperatoradmissionwebhookconfig.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+
+
+Appears in: link:clustermonitoringconfiguration.adoc[ClusterMonitoringConfiguration]
+
+[options="header"]
+|===
+| Property | Type | Description 
+|resources|*v1.ResourceRequirements|Defines resource requests and limits for the console-plugin container.
```
juzhao, line 21:
```suggestion
|resources|*v1.ResourceRequirements|Defines resource requests and limits for the prometheus-operator-admission-webhook container.
```

Documentation/api.md
```diff
@@ -430,6 +436,21 @@ The `PrometheusK8sConfig` resource defines settings for the Prometheus component
...
+
+
+<em>appears in: [ClusterMonitoringConfiguration](#clustermonitoringconfiguration)</em>
+
+| Property | Type | Description |
+| -------- | ---- | ----------- |
+| resources | *[v1.ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#resourcerequirements-v1-core) | Defines resource requests and limits for the console-plugin container. |
```
juzhao, line 450:
```suggestion
| resources | *[v1.ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#resourcerequirements-v1-core) | Defines resource requests and limits for the prometheus-operator-admission-webhook container. |
```

Documentation/api.md
```diff
@@ -545,6 +567,7 @@ The `TLSConfig` resource configures the settings for TLS connections.
 | Property | Type | Description |
 | -------- | ---- | ----------- |
 | nodeSelector | map[string]string | Defines the nodes on which the pods are scheduled. |
+| resources | *[v1.ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#resourcerequirements-v1-core) | Defines resource requests and limits for the Telemeter container. |
```
juzhao, line 570:
```suggestion
| resources | *[v1.ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#resourcerequirements-v1-core) | Defines resource requests and limits for the Telemeter Client container. |
```

pkg/manifests/types.go
```diff
@@ -248,6 +258,8 @@ type TelemeterClientConfig struct {
 	Enabled *bool `json:"enabled,omitempty"`
 	// Defines the nodes on which the pods are scheduled.
 	NodeSelector map[string]string `json:"nodeSelector"`
+	// Defines resource requests and limits for the Telemeter container.
```
juzhao, line 261:
```suggestion
	// Defines resource requests and limits for the Telemeter Client container.
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - [#1950](https://github.com/openshift/cluster-monitoring-operator/pull/1950) Disable CORS headers on Thanos querier by default and add a flag to enable them back.
 - [#1963](https://github.com/openshift/cluster-monitoring-operator/pull/1963) Add nodeExporter settings for network devices list.
 - [#2049](https://github.com/openshift/cluster-monitoring-operator/pull/2049) Remove Kube*QuotaOvercommit alerts.
+- [#2067](https://github.com/openshift/cluster-monitoring-operator/pull/2067) Add options to specify resource requests and limits for all components.
```
machine424, line 10: Cool, finally :)

but IIUC, these new fields would override the default requests we already set.

Is it really what we want to do? Force users to define both requests and limits even when they only want to set limits?

IIRC, when only limits are defined, Kubernetes will set requests to the same values than limits.

juzhao: 1. if we don't set requests for the components, it will use the default requests as defined in each deployment/daemonset
2. https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
```
Note: If you specify a limit for a resource, but do not specify any request, and no admission-time mechanism has applied a default request for that resource, then Kubernetes copies the limit you specified and uses it as the requested value for the resource.
```
3. specify a request/limit, the component should use the specified settings

machine424: Thanks @juzhao for confirming that "if only limits are set, requests are set equal to limits".

Here are my concerns in two scenarios:

- User only sets the limits via this feature field without knowing that the default requests are not taken into account anymore and that the requests will be equal to the limits they provided (they may have some capacity issues) (unexpected behaviour)
- User took the care to copy the default requests in this feature field, but I a change was done that makes a component need less or more requests, we take care of changing the default requests, but the requests in the feature field are not up to date anymore. (maintenance burden)

pkg/manifests/manifests_test.go
```diff
@@ -3365,12 +3364,33 @@ nodeExporter:
...
+	}
+	return nil
+}
+
 func TestKubeStateMetrics(t *testing.T) {
 	config := `kubeStateMetrics:
   resources:
-    requests:
-      cpu: 100m
-      memory: 100Mi
```
rexagod: `requests` here will be now replaced with defaults, in case only `limits` were specified.

@machine424 WDYT?

machine424: Yep, merging the config may be more practical for users.

We'll just need to see at which level should the merging stop (if the user only provides `requests.cpu` e.g. should we keep `defaultRequests.memory` or set it no nil?)

Maybe we can check for the behavior that the other operators opted for, to keep consistency.

Maybe we can ask the others for their opinions.

rexagod: It seems PO adheres to [the same approach](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/k8sutil/merge.go#L56), i.e., only overriding `requests.cpu` (if the user specified just that) while using the default value for `requests.memory`.

CHANGELOG.md
juzhao: > Thanks @juzhao for confirming that "if only limits are set, requests are set equal to limits".
> 
> Here are my concerns in two scenarios:
> 
>     * User only sets the limits via this feature field without knowing that the default requests are not taken into account anymore and that the requests will be equal to the limits they provided (they may have some capacity issues) (unexpected behaviour)
I think this is expected behavior, by design of Kubernetes, user should know it

>     * User took the care to copy the default requests in this feature field, but I a change was done that makes a component need less or more requests, we take care of changing the default requests, but the requests in the feature field are not up to date anymore. (maintenance burden)

prometheus-adapter default resources.requests
https://github.com/openshift/cluster-monitoring-operator/blob/release-4.14/assets/prometheus-adapter/deployment.yaml#L73-L76
1. if use does not set resources.requests in cluster-monitoring-config configmap, prometheus-adapter will sue the  default resources.requests(if the default value can not let prometheus-adapter pod start up, then it's a bug, should be fixed in CMO)
2. if user only set resources.limits, resources.requests use the same value as resources.limits,
```
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    k8sPrometheusAdapter:
      resources:
        limits:
          cpu: 200m
          memory: 500Mi
```
3. set resources.limits and resources.requests, pod should use the  specified value
```
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    k8sPrometheusAdapter:
      resources:
        requests:
          cpu: 20m
          memory: 200Mi
        limits:
          cpu: 200m
          memory: 500Mi
```
when set resources.limits and resources.requests, user should know clearly of the prometheus-adapter cpu/memory usage, then set resources.limits and resources.requests to a proper value, otherwise, pod would not start up

pkg/manifests/manifests_test.go
jan--f: We currently overwrite any existing defaults, even if the user only sets limits. So in that sense the merge proposal is functionality change. We should think through all upgrade scenarios and make sure nothing obviously bad can happen.
For example if a user has set a limit lower then the default request, this might work for them currently. If we now start merging the `Resources` fields, they will run into very annoying errors when they upgrade.

I think my preference would be to keep the old behavior (overwrite everything) and rather document this explicitly.

machine424: I didn't know (or maybe I forgot) we already provide to customize resources for some components.
I don't know how much that customization is used, but apparently the two scenarios I mentioned don't seem to happen a lot (or we don't know about it).
Yep of course it's better to keep the same logic.

rexagod: Removed the latest commit, the PR implements the overriding behaviour now.

PR #2071, opened 2023-08-11
MON-988: remove alert "MultipleContainersOOMKilled"

raptorsun:
The alert `MultipleContainersOOMKilled` is removed. 
The metric `container_runtime_crio_containers_oom_total` is available in OCP now.
Moreover, the alert  `MultipleContainersOOMKilled` is almost always accompanied by the alert `SystemMemoryExceedsReservation` of warning level from [MachineConfigOperator](https://github.com/openshift/machine-config-operator/blob/4048c667b54967ecd62b1e2bed886b4f54b78732/install/0000_90_machine-config-operator_01_prometheus-rules.yaml#L91). 
image

### DIFF
--- a/assets/cluster-monitoring-operator/prometheus-rule.yaml
+++ b/assets/cluster-monitoring-operator/prometheus-rule.yaml
@@ -256,15 +256,6 @@ spec:
       for: 15m
       labels:
         severity: warning
-    - alert: MultipleContainersOOMKilled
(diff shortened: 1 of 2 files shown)


jan--f:
Given that the node team decided not to add an equivalent alert (see https://issues.redhat.com/browse/OCPNODE-600), this

/lgtm

PR #2072, opened 2023-08-14
OCPBUGS-17682: add topologySpreadConstraints to UWM prometheus operator

marioferh:
Add topologySpreadConstraints to UWM prometheus operator
Add test


pkg/manifests/manifests_test.go
```diff
@@ -4322,9 +4322,19 @@ func TestPodDisruptionBudget(t *testing.T) {
 }
 
 func TestPrometheusOperatorUserWorkloadConfiguration(t *testing.T) {
-	c, err := NewConfigFromString(`
-enableUserWorkload: true
-`, false)
+	c := NewDefaultConfig()
+	uwc, err := NewUserConfigFromString(`prometheusOperator:
+  enableUserWorkload: true
```
juzhao, line 4327: `enableUserWorkload: true` is not needed, since this func is test UWM prometheus-operator

marioferh: right, thanks

PR #2078, opened 2023-08-29
MON-3163: support VPAs

rexagod:
**Note:** This requires the latest KSMv2.10 release to work.
***
Introduce and leverage KSM's CRS featureset and support VPAs using that.

<details>
<summary>Details</summary>

### Logs

```
I1124 10:31:19.440466       1 wrapper.go:98] "Starting kube-state-metrics"
W1124 10:31:19.440711       1 client_config.go:618] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I1124 10:31:19.440994       1 server.go:192] "Used default resources"
I1124 10:31:19.441025       1 types.go:184] "Using all namespaces"
I1124 10:31:19.441203       1 server.go:225] "Metric allow-denylisting" allowDenyStatus="Excluding the following lists that were on denylist: ^kube_pod_completion_time$, ^kube_pod_status_scheduled$, ^kube_secret_labels$, ^kube_.+_annotations$, ^
W1124 10:31:19.441221       1 client_config.go:618] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I1124 10:31:19.441812       1 utils.go:70] "Tested communication with server"
I1124 10:31:19.451429       1 utils.go:75] "Run with Kubernetes cluster version" major="1" minor="27" gitVersion="v1.27.6+d548052" gitTreeState="clean" gitCommit="f3ec0ed759cde48849b6e3117c091b7db90c95fa" platform="linux/amd64"
I1124 10:31:19.451470       1 utils.go:76] "Communication with server successful"
I1124 10:31:19.454001       1 server.go:347] "Started metrics server" metricsServerAddress="127.0.0.1:8081"
I1124 10:31:19.454002       1 server.go:336] "Started kube-state-metrics self metrics server" telemetryAddress="127.0.0.1:8082"
I1124 10:31:19.454187       1 server.go:72] levelinfomsgListening onaddress127.0.0.1:8081
I1124 10:31:19.454229       1 server.go:72] levelinfomsgTLS is disabled.http2falseaddress127.0.0.1:8081
I1124 10:31:19.454188       1 server.go:72] levelinfomsgListening onaddress127.0.0.1:8082
I1124 10:31:19.454271       1 server.go:72] levelinfomsgTLS is disabled.http2falseaddress127.0.0.1:8082
I1124 10:31:22.455661       1 config.go:82] "Using custom resource plural" resource="autoscaling.k8s.io_v1_VerticalPodAutoscaler" plural="verticalpodautoscalers"
I1124 10:31:22.455768       1 discovery.go:274] "discovery finished, cache updated"
I1124 10:31:22.455810       1 metrics_handler.go:99] "Autosharding disabled"
I1124 10:31:22.457445       1 custom_resource_metrics.go:79] "Custom resource state added metrics" familyNames=["kube_customresource_kube_verticalpodautoscaler_annotations_info","kube_customresource_kube_verticalpodautoscaler_labels_info","kube_
I1124 10:31:22.568913       1 builder.go:275] "Active resources" activeStoreNames="certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,ingresses,jobs,leases,limitranges,mutatingwebhookconfigur
```

### Query

Screenshot 2023-11-24 at 16 05 42

Screenshot 2023-11-24 at 16 06 44

### Generated `ConfigMap`

```
Name:         kube-state-metrics-custom-resource-state-configmap
Namespace:    openshift-monitoring
Labels:       app.kubernetes.io/managed-by=cluster-monitoring-operator
              app.kubernetes.io/part-of=openshift-monitoring
Annotations:  <none>

Data
====
custom-resource-state-configmap.yaml:
----
"kind": "CustomResourceStateMetrics"
"spec":
  "resources":
  - "groupVersionKind":
      "group": "autoscaling.k8s.io"
      "kind": "VerticalPodAutoscaler"
      "version": "v1"
    "metrics":
    - "commonLabels": null
      "each":
        "info":
          "labelsFromPath":
            "*":
            - "metadata"
            - "annotations"
            "container":
            - "containerName"
          "valueFrom": 1
        "type": "Info"
      "help": "Kubernetes annotations converted to Prometheus labels."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_annotations_info"
    - "commonLabels": null
      "each":
        "info":
          "labelsFromPath":
            "*":
            - "metadata"
            - "labels"
            "container":
            - "containerName"
          "valueFrom": 1
        "type": "Info"
      "help": "Kubernetes labels converted to Prometheus labels."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_labels_info"
    - "commonLabels": null
      "each":
        "stateSet":
          "labelName": "updatemode"
          "list":
          - "Off"
          - "Initial"
          - "Recreate"
          - "Auto"
          "path":
          - "spec"
          - "updatePolicy"
          - "updateMode"
        "type": "StateSet"
      "help": "Update mode of the VerticalPodAutoscaler."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_spec_updatepolicy_updatemode"
    - "commonLabels":
        "resource": "cpu"
        "unit": "cores"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "status"
          - "recommendation"
          - "containerRecommendations"
          "valueFrom":
          - "lowerBound"
          - "cpu"
        "type": "Gauge"
      "help": "Minimum cpu resources the container can use before the VerticalPodAutoscaler updater evicts it."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_status_recommendation_containerrecommendations_lowerbound_cpu"
    - "commonLabels":
        "resource": "memory"
        "unit": "bytes"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "status"
          - "recommendation"
          - "containerRecommendations"
          "valueFrom":
          - "lowerBound"
          - "memory"
        "type": "Gauge"
      "help": "Minimum memory resources the container can use before the VerticalPodAutoscaler updater evicts it."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_status_recommendation_containerrecommendations_lowerbound_memory"
    - "commonLabels":
        "resource": "cpu"
        "unit": "cores"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "status"
          - "recommendation"
          - "containerRecommendations"
          "valueFrom":
          - "upperBound"
          - "cpu"
        "type": "Gauge"
      "help": "Maximum cpu resources the container can use before the VerticalPodAutoscaler updater evicts it."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_status_recommendation_containerrecommendations_upperbound_cpu"
    - "commonLabels":
        "resource": "memory"
        "unit": "bytes"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "status"
          - "recommendation"
          - "containerRecommendations"
          "valueFrom":
          - "upperBound"
          - "memory"
        "type": "Gauge"
      "help": "Maximum memory resources the container can use before the VerticalPodAutoscaler updater evicts it."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_status_recommendation_containerrecommendations_upperbound_memory"
    - "commonLabels":
        "resource": "cpu"
        "unit": "cores"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "status"
          - "recommendation"
          - "containerRecommendations"
          "valueFrom":
          - "target"
          - "cpu"
        "type": "Gauge"
      "help": "Target cpu resources the VerticalPodAutoscaler recommends for the container."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_status_recommendation_containerrecommendations_target_cpu"
    - "commonLabels":
        "resource": "memory"
        "unit": "bytes"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "status"
          - "recommendation"
          - "containerRecommendations"
          "valueFrom":
          - "target"
          - "memory"
        "type": "Gauge"
      "help": "Target memory resources the VerticalPodAutoscaler recommends for the container."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_status_recommendation_containerrecommendations_target_memory"
    - "commonLabels":
        "resource": "cpu"
        "unit": "cores"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "status"
          - "recommendation"
          - "containerRecommendations"
          "valueFrom":
          - "uncappedTarget"
          - "cpu"
        "type": "Gauge"
      "help": "Target cpu resources the VerticalPodAutoscaler recommends for the container ignoring bounds."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_status_recommendation_containerrecommendations_uncappedtarget_cpu"
    - "commonLabels":
        "resource": "memory"
        "unit": "bytes"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "status"
          - "recommendation"
          - "containerRecommendations"
          "valueFrom":
          - "uncappedTarget"
          - "memory"
        "type": "Gauge"
      "help": "Target memory resources the VerticalPodAutoscaler recommends for the container ignoring bounds."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_status_recommendation_containerrecommendations_uncappedtarget_memory"
    - "commonLabels":
        "resource": "cpu"
        "unit": "cores"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "spec"
          - "resourcePolicy"
          - "containerPolicies"
          "valueFrom":
          - "minAllowed"
          - "cpu"
        "type": "Gauge"
      "help": "Minimum cpu resources the VerticalPodAutoscaler can set for containers matching the name."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_spec_resourcepolicy_container_policies_minallowed_cpu"
    - "commonLabels":
        "resource": "memory"
        "unit": "bytes"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "spec"
          - "resourcePolicy"
          - "containerPolicies"
          "valueFrom":
          - "minAllowed"
          - "memory"
        "type": "Gauge"
      "help": "Minimum memory resources the VerticalPodAutoscaler can set for containers matching the name."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_spec_resourcepolicy_container_policies_minallowed_memory"
    - "commonLabels":
        "resource": "cpu"
        "unit": "cores"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "spec"
          - "resourcePolicy"
          - "containerPolicies"
          "valueFrom":
          - "maxAllowed"
          - "cpu"
        "type": "Gauge"
      "help": "Minimum cpu resources the VerticalPodAutoscaler can set for containers matching the name."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_spec_resourcepolicy_container_policies_maxallowed_cpu"
    - "commonLabels":
        "resource": "memory"
        "unit": "bytes"
      "each":
        "gauge":
          "labelsFromPath":
            "container":
            - "containerName"
          "path":
          - "spec"
          - "resourcePolicy"
          - "containerPolicies"
          "valueFrom":
          - "maxAllowed"
          - "memory"
        "type": "Gauge"
      "help": "Minimum memory resources the VerticalPodAutoscaler can set for containers matching the name."
      "labelsFromPath":
        "namespace":
        - "metadata"
        - "namespace"
        "target_api_version":
        - "spec"
        - "targetRef"
        - "apiVersion"
        "target_kind":
        - "spec"
        - "targetRef"
        - "kind"
        "target_name":
        - "spec"
        - "targetRef"
        - "name"
        "verticalpodautoscaler":
        - "metadata"
        - "name"
      "name": "kube_verticalpodautoscaler_spec_resourcepolicy_container_policies_maxallowed_memory"

BinaryData
====

Events:  <none>
```

</details>


jan--f:
Proposed some jsonnet changes in https://github.com/rexagod/cluster-monitoring-operator/pull/1

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -212,11 +243,18 @@ function(params)
...
+                      volumeMounts: [
+                        {
+                          mountPath: '/tmp',
+                          name: tmpVolumeName,
+                          readOnly: false,
+                        },
+                        {
+                          mountPath: utils.getDirectoryPath(crsAssetPath),
+                          name: crsVolumeName,
+                          readOnly: false,
```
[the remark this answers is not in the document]
rexagod: Right, only `read` should suffice here.

rexagod:
/retest
[Tests seem to be flaking.](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/2078/pull-ci-openshift-cluster-monitoring-operator-master-e2e-aws-ovn-techpreview/1727246318662848512#1:build-log.txt%3A26777)

assets/kube-state-metrics/custom-resource-state-config.yaml
```diff
@@ -0,0 +1,555 @@
+kind: CustomResourceStateMetrics
```
simonpasquier, line 1: I'd rather see a configmap here. At first I thought that it was a new CRD.

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -204,6 +234,7 @@ function(params)
                           ^kube_.+_annotations$
                         |||,
                         '--metric-labels-allowlist=pods=[*],nodes=[*],namespaces=[*],persistentvolumes=[*],persistentvolumeclaims=[*],poddisruptionbudgets=[*]',
+                        '--custom-resource-state-config-file=' + crsAssetPath,
```
simonpasquier, line 237: From my POV, hardcoding the paths is fine and more readable. Any typo/misconfiguration will be detected by the e2e tests.

```suggestion
                        '--custom-resource-state-config-file=/etc/kube-state-metrics/custom-resource-state-config.yaml',
```

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -212,11 +243,18 @@ function(params)
...
-                        readOnly: false,
-                      }],
+                      volumeMounts: [
+                        {
+                          mountPath: '/tmp',
+                          name: tmpVolumeName,
+                          readOnly: false,
+                        },
+                        {
+                          mountPath: utils.getDirectoryPath(crsAssetPath),
```
simonpasquier, line 253:
```suggestion
                          mountPath: '/etc/kube-state-metrics',
```

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -243,11 +281,19 @@ function(params)
...
+              },
             ],
             securityContext: {},
             priorityClassName: 'system-cluster-critical',
           },
         },
       },
     },
+
+    customResourceStateConfig: crsConfig,
```
simonpasquier, line 298:
```suggestion
    customResourceStateConfigmap: crsConfig,
```

jsonnet/components/kube-state-metrics-custom-resource-state.libsonnet
```diff
@@ -0,0 +1,119 @@
...
+          version: 'v1',
+          kind: 'VerticalPodAutoscaler',
+        },
+        metrics: vpaMetrics,
+      },
+    ],
+  },
+};
+
+local configMap = {
```
simonpasquier, line 104: (nit) I would move the configmap "envelope" to kube-state-metrics.libsonnet

jsonnet/components/kube-state-metrics-custom-resource-state.libsonnet
simonpasquier, line 1: (nit) usually 1 file in jsonnet/components = 1 monitoring component. I'd move this file to jsonnet/utils

pkg/manifests/manifests.go
```diff
@@ -793,6 +794,25 @@ func (f *Factory) KubeStateMetricsPrometheusRule() (*monv1.PrometheusRule, error
...
+	manifest, err := io.ReadAll(f.assets.MustNewAssetReader(KubeStateMetricsCRSConfig))
+	if err != nil {
+		return nil, err
+	}
+	cm := &v1.ConfigMap{}
+	if err := yaml.Unmarshal(manifest, cm); err != nil {
+		return nil, err
+	}
+	// Populate KSM's CRS configmap's fields.
+	cm.Name = strings.ReplaceAll(strings.TrimSuffix(KubeStateMetricsCRSConfig, ".yaml"), "/", "-")
```
simonpasquier, line 807: why not using the name from the static manifest?

pkg/manifests/manifests.go
```diff
@@ -793,6 +794,25 @@ func (f *Factory) KubeStateMetricsPrometheusRule() (*monv1.PrometheusRule, error
...
+	cm.Namespace = f.namespace
+	cm.Labels = map[string]string{
```
simonpasquier, line 809: it should already be set in the static manifest?

pkg/tasks/kubestatemetrics.go
```diff
@@ -35,6 +35,16 @@ func NewKubeStateMetricsTask(client *client.Client, factory *manifests.Factory)
...
 
 func (t *KubeStateMetricsTask) Run(ctx context.Context) error {
+	cm, err := t.factory.KubeStateMetricsCRSConfigMap()
+	if err != nil {
+		return errors.Wrap(err, "initializing kube-state-metrics CRS ConfigMap failed")
+	}
+
+	err = t.client.CreateOrUpdateConfigMap(ctx, cm)
+	if err != nil {
+		return errors.Wrap(err, "reconciling kube-state-metrics CRS ConfigMap failed")
```
simonpasquier, line 45:
```suggestion
		return errors.Wrap(err, "reconciling kube-state-metrics custom resource state ConfigMap failed")
```

pkg/tasks/kubestatemetrics.go
```diff
@@ -35,6 +35,16 @@ func NewKubeStateMetricsTask(client *client.Client, factory *manifests.Factory)
 }
 
 func (t *KubeStateMetricsTask) Run(ctx context.Context) error {
+	cm, err := t.factory.KubeStateMetricsCRSConfigMap()
+	if err != nil {
+		return errors.Wrap(err, "initializing kube-state-metrics CRS ConfigMap failed")
```
simonpasquier, line 40:
```suggestion
		return errors.Wrap(err, "initializing kube-state-metrics custom resource state ConfigMap failed")
```

simonpasquier (commented):
@rexagod can you update the PR description with an example of the metrics being collected?

/cc @joelsmith
Joel, can we get this reviewed by someone from the VPA team?

rexagod:
Also updated the PR description with relevant data.

pkg/manifests/manifests.go
```diff
@@ -793,6 +794,18 @@ func (f *Factory) KubeStateMetricsPrometheusRule() (*monv1.PrometheusRule, error
...
+func (f *Factory) KubeStateMetricsCRSConfigMap() (*v1.ConfigMap, error) {
+	manifest, err := io.ReadAll(f.assets.MustNewAssetReader(KubeStateMetricsCRSConfig))
+	if err != nil {
+		return nil, err
+	}
+	cm := &v1.ConfigMap{}
+	if err := yaml.Unmarshal(manifest, cm); err != nil {
+		return nil, err
+	}
+	return cm, nil
```
simonpasquier, line 806: shorter version :)
```suggestion
	return f.NewConfigMap(f.assets.MustNewAssetReader(KubeStateMetricsCRSConfig))
```

rexagod:
The latest commit patch addresses an upstream issue where the config change triggered reloads, but eventually led to KSM crashing due to a race condition (reload and port acquiring procedure happening before ports were released).

<details>
<summary>Logs</summary>


```
┌[rexagod@nebuchadnezzar] [/dev/ttys002] [main]
└[~/repositories/oss/kube-state-metrics]> go install && kube-state-metrics --kubeconfig=$KUBECONFIG --custom-resource-state-config-file ../../misc/ksm-crs.yaml
I1127 16:28:00.925559   42065 wrapper.go:120] "Starting kube-state-metrics"
I1127 16:28:00.928076   42065 server.go:192] "Used default resources"
I1127 16:28:00.928137   42065 types.go:184] "Using all namespaces"
I1127 16:28:00.928178   42065 server.go:225] "Metric allow-denylisting" allowDenyStatus="Excluding the following lists that were on denylist: "
I1127 16:28:00.930107   42065 utils.go:70] "Tested communication with server"
I1127 16:28:00.951541   42065 utils.go:75] "Run with Kubernetes cluster version" major="1" minor="27" gitVersion="v1.27.3" gitTreeState="clean" gitCommit="25b4e43193bcda6c7328a6d147b1fb73a33f1598" platform="linux/arm64"
I1127 16:28:00.951559   42065 utils.go:76] "Communication with server successful"
I1127 16:28:00.952085   42065 server.go:349] "Started metrics server" metricsServerAddress="[::]:8080"
I1127 16:28:00.952090   42065 server.go:338] "Started kube-state-metrics self metrics server" telemetryAddress="[::]:8081"
I1127 16:28:00.952278   42065 server.go:72] levelinfomsgListening onaddress[::]:8080
I1127 16:28:00.952284   42065 server.go:72] levelinfomsgListening onaddress[::]:8081
I1127 16:28:00.952292   42065 server.go:72] levelinfomsgTLS is disabled.http2falseaddress[::]:8081
I1127 16:28:00.952285   42065 server.go:72] levelinfomsgTLS is disabled.http2falseaddress[::]:8080
I1127 16:28:29.842231   42065 wrapper.go:89] "Changes detected" name="../../misc/ksm-crs.yaml"
I1127 16:28:30.953733   42065 discovery.go:262] "context cancelled"
I1127 16:28:32.848284   42065 server.go:192] "Used default resources"
I1127 16:28:32.848378   42065 types.go:184] "Using all namespaces"
I1127 16:28:32.848411   42065 server.go:225] "Metric allow-denylisting" allowDenyStatus="Excluding the following lists that were on denylist: "
I1127 16:28:32.849526   42065 server.go:349] "Started metrics server" metricsServerAddress="[::]:8080"
I1127 16:28:32.849534   42065 server.go:338] "Started kube-state-metrics self metrics server" telemetryAddress="[::]:8081"
E1127 16:28:32.849878   42065 wrapper.go:40] "Failed to run kube-state-metrics" err="run server group error: listen tcp 0.0.0.0:8080: bind: address already in use"
```

</details>

pkg/manifests/manifests.go
```diff
@@ -806,6 +825,44 @@ func (f *Factory) KubeStateMetricsPrometheusRule() (*monv1.PrometheusRule, error
...
+func (f *Factory) KubeStateMetricsCRSConfigMap() (*v1.ConfigMap, error) {
+	configmap, err := f.NewConfigMap(f.assets.MustNewAssetReader(KubeStateMetricsCRSConfig))
+	if err != nil {
+		return nil, err
+	}
+
+	// Check CRS prerequisites before enabling CRS metrics in CMO.
+	// Currently, the prerequisites are:
+	// * The presence of `VerticalPodAutoscaler` CRD in the cluster: `KubeStateMetricsListErrors` will be fired if absent.
+	_, err = f.client.ApiExtensionsInterface().ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), verticalPodAutoscalersGR, metav1.GetOptions{})
```
simonpasquier, line 837: It breaks the (implicit) separation of concerns that Factory doesn't interact with the Kubernetes API.

simonpasquier (commented):
Instead of `Factory` querying the Kubernetes API, I'd recommend to add a parameter to `KubeStateMetricsTask` telling whether it should configure VPA metrics. The Operator would be in charge of checking the presence  of the VPA CRDs.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - [#1963](https://github.com/openshift/cluster-monitoring-operator/pull/1963) Add nodeExporter settings for network devices list.
 - [#2049](https://github.com/openshift/cluster-monitoring-operator/pull/2049) Remove Kube*QuotaOvercommit alerts.
 - [#2067](https://github.com/openshift/cluster-monitoring-operator/pull/2067) Add options to specify resource requests and limits for all components.
+- [#2078](https://github.com/openshift/cluster-monitoring-operator/pull/2078) Support exporting VPA metrics from KSM.
```
jan--f, line 11: this should probably be in the 4.15 section?

test/e2e/framework/framework.go
```diff
@@ -162,16 +162,7 @@ func New(kubeConfigPath string) (*Framework, CleanUpFunc, error) {
...
-			namespaceName,
-			userWorkloadNamespaceName,
-			nil,
-			nil,
-			nil,
-			manifests.NewAssets("../../assets"),
-			&manifests.APIServerConfig{},
-			&configv1.Console{},
-		),
+		ManifestsFactory:          manifests.NewFactory(namespaceName, userWorkloadNamespaceName, nil, nil, nil, manifests.NewAssets("../../assets"), &manifests.APIServerConfig{}, &configv1.Console{}),
```
jan--f, line 165: Just for the future, I'd prefer we avoid reformatting changes like this. It just makes `git blame` more cumbersome to use.

pkg/client/client.go
```diff
@@ -68,10 +68,11 @@ import (
...
 const (
-	deploymentCreateTimeout = 5 * time.Minute
-	deploymentDeleteTimeout = 5 * time.Minute
-	metadataPrefix          = "monitoring.openshift.io/"
-	clusterConsole          = "cluster"
+	deploymentCreateTimeout       = 5 * time.Minute
+	deploymentDeleteTimeout       = 5 * time.Minute
+	metadataPrefix                = "monitoring.openshift.io/"
+	clusterConsole                = "cluster"
+	VerticalPodAutoscalerGRString = "verticalpodautoscalers.autoscaling.k8s.io"
```
jan--f, line 75:
```suggestion
	VerticalPodAutoscalerCRString = "verticalpodautoscalers.autoscaling.k8s.io"
```
Just wondering if the name is supposed to be VerticalPodAutoscaler**GR**String. Feel free to ignore.

jan--f (approved):
/lgtm
Just a small nit that isn't clear to me, but nothing that needs to hold this up.

pkg/client/client.go
rexagod: I suffixed it GR to reflect the "resource.group" format of that string (analogous to GVK). Would something like `VerticalPodAutoscalerCRDMetadataName` (or `VerticalPodAutoscalerGroupResourceString`) be a better candidate instead?

jan--f: That's what I came up with too...so that should work. We can always change the name in a follow up though.

test/e2e/framework/framework.go
rexagod: Ah, I used this new replace feature in Goland [to remove the last parameter from all `NewFactory` calls](https://github.com/openshift/cluster-monitoring-operator/pull/2078/commits/441c0465c9620b7cf99f6473610df90a4376ca71#diff-c354587e36151f73f4a4da62c22d612510bee633ada8a52ed16b89c7185f0283L174) and it ended up refactoring them too. Let me revert this, as this could lead to unnecessary merge conflicts down the line.

pkg/operator/operator.go
```diff
@@ -405,6 +406,20 @@ func New(
 	}
 	o.informers = append(o.informers, informer)
 
+	informer = cache.NewSharedIndexInformer(
+		o.client.VerticalPodAutoscalerCRDListWatch(ctx),
+		&apiextv1.CustomResourceDefinition{}, resyncPeriod, cache.Indexers{},
+	)
+	_, err = informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    o.handleEvent,
+		UpdateFunc: func(_, newObj interface{}) { o.handleEvent(newObj) },
```
simonpasquier, line 415: do we care about update events? I think not.

pkg/operator/operator.go
```diff
@@ -728,7 +747,30 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 		klog.Warningf("Fail to load ConsoleConfig, AlertManager's externalURL may be outdated")
 	}
 
-	factory := manifests.NewFactory(o.namespace, o.namespaceUserWorkload, config, o.loadInfrastructureConfig(ctx), proxyConfig, o.assets, apiServerConfig, consoleConfig)
+	// Enable kube-state-metrics' custom-resource-state-based metrics if VPA CRD is installed within the cluster.
+	enableKSMCRSMetrics := false
+	_, err = o.client.ApiExtensionsInterface().ApiextensionsV1().CustomResourceDefinitions().Get(ctx, client.VerticalPodAutoscalerCRDMetadataName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Warningf("%s CRD not found, kube-state-metrics' custom-resource-state-based metrics will be disabled", client.VerticalPodAutoscalerCRDMetadataName)
```
simonpasquier, line 755: I wouldn't log anything (at least not a warning).

pkg/operator/operator.go
```diff
@@ -728,7 +747,30 @@ func (o *Operator) sync(ctx context.Context, key string) error {
...
+		} else {
+			klog.Errorf("Failed to get %s CRD, skipping kube-state-metrics' custom-resource-state-based metrics generation", client.VerticalPodAutoscalerCRDMetadataName)
```
simonpasquier, line 757: should be warning.

pkg/tasks/kubestatemetrics.go
```diff
@@ -85,11 +93,55 @@ func (t *KubeStateMetricsTask) Run(ctx context.Context) error {
...
 	dep, err := t.factory.KubeStateMetricsDeployment()
 	if err != nil {
 		return errors.Wrap(err, "initializing kube-state-metrics Deployment failed")
 	}
 
+	// Check if custom-resource-state metrics have been enabled.
+	// * If so, add the --custom-resource-state-config-file flag to the kube-state-metrics container, if it is absent.
+	// * If not, remove the --custom-resource-state-config-file flag from the kube-state-metrics container, if it is present.
+	// This will, in turn, also cause the kube-state-metrics container to be restarted with the new set of flags in effect.
+	if t.enableCRSMetrics {
```
simonpasquier, line 115: it should be moved to t.factory.KubeStateMetricsDeployment() as we do for other components. You can add tests to pkg/manifests/manifests_test.go

assets/kube-state-metrics/deployment.yaml
```diff
@@ -40,6 +40,7 @@ spec:
           ^kube_secret_labels$,
           ^kube_.+_annotations$
         - --metric-labels-allowlist=pods=[*],nodes=[*],namespaces=[*],persistentvolumes=[*],persistentvolumeclaims=[*],poddisruptionbudgets=[*]
+        - --custom-resource-state-config-file=/etc/kube-state-metrics/custom-resource-state-configmap.yaml
```
simonpasquier, line 43: I would remove the argument so you only need to add it.

pkg/manifests/manifests.go
```diff
@@ -761,7 +773,49 @@ func (f *Factory) KubeStateMetricsMinimalServiceMonitor() (*monv1.ServiceMonitor
 	return f.NewServiceMonitor(f.assets.MustNewAssetReader(KubeStateMetricsMinimalServiceMonitor))
 }
 
-func (f *Factory) KubeStateMetricsDeployment() (*appsv1.Deployment, error) {
+// Check if custom-resource-state metrics have been enabled.
+// * If so, add the --custom-resource-state-config-file flag to the kube-state-metrics container, if it is absent.
+// * If not, remove the --custom-resource-state-config-file flag from the kube-state-metrics container, if it is present.
+// This will, in turn, also cause the kube-state-metrics container to be restarted with the new set of flags in effect.
+func enableOrDisableCRSMetrics(d *appsv1.Deployment, enableCRSMetrics bool) {
```
simonpasquier, line 780: A separate function makes it harder to follow for me. Why not injecting the argument here?

https://github.com/openshift/cluster-monitoring-operator/blob/84be118af132567e6f84dfd6e72f3172877c5a3e/pkg/manifests/manifests.go#L774-L778

```go
		case "kube-state-metrics":
			d.Spec.Template.Spec.Containers[i].Image = f.config.Images.KubeStateMetrics
			if f.config.ClusterMonitoringConfiguration.KubeStateMetricsConfig.Resources != nil {
				d.Spec.Template.Spec.Containers[i].Resources = *f.config.ClusterMonitoringConfiguration.KubeStateMetricsConfig.Resources
			}
			if enableCRSMetrics {
				d.Spec.Template.Spec.Containers[containerIndex].Args = append(d.Spec.Template.Spec.Containers[containerIndex].Args, "--custom-resource-state-config-file=/etc/kube-state-metrics/custom-resource-state-configmap.yaml")
			}
		}
```

pkg/operator/operator.go
```diff
@@ -405,6 +406,19 @@ func New(
 	}
 	o.informers = append(o.informers, informer)
 
+	informer = cache.NewSharedIndexInformer(
+		o.client.VerticalPodAutoscalerCRDListWatch(ctx),
+		&apiextv1.CustomResourceDefinition{}, resyncPeriod, cache.Indexers{},
+	)
+	_, err = informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
simonpasquier, line 413:
```suggestion
	// The operator only needs to reconcile when the CRD is added/removed.
	_, err = informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```

pkg/operator/operator.go
```diff
@@ -728,7 +746,28 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 		klog.Warningf("Fail to load ConsoleConfig, AlertManager's externalURL may be outdated")
 	}
 
-	factory := manifests.NewFactory(o.namespace, o.namespaceUserWorkload, config, o.loadInfrastructureConfig(ctx), proxyConfig, o.assets, apiServerConfig, consoleConfig)
+	// Enable kube-state-metrics' custom-resource-state-based metrics if VPA CRD is installed within the cluster.
+	enableKSMCRSMetrics := false
+	_, err = o.client.ApiExtensionsInterface().ApiextensionsV1().CustomResourceDefinitions().Get(ctx, client.VerticalPodAutoscalerCRDMetadataName, metav1.GetOptions{})
```
simonpasquier, line 751: We might want to return an error if we get an error other than not found from the API, otherwise the operator might update the KSM deployment and revert the good config.
Something we did in other cases is to store the last know value in the operator struct and default to it in case of transient errors from the API.

(nit) I would encapsulate the logic in a new function in pkg/client (e.g. `func (c *Client) VerticalPodAutoscalerCRDPresent() (bool, error)`).

pkg/manifests/manifests.go
rexagod: Moving this out into its own function enabled me to test both of the `enableCRSMetrics` [branches](https://github.com/openshift/cluster-monitoring-operator/pull/2078/files#diff-1f9d2d18d4cabd8262e585b932330183f7cb9b544ac9f27c8bcb3ef9c8d14376R782-R816): https://github.com/openshift/cluster-monitoring-operator/pull/2078/files#diff-b388a528a3665a46f4dfa8eb4a8c02de116acb8f85e4611f12f11dc337d7f7b5R3564-R3568.

simonpasquier: Hmm, the static manifest doesn't have the flag in the first place and the factory loads that manifest so it's safe to assume that the argument will never present. To ensure that the assumption doesn't change in the future, we need an e2e test which is required anyway from my POV: there are many things that could go wrong and only e2e testing will validate that the metrics are ingested in Prometheus eventually.

FWIW this wouldn't be different from what we're doing elsewhere https://github.com/openshift/cluster-monitoring-operator/blob/e39355c5d44382b31937bf577da66883e835b01b/pkg/manifests/manifests.go#L2880-L2892

rexagod: Reverted.

test/e2e/kube_state_metrics_test.go
```diff
@@ -65,3 +72,138 @@ func TestKSMMetricsSuppression(t *testing.T) {
 	}
 
 }
+
+func TestKSMCRSMetrics(t *testing.T) {
+	assetsDir := "./assets/"
+	query := "kube_customresource_kube_verticalpodautoscaler_labels_info"
+
+	// Install a VPAv1 CRD.
+	manifest, err := f.ReadManifest(assetsDir + "verticalpodautoscalers-v1-crd.yaml")
```
simonpasquier, line 81: (nit)
```suggestion
	manifest, err := f.ReadManifest(path.Join(assetsDir, "verticalpodautoscalers-v1-crd.yaml"))
```

test/e2e/kube_state_metrics_test.go
```diff
@@ -65,3 +72,138 @@ func TestKSMMetricsSuppression(t *testing.T) {
...
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("failed to query Prometheus: %v", err)
+	}
+}
+
+func createVPACR(t *testing.T, vpaCR *vpav1.VerticalPodAutoscaler) {
+	_, err := f.VPAClient.VerticalPodAutoscalers(f.Ns).Create(ctx, vpaCR, metav1.CreateOptions{})
```
simonpasquier, line 144: I'd rather Poll on the custom resource creation. We don't expect any error eventually (e.g. "AlreadyExists" isn't an option) and we can exit as soon as the API told us that the creation was successful.

test/e2e/kube_state_metrics_test.go
```diff
@@ -65,3 +72,138 @@ func TestKSMMetricsSuppression(t *testing.T) {
...
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("failed to create VPA CR: %v", err)
+	}
+}
+
+func createVPACRD(t *testing.T, vpaCRD interface{}) {
+	vpaCRD, err := f.APIExtensionsClient.ApiextensionsV1().CustomResourceDefinitions().Create(ctx, vpaCRD.(*apiextv1.CustomResourceDefinition), metav1.CreateOptions{})
```
simonpasquier, line 161: same here

assets/kube-state-metrics/custom-resource-state-configmap.yaml
```diff
@@ -0,0 +1,561 @@
...
+            - "targetRef"
+            - "kind"
+            "target_name":
+            - "spec"
+            - "targetRef"
+            - "name"
+            "verticalpodautoscaler":
+            - "metadata"
+            - "name"
+          "name": "kube_verticalpodautoscaler_annotations_info"
```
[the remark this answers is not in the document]
rexagod: ACK, the `*` wildcard should expose all annotations (for eg., https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md#multiple-metricskitchen-sink).

assets/kube-state-metrics/custom-resource-state-configmap.yaml
```diff
@@ -0,0 +1,561 @@
...
+            - "targetRef"
+            - "kind"
+            "target_name":
+            - "spec"
+            - "targetRef"
+            - "name"
+            "verticalpodautoscaler":
+            - "metadata"
+            - "name"
+          "name": "kube_verticalpodautoscaler_labels_info"
```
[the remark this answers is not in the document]
rexagod: ACK, the `*` wildcard should expose all labels (for eg., https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md#multiple-metricskitchen-sink).

pkg/manifests/manifests.go
```diff
@@ -776,6 +789,33 @@ func (f *Factory) KubeStateMetricsDeployment() (*appsv1.Deployment, error) {
...
+					// value as the one set by the operator.
+					if arg == flagCRSConfigFile {
+						crsEnabled = true
+						break
+					}
+				}
+				if !crsEnabled {
+					d.Spec.Template.Spec.Containers[i].Args = append(container.Args, flagCRSConfigFile)
+				}
+			} else {
```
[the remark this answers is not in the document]
rexagod: Wouldn't this still be needed in the case where the CRS feature-set was previously enabled?

rexagod:
/hold
I need to test the latest changes, not able to get a cluster right now. Will unhold once I've tested the latest commit.

assets/kube-state-metrics/custom-resource-state-configmap.yaml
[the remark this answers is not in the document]
simonpasquier: For other `kube_*_labels_info` metrics, we don't expose all Kubernetes labels because of cardinality concerns (minus a few exceptions like `for pods, namespaces, ...). I'd probably go the same way for VPA metrics until we have a request/use case to do differently.

@jan--f WDYT?

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -212,11 +241,18 @@ function(params)
...
-                        name: tmpVolumeName,
-                        readOnly: false,
-                      }],
+                      volumeMounts: [
+                        {
+                          mountPath: '/tmp',
+                          name: tmpVolumeName,
+                          readOnly: false,
+                        },
+                        {
```
simonpasquier, line 250: (nit) a comment would clarify why the configmap is always mounted
```suggestion
                        // The custom resource state configmap is always mounted in the kube-state-metrics container and only when the VPA CRD is installed, CMO will add `--custom-resource-state-config-file` to the container arguments list.
                        {
```

pkg/client/client.go
```diff
@@ -1873,6 +1893,26 @@ func (c *Client) RegisterConsolePlugin(ctx context.Context, name string) error {
 	return errors.Wrapf(err, "registering console-plugin %q with console %q failed", name, clusterConsole)
 }
 
+// ShouldEnableKSMCRSMetrics checks if kube-state-metrics' custom-resource-state-based metrics generation should be enabled.
+func (c *Client) ShouldEnableKSMCRSMetrics(ctx context.Context, lastKnownEnableKSMCRSMetricsState bool) bool {
```
simonpasquier, line 1897: (nit) in general, the client package acts a wrapper to the Kube API and the operator package is in charge of interpreting the information, logging, ...
```suggestion
func (c *Client) VPACustomResourceDefinitionPresent(ctx context.Context) (bool, error) {
```

pkg/operator/operator.go
```diff
@@ -188,6 +189,8 @@ type Operator struct {
 
 	ruleController    *alert.RuleController
 	relabelController *alert.RelabelConfigController
+
+	lastKnownEnableKSMCRSMetricsState bool
```
simonpasquier, line 193:
```suggestion
	lastKnownVPACustomResourceDefinitionPresent bool
```

pkg/manifests/manifests.go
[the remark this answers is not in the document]
simonpasquier: No CMO reads the deployment manifest from the assets/ directory when reconciling. It doesn't read the current deployment object from the Kubernetes API so we're certain that the --custom-resource-state-config-file argument is absent. If it ever changes, the e2e test should catch the regression.

test/e2e/kube_state_metrics_test.go
```diff
@@ -15,6 +15,12 @@
 package e2e
 
 import (
+	v1 "k8s.io/api/autoscaling/v1"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	vpav1 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
+	"path"
```
simonpasquier, line 23: (nit) can you rearrange the import list to have the stdlib packages first and separated from the rest?

test/e2e/kube_state_metrics_test.go
```diff
@@ -65,3 +71,102 @@ func TestKSMMetricsSuppression(t *testing.T) {
...
+				APIVersion: "apps/v1",
+				Kind:       "Deployment",
+				Name:       "cluster-monitoring-operator",
+			},
+		},
+	}
+	createVPACR(t, vpaCR)
+
+	// Fetch KSM CRS metrics.
+	fetchKSMCRSMetrics(t, queryPresenceCheck)
```
simonpasquier, line 116: (nit) since fetchKSMCRSMetrics is a one-liner, I would inline it.

test/e2e/kube_state_metrics_test.go
```diff
@@ -65,3 +71,102 @@ func TestKSMMetricsSuppression(t *testing.T) {
...
+
+	// Cleanup.
+	deleteVPACRD(t, vpaCRD)
+
+	// Fetch KSM CRS metrics, but expect failure.
+	fetchKSMCRSMetrics(t, queryAbsenceCheck)
```
simonpasquier, line 122: same here.

test/e2e/kube_state_metrics_test.go
```diff
@@ -65,3 +71,102 @@ func TestKSMMetricsSuppression(t *testing.T) {
...
+	}
+	createVPACR(t, vpaCR)
+
+	// Fetch KSM CRS metrics.
+	fetchKSMCRSMetrics(t, queryPresenceCheck)
+
+	// Cleanup.
+	deleteVPACRD(t, vpaCRD)
+
+	// Fetch KSM CRS metrics, but expect failure.
```
simonpasquier, line 121:
```suggestion
	// Fetch KSM CRS metrics, but expect absence.
```

assets/kube-state-metrics/custom-resource-state-configmap.yaml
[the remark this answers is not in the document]
rexagod: ACK, I'll reopen and update https://issues.redhat.com/browse/MON-3191 with the discussion here.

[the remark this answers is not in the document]
rexagod: We may want to revisit the regex here (to define behaviors for all `kube_customresource_.+_annotation_info` metrics, and disable annotations (and labels) entirely: https://github.com/openshift/cluster-monitoring-operator/blob/8db648fa23ade5453b87a1f58206f07be5791110/assets/kube-state-metrics/deployment.yaml#L41

Joel's mail mentioned a particular ask for `verticalpodautoscaler_status_recommendation_containerrecommendations_target_cpu`, so leaving out the annotation and labels metrics should be safe from that POV. 
```This did exactly what I wanted:
              - name: "verticalpodautoscaler_status_recommendation_containerrecommendations_target_cpu"
                each:
                  type: Gauge
                  gauge:
                    path: [status, recommendation, containerRecommendations]
                    valueFrom: [target, cpu]
                    labelsFromPath:
                      container: [containerName]
                  commonLabels:
                    resource: "cpu"
                    unit: "core"
``` 
Although, for the future, I would want `metric-labels-allowlist` to support custom resources as well. Alternatively, we could add options to `KubeStateMetricsConfig` to allow users to explicitly enable all VPA labels and annotations metrics, albeit that could steer them into a crash state, so hardcoding on our end makes sense. That is, once `metric-labels-allowlist` supports custom resources upstream, [we can add `verticalpodautoscalers=[<specific labels>]`](https://github.com/openshift/cluster-monitoring-operator/blob/8db648fa23ade5453b87a1f58206f07be5791110/assets/kube-state-metrics/deployment.yaml#L42C13-L42C36) (a set of nominal ones and others on a case-by-case request basis).

rexagod:
The test seems to crash [here](https://github.com/openshift/cluster-monitoring-operator/blob/cf5d9a13643f83f3ca630f92af7476d8054d13b2/test/e2e/framework/client.go#L295) due to an empty `response` being returned, albeit with a `200` status, not sure why since this should return a `{} 1` time series. @simonpasquier Should we fallback to the previous implementation?

simonpasquier:
@rexgod maybe it should wait longer? It could be that the previous series isn't marked as stale and continues showing up in the result?

simonpasquier (commented):
FYI I tried (by accident) the PR with an hypershift-hosted cluster and the VPA pods couldn't be installed (because the components want to be installed in the master nodes) though the CRDs were installed.

````
E1206 14:24:27.329989 1 registry_factory.go:662] "kube_customresource_kube_verticalpodautoscaler_spec_updatepolicy_updatemode" err="[spec,updatePolicy,updateMode]: expected value for path to be string, got <nil>"
E1206 14:24:27.330032 1 registry_factory.go:662] "kube_customresource_kube_verticalpodautoscaler_status_recommendation_containerrecommendations_lowerbound_cpu" err="[status,recommendation,containerRecommendations]: got nil while resolving path"
E1206 14:24:27.330058 1 registry_factory.go:662] "kube_customresource_kube_verticalpodautoscaler_status_recommendation_containerrecommendations_lowerbound_memory" err="[status,recommendation,containerRecommendations]: got nil while resolving path"
E1206 14:24:27.330087 1 registry_factory.go:662] "kube_customresource_kube_verticalpodautoscaler_status_recommendation_containerrecommendations_upperbound_cpu" err="[status,recommendation,containerRecommendations]: got nil while resolving path"
E1206 14:24:27.330120 1 registry_factory.go:662] "kube_customresource_kube_verticalpodautoscaler_status_recommendation_containerrecommendations_upperbound_memory" err="[status,recommendation,containerRecommendations]: got nil while resolving path"
E1206 14:24:27.330154 1 registry_factory.go:662] "kube_customresource_kube_verticalpodautoscaler_status_recommendation_containerrecommendations_target_cpu" err="[status,recommendation,containerRecommendations]: got nil while resolving path"
E1206 14:24:27.330188 1 registry_factory.go:662] "kube_customresource_kube_verticalpodautoscaler_status_recommendation_containerrecommendations_target_memory" err="[status,recommendation,containerRecommendations]: got nil while resolving path"
````

**Review by simonpasquier — COMMENTED** [review]
I realize that the metrics are prefixed like this `kube_customresource_kube_...` which repeats a bit. Maybe it can be shorten? E.g. from `kube_customresource_kube_verticalpodautoscaler_...` to `kube_customresource_verticalpodautoscaler_...`

**Comment by rexagod** [comment]
Apologies for the push-ping, I pushed a bad commit and force-pushed again resetting the PR to the state before it. I'll rope in all the above suggestions into a commit and push.

**Comment by rexagod** [comment]
@simonpasquier Regarding https://github.com/openshift/cluster-monitoring-operator/pull/2078#pullrequestreview-1767734696, are you sure the VPA CRs were not installed? These errors should pop up whenever there's fields in the config cannot be populated from the CR, usually due to the absence of values. It is possible to create a VPA without a `targetRef` that will cause KSM to log these errors due to missing recommendations and other such values.

**Comment by rexagod** [comment]
Updated my comment above: https://github.com/openshift/cluster-monitoring-operator/pull/2078#issuecomment-1843256771.

**Comment by simonpasquier** [comment]
Thanks for the details @rexagod!  I did another test with a valid VPA object for which the controller updated the status subresource and indeed I see metrics now.

Does it mean that whenever kube-state-metrics fails to convert at least 1 metric of a VPA resource then no metrics at all will be exposed for this resource? As a side-note, it's not clear from the logs which VPA resource(s) were "dropped" by kube-state-metrics.

Otherwise I think that the e2e test fails because the `` metric isn't exposed anymore . I'd expect a `kube_customresource_verticalpodautoscaler_info` metric to be exposed instead?

File: jsonnet/components/kube-state-metrics.libsonnet [hunk 37]
```diff
@@ -202,6 +231,8 @@ function(params)
                           --metric-denylist=
                           ^kube_secret_labels$,
                           ^kube_.+_annotations$
+                          ^kube_customresource_.+_annotations_info$,
```
simonpasquier, line 234: it doesn't seem to filter annotations metrics?

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -202,6 +231,8 @@ function(params)
...
+                          ^kube_customresource_.+_labels_info$,
```
simonpasquier, line 235: not needed?

rexagod: IIUC Both of these will be needed to filter out the CRS `*_info` and `*_label` metrics: https://github.com/openshift/cluster-monitoring-operator/pull/2078#discussion_r1417208635.

jsonnet/components/kube-state-metrics.libsonnet
rexagod: Ah, I should've mentioned this. So, for testing purposes I didn't spin a cluster but used a local `kind` one, with the configuration in this PR. PTAL below for a set of metrics in an OpenShift 4.15 cluster with this patch applied.

rexagod: Ah, deny-listing doesn't work with CRS metrics. I've skipped both of these metric generations and pushed.

rexagod:
> Does it mean that whenever kube-state-metrics fails to convert at least 1 metric of a VPA resource then no metrics at all will be exposed for this resource?

Only the invalid metrics will be left out. I intentionally created a VPA with an invalid `targetRef`. In the screenshots below, it can be observed that only the erroneous metrics were dropped from the metrics final output.

<details>
<summary>Details</summary>

### KSM logs

Screenshot 2023-12-11 at 01 23 15

### SSH'ing into the KSM pod and querying metrics

Screenshot 2023-12-11 at 01 21 59

</details>

> As a side-note, it's not clear from the logs which VPA resource(s) were "dropped" by kube-state-metrics.

That's true. I think this was never addressed since we never took field modifications into the picture, eg., VPAs. I have opened an issue to resolve this: https://github.com/kubernetes/kube-state-metrics/issues/2262.

> Otherwise I think that the e2e test fails because the `` metric isn't exposed anymore . I'd expect a kube_customresource_verticalpodautoscaler_info metric to be exposed instead?

I've used the `verticalpodautoscaler_spec_updatepolicy_updatemode` metric for the tests in the latest patch. I'm not sure there's a `kube_customresource_verticalpodautoscaler_info` metric?

pkg/client/client_test.go
```diff
@@ -2238,7 +2238,8 @@ func TestPollUntil(t *testing.T) {
 
 	// the parent context times out.
 	var lastErr6 error
-	parentCtx, _ := context.WithTimeout(context.Background(), 10*time.Millisecond)
+	parentCtx, parentCancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
+	defer parentCancel()
```
rexagod, line 2242: PTAL https://github.com/openshift/cluster-monitoring-operator/pull/2234/files#diff-2dad64e28ce5e2eddf0adf303052d38e3b5b4fcba6b32fc4ca26d5e4d6b8fa4aR2373 (fails `go vet ./...`).

rexagod: Moved this (and the fix) to https://github.com/openshift/cluster-monitoring-operator/pull/2292.

rexagod:
Bump.

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -27,6 +40,22 @@ function(params)
...
     },
 
+    clusterRole+: {
+      rules+: [
+        {
+          apiGroups: ['autoscaling.k8s.io'],
+          resources: ['verticalpodautoscalers'],
+          verbs: ['list', 'watch'],
+        },
+        // https://github.com/kubernetes/kube-state-metrics/pull/1851/files#diff-916e6863e1245c673b4e5965c98dc27bafbd72650fdb38ce65ea73ee6304e027R45-R47
```
jan--f, line 50: Maybe add a comment what one can find behind this link? Something like `With VPA metrics KSM needs to watch customresourcedefinitions`?

pkg/client/client.go
```diff
@@ -1748,6 +1769,29 @@ func (c *Client) RegisterConsolePlugin(ctx context.Context, name string) error {
...
+		}
+
+		// If we don't, throw.
+		return nil, fmt.Errorf("failed to get %s CRD: %w", VerticalPodAutoscalerCRDMetadataName, err)
+	}
+
+	// VPA CRDs are present.
+	return ptr.To(true), nil
+}
+
```
jan--f, line 1794: I have two comments for this function:

1. Any reason not to return `(bool, err)`? Iiuc you make the bools to pointers here only to dereference them in [operator.go](https://github.com/openshift/cluster-monitoring-operator/pull/2078/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR767).
2. I'm not sure we should fall back to the last known return value here. Is adding/removing the VPA assets so expensive? I would argue that in an error scenario where resource retrieval doesn't work, our stack will likely have other problems.

rexagod: 1. I remember the function signature accommodates for the three states of `lastKnownVPACustomResourceDefinitionPresent` that may be returned, i.e., `nil` (we don't remember the last state, as it was never set), `false` (we do, and they were absent), and `true` (we do, and they were present) (this was discussed in a call with @simonpasquier way back).
2. So the reason we had to check for VPA CRD presence before enabling them is due to the fact that KSM will complain (PTAL at Simon's comment here: https://github.com/openshift/cluster-monitoring-operator/pull/2078#issuecomment-1825652771) about any enabled GVKs that are absent from the cluster.

Please correct me if I misunderstood either of these.

jsonnet/components/kube-state-metrics.libsonnet
rexagod: ACK, done.

jsonnet/components/kube-state-metrics.libsonnet
rexagod: ACK, done.

pkg/client/client.go
jan--f: Sure I get the need to check and that this check for nil as a signal that there is no cache value. Not sure I understand the need for caching this value though. If that has been discussed before, maybe we can add the reasoning as a comment for future readers?

rexagod: I was able to dig through the reviews and find a textual version of our discussion here: https://github.com/openshift/cluster-monitoring-operator/pull/2078#discussion_r1410364467.

Right, Kube API failing resource retrieval operations will most likely have more severe repercussions stack-wide than this. But I think, to the best of my understanding (and I maybe going out on a limb here), Simon meant the _transient_ error to occur within the span of a couple ticks (whilst this retrieval is happening) so it may not affect other concurrent tasks that are not retrieving resources during those same ticks.

LMK if you still think we should drop this and I'll patch it up.

PS. I've tried to sum things up in a comment here: https://github.com/openshift/cluster-monitoring-operator/pull/2078/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR194-R212.

jan--f: :+1: thanks for the context!

pkg/operator/operator.go
```diff
@@ -735,7 +775,31 @@ func (o *Operator) sync(ctx context.Context, key string) error {
...
-	factory := manifests.NewFactory(o.namespace, o.namespaceUserWorkload, config, o.loadInfrastructureConfig(ctx), proxyConfig, o.assets, apiServerConfig, consoleConfig)
+	// Enable kube-state-metrics' custom-resource-state-based metrics if VPA CRD is installed within the cluster.
+	enableKSMCRSMetrics := false
+	isVPACRDPresent, err := o.client.VPACustomResourceDefinitionPresent(ctx, o.lastKnownVPACustomResourceDefinitionPresent)
+	if err != nil {
+		// Throw on all transient errors.
+		return fmt.Errorf("unable to guess the desired state for kube-state-metrics' custom-resource-state metrics enablement: %w", err)
+	} else {
+		// If we didn't get an error, we can safely assume that the CRD is deterministically either present or absent.
+		if *isVPACRDPresent {
```
jan--f, line 786: Nit: Do we really need `enableKSMCRSMetrics` here? If `isVPACRDPresent != nil` update `o.lastKnownVPACustomResourceDefinitionPresent` and if its even true, log the message.

Makefile
```diff
@@ -18,6 +18,7 @@ GOLANG_FILES:=$(shell find . -name \*.go -print)
 # NOTE: grep -v %.yaml is needed  because "%s-policy.yaml" is used
 # in manifest.go and that isn't a valid asset.
 ASSETS=$(shell grep -oh '[^"]*/.*\.yaml' pkg/manifests/manifests.go \
+          | grep -v '^/etc' \
```
jan--f, line 21: Mind updating the comment above why this line was added.

jan--f:
Two more nits, if no one else has change requests we're good to go here.

jan--f (changes requested):

rexagod:
Rebased and squashed.

juzhao:
ci/prow/e2e-agnostic-operator failed, log: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2078/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-operator/1782600533010288640
```
 === RUN   TestKSMCRSMetrics
    kube_state_metrics_test.go:115: context deadline exceeded: error getting first value from response body "{\"status\":\"success\",\"data\":{\"resultType\":\"vector\",\"result\":[],\"analysis\":{}}}\n" for query "group(verticalpodautoscaler_spec_updatepolicy_updatemode{update_mode=\"Off\"} == 1)": expected body to contain single timeseries but got 0
--- FAIL: TestKSMCRSMetrics (69.17s)
```

rexagod:
Not a flake, this is failing due to a change I made in response to https://github.com/openshift/cluster-monitoring-operator/pull/2078#discussion_r1572390836. Working on it.

rexagod:
Checking within the container, I can see the metrics expected by the tests are generated, but the context still times out. I'll keep digging.

```
bash-5.1$ curl localhost:8081/metrics | grep verticalpodautoscaler_spec_updatepolicy_updatemode
# HELP kube_customresource_verticalpodautoscaler_spec_updatepolicy_updatemode Update mode of the VerticalPodAutoscaler.
# TYPE kube_customresource_verticalpodautoscaler_spec_updatepolicy_updatemode stateset
kube_customresource_verticalpodautoscaler_spec_updatepolicy_updatemode{customresource_group="autoscaling.k8s.io",customresource_kind="VerticalPodAutoscaler",customresource_version="v1",namespace="openshift-monitoring",target_api_version="apps/v1",target_kind="Deployment",target_name="cluster-monitoring-operator",updatemode="Auto",verticalpodautoscaler="test-vpa"} 0
kube_customresource_verticalpodautoscaler_spec_updatepolicy_updatemode{customresource_group="autoscaling.k8s.io",customresource_kind="VerticalPodAutoscaler",customresource_version="v1",namespace="openshift-monitoring",target_api_version="apps/v1",target_kind="Deployment",target_name="cluster-monitoring-operator",updatemode="Initial",verticalpodautoscaler="test-vpa"} 0
kube_customresource_verticalpodautoscaler_spec_updatepolicy_updatemode{customresource_group="autoscaling.k8s.io",customresource_kind="VerticalPodAutoscaler",customresource_version="v1",namespace="openshift-monitoring",target_api_version="apps/v1",target_kind="Deployment",target_name="cluster-monitoring-operator",updatemode="Off",verticalpodautoscaler="test-vpa"} 1
kube_customresource_verticalpodautoscaler_spec_updatepolicy_updatemode{customresource_group="autoscaling.k8s.io",customresource_kind="VerticalPodAutoscaler",customresource_version="v1",namespace="openshift-monitoring",target_api_version="apps/v1",target_kind="Deployment",target_name="cluster-monitoring-operator",updatemode="Recreate",verticalpodautoscaler="test-vpa"} 0
```

rexagod:
🤦🏼 

Screenshot 2024-04-23 at 11 28 02 PM

rexagod:
@jan--f Just to clarify, do you mean the Node team (Joel)? They were the ones with this request initially.

jan--f:
You added a hold a while back in https://github.com/openshift/cluster-monitoring-operator/pull/2078#issuecomment-1825559847

I just didn't want to preempt anything. Feel free to unhold unless you want anyone else's eyes on this.

rexagod:
ACK, I've pinged the autoscaling team to take a look. 🤞🏼

jan--f:
The single node test is not required...that won't block the merge. Only the `hold` label blocks this currently.

joelsmith:
LGTM, thanks!

rexagod:
/unhold

Thanks, Joel! Unholding now.

juzhao:
bug:https://issues.redhat.com/browse/OCPBUGS-32949

jan--f:
/hold
might be worth fixing the bug first

rexagod:
Simon and I went over this here: https://github.com/openshift/cluster-monitoring-operator/pull/2078#issuecomment-1843256771, PTAL.

juzhao:
> Simon and I went over this here: [#2078 (comment)](https://github.com/openshift/cluster-monitoring-operator/pull/2078#issuecomment-1843256771), PTAL.

thanks for the explanation, let the kube-state-metrics pod run for a few time, the error only shown once, delete the  kube-state-metrics pod, watch for a while, no such errors.

jan--f:
Thanks for the details.
/unhold
/retest

PR #2080, opened 2023-09-04
MON-2642: Improve e2e tests for alertrelabelconfigs CRD

machine424:
Check on Prometheus config secret instead of relabelConfig secret to test more things/parts.

### DIFF
--- a/test/e2e/alert_relabel_config_test.go
+++ b/test/e2e/alert_relabel_config_test.go
@@ -16,34 +16,37 @@ package e2e
 
 import (
 	"context"
-	"fmt"
-	"strings"
 	"testing"
 	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	osmv1 "github.com/openshift/api/monitoring/v1"
-	"github.com/pkg/errors"
-	"github.com/prometheus/prometheus/model/relabel"
-	"gopkg.in/yaml.v2"
-
 	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
+	"github.com/pkg/errors"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/model/relabel"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	relabelSecretName = "alert-relabel-configs"
-	relabelSecretKey  = "config.yaml"
+	relabelConfigName          = "test-relabel-config"
+	relabelSecretName          = "alert-relabel-configs"
+	relabelSecretKey           = "config.yaml"
+	prometheusConfigSecretName = "prometheus-k8s"
 )
 
-func TestAlertRelabelConfigTechPreview(t *testing.T) {
+func TestAlertRelabelConfig(t *testing.T) {
+	initialRelabelConfig := prometheusRelabelConfig(t)
+
+	// By default we drop prometheus_replica label + add openshift_io_alert_source = 2
+	require.Len(t, initialRelabelConfig, 2)
+
 	ctx := context.Background()
-
-	arcName := framework.E2eTestLabelValue
-
 	arc := &osmv1.AlertRelabelConfig{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      arcName,
+			Name:      relabelConfigName,
 			Namespace: f.Ns,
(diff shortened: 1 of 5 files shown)


danielmellado:
Nice addition of testify over there, my only concern would be that the usage of this is homogeneous all along the tests. I'd say yes to this and consider this a start PR ;)

test/e2e/framework/prometheus.go
```diff
@@ -133,3 +140,20 @@ func (f Framework) MakePrometheusServiceRoute(svc *v1.Service) *routev1.Route {
 		},
 	}
 }
+
+func (f Framework) PrometheusConfigFromSecret(t *testing.T, namespace, secretName string) (*promConfig.Config, error) {
+	rawConfig := f.MustGetSecret(t, secretName, namespace).Data["prometheus.yaml.gz"]
```
simonpasquier, line 145:
```suggestion
	t.Helper()

	rawConfig := f.MustGetSecret(t, secretName, namespace).Data["prometheus.yaml.gz"]
```

test/e2e/alert_relabel_config_test.go
```diff
@@ -56,124 +61,70 @@ func TestAlertRelabelConfigTechPreview(t *testing.T) {
...
-
-		return nil
-	})
+	// Check Prometheus config forgot about the deleted AlertRelabelConfig.
+	validateCurrentRelabelConfig(t, initialRelabelConfig)
+}
 
+// prometheusRelabelConfig returns the alert relabel configuration part used by Prometheus config
+func prometheusRelabelConfig(t *testing.T) ([]*relabel.Config, error) {
+	prometheusConfig, err := f.PrometheusConfigFromSecret(t, f.Ns, prometheusConfigSecretName)
```
simonpasquier, line 107: If the function gets a t parameter, we can check the error directly?
```suggestion
	t.Helper()

	prometheusConfig, err := f.PrometheusConfigFromSecret(t, f.Ns, prometheusConfigSecretName)
	require.NoError(t, err)
```

test/e2e/alert_relabel_config_test.go
```diff
@@ -56,124 +61,70 @@ func TestAlertRelabelConfigTechPreview(t *testing.T) {
...
-		return fmt.Errorf("TargetLabel does not match (%q != %q)", arc.TargetLabel, fromSecret.TargetLabel)
-	}
+		if !assert.ElementsMatch(t, expectedRelabelConfig, currentRelabelConfig) {
+			return errors.New("the expected relabel config is not applied yet.")
+		}
+		return nil
+	})
 
-	arcRegex, err := relabel.NewRegexp(arc.Regex)
 	if err != nil {
```
simonpasquier, line 127: (nit) could use `require.NoError(t, err)`?

simonpasquier (commented):
Will the test also check that the resulting alert in Alertmanager is modified as expected?

test/e2e/framework/prometheus.go
machine424: Cool for the `t.Helper()`

test/e2e/alert_relabel_config_test.go
machine424: Now that there is no objection to testify, I migrated this and some others.

test/e2e/alert_relabel_config_test.go
machine424: I thought about returning an error to be able to retry this as well if there are any intermittent problems, do you think we should be aware of those and fail from the first time?

simonpasquier: hmm right after reading again `PrometheusConfigFromSecret()` it doesn't seem to use to the `t` parameter so probably the best option is to remove it from the signature?

machine424: Actually we pass it to `MustGetSecret` and we use it for `t.Helper()` now.

simonpasquier: Ah yes. Then maybe `PrometheusConfigFromSecret()` shouldn't return an error? If the secret data is invalid (e.g. can't be unzipped/deserialized), we should fail hard with `t.Fail`?

machine424: That was my question :) I see that `MustGetSecret` already has retries that should deal with network/apiserver `intermittent problems`, we're left with unzipping/deserialization problems, which for sure we don't want to hide/we can never hide. Done.

simonpasquier: :+1:

simonpasquier (commented):
/lgtm

simonpasquier:

no impact on the release payload.

PR #2081, opened 2023-09-05
OCPBUGS-18390: jsonnet/rules: exclude -1 from etcd objects count

dgrisonnet:
The apiserver_storage_objects metric can take -1 values when there are
communication issues between the kube-apiserver and etcd. In such case,
the kube-apiserver can't get objects count from etcd and as such exposes
a metric value denoting the issue. However, this is polluting the data
exported to telemetry so they should be skipped. Doing that will not
cause the lose of the etcd issue information since we have alerts that
would trigger in such scenarios.

### DIFF
--- a/assets/cluster-monitoring-operator/prometheus-rule.yaml
+++ b/assets/cluster-monitoring-operator/prometheus-rule.yaml
@@ -194,9 +194,9 @@ spec:
           )
         )
       record: cluster:kubelet_volume_stats_used_bytes:provisioner:sum
-    - expr: sum by (instance) (apiserver_storage_objects)
+    - expr: sum by (instance) (apiserver_storage_objects != -1)
       record: instance:etcd_object_counts:sum
-    - expr: topk(500, max by(resource) (apiserver_storage_objects))
+    - expr: topk(500, max by(resource) (apiserver_storage_objects != -1))
       record: cluster:usage:resources:sum
     - expr: count(count (kube_pod_restart_policy{type!="Always",namespace!~"openshift-.+"}) by (namespace,pod))
       record: cluster:usage:pods:terminal:workload:sum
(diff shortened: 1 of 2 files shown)


assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -194,9 +194,9 @@ spec:
           )
         )
       record: cluster:kubelet_volume_stats_used_bytes:provisioner:sum
-    - expr: sum by (instance) (apiserver_storage_objects)
+    - expr: sum by (instance) (apiserver_storage_objects != -1)
```
simonpasquier, line 197: just to be safe?
```suggestion
    - expr: sum by (instance) (apiserver_storage_objects >= 0)
```

simonpasquier (commented):
Thanks for tackling it, we (monitoring team) could have fixed it :)

dgrisonnet:
> Is -1 the best value to return in that case?

I don't think so. Personally in that kind of scenario I would not return value for the metric, produce a log line and rely on other metrics to detect connection problems with etcd and the kube-apiserver. However, since this metric has been in upstream for quite a while, I'd rather keep that `-1` value in case people rely on it

machine424:
Yep, stop exposing the metric or returning an empty metric make more sense.
Should we document it then? I see other metrics talk about special cases here: https://kubernetes.io/docs/reference/instrumentation/metrics/ unless we don't want more users to rely on it to make any potential change easier.

dgrisonnet:
That's fine by me, do you want to send the PR?

machine424:
Yes, I can take care of it.

machine424:
(just to close the loop, the doc change was made here https://github.com/kubernetes/kubernetes/pull/120597#issuecomment-1715361581)
/lgtm

PR #2090, opened 2023-09-19
OCPBUGS-19059: Enable ipv6 on monitoring-plugin nginx

danielmellado:
This commit modifies the monitoring-plugin configmap so nginx listens on
both ipv4 and ipv6 addresses. The nginx image used supported ipv6 but it
wasn't enabled on the endpoint config.

### DIFF
--- a/assets/monitoring-plugin/config-map.yaml
+++ b/assets/monitoring-plugin/config-map.yaml
@@ -9,6 +9,7 @@ data:
       keepalive_timeout  65;
       server {
         listen              9443 ssl;
+        listen              [::]:9443 ssl;
         ssl_certificate     /var/cert/tls.crt;
         ssl_certificate_key /var/cert/tls.key;
         root                /usr/share/nginx/html;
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
/lgtm

Additional question: how can we ensure with e2e tests that the console -> monitoring plugin communication works as expected?

danielmellado:
> /lgtm
> 
> Additional question: how can we ensure with e2e tests that the console -> monitoring plugin communication works as expected?

I'll take a look at the current tests and see how feasible is to address that in a follow up patch, thanks!

simonpasquier:
/skip

I'd expect the test to live in openshift/origin. Probably need to sync with the console team too as it looks like a generic check for all console plugins.

Tai-RedHat:
> /lgtm
> 
> Additional question: how can we ensure with e2e tests that the console -> monitoring plugin communication works as expected?

I will do a regression test for that

PR #2094, opened 2023-09-22
Set Thanos version in the Thanos Ruler spec

simonpasquier:
The `.spec.version` field is used by the operator to adapt the statefulset configuration depending on which features are supported or not.

In practice, it shouldn't induce any change in functionality but it's always better to be explicit.


slashpai (commented):
/lgtm

PR #2096, opened 2023-09-25
Add alert to detect multiple operators accesing prometheus crd

marioferh:
New Alert AlertmanagerMultipleOperatorsAccesingPrometheusCRD

Alert to detect many pod restart dude to multiple operators accessing to the prometheus CRD

### DIFF
--- a/assets/cluster-monitoring-operator/prometheus-rule.yaml
+++ b/assets/cluster-monitoring-operator/prometheus-rule.yaml
@@ -277,6 +277,14 @@ spec:
       for: 30m
       labels:
         severity: warning
+    - alert: MultipleOperatorsAccesingPrometheusCRD
+      annotations:
+        description: Pod {{ $labels.namespace }}/{{ $labels.pod }} restarted more than 5 times in last 10m
+        summary: Detect multiple operators accessing to the prometheus CRD.
+      expr: sum(max by(namespace, container, pod) (increase(kube_pod_container_status_restarts_total[10m]))) > 5
+      for: 10m
+      labels:
+        severity: warning
   - interval: 30s
     name: kubernetes-recurring.rules
     rules:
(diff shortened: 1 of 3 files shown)


assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -277,6 +277,14 @@ spec:
       for: 30m
       labels:
         severity: warning
+    - alert: MultipleOperatorsAccesingPrometheusCRD
+      annotations:
+        description: Pod {{ $labels.namespace }}/{{ $labels.pod }} restarted more than 5 times in last 10m
+        summary: Detect multiple operators accessing to the prometheus CRD.
+      expr: sum(max by(namespace, container, pod) (increase(kube_pod_container_status_restarts_total[10m]))) > 5
```
simonpasquier, line 284: In practice, this creates a new alert for all pods in the cluster while we want a more targeted approach.

But the crux of the issue is that when a 3rd-party Prometheus operator messes up with the openshift-monitoring operator, the whole monitoring stack is down (because prometheus-k8s pods are repeatedly recreated) and the alert will never fire.

My suggestion would be to report the issue in the Prometheus status conditions and trickle it down in the ClusterOperator monitoring resource. The detection could be based on the statefulset generation number: if the operator records the last generation number that it has seen, it could detect that something/someone is stomping on the statefulset objects. At least a cluster admin could see the information with the oc command-line or in the console.

Something like this:

````
kind: Prometheus
metadata:
  name: k8s
spec:
  ...
status:
  conditions:
  - lastTransitionTime: "2023-11-29T14:04:20Z"
    message: "The statefulset(s) \"prometheus-k8s\" are updated by an external actor which prevents the pods from being ready. It can be that another Prometheus operator runs in the cluster and watches for the resource. To fix the issue, either stop the other operator or update its configuration to restrict the watched resources." # To be discussed in more details
    observedGeneration: 4
    reason: "ConcurrentUpdate"
    status: "Degraded" # or maybe "False"
    type: Reconciled
 - ...
 ````

PR #2097, opened 2023-09-26
WIP: OCPBUGS-18282: Reject reserved labels used as external labels

slashpai:
prometheus and prometheus_replica are reserved labels, if a user configures this as a label in externalLabels in cluster-monitoring-configmap warn user about this and make operator status Degraded

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -736,6 +736,9 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 	} else if config.HasInconsistentAlertmanagerConfigurations() {
 		degradedConditionMessage = client.UserAlermanagerConfigMisconfiguredMessage
 		degradedConditionReason = client.UserAlermanagerConfigMisconfiguredReason
+	} else if config.HasPrometheusReservedExternalLabelsConfigured() {
+		degradedConditionMessage = client.CannotUseReservedExternalLabelsMessage
+		degradedConditionReason = client.CannotUseReservedExternalLabelsReason
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/operator/operator.go
```diff
@@ -736,6 +736,9 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 	} else if config.HasInconsistentAlertmanagerConfigurations() {
 		degradedConditionMessage = client.UserAlermanagerConfigMisconfiguredMessage
 		degradedConditionReason = client.UserAlermanagerConfigMisconfiguredReason
+	} else if config.HasPrometheusReservedExternalLabelsConfigured() {
```
machine424, line 739: How about integrating this to the PrometheusTask? Make the whole task fail if the labels are set?

machine424: Yes, but this won't prevent the prometheus task from applying the faulty labels (which may in some setups break the delivery of alerts perhaps.).
Adding the check at the beginning of the prometheus task will prevent that.
Now that I think about it, doing so may also break legitimate/needed prometheus syncs...

By the way, the `SetRollOutDone` below sets `Degraded=false`, so no "default" alert will be triggered for this.

> One orthogonal issue though: This condition reporting here being a big else-if, doesn't this potentially hide issues. E.g. if no storage is configured and the forbidden external labels are set, we only ever see the storageNotConfiguredMessage and CannotUseReservedExternalLabelsMessage only pops up after storage was configured. Shouldn't we collect all conditions and report them all at once?

Good point, we could group them under some "umbrella"  reason and only join the messages, it's worth a ticket so we can discuss this.

PR #2100, opened 2023-10-03
MON-3398: Add RHACM telemetry metric

zlayne:

I am not sure if the addition of this metric requires a changelog update. If so please let me know.
Related issue: https://issues.redhat.com/browse/MON-3398


manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -694,6 +694,11 @@ data:
     # acm_managed_cluster_info provides Subscription watch and other information for the managed clusters for an ACM Hub cluster.
     - '{__name__="acm_managed_cluster_info"}'
     #
+    # owners: (https://github.com/orgs/stolostron/teams/search-admin, @acm-observability-search)
+    #
+    # acm_console_page_count:sum counts the total number of visits for each page in ACM console.
+    - '{__name__="acm_console_page_count:sum"}'
```
simonpasquier, line 700: I'd be more comfortable if the expression limits the possible values for the `page` label to avoid excessive cardinality in the future.

cc @jan--f

zlayne: Hey, in creating this metric we had purposely left it dynamic for new page labels. At the moment we only use a max of 7 pages - 1 for each of the main features of our product. 

If a current cardinality of 7 is too high, do you have any recommendations for alternatives that would satisfy the cardinality requirements? I suppose we could create a metric for each page, but this feels like it might be overkill for the purpose of the metric.

simonpasquier: In terms of cardinality, 1 metric per page is the same as 1 metric with 7 different label values. We need to be explicit about which label values are permitted otherwise we're at risk of blowing up the Telemeter server.
Another option would be to aggregate "pages" into a fixed set of groups.

zlayne: Thanks for the response. Could you provide a little more info/example for `aggregate "pages" into a fixed set of groups`? I am not sure I quite follow what you are asking. 

On our side, we only allow for 7 pages and sum the total counts so the metric data looks as follows. But to be even more explicit I guess we could add something like: `{page=~"page1,page2"}` to the metric collected here?


Screenshot 2023-10-31 at 10 25 11 AM

zlayne: > for example you could sum the "overview." label values into an "overview" value, "search." into "search" and the rest into "other".

I *think* that's what I have done? Unless you are meaning instead of having a page label at all, we use a label for Overview, search, etc?

If not, I can certainly update the config to explicitly list the pages.

simonpasquier: Right now the recording rule sums per individual page. What I was suggesting is to group related pages under a common/generic label value.

````
- expr: sum(acm_console_page_count{page=~"overview.*"})
  name: acm_console_page_count:sum
  labels:
    page: overview
- expr: sum(acm_console_page_count{page=~"search.*"})
  name: acm_console_page_count:sum
  labels:
    page: search
- expr: sum(acm_console_page_count{page!~"overview.*|search.*"})
  name: acm_console_page_count:sum
  labels:
    page: other
`````

simonpasquier:
/lgtm

I'll defer to @jan--f for approval

PR #2102, opened 2023-10-04
MON-3270: Add kube-metrics-server to image-references

slashpai:
Since ART build is completed inorder to get
included in payload it needs to be referred in
image-references.

### DIFF
--- a/hack/generate-versions.sh
+++ b/hack/generate-versions.sh
@@ -28,7 +28,13 @@ if [ "$PULL_BASE_REF" != "$MAIN_BRANCH" ]; then
 fi
 
 version_from_remote() {
-	curl --retry 5 --silent --fail "https://raw.githubusercontent.com/${1}/${MAIN_BRANCH}/VERSION"
+  # since kubernetes-metrics-server repo doesn't have a VERSION file
+  if [ "$1" == "openshift/kubernetes-metrics-server" ]; then
+    curl --retry 5 --silent --fail "https://raw.githubusercontent.com/${1}/${MAIN_BRANCH}/manifests/release/kustomization.yaml" | \
+      gojsontoyaml -yamltojson | jq '.images[0].newTag' | sed 's/"//g'
+  else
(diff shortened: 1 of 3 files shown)


hack/generate-versions.sh
```diff
@@ -28,7 +28,12 @@ if [ "$PULL_BASE_REF" != "$MAIN_BRANCH" ]; then
 fi
 
 version_from_remote() {
-	curl --retry 5 --silent --fail "https://raw.githubusercontent.com/${1}/${MAIN_BRANCH}/VERSION"
+	# since kubernetes-metrics-server repo doesn't have a VERSION file
+	if [ "$1" == "openshift/kubernetes-metrics-server" ]; then
+    curl --retry 5 --silent --fail  "https://raw.githubusercontent.com/${1}/${MAIN_BRANCH}/manifests/release/kustomization.yaml"|gojsontoyaml -yamltojson|jq '.images[0].newTag'|sed 's/"//g'
```
danielmellado, line 33: just a tiny nit here, could we format this expression so it ends like:

version_from_remote() {
  if [ "$1" == "openshift/kubernetes-metrics-server" ]; then
    curl --retry 5 --silent --fail "https://raw.githubusercontent.com/${1}/${MAIN_BRANCH}/manifests/release/kustomization.yaml" | \
      gojsontoyaml -yamltojson | jq '.images[0].newTag' | sed 's/"//g'
  else
    curl --retry 5 --silent --fail "https://raw.githubusercontent.com/${1}/${MAIN_BRANCH}/VERSION"
  fi
}

slashpai: updated

jan--f:
+1 to Danis formatting request, otherwise the lgtm.
Might be worth to propose a `VERSION` file upstream? Seems like many projects do it but not enough to call it a convention.

slashpai:
> +1 to Danis formatting request, otherwise the lgtm. Might be worth to propose a `VERSION` file upstream? Seems like many projects do it but not enough to call it a convention.

Since for upstream VERSION file didn't made much sense it was hard to push for that one :)

Addressed Dani comments

PR #2105, opened 2023-10-05
OCPVE-711: feat: add olm capability annotation

eggfoobar:
OLM is an optional operator, we exclude OLM resources with this annotation if the monitoring operator ever becomes a capability, using a <monitoring_label>+OperatorLifecycleManager will work for both

Openshift has the capabilities feature documented in this [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/installer/component-selection.md), This allows us to annotate manifests for specific operators and make them optional during install. The reason that this PR exists is because the OperatorLifecycleManager is going to be optional in 4.15, the deployment of OperatorGroup fails during install, so we add this annotation and CVO will know how to handle it correctly.

/hold

Wait for CVO to be merged in first https://github.com/openshift/cluster-version-operator/pull/971

### DIFF
--- a/manifests/0000_90_cluster-monitoring-operator_00-operatorgroup.yaml
+++ b/manifests/0000_90_cluster-monitoring-operator_00-operatorgroup.yaml
@@ -8,6 +8,7 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     olm.providedAPIs: Alertmanager.v1.monitoring.coreos.com,PodMonitor.v1.monitoring.coreos.com,Probe.v1.monitoring.coreos.com,Prometheus.v1.monitoring.coreos.com,PrometheusRule.v1.monitoring.coreos.com,ServiceMonitor.v1.monitoring.coreos.com,ThanosRuler.v1.monitoring.coreos.com,AlertmanagerConfig.v1alpha1.monitoring.coreos.com
     include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: "OperatorLifecycleManager"
 spec:
   staticProvidedAPIs: true
   selector:


danielmellado (commented):
Hi, althought it seems pretty much straightforward... would you mind clarifying the need for this? I'd love to get the PR message expanded with some more context, thanks!

eggfoobar:
Hey @danielmellado, Sure thing! Openshift has the capabilities feature documented in this [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/installer/component-selection.md), This allows us to annotate manifests for specific operators and make them optional during install. The reason that this PR exists is because the OperatorLifecycleManager is going to be optional in 4.15, the deployment of OperatorGroup fails during install, so we add this annotation and CVO will know how to handle it correctly.

If all goods from your end, feel free to slap a lgtm/approve, we're getting final validation and I'll go ahead and unhold it once the change is in CVO.

danielmellado:
@eggfoobar thanks for addressing my comment there and adding the clarification, it does lgtm but I'd like for @jan--f or @simonpasquier to take a look at this too, thanks!

machine424:
Are you sure this is sufficient to avoid deploying that resource once a cluster disable OLM? I'm afraid this would implicitly re-enable OLM back:
cf https://issues.redhat.com/browse/OCPBUGS-18326?focusedId=23265807&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23265807

I think this is documented in here https://github.com/openshift/enhancements/blob/cb81452fddf86c1099acd87610b88369cd6192db/enhancements/installer/component-selection.md#updates but maybe it's something worth mentioning in https://github.com/openshift/enhancements/tree/master/dev-guide/cluster-version-operator as well (I'm not sure, maybe I'm getting it wrong, but a more explicit doc will in all cases)

PR #2106, opened 2023-10-05
WIP: Detect static pods from metrics

dgrisonnet:
The kubernetes.io/config.seen annotation on pods is set by kubelet and
can be helpful to identify static pods. For static pods, this annotation
will be equal to file since their configurations are files on disk.
Exposing this information via kube-state-metrics would allow correlating
the annotation metrics with other metric to build alerts specific to
static pods.

### DIFF
--- a/jsonnet/components/kube-state-metrics.libsonnet
+++ b/jsonnet/components/kube-state-metrics.libsonnet
@@ -201,9 +201,9 @@ function(params)
                         |||
                           --metric-denylist=
                           ^kube_secret_labels$,
-                          ^kube_.+_annotations$
                         |||,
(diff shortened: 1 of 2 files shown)


jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -201,9 +201,9 @@ function(params)
                         |||
                           --metric-denylist=
                           ^kube_secret_labels$,
-                          ^kube_.+_annotations$
```
dgrisonnet, line 204: This will not have any impact once o/ksm is bumped to v2.10.0 since we remove these metrics by default now: https://github.com/kubernetes/kube-state-metrics/releases/tag/v2.10.0

PR #2112, opened 2023-10-09
build: add `-e` to `go list` command to install tools

jan--f:
we rely on go imports to install our build tooling via hack/tools/go.mod. Some of these imports don't actually import valid modules since `package main` is used in some of them (e.g.
https://github.com/brancz/gojsontoyaml/blob/6ffa820a00533125ad1c3f77d0429e4cac6f7072/main.go#L1). With go1.21 the `list` command will fail on those. Adding `-e` fixes this. I think this flag exists in older go version too, so this should work without being aware of the version.

\cc @simonpasquier 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -258,6 +258,6 @@ $(JSON_MANIFESTS_DIR):
 
 $(TOOLING): $(BIN_DIR)
 	@echo Installing tools from hack/tools.go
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
/lgtm

PR #1875, opened 2023-01-31
OCPBUGS-4069: Prometheus Adatper takes metrics from "kubelet" job only

raptorsun:
* [x] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #2117, opened 2023-10-12
OCPBUGS-4069: Prometheus Adatper takes metrics from kubelet job only.

raptorsun:
The same modification in CMO as https://github.com/openshift/cluster-monitoring-operator/pull/1875 but using the new option in upstream kube-prometheus introduced by [this PR](https://github.com/prometheus-operator/kube-prometheus/pull/2003).

### DIFF
--- a/assets/prometheus-adapter/config-map-dedicated-service-monitors.yaml
+++ b/assets/prometheus-adapter/config-map-dedicated-service-monitors.yaml
@@ -7,7 +7,7 @@ data:
         "containerQuery": |
           sum by (<<.GroupBy>>) (
             irate (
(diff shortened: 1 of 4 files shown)


marioferh (commented):
/lgtm

simonpasquier:
@raptorsun I fear that the change breaks when the user configures a [dedicated](https://docs.openshift.com/container-platform/4.13/monitoring/config-map-reference-for-the-cluster-monitoring-operator.html#dedicatedservicemonitors) service monitor for Prometheus adapter.

simonpasquier:
@machine424 you're correct! And we have an e2e test validating the dedicated service monitor so all should be good.

PR #2118, opened 2023-10-13
OCPBUGS-18326: revert previous fix to avoid activating console capability during update

raptorsun:
Revert PR#2101
As Trevor commented in https://issues.redhat.com/browse/OCPBUGS-18326 , the capability annotation will activate the required capability implicitly on clusters during an update. We should avoid activating capabilities during update.


raptorsun:
/close 
Will look for better solutions in the 4.15 development cycle.

PR #2128, opened 2023-10-19
fix: force HTTP/1.1 connections

simonpasquier:
Commit b3b011b0 was not enough to prevent HTTP2 connections as demonstrated in https://github.com/kubernetes/kubernetes/issues/121197.

To prevent any risk with HTTP2, this change disables HTTP2 at the server level. There's no expected performance penalty because the web server is only used for scraping metrics.

### DIFF
--- a/pkg/metrics/server.go
+++ b/pkg/metrics/server.go
@@ -78,6 +78,11 @@ func (s *Server) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	// Mitigate CVE-2023-44487 by disabling HTTP2 until the Go standard library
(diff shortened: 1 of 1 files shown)


machine424:
I suppose we don't want to do the same for the `/debug` server https://github.com/openshift/cluster-monitoring-operator/blob/3cc8dd9d1e0196377fbedf1b0b857f5a6057ceb0/cmd/operator/main.go#L190-L197 as it listens on localhost (if you can reach it you can make more damage than OOM)

/lgtm

simonpasquier:
@machine424 yes localhost listeners aren't an issue

PR #2129, opened 2023-10-19
MON-3421: improve the detection of CMO unavailability/degradation due to U…

machine424:
…WM components

see the proposal doc attached to the ticket for more details.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -27,7 +27,6 @@ import (
 	"github.com/openshift/cluster-monitoring-operator/pkg/client"
 	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
 	"github.com/openshift/cluster-monitoring-operator/pkg/metrics"
-	cmostr "github.com/openshift/cluster-monitoring-operator/pkg/strings"
 	"github.com/openshift/cluster-monitoring-operator/pkg/tasks"
 	"github.com/openshift/library-go/pkg/operator/csr"
 	"github.com/openshift/library-go/pkg/operator/events"
@@ -151,6 +150,8 @@ const (
 
 	// Canonical name of the cluster-wide infrastructure resource.
 	clusterResourceName = "cluster"
+
+	UWMTaskPrefix = "UpdatingUserWorkload"
 )
 
 type Operator struct {
@@ -654,6 +655,14 @@ func getProxyReader(ctx context.Context, config *manifests.Config, proxyConfigSu
 	return clusterProxyConfig
 }
 
+func newTaskSpec(targetName string, task tasks.Task) *tasks.TaskSpec {
+	return tasks.NewTaskSpec("Updating"+targetName, task)
+}
+
+func newUWMTaskSpec(targetName string, task tasks.Task) *tasks.TaskSpec {
+	return tasks.NewTaskSpec(UWMTaskPrefix+targetName, task)
+}
+
 func (o *Operator) sync(ctx context.Context, key string) error {
 	config, err := o.Config(ctx, key)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/tasks/tasks.go
```diff
@@ -102,25 +102,40 @@ type TaskGroup struct {
...
-		Name: name,
-		Task: task,
+		Name:       fmt.Sprintf("Updating%s", target),
+		UWMRelated: false,
+		Task:       task,
+	}
+}
+
+// NewUWMTaskSpec creates a task that has UWM components as a target
+func NewUWMTaskSpec(target string, task Task) *TaskSpec {
```
machine424, line 115: Maybe we want to keep `tasks.go` UWM stuff-agnostic? I can try another approach

pkg/operator/operator.go
```diff
@@ -689,26 +688,27 @@ func (o *Operator) sync(ctx context.Context, key string) error {
...
+				tasks.NewTaskSpec("Alertmanager", tasks.NewAlertmanagerTask(o.client, factory, config)),
+				tasks.NewTaskSpec("NodeExporter", tasks.NewNodeExporterTask(o.client, factory)),
+				tasks.NewTaskSpec("KubeStateMetrics", tasks.NewKubeStateMetricsTask(o.client, factory)),
+				tasks.NewTaskSpec("OpenshiftStateMetrics", tasks.NewOpenShiftStateMetricsTask(o.client, factory)),
+				tasks.NewTaskSpec("PrometheusAdapter", tasks.NewPrometheusAdapterTask(ctx, o.namespace, o.client, factory, config)),
+				tasks.NewTaskSpec("TelemeterClient", tasks.NewTelemeterClientTask(o.client, factory, config)),
+				tasks.NewTaskSpec("ThanosQuerier", tasks.NewThanosQuerierTask(o.client, factory, config)),
+				tasks.NewTaskSpec("ControlPlaneComponents", tasks.NewControlPlaneTask(o.client, factory, config)),
+				tasks.NewTaskSpec("ConsolePluginComponents", tasks.NewMonitoringPluginTask(o.client, factory, config)),
+				// TODO: maybe put the operator in the first group
```
machine424, line 707: WDYT?

Also tell me if I should leave the reordering to another PR

pkg/operator/operator_test.go
```diff
@@ -281,3 +285,292 @@ func TestRunReport(t *testing.T) {
...
+			expectedReport: runReport{
+				degraded: &stateInfo{
+					reason:   "UserWorkloadTasksFailed",
+					status:   "True",
+					messages: []string{"UpdatingUserWorkloadFoo: Foo failed", "UpdatingUserWorkloadBar: Bar failed"},
+				},
+				available: &stateInfo{
+					reason:   "UserWorkloadTasksFailed",
+					status:   "False",
+					messages: []string{"UpdatingUserWorkloadFoo: Foo failed", "UpdatingUserWorkloadBar: Bar failed"},
```
machine424, line 433: Added the task name to the `messages` to make it clearer in case of multiple tasks failing (also in case of 1 task to keep the code simple)

machine424:
Made some manual e2e tests as well, was thinking about add some e2e tests, but the CO reconciliation takes time and sometimes it never occurs (when e.g. the `conditions[].message` changes but not the `conditions[].reason`).
I'll create a ticket for that bug and maybe I'll add some e2e tests for this, but in another PR.

I'll also look into adding some new lines into to `conditions[].message` make them more readable https://github.com/openshift/api/blob/f404f2bc3524e03f3bc29a7b3cd0f624cd36eae7/config/v1/types_cluster_operator.go#L138-L142

/cc @simonpasquier 
/cc @jan--f 

since you participated in the conversation from the beginning.

pkg/tasks/tasks.go
danielmellado: I think it's fine as long as we mark whatever comes from UWM as such, as you're stating there.

jan--f: We don't use `tasks.go` anywhere else so there is generally no need to keep things decoupled. One observation though: `operator.go` is really the only component that needs to know which part (platform or UWM) a specific task updates. `NewUWMTaskSpec` is really just a convenience method that sets `UWMRelated: true` and a specfic name prefix.

Couldn't all this be handled from `operator.go`? We could leave `tasks.go` unchanged and have `operator.go` gain a convenience method that adds a UWM specific prefix to a task. Based on that prefix we can also alter the reporting to differentiate between platform and UWM errors.

machine424: Sure, here is a version https://github.com/openshift/cluster-monitoring-operator/compare/master...machine424:cluster-monitoring-operator:uwm-tasks-2?expand=1

The only concern I have is that using/relying on prefixes always seem hacky/error-prone even though we control the whole chain...

I think we can reduce the functions composition `tasks.NewTaskSpec(TaskName("...` e.g., by wrapping the whole in a function, sth like https://github.com/openshift/cluster-monitoring-operator/compare/master...machine424:cluster-monitoring-operator:uwm-tasks-3?expand=1

WDYT?

machine424: @jan--f (in case this went unnoticed)

jan--f:
Apologies for the long lag!
I quite like the `uwm-tasks-3` version. To me this is the change that is straight forward to read and follow. The only comment there I'd have is to not export the three convenience function `NewTaskSpec`, `NewUWMTaskSpec` and `IsUWMTaskErr`. 
One thing to consider is maybe to change an interface: `func IsUWMTaskErr(task task.TaskErr) bool`. That might make it easier to change the implementation of what is a UWM error later. But at this point this is really bikeshedding :)

pkg/operator/operator.go
jan--f: I think this can be in here. Is this an issue currently? If so a description of how this improves things in a separate commit would be good to have.

machine424: done

pkg/operator/operator.go
```diff
@@ -996,50 +1005,72 @@ func toStateErrors(err error) []*client.StateError {
...
+		case len(tge) == 1:
+			return tge[0].Name + "Failed"
+		case uwmErrCount == totalErrCount:
+			return "UserWorkloadTasksFailed"
+		case uwmErrCount == 0:
+			return "PlatformTasksFailed"
+		default:
+			// If no errors to report, this will be ignored anyway
+			return defaultReason
+		}
```
jan--f, line 1062: I find it tricky to follow all the implicit assumptions for this function. Purely for the sake of readability, what do you think about:
- Set a `defaultReason := tge[0].Name + "Failed"

and 
```suggestion
	inferReason := func(uwmErrCount, totalErrCount int) string {
		switch {
		case uwmErrCount == totalErrCount:
			return "UserWorkloadTasksFailed"
		case uwmErrCount == 0:
			return "PlatformTasksFailed"
	        case totalErrCount != 0:
	                return "MultipleTasksFailed"
		default:
			return defaultReason
		}
```

I think this requires less context (no need to remember what `tge` is.

machine424: In case only one task is failing (len(tge)=1), we want to return its name, with the proposed change, this will override the "default" with UserWorkloadTasksFailed or PlatformTasksFailed

`one_failing_UWM_task_with_a_generic_error` case in tests e.g. will fail, you can run tests with `go test -v ./pkg/operator -run TestGenerateRunReportFromTaskErrors` (after removing the `No errors returned` case)

jan--f: ahh you are right of course. This _is_ tricky to understand and change :laughing: 
```suggestion
	inferReason := func(uwmErrCount, totalErrCount int) string {
		switch {
		case totalErrCount == 1:
			return defaultReason
		case uwmErrCount == totalErrCount:
			return "UserWorkloadTasksFailed"
		case uwmErrCount == 0:
			return "PlatformTasksFailed"
		default:
			return "MultipleTasksFailed"
		}
```
Might work?

But I'd also be happy with more comments to point out the assumptions that are baked in.

machine424: This makes `multiple_failing_tasks_with_UWM_Degraded_StateError_and_Platform_Unavailable_StateError` fail,
As we can have a UWM degraded error and a Platform unavailable error. Now I remember, we don't keep track of the overall errors (only len(tge) can provide that, and I don't think adding a variable is worth it + what if ) but of the number of errors per state degraded or unavailable... I'll add this as comments + rename tge + maybe use some variables for readability...

 at least we're doing some mutation testing https://en.wikipedia.org/wiki/Mutation_testing and the tests are doing good :)

machine424: I pushed some changes https://github.com/openshift/cluster-monitoring-operator/compare/be41464a356b83eb39d8d905df140773bad51a9f..c0e6d61567b7c59e9ccb553d3666205fc55ba3a8 tel me if it's ok

jan--f: ah the default reason is the issue with that :roll_eyes: gotcha. Ok lets go ahead with your additional change from 
> https://github.com/openshift/cluster-monitoring-operator/compare/be41464a356b83eb39d8d905df140773bad51a9f..c0e6d61567b7c59e9ccb553d3666205fc55ba3a8

I have a feeling this is not the last time we'll look at this code :)

machine424:
I'll take look at the test, maybe it's the reordering of tasks...

machine424:
@jan--f, I put back the UWM prom operator task into the second tasks group. Some e2e tests were failing with some weird errors, they should pass now.

PR #2099, opened 2023-09-29
OCPBUGS-17850: replace nonResourceURLs permissions by subresource

marioferh:
Don't allow common users access to restricted resources

PR #2131, opened 2023-10-20
Fix UWM alert access permissions

marioferh:
Fix https://github.com/openshift/cluster-monitoring-operator/pull/2099


raptorsun (approved):
/lgtm

PR #2138, opened 2023-11-02
OCPBUGS-22742: go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp to 0.44.0

marioferh:


jan--f:
The impacted dependecies are all listed as `indirect`. Iiuc this means that a direct dependency uses the vulnerable otel library.
Going by
```
cluster-monitoring-operator (git)-[master] % go mod why go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
# go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
github.com/openshift/cluster-monitoring-operator/pkg/metrics
k8s.io/apiserver/pkg/server
k8s.io/apiserver/pkg/endpoints/filters
go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
```
We get this via `k8s.io/apiserver`. They have a fix in https://github.com/kubernetes/apiserver/commit/7bd34035e711e0fdfd31fe6a7e20123ef84421fa but I'm not sure if that was tagged yet.

PR #2144, opened 2023-11-03
OCPBUGS-22920: jsonnet: pin commits

jan--f:


raptorsun (approved):
/lgtm

PR #2151, opened 2023-11-10
MON-3476: Add method to check FeatureGate status

slashpai:
Add method to check FeatureGate status

Upgrading github.com/openshift/api to latest available as well to get latest features

Related-to https://github.com/openshift/cluster-monitoring-operator/pull/2022 as pre-requiste


slashpai:
upgrading api module causes failure, checking this

pkg/operator/operator.go
```diff
@@ -401,6 +412,39 @@ func New(
...
+		klog.Errorf("timed out waiting for FeatureGate detection")
+		return nil, fmt.Errorf("timed out waiting for FeatureGate detection")
+	}
+
+	featureGates, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return nil, err
+	}
+	// read featuregate read and usage to set a variable to pass to a controller
+	o.metricsServerEnabled = featureGates.Enabled(configv1.FeatureGateMetricsServer)
```
slashpai, line 446: I see newly added FeatureGate is available in https://pkg.go.dev/github.com/openshift/api@v0.0.0-20231109185848-6cd72e415ddb/config/v1 which is current module we using

slashpai:
This should be good for initial review

pkg/operator/operator.go
slashpai: MetricsServer featuregate is not registered in FeatureGates in yesterday's nightly image so cluster-bot test was giving error on that

slashpai: I realised I needed to vendor change in CCO https://github.com/openshift/cluster-config-operator/pull/377/files

slashpai: The idea is to pass o.metricsServerEnabled to metrics server and prometheus-adapter tasks in implementation PR

slashpai:
cc @simonpasquier could you please review when you get chance

pkg/metrics/server.go
```diff
@@ -73,7 +73,8 @@ func (s *Server) Run(ctx context.Context) error {
 		operatorv1alpha1.DelegatedAuthorization{},
 		s.kubeConfig,
 		s.kubeClient,
-		nil, // disable leader election
+		nil,   // disable leader election
+		false, // disable http2
```
jan--f, line 77: Is this actually part of the PR? I have nothing against this change, just wondering if this slipped in here by accident?

pkg/operator/operator.go
```diff
@@ -401,6 +412,39 @@ func New(
...
+	)
+	go featureGateAccessor.Run(ctx)
+	go configInformers.Start(ctx.Done())
+
+	select {
+	case <-featureGateAccessor.InitialFeatureGatesObserved():
+		featureGates, _ := featureGateAccessor.CurrentFeatureGates()
+		klog.Infof("FeatureGates initialized: %v", featureGates.KnownFeatures())
+	case <-time.After(1 * time.Minute):
+		klog.Errorf("timed out waiting for FeatureGate detection")
```
simonpasquier, line 437: it will also be logged by the caller
```suggestion
```

pkg/metrics/server.go
slashpai: Ya, because change in func signature

pkg/operator/operator.go
```diff
@@ -401,6 +412,39 @@ func New(
...
+		configInformers.Config().V1().FeatureGates(),
+		eventRecorder,
+	)
+	go featureGateAccessor.Run(ctx)
+	go configInformers.Start(ctx.Done())
+
+	select {
+	case <-featureGateAccessor.InitialFeatureGatesObserved():
+		featureGates, _ := featureGateAccessor.CurrentFeatureGates()
+		klog.Infof("FeatureGates initialized: %v", featureGates.KnownFeatures())
```
simonpasquier, line 435: the event recorder also logs the information
```suggestion
```

pkg/operator/operator.go
simonpasquier, line 446: It might be good to have an info log here when the metrics server gate is enabled:
```suggestion
	o.metricsServerEnabled = featureGates.Enabled(configv1.FeatureGateMetricsServer)
	if o.metricsServerEnabled {
		klog.Info("Metrics server enabled")
	}
```

pkg/operator/operator.go
```diff
@@ -401,6 +412,39 @@ func New(
...
+
+	// By default, this will exit(0) if the featuregates change
+	featureGateAccessor := featuregates.NewFeatureGateAccess(
+		desiredVersion, missingVersion,
+		configInformers.Config().V1().ClusterVersions(),
+		configInformers.Config().V1().FeatureGates(),
+		eventRecorder,
+	)
+	go featureGateAccessor.Run(ctx)
+	go configInformers.Start(ctx.Done())
```
jan--f, line 430: I'm wondering if we shouldn't run the informer just like the others (by appending to `o.informerFactories`). Any reason not to do that?

Similarly for the `featureGateAccessor`: afaiu assigning o.metricsServerEnabled`` only runs once now `operator.New()`. Doesn't this mean that we can't detect a later change in the featuregates? Or is that intentional? I don't fully understand how `featuregates` work yet, so I might be missing something.

pkg/metrics/server.go
jan--f: :facepalm: of course.

pkg/operator/operator.go
jan--f: Oh I see now, the accessor restarts the operator when feature gates change. In that case I'd kindly ask to extend the comment in line 422 a bit :)

slashpai: ya thats my understanding too. I updated the comment :)

pkg/operator/operator.go
```diff
@@ -401,6 +412,40 @@ func New(
...
+
+	select {
+	case <-featureGateAccessor.InitialFeatureGatesObserved():
+		featureGateAccessor.CurrentFeatureGates()
```
simonpasquier, line 434:
```suggestion
		featureGates, err := featureGateAccessor.CurrentFeatureGates()
		if err != nil {
			return nil, err
		}
		o.metricsServerEnabled = featureGates.Enabled(configv1.FeatureGateMetricsServer)
```

pkg/operator/operator.go
```diff
@@ -401,6 +412,39 @@ func New(
...
+	configInformers := configv1informers.NewSharedInformerFactory(configClient, 10*time.Minute)
+	missingVersion := "0.0.1-snapshot"
+	desiredVersion := missingVersion
+	if envVersion, exists := os.LookupEnv("RELEASE_VERSION"); exists {
+		desiredVersion = envVersion
+	}
+
+	// By default when the enabled/disabled list of featuregates changes, os.Exit is called.
+	featureGateAccessor := featuregates.NewFeatureGateAccess(
+		desiredVersion, missingVersion,
```
simonpasquier, line 424: `desiredVersion`  should be the same as the `version` input parameter?
```suggestion
		version, missingVersion,
```

pkg/operator/operator.go
```diff
@@ -401,6 +412,39 @@ func New(
...
 	o.informerFactories = append(o.informerFactories, kubeInformersOperatorNS)
 
+	configInformers := configv1informers.NewSharedInformerFactory(configClient, 10*time.Minute)
+	missingVersion := "0.0.1-snapshot"
+	desiredVersion := missingVersion
+	if envVersion, exists := os.LookupEnv("RELEASE_VERSION"); exists {
+		desiredVersion = envVersion
+	}
+
+	// By default when the enabled/disabled list of featuregates changes, os.Exit is called.
```
simonpasquier, line 422:
```suggestion
	// By default when the enabled/disabled list of featuregates changes, os.Exit is called which will trigger a restart of the container and the new container will get the updated value.
```

pkg/operator/operator.go
```diff
@@ -401,6 +412,39 @@ func New(
...
+	go featureGateAccessor.Run(ctx)
+	go configInformers.Start(ctx.Done())
+
+	select {
+	case <-featureGateAccessor.InitialFeatureGatesObserved():
+		featureGates, err := featureGateAccessor.CurrentFeatureGates()
+		if err != nil {
+			return nil, err
+		}
+		// read featuregate read and usage to set a variable to pass to a controller
```
simonpasquier, line 438:
```suggestion
```

slashpai:
@simonpasquier addressed comments, could you please take a look again?

simonpasquier (commented):
/lgtm

slashpai:
For this change we won't need docs update since this is a pre-requisite code block for actual implementation https://github.com/openshift/cluster-monitoring-operator/pull/2022

/label docs-approved

PR #2152, opened 2023-11-10
MON-3479: update Prometheus operator assets to v0.69.1

simonpasquier:
This commit also pins the prometheus-operator jsonnet to the current stable release branch of the operator. The reason is that pinning to the main branch has the drawback of pulling CRD changes which are not yet supported by the downstream operator.

The commit also updates the different jsonnet dependencies to their latest versions.

### DIFF
--- a/assets/thanos-querier/deployment.yaml
+++ b/assets/thanos-querier/deployment.yaml
@@ -76,6 +76,15 @@ spec:
           requests:
             cpu: 10m
             memory: 12Mi
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+          seccompProfile:
+            type: RuntimeDefault
         terminationMessagePolicy: FallbackToLogsOnError
         volumeMounts:
         - mountPath: /etc/tls/grpc
--- a/assets/control-plane/prometheus-rule.yaml
+++ b/assets/control-plane/prometheus-rule.yaml
@@ -530,7 +530,7 @@ spec:
       labels:
         namespace: kube-system
         severity: critical
-  - name: k8s.rules
+  - name: k8s.rules.container_cpu_usage_seconds_total
     rules:
     - expr: |
         sum by (cluster, namespace, pod, container) (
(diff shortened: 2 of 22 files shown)


simonpasquier:
/hold
no need to retest, the failures are valid.

raptorsun:
Shall we limit the blast radius to Prometheus Operator repository only?
in the file `jsonnet/jsonnetfile.lock.json` we keep the old version for other repositories. 
For example the [kube-thanos have a recent change ](https://github.com/thanos-io/kube-thanos/commit/48639958ccd4fa81fbb261ce4f9e790d69c71e2e#diff-d60084b80abfb3aa8e6133cd73a9b88e0e58811e30f752cd6d091fa7e4fe0c6a)on security context that breaks the security rules.

We update the components one by one with required patches on the security settings.

simonpasquier:
> Shall we limit the blast radius to Prometheus Operator repository only?

The main reason for having a different rule for Prometheus operator is that it ships the coreos.monitoring.com CRDs which are a customer-facing interface via user-defined monitoring.

danielmellado:
while this LGTM, I was wondering if we should just update this PR and update P-O to v.069.1, as I've seen that there's a new tag there https://github.com/prometheus-operator/prometheus-operator/tree/v0.69.1

simonpasquier:
@danielmellado this is updating to v0.69.1: the jsonnetfile pins to release-0.69 which is equivalent to v0.69.1.

assets/thanos-querier/deployment.yaml
```diff
@@ -76,6 +76,15 @@ spec:
           requests:
             cpu: 10m
             memory: 12Mi
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
```
rexagod, line 83: I think the containers below can benefit from this as well.
- `prometheus-operator-admission-webhook`
- `kube-state-metrics`
- `kube-rbac-proxy-{main,self}`
- `init-textfile`

rexagod (commented):
Some commonplace suggestions, otherwise LGTM.

assets/thanos-querier/deployment.yaml
raptorsun: maybe in a follow up PR?

simonpasquier:
@rexagod all good comments but for a follow-up PR, this one is about syncing manifests from upstream kube-prometheus.

assets/thanos-querier/deployment.yaml
jan--f: @rexagod Can you file a JIRA ticket for this please? Might even be worth a bug?

PR #2153, opened 2023-11-17
MON-3487: base CP enablement on dedicated feature gate

rexagod:
Base Collection Profile enablement on dedicated feature gate introduced in: https://github.com/openshift/api/pull/1670. The effort entails utilizing https://github.com/openshift/cluster-monitoring-operator/pull/2151 machinery.

***

Blocked by https://github.com/openshift/api/pull/1670.


slashpai:
Once https://github.com/openshift/api/pull/1670 is merged we need to vendor it in https://github.com/openshift/cluster-config-operator and bump api version in cmo

Once https://github.com/openshift/cluster-monitoring-operator/pull/2151 is merged it would be easier :)

pkg/operator/operator.go
```diff
@@ -435,6 +436,7 @@ func New(
 			return nil, err
 		}
 		o.metricsServerEnabled = featureGates.Enabled(configv1.FeatureGateMetricsServer)
+		o.collectionProfilesEnabled = featureGates.Enabled(configv1.FeatureGateCollectionProfiles)
```
rexagod, line 439: Depends on https://github.com/openshift/api/pull/1670#issue-1999674333.

rexagod: Also, `s/FeatureGateCollectionProfiles/FeatureGateMetricsCollectionProfiles`.

rexagod:
@slashpai WRT https://github.com/openshift/cluster-monitoring-operator/pull/2153#issuecomment-1818447769, please correct me if I'm wrong but IIUC the CCO vendoring is [not needed anymore](https://github.com/openshift/cluster-config-operator/pull/413#issuecomment-2025142581)?

I was able to make this work by:
* merging the feature-gate in `/api`,
* `go get`-ing the commit here, and,
* updating and vendoring indirect dependancies.

If this workflow looks good to you, I'll go ahead and close my CCO PR (https://github.com/openshift/cluster-config-operator/pull/413).

rexagod:
PS. I took the liberty of replacing the TP checks with FG ones instead of adding them checks alongside the TP ones, since we'll have all FGs within the TP mode (and for GA'd features, the default mode as well) anyway, making the TP checks redundant. The FG checks need not be removed after enabling this in the default mode, since the FG will still be present.

If that doesn't make sense, I'll add them back in.

slashpai:
> @slashpai WRT https://github.com/openshift/cluster-monitoring-operator/pull/2153#issuecomment-1818447769, please correct me if I'm wrong but IIUC the CCO vendoring is [not needed anymore](https://github.com/openshift/cluster-config-operator/pull/413#issuecomment-2025142581)?
> 
> 
> 
> I was able to make this work by:
> 
> * merging the feature-gate in `/api`,
> 
> * `go get`-ing the commit here, and,
> 
> * updating and vendoring indirect dependancies.
> 
> 
> 
> If this workflow looks good to you, I'll go ahead and close my CCO PR (https://github.com/openshift/cluster-config-operator/pull/413).

May be that's not needed anymore for Metrics Server it was needed as per FAQ doc that time

rexagod:
```
make: *** No rule to make target `vendor/github.com/openshift/api/monitoring/v1/0000_50_monitoring_01_alertingrules.crd.yaml', needed by `jsonnet/crds/alertingrules-custom-resource-definition.json'.  Stop.
```

Needs https://github.com/openshift/cluster-monitoring-operator/pull/2290.

jan--f:
Looks like you'll need to run `make generate` and commit the results.

pkg/metrics/metrics.go
```diff
@@ -5,6 +5,12 @@ import (
 	"k8s.io/component-base/metrics/legacyregistry"
 )
 
+// The various stability levels used here are analogous to the ones used in the Kubernetes project. They are:
+// INTERNAL metrics have no stability guarantees, as such, labels may be arbitrarily added/removed and the metric may be deleted at any time.
+// ALPHA metrics have no stability guarantees, as such, labels may be arbitrarily added/removed, while metrics undergo a deprecation cycle before deletion.
+// BETA metrics have some stability guarantees, as such, labels may be added but not removed, while metrics undergo a deprecation cycle before deletion.
+// STABLE metrics are guaranteed not be mutated.
+
```
jan--f, line 13: Not sure we want to promise any deprecation cycle or stability guaranty. We typically give no stability guaranty for metrics or labels.

rexagod: ACK, in that case, do you think we can go with [`metrics.INTERNAL`](https://github.com/openshift/cluster-monitoring-operator/pull/2153/files#diff-4bc91a74bc47b8467cdcd25a7e3fe1651dbe44907f478377eca31b982f444f23R9) for all meta metrics defined here? I say this because of two reasons.

* `ALPHA` metrics undergo a deprecation cycle, unlike `INTERNAL` ones.
* `ALPHA` is the [default](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/metrics/opts.go#L92) stability level.

Owing to these we could move `BETA` and `ALPHA` meta metrics to an `INTERNAL` stability level (no deprecation cycles and no stability guarantees), while being compliant with `component-base`.

rexagod: (bump)

rexagod: (ping) Requesting inputs on this.

jan--f: Sure I'm open to that, it makes sense. But I'd suggest we don't solve this in this PR here. This can probably wait for the 4.17 dev cycle. Until then we can just go with ALPHA like we have in the past?

rexagod: Sure, I've pushed the changes.

rexagod:
Depends on https://github.com/openshift/cluster-monitoring-operator/pull/2328 and https://github.com/openshift/api/pull/1863. Will `/retest` once those are in.

slashpai:
> #2328

only https://github.com/openshift/cluster-monitoring-operator/pull/2328 is required since 
ci/prow/e2e-agnostic-operator  test is failing due to prometheus adapter

rexagod:
Ah, right.

rexagod:
@slashpai I see that the `single-node` test is failing too, and most likely dependent on https://github.com/openshift/api/pull/1863's resolution?

jan--f:
single node is not a required test gate.

rexagod:
Ah, makes sense. I see the `required` badge isn't present on `single-node` checks. Apologies for the extra pings.

rexagod:
/skip e2e-aws-ovn-single-node

Skipping these since they aren't required for now.

PR #2156, opened 2023-11-20
MON-3402: Add Konveyor metrics

mansam:
Add usage metrics for Konveyor to telemetry. See https://issues.redhat.com/browse/MON-3402

cc @dymurray 


jan--f:
/label ok-to-test

Thanks, @mansam can you link me to where these metrics are exposed or rules defined if applicable? Just for due diligence regarding the label set.

PR #2157, opened 2023-11-21
a PoC of a workaround to address https://issues.redhat.com/browse/OCP…

machine424:
…BUGS-21610


machine424:
Will be done in https://github.com/openshift/cluster-monitoring-operator/pull/2166

PR #2158, opened 2023-11-21
Add ownership annotation for certificates

vrutkovs:


raptorsun:
How these new annotation will be used?

vrutkovs:
This would help customers file correct bug reports. This case is fairly obvious, but its not the case for other components. Also it can help us find situations when a team adds a new unregistered TLS artifact. More metadata would be added to these secrets later.

See https://github.com/openshift/enhancements/pull/1502 for more information

rexagod:
I see the changes introduced here are in-line with: ["As an Openshift developer, I want to be able to find which team manages a particular certificate."](https://github.com/openshift/enhancements/pull/1502/files#diff-6f4d97397f99cc1df15a5a6cffbccff4afd182e9e267b22f79d7c21f1fd42798R42-R43). While this looks fine, we should hold off on merging this until the EP gets in.

deads2k:
Ownership annotations are already in the openshif/api. The enhancement is about building an automatic report against those annotations, not the annotations themselves.  IOW: the annotations are ready for merge now.

jan--f:
We pre-generate static assets and I think the annotation can be added there already. See https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/utils/generate-certificate-injection.libsonnet. Adding the annotation there will probably do the trick. That'll keep things consistent for future ca-bundle ConfigMaps.
Is there a JIRA we can attach to this?

simonpasquier:
> IIUC these will not be applied to the upgraded clusters?

They will/should be applied on upgrades.

vrutkovs:
Ah, certs with `inject-trusted-cabundle` are not really owned by monitoring team - cluster-network-operator injects CA bundle there and it annotates them already. 

Also, more cert-specific metadata would be added there (i.e. cert description) so we still need a place where its being set per-cert/configmap

simonpasquier:
Adding the "openshift.io/owning-component" annotation in the jsonnet code is the right thing to do IMHO because the information is static and it's way easier for future ourselves.

IIUC we don't have to do anything for certificates which are issued by the service-ca operator and the cluster-network-operator?
AFAICT CMO manages directly only the gRPC TLS certificates for thanos-querier, prometheus-k8s, thanos-ruler and prometheus-user-workload (assets/*/grpc-tls-secret.yaml).

vrutkovs:
@simonpasquier I don't mind adding the annotations to jsonnet code, but I can't seem to find where these are defined. Any pointers?

raptorsun:
> @simonpasquier I don't mind adding the annotations to jsonnet code, but I can't seem to find where these are defined. Any pointers?

The labels can [be added here ](https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/utils/generate-certificate-injection.libsonnet#L10)and it will be applied to all ca bundle config maps generated from this jsonnet function. [Here is where the CA bundle is generate for alertmanager.
](https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/components/alertmanager.libsonnet#L13C44-L13C64)

vrutkovs:
Done, now component annotation is set in jsonnet where possible

PR #2160, opened 2023-11-22
MON-3454: Add track timestamps staleness

jan--f:
Marking this as draft for now since we need the prometheus-operator 0.70 first.
Happy to discuss this change though.


pkg/manifests/types.go
```diff
@@ -121,6 +121,8 @@ type K8sPrometheusAdapter struct {
 
 // You can use the `DedicatedServiceMonitors` resource to configure dedicated
 // Service Monitors for the Prometheus Adapter
+// This setting is DEPRECATED and will be removed in a future version. In this
+// version this setting has no effect.
 type DedicatedServiceMonitors struct {
 	// When `enabled` is set to `true`, the Cluster Monitoring Operator (CMO)
```
simonpasquier, line 127: should we remove the whole comment and just reiterate that the setting is deprecated?

pkg/tasks/controlplane.go
```diff
@@ -64,6 +65,7 @@ func (t *ControlPlaneTask) Run(ctx context.Context) error {
 	}
 
 	if t.config.ClusterMonitoringConfiguration.K8sPrometheusAdapter.DedicatedServiceMonitors.Enabled {
+		klog.V(4).Info("Dedicated ServiceMonitors are deprecated, this setting has no effect in this version and will be removed in a future release")
```
simonpasquier, line 68: we'd need to remove the service monitor?

pkg/manifests/types.go
jan--f: Yes good point, makes a lot of sense. Guess we'll need Brians approval ultimately though.

pkg/tasks/controlplane.go
jan--f: I left it in for now since a user might actually use the provided metric outside of just prometheus adapter (it has its own name). So removing the SM would be something for next release?

Or should we refer to our lack of guarantee for metric name stability?

simonpasquier: I'd be fine with dropping the metrics given we provide no guarantee.

machine424: I think we can go further and delete the related resources from `assets/` to avoid any confusion and just use their names to delete them.
(add a TODO remove the delete code or even a ticket to remove it in future releases will be great)

pkg/manifests/types.go
simonpasquier: bike-shedding but we could leave the struct forever (or at least as long as we use the CMO configmap) but not document it. There's probably work involved as I don't think that the doc generation tool supports this easily.

jsonnet/components/control-plane.libsonnet
```diff
@@ -42,6 +42,8 @@ function(params)
               if 'path' in e && e.path == '/metrics/cadvisor' then
                 // Drop cAdvisor metrics with excessive cardinality.
                 {
+                  honorTimestamps: true,
```
simonpasquier, line 45: (suggestion) add a comment explaining the settings. Also should we push the  change upstream eventually?

pkg/manifests/types.go
jan--f: yeah that would also work. Maybe we add the deprecation notice for now and decide next release if we'll just drop it or make it invisible?

pkg/tasks/controlplane.go
jan--f: Added the the resource deletion.

pkg/tasks/prometheusadapter.go
```diff
@@ -255,6 +231,14 @@ func (t *PrometheusAdapterTask) create(ctx context.Context) error {
 		}
 	}
 
+	{
+		// TODO: Remove this in 4.16
+		err := t.client.DeleteServiceMonitorByNamespaceAndName(ctx, t.client.Namespace(), "kubelet-resource-metrics")
```
machine424, line 236: unfortunate copy/past :)
You meant to delete the `adapter-config-dedicated-sm` configmap I think
(I think Daniel has added some checks for unused fcts?)
For being too picky I'll create a ticket to get rid of this code in 4.16 :)

jan--f: oh dear...unfortunate? I'd rather go with vim fat fingering on my part. Thanks for cleaning up after me :grimacing: .

Fixed now.
I'm also a little surprised that golang-ci didn't catch that.

machine424: I'll ping him here if he has an idea (cc @danielmellado), otherwise I'll dig it later.

PR #2161, opened 2023-11-23
MON-3500: Enable sending exemplars over RW in UWM

rexagod:

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -49,6 +49,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
+	"k8s.io/klog/v2"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 )
 
@@ -69,6 +70,9 @@ const (
 	telemetryTokenSecretKey = "token"
 
 	collectionProfileLabel = "monitoring.openshift.io/collection-profile"
+
+	// --enable-feature=exemplar-storage: https://prometheus.io/docs/prometheus/latest/feature_flags/#exemplars-storage
+	EnableFeatureExemplarStorageString = "exemplar-storage"
 )
 
 var (
@@ -1455,6 +1459,14 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 
 	if len(f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RemoteWrite) > 0 {
 		p.Spec.RemoteWrite = addRemoteWriteConfigs(clusterID, p.Spec.RemoteWrite, f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RemoteWrite...)
+
+		// Explicitly warn that we are ignoring the `SendExemplars` configuration for in-cluster scenarios.
+		for _, rws := range f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RemoteWrite {
+			if rws.SendExemplars != nil && *rws.SendExemplars {
+				klog.Warningln("Enabling `SendExemplars` in the in-cluster Prometheus configuration has no affect, as it is only supported for UWM configurations.")
+				break
+			}
+		}
 	}
 
 	for _, rw := range p.Spec.RemoteWrite {
@@ -1759,6 +1771,15 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret, trustedCABundleCM *
 			f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.ClusterID,
 			p.Spec.RemoteWrite,
 			f.config.UserWorkloadConfiguration.Prometheus.RemoteWrite...)
+
+		// Since `SendExemplars` is experimental currently, we need to enable "exemplar-storage" explicitly to make sure
+		// CMO turns this on automatically in Prometheus if any *UWM* RemoteWrite[] enables this.
+		for _, rws := range f.config.UserWorkloadConfiguration.Prometheus.RemoteWrite {
+			if *rws.SendExemplars {
+				p.Spec.EnableFeatures = append(p.Spec.EnableFeatures, EnableFeatureExemplarStorageString)
+				break
+			}
+		}
 	}
 
 	if f.config.UserWorkloadConfiguration.Prometheus.EnforcedSampleLimit != nil {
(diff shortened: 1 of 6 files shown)


pkg/manifests/types.go
```diff
@@ -753,6 +753,10 @@ type RemoteWriteSpec struct {
 	QueueConfig *monv1.QueueConfig `json:"queueConfig,omitempty"`
 	// Defines the timeout value for requests to the remote write endpoint.
 	RemoteTimeout string `json:"remoteTimeout,omitempty"`
+	// Enables sending of exemplars over remote write. Note that
+	// exemplar-storage itself must be enabled using the `spec.enableFeature`
```
simonpasquier, line 757: This should be managed automatically by CMO: if one remote write config enables exemplars then CMO should turn on the feature in Prometheus.

pkg/manifests/types.go
```diff
@@ -753,6 +753,11 @@ type RemoteWriteSpec struct {
 	QueueConfig *monv1.QueueConfig `json:"queueConfig,omitempty"`
 	// Defines the timeout value for requests to the remote write endpoint.
 	RemoteTimeout string `json:"remoteTimeout,omitempty"`
+	// Enables sending of exemplars over remote write. Note that the experimental
+	// "exemplar-storage" feature must be enabled using the `spec.enableFeature`
+	// option for exemplars to be scraped in the first place. Only supported for
+	// UWM configurations.
```
simonpasquier, line 759: I'll omit the implementation details about the Prometheus command-line argument and rather focus on the capabilities and "limitations".

cc @bburt-rh for review.

```suggestion
	// Enables sending of exemplars over remote write. Note when enabled, Prometheus is
	// configured to store exemplars in memory with a maximum number of 100,000 exemplars. The setting is only effective for
	// user-defined monitoring and has no effect for cluster monitoring.
```

pkg/manifests/manifests.go
```diff
@@ -1759,6 +1771,15 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret, trustedCABundleCM *
 			f.config.ClusterMonitoringConfiguration.TelemeterClientConfig.ClusterID,
 			p.Spec.RemoteWrite,
 			f.config.UserWorkloadConfiguration.Prometheus.RemoteWrite...)
+
+		// Since `SendExemplars` is experimental currently, we need to enable "exemplar-storage" explicitly to make sure
+		// CMO turns this on automatically in Prometheus if any *UWM* RemoteWrite[] enables this.
+		for _, rws := range f.config.UserWorkloadConfiguration.Prometheus.RemoteWrite {
+			if *rws.SendExemplars {
```
simonpasquier, line 1778:
```suggestion
			if ptr.Deref(rws.SendExemplars, false) {
```

pkg/manifests/manifests.go
```diff
@@ -1455,6 +1459,14 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 
 	if len(f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RemoteWrite) > 0 {
 		p.Spec.RemoteWrite = addRemoteWriteConfigs(clusterID, p.Spec.RemoteWrite, f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RemoteWrite...)
+
+		// Explicitly warn that we are ignoring the `SendExemplars` configuration for in-cluster scenarios.
+		for _, rws := range f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RemoteWrite {
+			if rws.SendExemplars != nil && *rws.SendExemplars {
+				klog.Warningln("Enabling `SendExemplars` in the in-cluster Prometheus configuration has no affect, as it is only supported for UWM configurations.")
```
simonpasquier, line 1466: I'm not sure we need a warning log. IMHO it would be lost in the existing output + there's no harm being done by having the flag enabled here.

rexagod: @danielmellado Ah, it seems I missed your comment (I reloaded this page after >1h). Do you suppose we replace `Warningln` with a `V(x)` level call?

simonpasquier (commented):
It needs an e2e test, at least to check that the feature gets enabled.
Ideally we add exemplars support in https://github.com/rhobs/prometheus-example-app and check that they can be queried via the Prometheus/Thanos APIs.

test/e2e/config_test.go
```diff
@@ -619,6 +619,7 @@ func TestUserWorkloadMonitorPrometheusK8Config(t *testing.T) {
     datacenter: eu-west
   remoteWrite:
   - url: "https://test.remotewrite.com/api/write"
+	sendExemplars: true
```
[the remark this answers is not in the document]
rexagod: Hopefully no more of these in the future. 😅 

Screenshot 2023-11-28 at 16 23 52

pkg/manifests/types.go
```diff
@@ -753,6 +753,11 @@ type RemoteWriteSpec struct {
 	QueueConfig *monv1.QueueConfig `json:"queueConfig,omitempty"`
 	// Defines the timeout value for requests to the remote write endpoint.
 	RemoteTimeout string `json:"remoteTimeout,omitempty"`
+	// Enables sending of exemplars over remote write. Note when enabled, Prometheus is
+	// configured to store exemplars in memory with a maximum number of 100,000 exemplars.
+	// The setting is only effective for user-defined monitoring and has no effect for
+	// cluster monitoring.
```
bburt-rh, line 759:
```suggestion
	// Enables using remote write to send exemplars. When enabled, Prometheus is
	// configured to store a maximum of 100,000 exemplars in memory.
	// Note that this setting only applies to user-defined monitoring. It is not applicable 
	// to default in-cluster monitoring.
```

pkg/manifests/types.go
```diff
@@ -753,6 +753,11 @@ type RemoteWriteSpec struct {
 	QueueConfig *monv1.QueueConfig `json:"queueConfig,omitempty"`
 	// Defines the timeout value for requests to the remote write endpoint.
 	RemoteTimeout string `json:"remoteTimeout,omitempty"`
+	// Enables sending exemplars via remote write. When enabled, Prometheus is
+	// configured to store a maximum of 100,000 exemplars in memory.
+	// Note that this setting only applies to user-defined monitoring. It is not applicable
+	// to default in-cluster monitoring.
+	SendExemplars *bool `json:"sendExemplars,omitempty"`
```
jan--f, line 760: I though we could also have a
```
type UWMRemoteWriteSpec struct {
        RemoteWriteSpec
        // Enables sending exemplars via remote write. When enabled, Prometheus is
	// configured to store a maximum of 100,000 exemplars in memory.
	SendExemplars *bool `json:"sendExemplars,omitempty"`
}
```
and then use that in `PrometheusRestrictedConfig`. That would make it clearer what option can be used where. Though not sure how that affects the API doc generation.

jan--f: just to be sure, this is not a blocker for me and we could go ahead as is.

PR #2163, opened 2023-11-23
MON-3505: move aggregated-metrics-reader role to cmo jsonnet

slashpai:
This change came as part of reviewing #2022
To avoid duplicated code in metrics-server
and prometheus-adapter moving this role
to cluster-monitoring-operator.jsonnet

cc: @simonpasquier 

### DIFF
--- a/jsonnet/components/prometheus-adapter.libsonnet
+++ b/jsonnet/components/prometheus-adapter.libsonnet
@@ -39,14 +39,9 @@ function(params)
       },
     },
 
-    clusterRoleAggregatedMetricsReader+:
-      {
-        metadata+: {
-          labels+: {
-            'rbac.authorization.k8s.io/aggregate-to-cluster-reader': 'true',
-          },
-        },
-      },
+    // This role is moved to cluster-monitoring-operator.libsonnet
+    // Hence hiding it in prometheus-adapter.
+    clusterRoleAggregatedMetricsReader:: null,
 
     apiService+:
       {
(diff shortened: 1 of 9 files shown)


raptorsun:
It looks good to me, just leave a question about why not remove the `clusterRoleAggregatedMetricsReader` in prometheus adapter file.

jsonnet/components/prometheus-adapter.libsonnet
```diff
@@ -39,7 +39,9 @@ function(params)
       },
     },
 
-    clusterRoleAggregatedMetricsReader+:
+    // This role is moved to cluster-monitoring-operator.libsonnet
+    // Hence hidding it in prometheus-adapter
```
simonpasquier, line 43: (nit)
```suggestion
    // Hence hiding it in prometheus-adapter.
```

jsonnet/components/prometheus-adapter.libsonnet
```diff
@@ -39,7 +39,9 @@ function(params)
...
+    clusterRoleAggregatedMetricsReader::
```
simonpasquier, line 44: (nit) better to set to an empty map or null.
```suggestion
    clusterRoleAggregatedMetricsReader:: {},
```

slashpai:
> It looks good to me, just leave a question about why not remove the `clusterRoleAggregatedMetricsReader` in prometheus adapter file.

we needed to hide to avoid creating the asset file for same. I set it as null now :)

slashpai:
I will rebase once https://github.com/openshift/cluster-monitoring-operator/pull/2163 is merged. I thought that was merged as it was ready to merge state :(. CI seems to have issues

PR #2164, opened 2023-11-23
OCPBUGS-23495: Change UWM Prometheus' kube-rbac-proxy-thanos port num…

machine424:
…ber to avoid duplicate

The port name stays the same so this is not breaking.


simonpasquier (commented):
/lgtm

PR #2165, opened 2023-11-23
MON-3287: Remove openshift-etcd related RBAC as they will be manager by

machine424:
cluster-etcd-operator now.

The resources may keep some of CMO related metadata, this is the trade-off to keep the same names and avoid unnecessary API calls in CMO.

/hold

Will be managed in CEO here https://github.com/openshift/cluster-etcd-operator/pull/1164

(tested installing a cluster with both PRs + upgrading with both PRs and didn't see any issues)

Will try to merge the CEO one first and this one just after (may get some transient clash that will not lead to downtime)


simonpasquier (commented):
/lgtm

PR #2166, opened 2023-11-23
OCPBUGS-21610: Change config to allow ipv6/4

danielmellado:
This commit uses the Kube Downward API to get the plugin's pod
status.podIP and bash to set it and run nginx accordingly. This is also
set in the jsonnet so it gets generated, even though it's some kind of
double templating.

### DIFF
--- a/assets/monitoring-plugin/config-map.yaml
+++ b/assets/monitoring-plugin/config-map.yaml
@@ -8,8 +8,7 @@ data:
       default_type       application/octet-stream;
       keepalive_timeout  65;
       server {
-        listen              9443 ssl;
-        listen              [::]:9443 ssl;
+        listen              LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME ssl;
         ssl_certificate     /var/cert/tls.crt;
         ssl_certificate_key /var/cert/tls.key;
         root                /usr/share/nginx/html;
--- a/assets/monitoring-plugin/deployment.yaml
+++ b/assets/monitoring-plugin/deployment.yaml
@@ -44,7 +44,23 @@ spec:
             topologyKey: kubernetes.io/hostname
       automountServiceAccountToken: false
       containers:
-      - image: quay.io/openshift/origin-monitoring-plugin:1.0.0
+      - command:
+        - /bin/sh
+        - -c
+        - |
+          if echo "$POD_IP" | grep -qE '^([0-9]{1,3}\.){3}[0-9]{1,3}$'; then
+            LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME="9943"
+          else
+            LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME="[::]:9443"
+          fi
+          sed "s/LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME/$LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME/g" /etc/nginx/nginx.conf > /tmp/nginx.conf
+          exec nginx -c /tmp/nginx.conf -g 'daemon off;'
+        env:
+        - name: POD_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
+        image: quay.io/openshift/origin-monitoring-plugin:1.0.0
         imagePullPolicy: IfNotPresent
         name: monitoring-plugin
         ports:
--- a/jsonnet/components/monitoring-plugin.libsonnet
+++ b/jsonnet/components/monitoring-plugin.libsonnet
@@ -22,8 +22,7 @@ function(params)
       default_type       application/octet-stream;
       keepalive_timeout  65;
       server {
-        listen              %(nginxPort)d ssl;
-        listen              [::]:%(nginxPort)d ssl;
+        listen              LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME ssl;
         ssl_certificate     %(tlsPath)s/tls.crt;
         ssl_certificate_key %(tlsPath)s/tls.key;
         root                /usr/share/nginx/html;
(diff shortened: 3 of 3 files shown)


assets/monitoring-plugin/deployment.yaml
```diff
@@ -44,7 +44,27 @@ spec:
             topologyKey: kubernetes.io/hostname
       automountServiceAccountToken: false
       containers:
-      - image: quay.io/openshift/origin-monitoring-plugin:1.0.0
+      - command:
+        - /bin/sh
+        - -c
+        - |
+          # TODO: Maybe just use IPV4 if empty?
```
machine424, line 51: I think we should address this.

assets/monitoring-plugin/config-map.yaml
```diff
@@ -9,7 +9,6 @@ data:
       keepalive_timeout  65;
       server {
         listen              9443 ssl;
-        listen              [::]:9443 ssl;
```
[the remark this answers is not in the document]
danielmellado: d'oh, it wasn't saved when I commited, fixing it!

assets/monitoring-plugin/config-map.yaml
```diff
@@ -8,8 +8,7 @@ data:
       default_type       application/octet-stream;
       keepalive_timeout  65;
       server {
-        listen              9443 ssl;
-        listen              [::]:9443 ssl;
+        listen              LISTEN_ADDR_PORT ssl;
```
simonpasquier, line 11: just to make it clear that it's not the final value:
```suggestion
        listen              LISTEN_ADDRESS_PORT_REPLACED_AT_TIME ssl;
```

assets/monitoring-plugin/deployment.yaml
```diff
@@ -44,7 +44,27 @@ spec:
...
+          if [ -z "$POD_IP" ]; then
+            echo "POD_IP env variable is not set or is empty"
```
simonpasquier, line 53: is it even possible?

danielmellado: Theoretically if the CNI plugin is broken, yeah, but we'll have some bigger problems then :D . I guess I'll just skip it

assets/monitoring-plugin/config-map.yaml
danielmellado: Yeah, it can get confusing, I'll change this

assets/monitoring-plugin/deployment.yaml
```diff
@@ -44,7 +44,27 @@ spec:
...
+          else
+            LISTEN_ADDR_PORT="[::]:9443"
+          fi
+          sed "s/LISTEN_ADDR_PORT/$LISTEN_ADDR_PORT/g" /etc/nginx/nginx.conf > /tmp/nginx.conf
+          exec nginx -c /tmp/nginx.conf -g 'daemon off;'
+        env:
+        - name: POD_IP
+          valueFrom:
+            fieldRef:
+              fieldfPath: status.podIP
```
machine424, line 66: `fieldPath` (extra `f` this is for not adding a remote haha)

danielmellado: just noticed this as debugging the PR lol 🗡️

assets/monitoring-plugin/config-map.yaml
machine424: (I think he meant `..._AT_RUNTIME`)

danielmellado: I can change it, it wouldn't hurt and may be more clear, thanks!

assets/monitoring-plugin/deployment.yaml
```diff
@@ -44,7 +44,23 @@ spec:
             topologyKey: kubernetes.io/hostname
       automountServiceAccountToken: false
       containers:
-      - image: quay.io/openshift/origin-monitoring-plugin:1.0.0
+      - command:
+        - /bin/sh
+        - -c
+        - |
+          if echo "$POD_IP" | grep -qE '^([0-9]{1,3}\.){3}[0-9]{1,3}$'; then
```
raptorsun, line 51: This logic checks whether the `$POD_IP` is an IPv4 address or not. If not, it may not mean the pod has an IPv6 address.
It could be better to check whether  the `$POD_IP` contains an IPv6 address.
```bash
# Regular expression for matching IPv6 addresses
ipv6_regex='^([0-9a-fA-F]{1,4}:){7}([0-9a-fA-F]{1,4})|(([0-9a-fA-F]{1,4}:){1,7}:)|(:([0-9a-fA-F]{1,4}:){1,7})|(::([0-9a-fA-F]{1,4}:){0,6}[0-9a-fA-F]{1,4})|(([0-9a-fA-F]{1,4}:){1,6}:([0-9a-fA-F]{1,4}:){1,6})$'

# Split the string into individual addresses and check each one
IFS=',' read -ra ADDR <<<$POD_IP
for ip in "${ADDR[@]}"; do
  if echo "$ip" | grep -Eq "$ipv6_regex"; then
    echo "$ip is a valid IPv6 address."
    LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME="[::]:9443"
    break
  else
    echo "$ip is not a valid IPv6 address."
  fi
done
LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME="9943"

```

danielmellado: Hi Haoyu, thanks for taking a look! First of all I see that your code would apply to `status.podIPs` but not to `status.podIP`. In fact `status.podIP==status.podIPs[0]`. What you're saying in `The property status.podIP could contain more than one IP address delimited by comma` isn't really true. (check https://kubernetes.io/docs/concepts/workloads/pods/downward-api/)

I'm stating that as no real need to split that out with IFS. That would also be a limmitation if you see that as assuming a `podIPs==10.200.16.10,2041:0000:140F::875B:131B` it'll go there and first set that to `9943` and later on overwrite it with `[::]:9443`

We do only have one listening address, which I'd map to the main IP family. There's a lot behind this as you can read in:

https://github.com/kubernetes/kubernetes/issues/95251
https://github.com/kubernetes/kubernetes/pull/97979

So I'd keep this as simple as possible. TL;DR `status.PodIP[0]` will always be either ipv4 or ipv6, so I think it's just simpler to check for the ipv4.

raptorsun: Oh yes I confused `status.podIP` with `status.podIPs` 😅
From the https://github.com/kubernetes/kubernetes/pull/97979 , can we say that in a dual stack setup the IPv6 always go first as `status.podIP` ?

danielmellado: It'll be based on the primary node IP. quoting the comments on that 

> This PR fixes kubelet to always pick an IP of the same family as its --node-ip for the first pod IP, and an IP of the opposite family (if any) for the second pod IP, and ignore any other IPs from the runtime.

so if you check the code, it really gets it from the family of the `nodeIP`but in any case a dual stack would just work with any of the families and it'll be basically cosmetic. 

The bug here would be basically for a single IPv4 stack cluster trying to load an nginx config binding an ipv6 cluster, or the other way around, and this would just sort that out :D 

In any case there's a lot of interactios, CRI, CNI, Kubelet, and so on and so on so I think that we should intentionally keep this the simplest we can for now.

assets/monitoring-plugin/deployment.yaml
```diff
@@ -44,7 +44,23 @@ spec:
...
+            LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME="9943"
```
simonpasquier, line 52: 9943 should have been 9443

PR #2167, opened 2023-11-23
WIP: remove kube-rbac-proxy sidecar for metrics

simonpasquier:


simonpasquier (commented):
/close

we still need https://github.com/brancz/kube-rbac-proxy/issues/146 to be fixed... In the mean time, I realize that the CMO e2e tests don't verify that the alertmanager metrics are scraped correctly...

PR #2168, opened 2023-11-24
MON-3511: hack/local-cmo.sh: pass desired version in cmo run command

slashpai:
After FeatureGate is introduced in #2151, local run doesn't correctly since it cannot get the desired version. This commit pass release-version in cmo run

### DIFF
--- a/hack/local-cmo.sh
+++ b/hack/local-cmo.sh
@@ -201,6 +201,9 @@ main() {
 	local operator_config=manifests/0000_50_cluster-monitoring-operator_04-config.yaml
 	local telemetry_conf=/tmp/telemetry-config.yaml
 
+	local desired_version
+	desired_version=$(kubectl get clusterversion version -o jsonpath='{ .status.desired.version }')
+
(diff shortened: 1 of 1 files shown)


hack/local-cmo.sh
```diff
@@ -201,6 +201,9 @@ main() {
 	local operator_config=manifests/0000_50_cluster-monitoring-operator_04-config.yaml
 	local telemetry_conf=/tmp/telemetry-config.yaml
 
+	local desired_version
+	desired_version=$(kubectl get clusterversion version -o json|jq .status.desired.version|sed 's/"//g')
```
simonpasquier, line 205: (nit) we should be able to use jsonpath directly (untested)
```suggestion
	desired_version=$(kubectl get clusterversion version -o jsonpath='{ .status.desired.version }')
```

slashpai: indeed it works :) updated

slashpai:
Since this is needed to run tests locally

PR #2179, opened 2023-12-04
OCPBUGS-23745: Wait for 3 (instead of 2) consecutive failing reconcil…

machine424:
…iations before reporting failures.

The rate limiting was added in https://github.com/openshift/cluster-monitoring-operator/pull/1193 The threshold was changed by mistake in https://github.com/openshift/cluster-monitoring-operator/pull/1558

Add comments and change logs in attempt to make clearer.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -139,7 +139,6 @@ func (pc *ProxyConfig) NoProxy() string {
 const (
 	resyncPeriod         = 15 * time.Minute
 	reconciliationPeriod = 5 * time.Minute
-	maxFailCount         = 2
 
 	// see https://github.com/kubernetes/apiserver/blob/b571c70e6e823fd78910c3f5b9be895a756f4cbb/pkg/server/options/authentication.go#L239
 	apiAuthenticationConfigMap    = "kube-system/extension-apiserver-authentication"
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
/lgtm

pkg/operator/operator.go
```diff
@@ -139,7 +139,6 @@ func (pc *ProxyConfig) NoProxy() string {
 const (
 	resyncPeriod         = 15 * time.Minute
 	reconciliationPeriod = 5 * time.Minute
-	maxFailCount         = 2
```
raptorsun, line 142: I prefer keep it as constant. no need to put it into the stack if it does not change ;)

rexagod: +1, we definitely should take measure to prevent this from changing.

machine424: I think the performance/resource consumption differences between the two approaches are negligible here.

I think having it defined inside the function would make the code more readable (no need to jump into the variable to get the value + we centralize the comments) and less easier to be changed b mistake

jan--f: I think the relocation makes sense, since this is not used anywhere else. With the new scope I think making it a var is acceptable.

PR #2180, opened 2023-12-04
MON-3551: fix: simplify jq script avoids error with jq1.7

jan--f:
According to the documentation both the existing version and this change should be legal syntax. With this change I can run a local CMO with jq 1.7.
I think regardless fewer pipes make this more readable.

### DIFF
--- a/hack/local-cmo.sh
+++ b/hack/local-cmo.sh
@@ -80,7 +80,7 @@ disable_managed_cmo() {
 			    {
 			      "spec": {
 			        "overrides": [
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
/lgtm

slashpai:
No code change  just a local run script fix
/label docs-approved
/label qe-approved
/label px-approved

slashpai:
/retitle MON-3551: fix: simplify jq script avoids error with jq1.7

PR #2182, opened 2023-12-05
MON-3544: Adjust NodeClock* alerting rules to work with PTP operator

simonpasquier:
This commit adapts the upstream NodeClockNotSynchronising and NodeClockSkewDetected rules to be always inactive when the PTP operator is installed. The PTP operator ships a more robust rule to detect unsynchronised clocks and the default rules are redundant in this case.

### DIFF
--- a/assets/node-exporter/prometheus-rule.yaml
+++ b/assets/node-exporter/prometheus-rule.yaml
@@ -173,7 +173,8 @@ spec:
       annotations:
         description: Clock at {{ $labels.instance }} is out of sync by more than 0.05s. Ensure NTP is configured correctly on this host.
         summary: Clock skew detected.
-      expr: |
+      expr: |-
+        (
         (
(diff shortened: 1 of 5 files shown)


assets/node-exporter/prometheus-rule.yaml
```diff
@@ -185,6 +186,7 @@ spec:
         and
           deriv(node_timex_offset_seconds{job="node-exporter"}[5m]) <= 0
         )
+        ) and on() absent(up{job="ptp-monitor-service"})
```
simonpasquier, line 189: @midu16 can you validate the job label value here? Looking at the telemetry data, I see that the `NodeOutOfPtpSync` alert fires with this value.

simonpasquier:
@midu16 this PR patches also the NodeClockSkewDetected alerting rule which wasn't mentioned in the original ticket but I think that it makes sense. Can you validate?

PR #2185, opened 2023-12-06
OCPBUGS-23516: prevent plugin entry assets from caching

simonpasquier:


simonpasquier:
/close
in favor of #2186

PR #2189, opened 2023-12-08
OCPBUGS-24630: additionalArgs: list items must have named fields

jan--f:


jsonnet/components/prometheus.libsonnet
```diff
@@ -542,15 +554,6 @@ function(params)
           },
           {
```
simonpasquier, line 555: remove the object?

simonpasquier: Right this is needed :+1:

simonpasquier (commented):
/lgtm

PR #2194, opened 2023-12-12
MON-3589: Refactor assertExemplarsEnabled check for UWM Prometheus xand add a check for --scrape.timestamp-tolerance on Platform Prometheus

machine424:
…rometheus xand add a check for --scrape.timestamp-tolerance on Platform Prometheus


jan--f:
Nice, thanks!
/lgtm

machine424:
(this only touches e2e tests and `ci/prow/e2e-agnostic-operator ` isn't yelling, I think it's safe to skip the other ones)

PR #2195, opened 2023-12-12
MON-3592: Use strict yaml unmarshaling

jan--f:
This PR adds strict yaml unmarshaling to avoid silently dropping fields from our static yaml assets, that are not part of the actual type. The old behaviour as caused two bugs in the 4.15 development cycle.

Additionally I added two refactor commits, which imo improve readability. They can however be easily dropped if there are objections. I recommend reviewing commits individually.

### DIFF
--- a/assets/prometheus-k8s/role-binding-specific-namespaces.yaml
+++ b/assets/prometheus-k8s/role-binding-specific-namespaces.yaml
@@ -77,7 +77,3 @@ items:
     name: prometheus-k8s
     namespace: openshift-monitoring
 kind: RoleBindingList
-metadata:
-  labels:
-    app.kubernetes.io/managed-by: cluster-monitoring-operator
-    app.kubernetes.io/part-of: openshift-monitoring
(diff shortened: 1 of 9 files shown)


assets/prometheus-k8s/role-binding-specific-namespaces.yaml
```diff
@@ -77,7 +77,3 @@ items:
     name: prometheus-k8s
     namespace: openshift-monitoring
 kind: RoleBindingList
-metadata:
```
simonpasquier, line 80: oops :D

simonpasquier (commented):
/lgtm

Nice refactoring!

pkg/manifests/assets.go
```diff
@@ -62,10 +60,10 @@ func (a *Assets) GetAsset(name string) ([]byte, error) {
 	return f, nil
 }
 
-func (a *Assets) MustNewAssetReader(name string) io.Reader {
+func (a *Assets) MustNewAssetSlice(name string) []byte {
```
machine424, line 63: (nit) how about just `MustNewAsset`?

machine424:
Thanks Jan
/lgtm

PR #2196, opened 2023-12-13
MON-3498: Modify prometheus-adapter e2e tests to compact with metrics server

slashpai:
Modify prometheus-adapter e2e tests to compact with metrics server since both are providing resource Metrics API.
When MetricsServer FeatureGate is enabled it will check for metrics-server backend is configured in APIService

### DIFF
--- a/test/e2e/framework/helpers.go
+++ b/test/e2e/framework/helpers.go
@@ -1,9 +1,11 @@
 package framework
 
 import (
+	"context"
 	"testing"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
 	"gopkg.in/yaml.v2"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
@@ -166,6 +168,45 @@ func (f *Framework) MustGetPods(t *testing.T, namespace string) *v1.PodList {
 		t.Fatalf("failed to get pods in namespace %s - %s", namespace, err.Error())
 	}
 	return pods
+}
+
+func (f *Framework) GetClusterVersion(name string) (*configv1.ClusterVersion, error) {
+	return f.OpenShiftConfigClient.ConfigV1().ClusterVersions().Get(context.Background(), name, metav1.GetOptions{})
+}
+
+func (f *Framework) IsFeatureGateEnabled(t *testing.T, name string) bool {
+	t.Helper()
+	var enabledFeatureGates []configv1.FeatureGateAttributes
+	// Get cluster version
+	clusterVersion, err := f.GetClusterVersion("version")
+	if err != nil {
+		t.Fatalf("failed to get cluster version: %s", err.Error())
+	}
+
+	// Get the desired cluster version
+	version := clusterVersion.Status.Desired.Version
+
+	// Get FeatureGates
+	featureGates, err := f.OpenShiftConfigClient.ConfigV1().FeatureGates().Get(context.Background(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("failed to get featuregate `cluster`: %s", err.Error())
+	}
+
+	// Get Enabled FeatureGates for the desired cluster version
+	for _, fg := range featureGates.Status.FeatureGates {
+		if fg.Version == version {
+			enabledFeatureGates = fg.Enabled
+			break
+		}
+	}
+
+	for _, g := range enabledFeatureGates {
+		if g.Name == configv1.FeatureGateName(name) {
+			return true
+		}
+	}
+
+	return false
 }
 
 func ensureCreatedByTestLabel(obj metav1.Object) {
--- a/test/e2e/metrics_adapter_test.go
+++ b/test/e2e/metrics_adapter_test.go
@@ -30,6 +30,27 @@ import (
 	apiservicesv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 )
 
+const MetricsServerFeatureGate string = "MetricsServer"
+
+func skipMetricsServerTests(t *testing.T) {
+	if !f.IsFeatureGateEnabled(t, MetricsServerFeatureGate) {
+		t.Skip("Skipping Metrics Server test")
+	}
+}
+
+func skipPrometheusAdapterTests(t *testing.T) {
+	if f.IsFeatureGateEnabled(t, MetricsServerFeatureGate) {
+		t.Skip("Skipping Prometheus Adapter test")
+	}
+}
+
+func isAPIServicePointingToRightMetricsService(t *testing.T, metricsService *apiservicesv1.APIService) bool {
+	if f.IsFeatureGateEnabled(t, MetricsServerFeatureGate) {
+		return metricsService.Spec.Service.Name == "metrics-server"
+	}
+	return metricsService.Spec.Service.Name == "prometheus-adapter"
+}
+
 func isNodeInNodesList(node string, nodes []corev1.Node) bool {
 	for _, n := range nodes {
 		if n.Name == node {
@@ -68,6 +89,10 @@ func TestMetricsAPIAvailability(t *testing.T) {
 		}
 		if !isAPIServiceAvailable(metricsService.Status.Conditions) {
 			lastErr = errors.New("v1beta1.metrics.k8s.io apiservice is not available")
+			return false, nil
+		}
+		if !isAPIServicePointingToRightMetricsService(t, metricsService) {
+			lastErr = errors.New("v1beta1.metrics.k8s.io apiservice is not pointing to right metrics api service")
 			return false, nil
 		}
 		return true, nil
@@ -226,6 +251,7 @@ func TestAggregatedMetricPermissions(t *testing.T) {
 }
 
 func TestPrometheusAdapterCARotation(t *testing.T) {
+	skipPrometheusAdapterTests(t)
 	ctx := context.Background()
 	// Wait for prometheus-adapter deployment
 	f.AssertDeploymentExistsAndRollout("prometheus-adapter", f.Ns)(t)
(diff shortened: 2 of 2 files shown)


slashpai:
I will add the deployment assertion test as well for metrics-server

slashpai:
> I will add the deployment assertion test as well for metrics-server

This is added now

test/e2e/framework/helpers.go
```diff
@@ -168,6 +170,23 @@ func (f *Framework) MustGetPods(t *testing.T, namespace string) *v1.PodList {
...
+
+	enabledFeatureGates := fg.Status.FeatureGates[0].Enabled
+
+	for _, g := range enabledFeatureGates {
+		if g.Name == configv1.FeatureGateName(name) {
+			return true, nil
+		}
+	}
+
+	return false, nil
```
jan--f, line 187: Two suggestions here:

1. You could pass `t *testing.T` and add a line `t.Helper()` and then call `t.Fatal[f]` on an error. This is in line with the other helpers (see `MustGetPods` for example) and allows you to drop the additional layer in the test file.
2. Do we need or want to poll on `.FeatureGates().Get()`? We do this for other resources and I tihnk this might the test a bit more resilient against temporary cluster issues.

test/e2e/metrics_server_adapter_test.go
jan--f, line 1: Should this maybe be called `metrics_adapter_test.go` since this also tests `prometheus-adapter`?

slashpai: Ya that was I named initially. But   ok to change to same 🙂

test/e2e/framework/helpers.go
slashpai: I didn't use poll here since featuregate 'cluster' would be already existing on cluster after its installed. Do we want to poll in that case?

slashpai:
@jan--f could you please review again?

test/e2e/framework/helpers.go
jan--f: It could be that there is an intermittent network issue that prevents the call from succeeding the first time. But we can go ahead with this and add a poll later if that turns out to be needed.

test/e2e/metrics_adapter_test.go
```diff
@@ -31,6 +31,12 @@ import (
 	apiservicesv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 )
 
+const MetricsServerFeatureGate string = "MetricsServer"
+
+func isMetricsServerFeatureGateEnabled(t *testing.T) bool {
+	return f.IsFeatureGateEnabled(t, MetricsServerFeatureGate)
+}
+
```
jan--f, line 39: polish: I think this can be dropped? Just using `f.IsFeatureGateEnabled(t, MetricsServerFeatureGate)` should work the same and is pretty readable too.

slashpai: right, updated :)

slashpai:
@jan--f updated :)

test/e2e/metrics_adapter_test.go
```diff
@@ -58,6 +60,13 @@ func isAPIServiceAvailable(conditions []apiservicesv1.APIServiceCondition) bool
...
 }
 
+func isAPIServicePointingToRightMetricsService(t *testing.T, metricsService *apiservicesv1.APIService) bool {
+	if f.IsFeatureGateEnabled(t, MetricsServerFeatureGate) {
+		return metricsService.Spec.Service.Name == "metrics-server"
+	}
+	return metricsService.Spec.Service.Name == "prometheus-adapter"
+}
+
 func TestMetricsAPIAvailability(t *testing.T) {
```
machine424, line 70: I don't think we need those dedicatedServiceMonitors tests after https://github.com/openshift/cluster-monitoring-operator/pull/2160

I'll prepare a PR to clean that up to simplify the tests.

test/e2e/metrics_adapter_test.go
```diff
@@ -79,7 +88,9 @@ func TestMetricsAPIAvailability(t *testing.T) {
 `,
 		},
 	} {
-		f.MustCreateOrUpdateConfigMap(t, f.BuildCMOConfigMap(t, tc.cmoConfig))
+		if !f.IsFeatureGateEnabled(t, MetricsServerFeatureGate) {
```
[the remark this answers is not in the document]
machine424: I'll prepare a PR to clean up the tests.

test/e2e/metrics_adapter_test.go
```diff
@@ -319,7 +335,54 @@ func TestAggregatedMetricPermissions(t *testing.T) {
 	}
 }
 
+func TestMetricsServerRollout(t *testing.T) {
```
machine424, line 338: Will this run on a cluster with prom-adapter and where we'll switch on mertricsserver? or will it run on a cluster that never knew about prom-adapter?

test/e2e/framework/helpers.go
```diff
@@ -168,6 +170,24 @@ func (f *Framework) MustGetPods(t *testing.T, namespace string) *v1.PodList {
 	return pods
 }
 
+func (f *Framework) IsFeatureGateEnabled(t *testing.T, name string) bool {
```
machine424, line 173: suggestion: could we refactor and reuse the code at https://github.com/openshift/cluster-monitoring-operator/blob/0e876ff8f315847f3699af1ed9e8cbce3e34d6c3/pkg/operator/operator.go#L420? that would also allow testing it.

test/e2e/metrics_adapter_test.go
machine424: here we go https://github.com/openshift/cluster-monitoring-operator/pull/2210

test/e2e/metrics_adapter_test.go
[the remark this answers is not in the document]
slashpai: Ya I was trying to avoid unnecessary func calls

test/e2e/framework/helpers.go
slashpai: That looks a bit complicated since it depends on informers but as part as this test is concerned IIUC checking feature gate status is good enough for us since we don't have a long running cluster in CI and we don't need to keep a informer to check the featuregate change for test.

test/e2e/metrics_adapter_test.go
slashpai: We add a check to verify if MetricsServer featuregate enabled inside this func, it will proceed only if featuregate enabled so currently only on TechPreview cluster

```
if !f.IsFeatureGateEnabled(t, MetricsServerFeatureGate) {
		return
	}
```

test/e2e/metrics_adapter_test.go
```diff
@@ -319,7 +335,54 @@ func TestAggregatedMetricPermissions(t *testing.T) {
...
+	if !f.IsFeatureGateEnabled(t, MetricsServerFeatureGate) {
+		return
```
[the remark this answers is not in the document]
slashpai: good suggestion, updating it :)

test/e2e/metrics_adapter_test.go
slashpai: I updated using t.skip() instead of return

slashpai:
@machine424 @jan--f updated per comments

test/e2e/framework/helpers.go
```diff
@@ -168,6 +170,24 @@ func (f *Framework) MustGetPods(t *testing.T, namespace string) *v1.PodList {
...
+	t.Helper()
+	fg, err := f.OpenShiftConfigClient.ConfigV1().FeatureGates().Get(context.Background(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("failed to get featuregate `cluster`: %s", err.Error())
+	}
+
+	enabledFeatureGates := fg.Status.FeatureGates[0].Enabled
```
raptorsun, line 180: just curious, is there always one and only one `featuregate` resource in every cluster?

simonpasquier:
To be honest and after reviewing the existing code, I'm not sure that the end-to-end tests care about which implementation is being deployed. What we want to verify is that the metrics.k8s.io API works as expected.

test/e2e/framework/helpers.go
slashpai: For `cluster` I have seen only. Not sure why it is an array though :)

slashpai:
> To be honest and after reviewing the existing code, I'm not sure that the end-to-end tests care about which implementation is being deployed. What we want to verify is that the metrics.k8s.io API works as expected.

IIUC that gets verified already with existing test. We still need to verify if switch happened between service right from prometheus-adapter to metrics-server service.

test/e2e/metrics_adapter_test.go
machine424: Just the term `Rollout` + `...is deleted` make me think that the test would run after a migration from prom adapter to metricsserver.

test/e2e/framework/helpers.go
machine424: (Now that I think about it, I don't see any safeguard on our side making sure that the techpreview clusters really have metricsserver enabled, we could end up always running prom-adapter tests... Let's continue discussing this somewhere else :))

test/e2e/framework/helpers.go
jan--f: https://github.com/openshift/api/blob/081317a013f87bfe55655c0441c07cfbd7087c54/config/v1/types_feature.go#L85-L102

There can be multiple `FeatureGates` and iiuc they should differ by version. 
Reading the doc string I do wonder if we need to actually look at the versions here (and in the operator code too)?

jan--f: I suppose this could list feature gates for multiple versions that change the clusters behavior when the cluster is updated...

slashpai: updated code to include the version check as well

rexagod:
> What we want to verify is that the metrics.k8s.io API works as expected.

Just thinking out loud here, but maybe we could additionally check if the `{Node,Pod}Metrics` resource updates (if the resource `timestamp` exceeds the time at which the `metrics-server` was [up](https://github.com/kubernetes-sigs/metrics-server/blob/af43f4066a7113812fa5ffc1c2e22731cda9b7ca/charts/metrics-server/values.yaml#L112)) after the replacement.

jan--f:
> > What we want to verify is that the metrics.k8s.io API works as expected.
> 
> Just thinking out loud here, but maybe we could additionally check if the `{Node,Pod}Metrics` resource updates (if the resource `timestamp` exceeds the time at which the `metrics-server` was [up](https://github.com/kubernetes-sigs/metrics-server/blob/af43f4066a7113812fa5ffc1c2e22731cda9b7ca/charts/metrics-server/values.yaml#L112)) after the replacement.

That would require us to add a test that enables a feature gate wouldn't it? Not sure we have such a test and if so it would probably be somewhere else. 
I think this PR is ready to go head.

slashpai:
@jan--f Can you take a look at this again :)

jan--f:
Thanks! This
/lgtm

PR #2199, opened 2023-12-13
OCPBUGS-24977: use 1.20 in go.mod

jan--f:


marioferh (commented):
/lgtm

jan--f:
/hold
I introduced a bug into the Makefile

go.mod
```diff
@@ -158,3 +158,5 @@ require (
 )
 
 replace k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20230928195430-ce36a0c3bb67
+
+replace github.com/prometheus-operator/prometheus-operator v0.70.0 => github.com/openshift/prometheus-operator v0.20.1-0.20231211152750-5bf04f3ca997
```
danielmellado, line 162: I asume that the replace is just using whatever commit was just before this https://github.com/prometheus-operator/prometheus-operator/commit/4e63f2fddb9bad1c682fea1f151c23f47dab75eb isn't it?

danielmellado:
Just a minor nit, lgtm overall!

go.mod
jan--f: Nope this replaces upstream with the `openshift` clone. See the commit message for details.

danielmellado: Ack then, I'd put a not on the PR but lgtm, thanks!

PR #2206, opened 2023-12-15
OCPBUGS-25378: No definition of InfoInhibitor is required in alertmanager.yaml

QuadJust:
The cluster-monitoring-operator does not have a definition for InfoInhibitor. To avoid confusion for the customer, it is safe to remove this from the alertmanager.yaml.

### DIFF
--- a/assets/alertmanager/secret.yaml
+++ b/assets/alertmanager/secret.yaml
@@ -29,17 +29,10 @@ stringData:
       - "severity = warning"
       "target_matchers":
       - "severity = info"
-    - "equal":
-      - "namespace"
-      "source_matchers":
-      - "alertname = InfoInhibitor"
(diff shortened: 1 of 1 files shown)


danielmellado:
@simonpasquier recently merged a PR referencing this, https://github.com/openshift/cluster-monitoring-operator/pull/2182/files so I'm not sure whether this would collide, let's test this and check.

simonpasquier:
`assets/alertmanager/secret.yaml` is generated from jsonnet/components/alertmanager.libsonnet which itself comes from github.com/promehteus-operator/kube-prometheus. The change isn't as easy as it looks.

QuadJust:
@simonpasquier Thanks for the clarification. Ummm, I got it. So, should I withdraw this PR?

PR #2210, opened 2023-12-18
MON-3552: remove temporary cleanup code and no longer needed tests

machine424:
… is not longer taken into account

triggered by https://github.com/openshift/cluster-monitoring-operator/pull/2196#discussion_r1430125733

/cc @slashpai 
/cc @jan--f 
/cc @simonpasquier


machine424:
Let's see what the CI has to say.

test/e2e/prometheusadapter_test.go
```diff
@@ -59,37 +59,11 @@ func isAPIServiceAvailable(conditions []apiservicesv1.APIServiceCondition) bool
...
-			cmoConfig: `prometheusK8s:
-  logLevel: debug
-  k8sPrometheusAdapter:
-    dedicatedServiceMonitors:
-      enabled: true
-`,
-		},
-	} {
-		f.MustCreateOrUpdateConfigMap(t, f.BuildCMOConfigMap(t, tc.cmoConfig))
+	// Check the prerequisites to make the test easier to debug.
```
machine424, line 62: or we can drop them :)
(If we could inject those pre-checks in the tests that rely on them, that would make debugging the tests easier...)

simonpasquier: The workflow was to modify the CMO config then check that CMO didn't go degraded/unavailable. We can remove the condition checks since the test don't modify the CMO config anymore.

test/e2e/prometheusadapter_test.go
```diff
@@ -113,37 +87,10 @@ func checkMetricsAPIAvailability(t *testing.T) {
...
-		f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionFalse)(t)
-		f.AssertOperatorCondition(configv1.OperatorProgressing, configv1.ConditionFalse)(t)
-		f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionTrue)(t)
-
-		checkNodeMetricsPresence(t)
-	}
-}
+	f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionFalse)(t)
+	f.AssertOperatorCondition(configv1.OperatorProgressing, configv1.ConditionFalse)(t)
+	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionTrue)(t)
```
simonpasquier, line 92: same here

test/e2e/prometheusadapter_test.go
```diff
@@ -59,37 +59,11 @@ func isAPIServiceAvailable(conditions []apiservicesv1.APIServiceCondition) bool
 }
 
 func TestMetricsAPIAvailability(t *testing.T) {
```
simonpasquier, line 61: I'm not sure that we need this test anymore since TestPodMetricsPresence and TestNodeMetricsPresence already validate that the metrics API service is working.

simonpasquier (commented):
We should also remove the DedicatedServiceMonitors section from the generated documentation.

test/e2e/prometheusadapter_test.go
slashpai: I think this test helps us to verify APIService

simonpasquier: If the API service isn't properly deployed then the metrics API will not work and the e2e tests exercising the API will fail. I feel that this particular test is redundant and I'd rather focus on user-facing scenarios.

jan--f:
> We should also remove the DedicatedServiceMonitors section from the generated documentation.

We also discussed removing the config map field eventually. Iiuc the main counter argument against removing the field is that we'd break users configurations when they missed (or failed to act on) the deprecation notice.
Otoh removing the doc but leaving the field in place seems like we encourage superfluous config fragments, that make user configs harder to understand.
Should we either leave the documentation (with deprecation notice) in place or remove both config and field?

test/e2e/prometheusadapter_test.go
machine424: yep, I just find it useful to validate these preconditions before running a test, it helps debugging: maybe the component we're willing to test is not well deployed, or was broken by another test etc.
I'll remove them for now, I'll dig into this in the future.

test/e2e/prometheusadapter_test.go
machine424: I see that in https://github.com/openshift/cluster-monitoring-operator/pull/2196, JP added a check into it to make sure it's pointing to the right service, that could be a good debugging test. But I agree that the check on the resource's status is redundant.

The test existed before `DedicatedServiceMonitors`, I think it's better to deal with it in another PR.

machine424:
> > We should also remove the DedicatedServiceMonitors section from the generated documentation.
> 
> We also discussed removing the config map field eventually. Iiuc the main counter argument against removing the field is that we'd break users configurations when they missed (or failed to act on) the deprecation notice. Otoh removing the doc but leaving the field in place seems like we encourage superfluous config fragments, that make user configs harder to understand. Should we either leave the documentation (with deprecation notice) in place or remove both config and field?

Forgot that we're already dev for 4.16 :expressionless:, if we want to keep the deprecation notice until 4.17 or more (don't know how much versions we should wait for this), +1 to keep it on the doc, if not, let's just remove everything.

I'll address the `// TODO: Remove this in 4.16` here as well.

machine424:
/retitle MON-3552: remove temporary cleanup code and no longer needed tests

pkg/manifests/config.go
```diff
@@ -291,9 +291,8 @@ func (c *Config) applyDefaults() {
 	if c.ClusterMonitoringConfiguration.K8sPrometheusAdapter == nil {
 		c.ClusterMonitoringConfiguration.K8sPrometheusAdapter = &K8sPrometheusAdapter{}
 	}
-	if c.ClusterMonitoringConfiguration.K8sPrometheusAdapter.DedicatedServiceMonitors == nil {
-		c.ClusterMonitoringConfiguration.K8sPrometheusAdapter.DedicatedServiceMonitors = &DedicatedServiceMonitors{}
-	}
+	// TODO: Remove this in 4.XX?
```
machine424, line 294: waiting for decision https://github.com/openshift/cluster-monitoring-operator/pull/2210#issuecomment-1863835566

(specify a version or get rid of this + its doc)

pkg/manifests/config_test.go
```diff
@@ -174,27 +174,44 @@ func TestTelemeterClientConfig(t *testing.T) {
 	}
 }
 
+// TODO: Remove this in 4.XX?
```
machine424, line 177: same as above.

pkg/manifests/config.go
simonpasquier: My recommendation would be:
* remove the field from the documentation
* keep the field in the config struct "forever"
* have a metric + alert (info severity) in CMO reporting that the field is in use but has no effect.

jan--f: Works for me, happy to defers to Simons experience here :)
I suppose we don't have a metric like that yet? Maybe something like `cluster_monitoring_operator_deprecated_config_use{deprecated_field="$FIELD_NAME"}`? Is it worth including the version this was deprecated?

machine424: We'll be able to get rid of it during the migration into the CRD :)

I'd say that a log (to ease any potential debugging related to the config) + the deprecation notice  in the release notes are sufficient especially that the change should be transparent to users. If users have nothing to worry about (config kept forever), why warn them?

We could give the metric a try, it could be helpful during the migration to CRD, but once the CRD are in place, it'd be easier for CMO to prune those deprecated fields.

machine424: Yes, we could add the deprecation version as well, I'll give it a try.

test/e2e/config_test.go
```diff
@@ -800,6 +800,35 @@ monitoringPlugin:
 	}
 }
 
+func TestClusterMonitoringDeprecatedConfig(t *testing.T) {
+	metricName := "cluster_monitoring_operator_deprecated_config_in_use"
+	checkMetricValue := func(value float64) {
+		f.PrometheusK8sClient.WaitForQueryReturn(
+			t, 5*time.Minute, fmt.Sprintf(`%s{configmap="openshift-monitoring/cluster-monitoring-config", field="k8sPrometheusAdapter.dedicatedServiceMonitors", deprecation_version="4.15"}`, metricName),
```
machine424, line 807: check the alert itself?

pkg/manifests/config.go
```diff
@@ -446,6 +427,35 @@ func (c *Config) LoadEnforcedBodySizeLimit(pcr PodCapacityReader, ctx context.Co
...
+			metrics.CollectionProfile.WithLabelValues(string(profile)).Set(v)
+		}
+		if !slices.Contains(SupportedCollectionProfiles, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile) {
+			return errors.Wrap(ErrConfigValidation, fmt.Sprintf(`%q is not supported, supported collection profiles are: %q`, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile, SupportedCollectionProfiles.String()))
+		}
+	}
+
+	var d float64
+	// Highlight deprecated config fields.
+	if c.ClusterMonitoringConfiguration.K8sPrometheusAdapter.DedicatedServiceMonitors != nil {
```
machine424, line 451: This will not cover:
```
k8sPrometheusAdapter:
  dedicatedServiceMonitors:
``` 
(where `dedicatedServiceMonitors` is empty)
I'll see if we could cover it as well

machine424: I don't think it's easy to catch that without custom unmarshalling which is not worth it...

jan--f: I think that's fine. We want to alert users when they use a deprecated field, even if empty.

pkg/manifests/config.go
```diff
@@ -446,6 +427,35 @@ func (c *Config) LoadEnforcedBodySizeLimit(pcr PodCapacityReader, ctx context.Co
...
+		d = 1
+	}
+	// Deprecated in https://github.com/openshift/cluster-monitoring-operator/pull/2160
+	metrics.DeprecatedConfig.WithLabelValues("openshift-monitoring/cluster-monitoring-config", "k8sPrometheusAdapter.dedicatedServiceMonitors", "4.15").Set(d)
```
jan--f, line 455: Do we actually emit the 0 values? Instead we could just omit/remove the metric. That is the happy path and metric absence signals the same as the 0, but saves us a metric.

pkg/manifests/config.go
machine424: Actually both
```
k8sPrometheusAdapter:
  dedicatedServiceMonitors:
```
and
```
k8sPrometheusAdapter:
```

result in `K8sPrometheusAdapter.DedicatedServiceMonitors=nil`, so we'll not be able to warn on empty `dedicatedServiceMonitors`.

pkg/manifests/config.go
machine424: yes, if the user gets rid of the deprecated config we need to stop emitting 1 (as in the test) and I don't think it's a good idea to just drop the metric, I don't think that`metrics` allows even doing that.
Plus it's good to be able to always list deprecated fields...

pkg/manifests/config_test.go
```diff
@@ -342,56 +318,104 @@ func TestLoadEnforcedBodySizeLimit(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, tc.expected, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile)
+		})
+	}
+}
 
-			if tc.expectedsp != c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile {
-				t.Fatalf("incorrect collection profile set, expected %s got %s", tc.expectedsp, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile)
-			}
+func TestDeprecatedConfig(t *testing.T) {
```
machine424, line 372: we can do without if you want.

pkg/manifests/config.go
jan--f: fair point.

jan--f:
Given that I have no further comments and others have been silent for a few weeks, this

/lgtm

We can always address further concerns in a follow-up

PR #2213, opened 2023-12-21
OCPBUGS-24977: Updating cluster-monitoring-operator-container image to be consistent with ART

simonpasquier:
Supersedes #2207


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/openshift/cluster-monitoring-operator
 
-go 1.20
+go 1.21
```
machine424, line 3: Maybe we could automate the upgrade of this as well.

simonpasquier:
/close
https://github.com/openshift/cluster-monitoring-operator/pull/2211 is the one we want to merge.

PR #2218, opened 2023-12-22
OCPBUGS-25849: make PrometheusAdapter and MetricsServer tasks less conflict prone

machine424:
…dateAPIService to avoid code duplication.

The function also avoids running update queries when nothing needs to change from CMO's side which reduces the probability of conflicts with other controllers and consumes resources for nothing.

### DIFF
--- a/assets/metrics-server/api-service.yaml
+++ b/assets/metrics-server/api-service.yaml
@@ -16,5 +16,6 @@ spec:
   service:
     name: metrics-server
     namespace: openshift-monitoring
+    port: 443
   version: v1beta1
   versionPriority: 100
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1831,31 +1831,8 @@ func (c *Client) CreateOrUpdateServiceMonitor(ctx context.Context, sm *monv1.Ser
 }
 
 func (c *Client) CreateOrUpdateAPIService(ctx context.Context, apiService *apiregistrationv1.APIService) error {
-	apsc := c.aggclient.ApiregistrationV1().APIServices()
-	existing, err := apsc.Get(ctx, apiService.GetName(), metav1.GetOptions{})
-	if apierrors.IsNotFound(err) {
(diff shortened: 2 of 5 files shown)


machine424:
/retitle OCPBUGS-25849: make PrometheusAdapter and MetricsServer tasks less conflict prone

assets/metrics-server/api-service.yaml
```diff
@@ -16,5 +16,6 @@ spec:
   service:
     name: metrics-server
     namespace: openshift-monitoring
+    port: 443
```
machine424, line 19: maybe in the future we can change `ApplyAPIService` so it takes the default value of the port (443) into consideration
(a warning would be useful as well for other users)

pkg/client/client.go
```diff
@@ -1723,25 +1723,8 @@ func (c *Client) CreateOrUpdateServiceMonitor(ctx context.Context, sm *monv1.Ser
...
-	required := apiService.DeepCopy()
-	required.ResourceVersion = existing.ResourceVersion
-	if val, ok := required.Annotations["service.beta.openshift.io/inject-cabundle"]; ok && val == "true" {
-		if len(existing.Spec.CABundle) > 0 {
-			required.Spec.CABundle = existing.Spec.CABundle
-		}
-	}
-	_, err = apsc.Update(ctx, required, metav1.UpdateOptions{})
-	return errors.Wrap(err, "updating APIService object failed")
+	_, _, err := resourceapply.ApplyAPIService(ctx, c.aggclient.ApiregistrationV1(), c.eventRecorder, apiService)
```
machine424, line 1726: deleting a label/annotation, would require two steps https://github.com/openshift/library-go/blob/5674ec6ced1c6e6ddc86201e5232d6713c7a262c/pkg/operator/resource/resourcemerge/object_merger.go#L141

+

labels + annotations will be merged now, as we do in `CreateOrUpdateServiceAccount` e.g.

PR #2220, opened 2023-12-29
MON-3644: Ease the tracking of monitoring components versions.

machine424:
…ate_versions.go and provide components versions for multiple OCP versions


hack/tools/generate_versions.go
```diff
@@ -0,0 +1,212 @@
+package main
```
machine424, line 1: It'd be great if we gradually migrate the other `.sh` scripts into Go for more readability and sanity :)
I'll create a ticket

Tell me if comments are needed for thei one.

machine424:
/retitle MON-3644: Ease the tracking of monitoring components versions.

hack/tools/generate_versions.go
simonpasquier: I'd rather keep hack/tools for tracking/installing external Go binaries that are needed by the build process. Can we move the program to a dedicated subdirectory?

hack/tools/generate_versions.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/olekukonko/tablewriter"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	mainBranch        = "master"
+	metricsServerRepo = "openshift/kubernetes-metrics-server"
+	versionFile       = "../../jsonnet/versions.yaml"
+	// This representation works for now.
+	minOCPVersion       = float32(4.10)
```
simonpasquier, line 22: In prometheus-operator, we use `github.com/blang/semver/v4` which is ok.

Documentation/deps-versions.md
```diff
@@ -0,0 +1,9 @@
+|  OCP Version |                                       alertmanager                                       |                                  kubeRbacProxy                                   |                                  kubeStateMetrics                                   |                                                kubernetesMetricsServer                                                 |                                 monitoringPlugin                                  |                                 nodeExporter                                  |                                  promLabelProxy                                  |                                 prometheus                                  |                                    prometheusAdapter                                    |                                  prometheusOperator                                  |                                 thanos                                  |
```
simonpasquier, line 1: I wonder if it wouldn't be more readable to have dedicated sections per minor?

## 4.16

| Component | Version |
|--------|--------|
| Alertmanager | 0.26.0 |
| kube-rbac-proxy | 0.15.0 |
| kube-state-metrics | 2.10.1 | 
| ... | ... | 

## 4.15

...

Documentation/deps-versions.md
```diff
@@ -0,0 +1,9 @@
...
+| release-4.10 | [0.23.0](https://github.com/openshift/prometheus-alertmanager/blob/release-4.10/VERSION) | [0.11.0](https://github.com/openshift/kube-rbac-proxy/blob/release-4.10/VERSION) | [2.3.0](https://github.com/openshift/kube-state-metrics/blob/release-4.10/VERSION)  | N/A                                                                                                                    | N/A                                                                               | [1.3.1](https://github.com/openshift/node_exporter/blob/release-4.10/VERSION) | [0.4.0](https://github.com/openshift/prom-label-proxy/blob/release-4.10/VERSION) | [2.32.1](https://github.com/openshift/prometheus/blob/release-4.10/VERSION) | [0.9.1](https://github.com/openshift/k8s-prometheus-adapter/blob/release-4.10/VERSION)  | [0.53.1](https://github.com/openshift/prometheus-operator/blob/release-4.10/VERSION) | [0.23.2](https://github.com/openshift/thanos/blob/release-4.10/VERSION) |
```
simonpasquier, line 3: (nit) The link could be to the branch's root directory (e.g. https://github.com/openshift/prometheus-alertmanager/blob/release-4.10).

machine424: I point it to the file containing the version to ease the retrieval of the commit that upgraded the version.

hack/tools/generate_versions.go
machine424: `4.10` isn't really a semver... But I pushed a version using it by ignoring the `patch`

Documentation/deps-versions.md
machine424: You'd need to scroll down a lot for older versions and the unique table easier to generate...

hack/tools/generate_versions.go
machine424: You're right, I thought I put them next to the sh ones...
I put them in a go/ (not too creative :)) dir now.

jsonnet/versions.yaml
```diff
@@ -1,28 +1,28 @@
 ---
-# This file is meant to be managed by hack/generate-versions.sh script
+# This file is meant to be managed by hack/tools/generate_versions.go script
```
jan--f, line 2:
```suggestion
# This file is meant to be managed by hack/go/generate_versions.go script
```

Documentation/deps-versions.md
jan--f: If we go with a single table (no strong opinion on that) can we have the newest version on top?

Documentation/deps-versions.md
jan--f: +1 for linking the repo root. For retrieval you could just hardcode the `VERSION` path component, no?

Documentation/deps-versions.md
machine424: newest versions are on top now.

Documentation/deps-versions.md
machine424: done.

jsonnet/versions.yaml
machine424: thanks

jan--f:
Thanks! This
/lgtm

PR #2223, opened 2024-01-08
WIP: MON-3352: Add error msg when multiple operators try to access same prometheus resource

marioferh:
## Description
If multiple operators try to access the same resource, it will cause the creation of the resource to fail indefinitely.

## Solution

By checking the resource generation and the replicas that are currently running, we can detect the problem.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -937,8 +937,9 @@ func (c *Client) ValidatePrometheus(ctx context.Context, promNsName types.Namesp
 	})
 
 	s, _ := c.kclient.AppsV1().StatefulSets("openshift-monitoring").Get(ctx, "prometheus-k8s", metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


pkg/client/client.go
```diff
@@ -932,6 +936,12 @@ func (c *Client) ValidatePrometheus(ctx context.Context, promNsName types.Namesp
 		return stop, nil
 	})
 
+	s, _ := c.kclient.AppsV1().StatefulSets("openshift-monitoring").Get(ctx, "prometheus-k8s", metav1.GetOptions{})
+	if c.lastGenerationNumber+10 < int(s.Status.ObservedGeneration) && s.Status.ReadyReplicas != s.Status.Replicas {
```
danielmellado, line 940: I would leave a comment here explaining the reason behind the '+10' in the generation number. Probably also worth adding something in the Error message.

marioferh: You are right, I think now is clearer

PR #2224, opened 2024-01-08
MON-3654: use Go standard errors package instead of github.com/pkg/errors

machine424:


.golangci.yaml
```diff
@@ -27,8 +29,16 @@ linters:
     - whitespace
     - gci
 
-linter-settings:
```
machine424, line 30: I'll see if we could enable strict unmarshalling got golangci

jan--f:
Nice improvement, thanks.
I'm a bit confused about the import reordering. If there is a string reason, sure let's do it. But then lets add a comment (or even better commit message) about the why.

machine424:
> I'm a bit confused about the import reordering. If there is a string reason, sure let's do it. But then lets add a comment (or even better commit message) about the why.

It's done in the second commit:

```
chore: add errorlint and depguard linters and fix gci linter config

errorlint: 'check for code that could be problematic with the error wrapping scheme'
depguard: to prevent the use of the deprecated 'github.com/pkg/errors'.
```

I needed to configure `depguard` and discovered that the config of `cgi` which handles import ordering is broken, so had to fix it.

jan--f:
Thanks!
/lgtm

PR #2225, opened 2024-01-08
OCPBUGS-25378: drop InfoInhibitor from default Alertmanager config

simonpasquier:
OCP doesn't include the InfoInhibitor alerting rule out of the box. To avoid confusion, it is desired to remove the alert from the default Alertmanager configuration.

Note that the change will only apply to newly created clusters because the Cluster Monitoring Operator only handles the creation of the Alertmanager configuration secret and it never updates when the secret already exists.

### DIFF
--- a/assets/alertmanager/secret.yaml
+++ b/assets/alertmanager/secret.yaml
@@ -12,8 +12,6 @@ metadata:
   namespace: openshift-monitoring
 stringData:
   alertmanager.yaml: |-
-    "global":
-      "resolve_timeout": "5m"
     "inhibit_rules":
     - "equal":
       - "namespace"
@@ -29,17 +27,10 @@ stringData:
       - "severity = warning"
       "target_matchers":
       - "severity = info"
-    - "equal":
-      - "namespace"
-      "source_matchers":
-      - "alertname = InfoInhibitor"
-      "target_matchers":
-      - "severity = info"
     "receivers":
     - "name": "Default"
     - "name": "Watchdog"
     - "name": "Critical"
-    - "name": "null"
     "route":
       "group_by":
       - "namespace"
@@ -51,9 +42,6 @@ stringData:
       - "matchers":
         - "alertname = Watchdog"
         "receiver": "Watchdog"
-      - "matchers":
-        - "alertname = InfoInhibitor"
-        "receiver": "null"
       - "matchers":
         - "severity = critical"
         "receiver": "Critical"
--- a/jsonnet/main.jsonnet
+++ b/jsonnet/main.jsonnet
@@ -123,6 +123,33 @@ local inCluster =
         },
         kubeRbacProxyImage: $.values.common.images.kubeRbacProxy,
         promLabelProxyImage: $.values.common.images.promLabelProxy,
+        config: {
+          inhibit_rules: [{
+            source_matchers: ['severity = critical'],
+            target_matchers: ['severity =~ warning|info'],
+            equal: ['namespace', 'alertname'],
+          }, {
+            source_matchers: ['severity = warning'],
+            target_matchers: ['severity = info'],
+            equal: ['namespace', 'alertname'],
+          }],
+          route: {
+            group_by: ['namespace'],
+            group_wait: '30s',
+            group_interval: '5m',
+            repeat_interval: '12h',
+            receiver: 'Default',
+            routes: [
+              { receiver: 'Watchdog', matchers: ['alertname = Watchdog'] },
+              { receiver: 'Critical', matchers: ['severity = critical'] },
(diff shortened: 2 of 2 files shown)


assets/alertmanager/secret.yaml
```diff
@@ -29,17 +29,10 @@ stringData:
...
-      - "namespace"
-      "source_matchers":
-      - "alertname = InfoInhibitor"
-      "target_matchers":
-      - "severity = info"
     "receivers":
     - "name": "Default"
     - "name": "Watchdog"
     - "name": "Critical"
-    - "name": "null"
```
machine424, line 42: I'm thinking about some gitops scripts that get applied to all new clusters and that suppose the existence of the `null` receiver on openshfit..
(same thing for the `InfoInhibitor` rule, "no need to add it, openshift already does..." we don't state that we support that explicitly, but still)

machine424: (users would get alerts in these two cases though)

simonpasquier: I'm not sure to follow but I expect that users managing lots of new clusters have their own default Alertmanager configuration and don't rely on the default one.

jsonnet/main.jsonnet
```diff
@@ -123,6 +123,36 @@ local inCluster =
         },
         kubeRbacProxyImage: $.values.common.images.kubeRbacProxy,
         promLabelProxyImage: $.values.common.images.promLabelProxy,
+        config: {
```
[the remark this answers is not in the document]
simonpasquier: It was my initial idea but I couldn't make it work.

assets/alertmanager/secret.yaml
machine424: I was thinking about users that will only modify the `route` field of this config to add routes to`null` assuming that the receiver `null` is there (but I think that alertmanager will warn about this).

I was also thinking about `AlertmanagerConfig` that would send to `null` in the default config, but I think `AlertmanagerConfig` should define their own receivers: https://github.com/prometheus-operator/prometheus-operator/blob/4cf0f5eae722add3d17530004abbef002e6dad08/pkg/alertmanager/validation/v1beta1/validation.go#L45,

simonpasquier: Discussed offline: we provide no stability guarantee on the default Alertmanager configuration. The assumption is that users would overwrite the out-of-the-box secret instead of patching it so altering the default config shouldn't break users.

jsonnet/main.jsonnet
```diff
@@ -123,6 +123,36 @@ local inCluster =
...
+              { receiver: 'Watchdog', matchers: ['alertname = Watchdog'] },
+              { receiver: 'Critical', matchers: ['severity = critical'] },
+            ],
+          },
+          receivers: [
+            { name: 'Default' },
+            { name: 'Watchdog' },
+            { name: 'Critical' },
+          ],
+        },
```
machine424, line 155: How about just combining `config` and `route` with `+`? This way we'll still get new fields + updates for the others fields from upstream.
(Of course we may get "broken/incomplete" updates as we'll be overwriting some of the fields...)

```suggestion
        config+: {
          inhibit_rules: [{
            source_matchers: ['severity = critical'],
            target_matchers: ['severity =~ warning|info'],
            equal: ['namespace', 'alertname'],
          }, {
            source_matchers: ['severity = warning'],
            target_matchers: ['severity = info'],
            equal: ['namespace', 'alertname'],
          }],
          route+: {
            routes: [
              { receiver: 'Watchdog', matchers: ['alertname = Watchdog'] },
              { receiver: 'Critical', matchers: ['severity = critical'] },
            ],
          },
          receivers: [
            { name: 'Default' },
            { name: 'Watchdog' },
            { name: 'Critical' },
          ],
        },
```

simonpasquier: I'm on the fence: I find it clearer to have the full config here than reviewing the merged config. Again the configuration provisioned on cluster creation is only informative.

machine424: I had a feeling :)

assets/alertmanager/secret.yaml
```diff
@@ -12,8 +12,6 @@ metadata:
   namespace: openshift-monitoring
 stringData:
   alertmanager.yaml: |-
-    "global":
-      "resolve_timeout": "5m"
```
simonpasquier, line 16: this is already the default value from Alertmanager + the setting has no impact in a traditional Prometheus -> Alertmanager workflow (it only applies if the received alerts have no end date).

machine424:
/lgtm
Thanks Simon,

PR #2226, opened 2024-01-08
MON-3717: pkg/client.go: make some CreateOrUpdateXXX functions use library-go’s resourceapply utils.

machine424:

See https://docs.google.com/document/d/1fm2SZs8HroexPQnqI0Ua85Y31-lars8gCsdwSJC3ngo/edit?usp=sharing for benchmarks results.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1431,60 +1431,8 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
 }
 
 func (c *Client) CreateOrUpdateSecret(ctx context.Context, s *v1.Secret) error {
-	sClient := c.kclient.CoreV1().Secrets(s.GetNamespace())
-	existing, err := sClient.Get(ctx, s.GetName(), metav1.GetOptions{})
-	if apierrors.IsNotFound(err) {
-		_, err := sClient.Create(ctx, s, metav1.CreateOptions{})
-		if err != nil {
-			return fmt.Errorf("creating Secret object failed: %w", err)
-		}
-		return nil
-	}
-	if err != nil {
-		return fmt.Errorf("retrieving Secret object failed: %w", err)
-	}
-
-	required := s.DeepCopy()
-	mergeMetadata(&required.ObjectMeta, existing.ObjectMeta)
-	// Check if the Secret has an owner reference to a Service, that carries
-	// the annotation with key
-	// service.beta.openshift.io/serving-cert-secret-name and the Secrets
-	// name as the value.
-	// This means that service-ca-operator controls and populates the two
-	// data fields tls.crt and tls.key. We want to retain those on updates
-	// if they exist and are not empty.
-	if c.maybeHasServiceCAData(ctx, required) {
-		if v, ok := existing.Data["tls.crt"]; ok && len(v) > 0 {
-			required.Data["tls.crt"] = v
-		}
-		if v, ok := existing.Data["tls.key"]; ok && len(v) > 0 {
-			required.Data["tls.key"] = v
-		}
-	}
-	_, err = sClient.Update(ctx, required, metav1.UpdateOptions{})
-	if err != nil {
-		return fmt.Errorf("updating Secret object failed: %w", err)
-	}
-	return nil
-}
-
-// maybeHasServiceCAData checks if the passed Secret s has at least one owner reference that
-// points to a Service with the annotation service.beta.openshift.io/serving-cert-secret-name: s.name
-func (c *Client) maybeHasServiceCAData(ctx context.Context, s *v1.Secret) bool {
-	for _, owner := range s.OwnerReferences {
-		if owner.Kind != "Service" {
-			continue
-		}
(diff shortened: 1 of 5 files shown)


pkg/client/client.go
```diff
@@ -1417,7 +1417,18 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
...
+	if c.maybeHasServiceCAData(ctx, s) {
+		panic(fmt.Sprintf("Why am I touching %s/%s?", s.Namespace, s.Name))
+	}
+	_, _, err := resourceapply.ApplySecret(ctx, c.kclient.CoreV1(), c.eventRecorder, s)
+	return err
+
+
+	// IIUC, we should never touch secrets created by service-ca-operator.
+
+/* 	sClient := c.kclient.CoreV1().Secrets(s.GetNamespace())
```
machine424, line 1431: the diff at the left is weird, why is the code commented??

pkg/client/client.go
```diff
@@ -1431,7 +1437,16 @@ func (c *Client) WaitForDaemonSetRollout(ctx context.Context, ds *appsv1.DaemonS
 }
 
 func (c *Client) CreateOrUpdateSecret(ctx context.Context, s *v1.Secret) error {
-	sClient := c.kclient.CoreV1().Secrets(s.GetNamespace())
+	// TEMP: just to make sure we never touch secrets created by service-ca-operator.
+	if c.maybeHasServiceCAData(ctx, s) {
```
machine424, line 1441: @jan--f, I'd like your input on this.
I don't think CMO touches secrets created by `service-ca-operator` or should do so.
Do you remember for which cases the `maybeHasServiceCAData` check was introduced.
(A lot of things have changed since then :))

jan--f: I think you're right.
In https://github.com/openshift/cluster-monitoring-operator/pull/1495 we discussed this and there is a discussion that menions CMO adding/removing metadata, but the example linked doesn't do that for secrets. 
And I can't find code that touches (or even knows) about the service-ca-op created secrets.
@simonpasquier can you think of anything we're missing here?

machine424: Ok, thanks to you two for the confirmation, I'll clean up the code then.
For the configmaps, https://github.com/openshift/library-go/pull/1676 should help CMO reduce unnecessary competition.
for ValidatingWebhookConfiguration, the existing ressourceapply should be sufficient.

pkg/client/client.go
machine424: Checked with gihtub support, it seems to be a bug

machine424:
/hold
needs https://github.com/openshift/library-go/pull/1676

machine424:
/retitle MON-3717: pkg/client.go: make some CreateOrUpdateXXX functions use library-go’s resourceapply utils.

test/e2e/framework/framework.go
```diff
@@ -117,7 +118,7 @@ func New(kubeConfigPath string) (*Framework, CleanUpFunc, error) {
 		return nil, nil, fmt.Errorf("creating monitoring beta client failed: %w", err)
 	}
 
-	operatorClient, err := client.NewForConfig(config, "", namespaceName, userWorkloadNamespaceName)
+	operatorClient, err := client.NewForConfig(config, "", namespaceName, userWorkloadNamespaceName, client.EventRecorder(events.NewInMemoryRecorder("cluster-monitoring-operator")))
```
jan--f, line 121: Minor improvement for future readers: Maybe leave a small comment as to what the event recorder does for us here?

jan--f:
Very nice work, thanks! Looking forward to merging this.

go.mod
```diff
@@ -138,3 +138,5 @@ require (
 	sigs.k8s.io/kube-storage-version-migrator v0.0.6-0.20230721195810-5c8923c5ff96 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
+
+replace github.com/openshift/library-go => github.com/machine424/library-go v0.0.0-20240208174128-e2b66c5d004d
```
danielmellado, line 142: just saw this. Unless this is something that it's a draft or it's depending on a change from a PR to openshift/library-go I'm against this replace here. Mind clarifying this'

danielmellado (changes requested):
Let's clarify the replacement from openshift's library go to your own one over there before moving forward with the change.

go.mod
danielmellado: I assume that that's because of needs https://github.com/openshift/library-go/pull/1676

machine424: yep https://github.com/openshift/cluster-monitoring-operator/pull/2226#issuecomment-1945681548

test/e2e/framework/framework.go
machine424: sure, I'll add that.
Maybe to [pkg/client/client_test.go](https://github.com/openshift/cluster-monitoring-operator/pull/2226/files#diff-2dad64e28ce5e2eddf0adf303052d38e3b5b4fcba6b32fc4ca26d5e4d6b8fa4a) as well (talking about it incase you missed it as it's collapsed by default)

PR #2227, opened 2024-01-09
MON-3352: Add error msg when multiple operators try to access same resource

marioferh:
## Description
If multiple operators try to access the same resource (prometheus, alertmanager), it will cause the creation of the resource to fail indefinitely.

## Solution

By checking the resource generation and the replicas that are currently running, we can detect the problem.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1024,6 +1024,9 @@ func validateMonitoringResource(expectedReplicas, updatedReplicas, availableRepl
 }
 
 func (c *Client) WaitForAlertmanager(ctx context.Context, a *monv1.Alertmanager) error {
+	s, _ := c.kclient.AppsV1().StatefulSets(a.GetNamespace()).Get(ctx, "alertmanager-main", metav1.GetOptions{})
+	lastGeneration := s.Status.ObservedGeneration
+
 	var lastErr error
 	if err := Poll(ctx, func(ctx context.Context) (bool, error) {
 		a, err := c.mclient.MonitoringV1().Alertmanagers(a.GetNamespace()).Get(ctx, a.GetName(), metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


pkg/client/client.go
```diff
@@ -932,6 +932,15 @@ func (c *Client) ValidatePrometheus(ctx context.Context, promNsName types.Namesp
 		return stop, nil
 	})
 
+	p, _ := c.mclient.MonitoringV1().Prometheuses(promNsName.Namespace).Get(ctx, promNsName.Name, metav1.GetOptions{})
+	s, _ := c.kclient.AppsV1().StatefulSets("openshift-monitoring").Get(ctx, "prometheus-k8s", metav1.GetOptions{})
+
+	generationAttemps := s.Status.ObservedGeneration - p.GetGeneration()
```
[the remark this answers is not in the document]
marioferh: Yep, I was trying to do something similar of this and I confused variables: https://github.com/openshift/cluster-monitoring-operator/blob/6b4252102a2aad5544be7f8b08419e73fd2621d9/pkg/client/client.go#L971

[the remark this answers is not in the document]
marioferh: @simonpasquier solved for prometheus, alertmanager validation work sin a different way https://github.com/openshift/cluster-monitoring-operator/blob/6b4252102a2aad5544be7f8b08419e73fd2621d9/pkg/tasks/alertmanager.go#L204C4-L204C4

pkg/client/client.go
```diff
@@ -1006,6 +1017,12 @@ func (c *Client) WaitForAlertmanager(ctx context.Context, a *monv1.Alertmanager)
 		if ctx.Err() != nil && lastErr != nil {
 			err = lastErr
 		}
+		s, _ := c.kclient.AppsV1().StatefulSets(a.GetNamespace()).Get(ctx, "alertmanager-main", metav1.GetOptions{})
+		generationAttemps := s.Status.ObservedGeneration - lastGeneration
+		if generationAttemps > 5 && s.Status.ReadyReplicas != s.Status.Replicas {
```
danielmellado, line 1022: I would leave a comment here explaining the reasoning about the 5 attempts.

marioferh: done

pkg/client/client.go
```diff
@@ -968,6 +970,12 @@ func (c *Client) ValidatePrometheus(ctx context.Context, promNsName types.Namesp
 	}, WithPollInterval(10*time.Second))
 
 	if pollErr != nil {
+		s, _ := c.kclient.AppsV1().StatefulSets(promNsName.Namespace).Get(ctx, "prometheus-k8s", metav1.GetOptions{})
+		generationAttemps := s.Status.ObservedGeneration - lastGeneration
```
rexagod, line 974:
```suggestion
		generationAttempts := s.Status.ObservedGeneration - lastGeneration
```
(nit)

danielmellado: +1 on the nit, actually we should have one linter for it :D

pkg/client/client.go
```diff
@@ -968,6 +970,12 @@ func (c *Client) ValidatePrometheus(ctx context.Context, promNsName types.Namesp
...
+		if generationAttemps > 5 && s.Status.ReadyReplicas != s.Status.Replicas {
```
rexagod, line 975: (from Simon's comment above)
```suggestion
		if generationAttemps > 10 && s.Status.ReadyReplicas != s.Status.Replicas {
```

pkg/client/client.go
rexagod: I believe the `misspell` and the `misspell-fix` targets work for us here, do you want to fail the CI if it errors out?

marioferh:
superseded by https://issues.redhat.com/browse/MON-3677

PR #2228, opened 2024-01-09
MON-1047: fix: add terminationMessagePolicy: FallbackToLogsOnError to all conta…

jan--f:
…iners

### DIFF
--- a/assets/prometheus-k8s/prometheus.yaml
+++ b/assets/prometheus-k8s/prometheus.yaml
@@ -153,10 +153,12 @@ spec:
       requests:
         cpu: 1m
         memory: 25Mi
+    terminationMessagePolicy: FallbackToLogsOnError
     volumeMounts:
     - mountPath: /etc/tls/grpc
       name: secret-grpc-tls
   - name: prometheus
+    terminationMessagePolicy: FallbackToLogsOnError
   enableFeatures: []
   externalLabels: {}
   externalURL: https://prometheus-k8s.openshift-monitoring.svc:9091
--- a/jsonnet/utils/set-terminationMessagePolicy.libsonnet
+++ b/jsonnet/utils/set-terminationMessagePolicy.libsonnet
@@ -1,0 +1,28 @@
+{
+  setTerminationMessagePolicy(o): o {
+    local addTerminationMessagePolicy(o) = o {
+      [if std.setMember(o.kind, ['DaemonSet', 'Deployment', 'ReplicaSet']) then 'spec']+: {
+        template+: {
+          spec+: {
+            containers: [
+              c {
+                terminationMessagePolicy: 'FallbackToLogsOnError',
+              }
+              for c in super.containers
+            ],
+          },
+        },
+      },
+      [if std.setMember(o.kind, ['Prometheus', 'Alertmanager', 'ThanosRuler']) then 'spec']+: {
+        containers: [
+          c {
+            terminationMessagePolicy: 'FallbackToLogsOnError',
+          }
+          for c in super.containers
+        ],
+      },
+    },
+    [k]: addTerminationMessagePolicy(o[k])
+    for k in std.objectFields(o)
+  },
+}
--- a/jsonnet/utils/set-terminationMessagePolicy.libsonnet
+++ b/jsonnet/utils/set-terminationMessagePolicy.libsonnet
@@ -8,17 +8,29 @@
               c {
                 terminationMessagePolicy: 'FallbackToLogsOnError',
               }
-              for c in super.containers
+              for c in o.spec.template.spec.containers
+            ],
+            [if 'initContainers' in o.spec.template.spec then 'initContainers']: [
+              c {
+                terminationMessagePolicy: 'FallbackToLogsOnError',
+              }
+              for c in o.spec.template.spec.initContainers
             ],
           },
         },
       },
-      [if std.setMember(o.kind, ['Prometheus', 'Alertmanager', 'ThanosRuler']) then 'spec']+: {
+      [if std.setMember(o.kind, ['Alertmanager', 'Prometheus', 'ThanosRuler']) then 'spec']+: {
         containers: [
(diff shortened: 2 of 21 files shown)


assets/prometheus-k8s/prometheus.yaml
```diff
@@ -153,10 +153,12 @@ spec:
       requests:
         cpu: 1m
         memory: 25Mi
+    terminationMessagePolicy: FallbackToLogsOnError
     volumeMounts:
     - mountPath: /etc/tls/grpc
       name: secret-grpc-tls
   - name: prometheus
+    terminationMessagePolicy: FallbackToLogsOnError
```
simonpasquier, line 161: the operator should already set the field

machine424: + we already set it for prom containers https://github.com/openshift/cluster-monitoring-operator/blob/6b4252102a2aad5544be7f8b08419e73fd2621d9/assets/prometheus-k8s/prometheus.yaml#L104 maybe we should remove that as well (we already manually set it for other containers as well: https://github.com/search?q=repo%3Aopenshift%2Fcluster-monitoring-operator%20terminationMessagePolicy&type=code, maybe remove that and let the jsonnet loop inject it)

machine424: (to test that this always stand + to test other properties we want to ensure for all "resource" on the 2 namespaces, origin monitor tests can be helpful https://github.com/openshift/origin/blob/master/pkg/monitortests/README.md: create a watch on the resource and run check on it at the end, we could create a ticket to add a check for terminationMessagePolicy and maybe others in the future)

jan--f: Yeah I missed to remove the individual sets. The idea here was to simply set it for all containers in a jsonnet util function instead of remembering to set it for all containers that need it.
I have removed the components level settings locally but something is not working right yet, so
/hold

But happy to discuss is this makes sense at all. I was just going off the old Jira linked to this. Good chance, this isn't the way.

jsonnet/utils/set-terminationMessagePolicy.libsonnet
```diff
@@ -0,0 +1,42 @@
...
+            [if 'initContainers' in super then 'initContainers']: [
+              c {
+                terminationMessagePolicy: 'FallbackToLogsOnError',
+              }
+              for c in super.initContainers
+            ],
+          },
+        },
+      },
+      [if std.setMember(o.kind, ['Prometheus', 'Alertmanager', 'ThanosRuler']) then 'spec']+: {
```
machine424, line 23: I think the set should be ordered (or we can convert the array to a set):

> Sets

> Sets are represented as ordered arrays without duplicates.

> Note that the std.set* functions rely on the uniqueness and ordering on arrays passed to them to work. This can be guaranteed by using std.set(arr). If that is not the case, the functions will quietly return non-meaningful results. 

from https://jsonnet.org/ref/stdlib.html

jsonnet/utils/set-terminationMessagePolicy.libsonnet
```diff
@@ -0,0 +1,42 @@
...
+        template+: {
+          spec+: {
+            containers: [
+              c {
+                terminationMessagePolicy: 'FallbackToLogsOnError',
+              }
+              for c in super.containers
+            ],
+          } + {
+            [if 'initContainers' in super then 'initContainers']: [
```
machine424, line 14: maybe `super` is considered to be `template` in here? (had similar behaviors before)

jsonnet/utils/set-terminationMessagePolicy.libsonnet
```diff
@@ -0,0 +1,42 @@
+{
+  setTerminationMessagePolicy(o): o {
+    local addTerminationMessagePolicy(o) = o {
+      [if std.setMember(o.kind, ['DaemonSet', 'Deployment', 'ReplicaSet']) then 'spec']+: {
+        template+: {
```
simonpasquier, line 5: this + using `spec` instead of `super` seems to work.
```suggestion
        template+: {
                local spec = super.spec,
```

jsonnet/utils/set-terminationMessagePolicy.libsonnet
```diff
@@ -0,0 +1,42 @@
+{
+  setTerminationMessagePolicy(o): o {
+    local addTerminationMessagePolicy(o) = o {
+      [if std.setMember(o.kind, ['DaemonSet', 'Deployment', 'ReplicaSet']) then 'spec']+: {
```
simonpasquier, line 4: (nit) we shouldn't have any ReplicaSet object?

jsonnet/utils/set-terminationMessagePolicy.libsonnet
jan--f: Thanks, that does fix it.

jsonnet/utils/set-terminationMessagePolicy.libsonnet
jan--f: yes I'm aware, though I figured we might as well throw it in their on the off change that we add one later.
The same is true for `initContainers`, we only have one in the node_exporter daemonset.

simonpasquier: ReplicaSet is a low-level object like Pod which I don't expect CMO to ever manage. But let's clean it up afterwards since it's also present in other jsonnet utilities.

jsonnet/utils/set-terminationMessagePolicy.libsonnet
jan--f: This works, though the same doesn't seem to work for the `Alertmanager` ... kinds. However just referencing `o.spec` or `o.spec.template.spec` seems to work well.

machine424:
/lgtm
Thanks Jan.

PR #2229, opened 2024-01-09
OCPNODE-2022: jsonnet: update crio port to TLS port 9637

rphillips:

### DIFF
--- a/jsonnet/components/control-plane.libsonnet
+++ b/jsonnet/components/control-plane.libsonnet
@@ -87,13 +87,18 @@ function(params)
           [{
             interval: '30s',
             port: 'https-metrics',
+            scheme: 'https',
+            tlsConfig+: {
+              serverName: 'machine-config-daemon.openshift-machine-config-operator.svc',
+              caFile: '/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt',
+            },
             relabelings: [
               {
                 sourceLabels: ['__address__'],
                 action: 'replace',
                 targetLabel: '__address__',
                 regex: '(.+)(?::\\d+)',
-                replacement: '$1:9537',
+                replacement: '$1:9637',
               },
               {
                 sourceLabels: ['endpoint'],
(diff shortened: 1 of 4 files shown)


jsonnet/components/control-plane.libsonnet
```diff
@@ -93,7 +93,7 @@ function(params)
                 action: 'replace',
                 targetLabel: '__address__',
                 regex: '(.+)(?::\\d+)',
-                replacement: '$1:9537',
+                replacement: 'https://$1:9637',
```
simonpasquier, line 96: you want to keep this line as-is because `__address__` is only `<address>:<port>`

what you need is to add `scheme: https` to the endpoint definition:

```
          // Collect metrics from CRI-O.
          [{
            interval: '30s',
            port: 'https-metrics',
            scheme: 'https',

```

rphillips:
@simonpasquier progress... That helped but the certificate doesn't have IP subject Alternative Names. I could use the FQDN of the machine.

```
Scrape failed" err="Get \"https://192.168.126.11:9637/metrics\": tls: failed to verify certificate: x509: cannot validate certificate for 192.168.126.11 because it doesn't contain any IP SANs"
```

rphillips:
Added a openshift/origin PR to change crio metrics to https.

https://github.com/openshift/origin/pull/28536

@simonpasquier or @jan--f Could you review this PR and 28536?

simonpasquier (commented):
The CRI-O metrics endpoint doesn't support client TLS certificate for authentication? Are there any plan to make it work?

rphillips:
@simonpasquier No plans yet... MCO also has a metrics endpoint for Machine Config Daemon. Are client certs wanted/needed by monitoring now? If they are is there a doc on how to enable it?

simonpasquier:
> MCO also has a metrics endpoint for Machine Config Daemon. Are client certs wanted/needed by monitoring now? If they are is there a doc on how to enable it?

yes it's highly recommended to use mTLS because it's more resilient to Kube API outages + more resource friendly:
* https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/client-cert-scraping.md
* https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#metrics
* https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/

simonpasquier:
Looking at https://github.com/openshift/machine-config-operator/blob/bee2519c0bc2ed42e76bbb7780de9d0d58e35b35/manifests/machineconfigdaemon/kube-rbac-proxy-config.yaml#L11-L17

The configuration should rather be something like:

```
    "authorization":
      "static":
      - "path": "/metrics"
        "resourceRequest": false
        "user":
        "verb": "get"    
```

PR #2231, opened 2024-01-12
MON-3664: chore: avoid issues with std.set* functions

simonpasquier:
Functions like std.setMember() expect the input array to be sorted without duplicates.


simonpasquier:
/ cc @machine424 
/cc @jan--f

jan--f:
:+1: 
/lgtm

machine424:
/lgtm
It even caught a bug, thanks!

simonpasquier:
/retitle MON-3664: chore: avoid issues with std.set* functions

PR #2233, opened 2024-01-15
OCPBUGS-26983: rollout monitoring plugin on TLS rotation

simonpasquier:
This commit ensures that the monitoring plugin's pods get redeployed when the TLS serving certificate is rotated.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -45,6 +45,7 @@ import (
 	policyv1 "k8s.io/api/policy/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	"k8s.io/utils/ptr"
@@ -2881,7 +2882,7 @@ func (f *Factory) MonitoringPlugin() (*consolev1.ConsolePlugin, error) {
 	return f.NewConsolePlugin(f.assets.MustNewAssetSlice(MonitoringPlugin))
 }
 
-func (f *Factory) MonitoringPluginDeployment() (*appsv1.Deployment, error) {
+func (f *Factory) MonitoringPluginDeployment(tlsSecret *v1.Secret) (*appsv1.Deployment, error) {
 	d, err := f.NewDeployment(f.assets.MustNewAssetSlice(MonitoringPluginDeployment))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/manifests/manifests.go
```diff
@@ -2900,6 +2901,15 @@ func (f *Factory) MonitoringPluginDeployment() (*appsv1.Deployment, error) {
 
 	containers[idx].Image = f.config.Images.MonitoringPlugin
 
+	// Hash the TLS secret and propagate it as an annotation to the
+	// deployment's pods to trigger a new rollout when the TLS certificate/key
+	// are rotated.
+	h := fnv.New64()
+	for _, k := range sets.StringKeySet[[]byte](tlsSecret.Data).List() {
+		h.Write(tlsSecret.Data[k])
```
machine424, line 2909: How about just?:

```suggestion
	for _, v := range tlsSecret.Data {
		h.Write(v)
	}
```

machine424: Some extreme scenarios I was thinking about, we could just ignore them:

I don't know if we should only consider the TLS keys, as we don't want to restart the plugin if anything else gets added to the secret by mistake (ca operator will correct that maybe but what if the change is too frequent/automated?)

Also, if the TLS content get missing or sth (content gets wiped by ca operator bcz of a bug or sth), should we tell the deployment about it or leave it run with the old certificate?

danielmellado: I was also thinking the same about the List but didn't think about the hashing

simonpasquier: selecting only tls.crt and tls.key is more error-prone IMHO but I'll add a comment to clarify.

machine424: if the ca operator changes that it'll be a breaking change, plus we already count on their stability https://github.com/search?q=repo%3Aopenshift%2Fcluster-monitoring-operator+tls.crt+language%3AGo&type=code&l=Go

But ok, let's keep it simple and see, thanks.

pkg/tasks/monitoring_plugin.go
```diff
@@ -102,10 +103,19 @@ func (t *MonitoringPluginTask) Run(ctx context.Context) error {
...
 		}
-	}
 
-	{ // deployment
-		d, err := t.factory.MonitoringPluginDeployment()
+		secret, err := t.client.WaitForSecretByNsName(
+			ctx,
+			types.NamespacedName{
+				Namespace: svc.Namespace,
+				Name:      svc.Annotations["service.beta.openshift.io/serving-cert-secret-name"],
```
machine424, line 111: @simonpasquier shouldn't we make sure CMO is listening for changes on that secret https://github.com/openshift/cluster-monitoring-operator/blob/a83f250fbf7efd2863a6faab471242322e563796/pkg/operator/operator.go#L665? otherwise the code that will change the sha may never run (if no other change triggers a sync).

simonpasquier: I missed the comment but yes you're probably right.

PR #2234, opened 2024-01-15
OCPBUGS-27213: Adjust lastError injection to wait.PollUntilContextTimeout's final error

machine424:
Stop checking for the parent context to be done as wait.PollUntilContextTimeout creates a child context with timeout (A child can't and shouldn't cancel the parent context).

Factorize that lastError injection logic in an util.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -68,10 +68,9 @@ import (
 )
 
 const (
-	deploymentCreateTimeout = 5 * time.Minute
-	deploymentDeleteTimeout = 5 * time.Minute
-	metadataPrefix          = "monitoring.openshift.io/"
-	clusterConsole          = "cluster"
+	deleteTimeout  = 10 * time.Minute
+	metadataPrefix = "monitoring.openshift.io/"
+	clusterConsole = "cluster"
 )
 
 type Client struct {
@@ -435,7 +434,7 @@ func (c *Client) EnsurePrometheusUserWorkloadConfigMapExists(ctx context.Context
 }
 
 func (c *Client) AssurePrometheusOperatorCRsExist(ctx context.Context) error {
-	return wait.PollUntilContextTimeout(ctx, time.Second, time.Minute*5, false, func(ctx context.Context) (bool, error) {
+	return Poll(ctx, func(ctx context.Context) (bool, error) {
 		_, err := c.mclient.MonitoringV1().Prometheuses(c.namespace).List(ctx, metav1.ListOptions{})
 		if err != nil {
 			klog.V(4).ErrorS(err, "AssurePrometheusOperatorCRsExist: failed to list Prometheuses")
@@ -699,7 +698,7 @@ func (c *Client) CreateOrUpdateAlertmanager(ctx context.Context, a *monv1.Alertm
 }
 
 func (c *Client) DeleteAlertmanager(ctx context.Context, a *monv1.Alertmanager) error {
-	return c.deleteResourceUntilGone(ctx, monv1.SchemeGroupVersion.WithResource("alertmanagers"), a, 10*time.Minute)
+	return c.deleteResourceUntilGone(ctx, monv1.SchemeGroupVersion.WithResource("alertmanagers"), a, deleteTimeout)
 }
 
 func (c *Client) CreateOrUpdateThanosRuler(ctx context.Context, t *monv1.ThanosRuler) error {
@@ -788,7 +787,7 @@ func (c *Client) DeleteValidatingWebhook(ctx context.Context, w *admissionv1.Val
 }
 
 func (c *Client) DeleteDeployment(ctx context.Context, d *appsv1.Deployment) error {
-	return c.deleteResourceUntilGone(ctx, appsv1.SchemeGroupVersion.WithResource("deployments"), d, deploymentDeleteTimeout)
+	return c.deleteResourceUntilGone(ctx, appsv1.SchemeGroupVersion.WithResource("deployments"), d, 5*time.Minute)
 }
 
 func (c *Client) DeletePodDisruptionBudget(ctx context.Context, pdb *policyv1.PodDisruptionBudget) error {
@@ -801,11 +800,11 @@ func (c *Client) DeletePodDisruptionBudget(ctx context.Context, pdb *policyv1.Po
 }
 
 func (c *Client) DeletePrometheus(ctx context.Context, p *monv1.Prometheus) error {
-	return c.deleteResourceUntilGone(ctx, monv1.SchemeGroupVersion.WithResource("prometheuses"), p, 10*time.Minute)
+	return c.deleteResourceUntilGone(ctx, monv1.SchemeGroupVersion.WithResource("prometheuses"), p, deleteTimeout)
 }
 
 func (c *Client) DeleteThanosRuler(ctx context.Context, tr *monv1.ThanosRuler) error {
-	return c.deleteResourceUntilGone(ctx, monv1.SchemeGroupVersion.WithResource("thanosrulers"), tr, 10*time.Minute)
+	return c.deleteResourceUntilGone(ctx, monv1.SchemeGroupVersion.WithResource("thanosrulers"), tr, deleteTimeout)
 }
 
 func (c *Client) DeleteDaemonSet(ctx context.Context, d *appsv1.DaemonSet) error {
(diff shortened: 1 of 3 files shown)


machine424:
/retest
/retitle  MON-3668: Adjust lastError injection to wait.PollUntilContextTimeout's final error

pkg/client/client.go
```diff
@@ -2055,3 +2023,24 @@ func deleteOptions(dp metav1.DeletionPropagation) metav1.DeleteOptions {
...
+		if lastError != nil && *lastError != nil {
+			err = fmt.Errorf("%w: lastError: %w", err, *lastError)
+		}
+		return err
+	}
+	return nil
+}
+
+func Poll(ctx context.Context, lastError *error, condition wait.ConditionWithContextFunc) error {
+	return pollUntil(ctx, 5*time.Minute, 3*time.Second, lastError, condition)
```
machine424, line 2040: This sets the same interval/timeout to all polls (which is not the case for interval currently, but it's better to have the same sane values set for all of them)

pkg/client/client_test.go
```diff
@@ -2304,3 +2306,59 @@ func assertErrorsMatch(t *testing.T, want, got []error) {
...
+	require.NoError(t, err)
+
+	// condition returns an error before timeout.
+	var lastErr3 error
+	err = pollUntil(context.Background(), timeout, interval, &lastErr3, func(ctx context.Context) (bool, error) {
+		insideError := fmt.Errorf("INSIDE ERROR")
+		lastErr3 = insideError
+		return false, insideError
+	})
+	require.ErrorContains(t, err, "INSIDE ERROR: lastError: INSIDE ERROR")
```
machine424, line 2342: Ok. otherwise code would be complicated and hacky.

pkg/client/client_test.go
```diff
@@ -2304,3 +2307,59 @@ func assertErrorsMatch(t *testing.T, want, got []error) {
...
 	}
 }
+
+func TestPollUntil(t *testing.T) {
+	timeout := 150 * time.Millisecond
+	interval := 100 * time.Millisecond
+
+	// condition is met.
+	var lastErr1 error
+	err := pollUntil(context.Background(), timeout, interval, &lastErr1, func(ctx context.Context) (bool, error) {
```
marioferh, line 2317: It is better that the tests verify Poll instead of helper?

pkg/client/client.go
```diff
@@ -2055,3 +2023,24 @@ func deleteOptions(dp metav1.DeletionPropagation) metav1.DeleteOptions {
...
+}
+
+// PollUntilDeleted has a larger timeout as it's acceptable for deletion to take longer.
+func PollUntilDeleted(ctx context.Context, lastError *error, condition wait.ConditionWithContextFunc) error {
```
raptorsun, line 2044: I think it better to use the PollUntil with the duration constants for creation, deletion, modification, etc.
The function `Poll` and `PollUntilDeleted` do not have additional instructions than `pollUntil`, while their names suggest they have dedicated actions for deletion or normal operation. It is confusing.

machine424: We can come up with better names or add more comments.
Passing the same args over and over could be cumbersome and less readable and more prone to inconsistencies.
Poll and PollUntilDeleted  are easier to use and give us more control over the interval/timeout.

pkg/client/client.go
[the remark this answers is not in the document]
machine424: Actually it's intentional to make the utils easier to use and to have control over the args.
I'll add some more comments

pkg/client/client_test.go
machine424: it'll not be possible to use it for all test cases as the timeout is large (5mins) which will reflect on the tests duration.

Using it only for some cases would be inconsistent  I think.

pkg/client/client_test.go
```diff
@@ -2304,3 +2307,59 @@ func assertErrorsMatch(t *testing.T, want, got []error) {
...
+	// condition is met, lastErr isn't used.
+	err = pollUntil(context.Background(), timeout, interval, nil, func(ctx context.Context) (bool, error) {
+		return true, nil
+	})
+	require.NoError(t, err)
+
+	// condition is eventually met.
+	var lastErr2 error
+	err = pollUntil(context.Background(), timeout, interval, &lastErr2, func(ctx context.Context) (bool, error) {
+		lastErr1 = fmt.Errorf("INSIDE ERROR")
```
machine424, line 2331: should be lastErr2

pkg/client/client.go
simonpasquier: I'd prefer to keep the same values as before to avoid unintended changes (especially regarding timeouts).

One solution could be to extend the signature with functional options:

```go
type pollOptions struct {
  timeout time.Duration
  interval time.Duration
  lastError *error
}

func withPollTimeout(d time.Duration) func(o *pollOptions) {
    return func(o *pollOptions) {
        o.timeout = d
    }
}

func withPollInterval(d time.Duration) func(o *pollOptions) {
    return func(o *pollOptions) {
        o.interval = d
    }
}

func withLastError(e *error) func(o *pollOptions) {
    return func(o *pollOptions) {
        o.lastError = e
    }
}

func Poll(ctx context.Context, condition wait.ConditionWithContextFunc, options ...pollOptions) error {
    opts := pollOptions{timeout: 5 * time.Minute, interval: time.Second}
    for _, o := range options {
       o(&opts)
    }
...
}

```

pkg/client/client_test.go
simonpasquier: See my comment above about having functional opts for Poll

pkg/client/client.go
machine424: Actually, the timeouts weren't changed (5mins for "deploy" requests and 10 for delete)
I just aligned the intervals to 3s (was aiming for 5s as well) for deploy and 5s for delete to simplify things without affecting the api server.
But if you think we should keep intervals as is, we can allow customizing.

simonpasquier: yes I'd prefer to keep things as-is for now. We can still look at refactoring afterwards.

machine424: okk, if you could take a look now.

pkg/client/client.go
```diff
@@ -2032,3 +2001,44 @@ func deleteOptions(dp metav1.DeletionPropagation) metav1.DeleteOptions {
...
+// the final error if if set by the condition, this would add more context to the "context deadline exceeded" error e.g..
+func Poll(ctx context.Context, condition wait.ConditionWithContextFunc, options ...func(o *pollOptions)) error {
+	opts := pollOptions{timeout: 5 * time.Minute, interval: time.Second}
+	for _, o := range options {
+		o(&opts)
+	}
+
+	if err := wait.PollUntilContextTimeout(ctx, opts.interval, opts.timeout, false, condition); err != nil {
+		if opts.lastError != nil && *opts.lastError != nil {
+			err = fmt.Errorf("%w: lastError: %w", err, *opts.lastError)
```
simonpasquier, line 2039: (nit) I'm wondering if we need the intermediate `lastError: ` string?
```suggestion
			err = fmt.Errorf("%w: %w", err, *opts.lastError)
```

pkg/client/client_test.go
```diff
@@ -2304,3 +2308,60 @@ func assertErrorsMatch(t *testing.T, want, got []error) {
...
+	require.NoError(t, err)
+
+	// condition returns an error before timeout.
+	var lastErr3 error
+	err = testPoll(context.Background(), func(ctx context.Context) (bool, error) {
+		insideError := fmt.Errorf("INSIDE ERROR")
+		lastErr3 = insideError
+		return false, insideError
+	}, &lastErr3)
+	require.ErrorContains(t, err, "INSIDE ERROR: lastError: INSIDE ERROR")
```
simonpasquier, line 2345: (nit) it might more readable/plausible to have 2 different errors (e.g. `INSIDE ERROR` and `OUTSIDE ERROR`)

machine424: Do you suggest we call this error "OUTSIDE ERROR"?

(Note that in these "rare" situations, the error returned by `PollUntilContextTimeout` and `lastError` will be the same, and I don't think the logic (I haven't thought of any) to avoid this duplication is worth it)

machine424: I think a simple comparison should do it, pushed a new version.

simonpasquier (commented):
/lgtm

machine424:
/retitle OCPBUGS-27213: Adjust lastError injection to wait.PollUntilContextTimeout's final error

PR #2235, opened 2024-01-16
STOR-1289: Move vSphere prometheus rules to cluster-storage-operator

jsafrane:
Move rules that are used to sanitize OCP metric for telemetry to cluster-storage-operator. These rules are used only for storage metrics and should live with storage code.

cluster-storage-operator PR that adds the rules: https://github.com/openshift/cluster-storage-operator/pull/442


jsafrane:
/hold
to coordinate merge with https://github.com/openshift/cluster-storage-operator/pull/442

jsafrane:
/px-approved
/docs-approved
This change is not facing users.

jsafrane:
I'm also removing `cluster:vsphere_csi_migration:max` from telemetry configuration, that metric was available only in 4.13.

PR #2237, opened 2024-01-17
MON-2853: add runbook link to TargetDown alert

machine424:


jan--f:
Thanks!
/lgtm

PR #2242, opened 2024-01-23
OCPBUGS-27289: rollout metrics-server on cert rotations

slashpai:
This commit ensures that the metrics-server's pods are redeployed when the Kubelet CA cert is rotated.

Similar fix like https://github.com/openshift/cluster-monitoring-operator/pull/2233 but for config map change

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -2122,7 +2122,7 @@ func (f *Factory) MetricsServerRoleBindingAuthReader() (*rbacv1.RoleBinding, err
 	return f.NewRoleBinding(f.assets.MustNewAssetSlice(MetricsServerRoleBindingAuthReader))
 }
 
-func (f *Factory) MetricsServerDeployment() (*appsv1.Deployment, error) {
+func (f *Factory) MetricsServerDeployment(kubeletCABundle *v1.ConfigMap, tlsSecret, metricsClientCert *v1.Secret) (*appsv1.Deployment, error) {
 	dep, err := f.NewDeployment(f.assets.MustNewAssetSlice(MetricsServerDeployment))
 	if err != nil {
 		return nil, err
@@ -2139,6 +2139,20 @@ func (f *Factory) MetricsServerDeployment() (*appsv1.Deployment, error) {
 	containers[idx].Image = f.config.Images.MetricsServer
 	containers[idx].Args = f.setTLSSecurityConfiguration(podSpec.Containers[0].Args,
 		MetricsServerTLSCipherSuitesFlag, MetricsServerTLSMinTLSVersionFlag)
+
+	// Hash the Kubelet Serving CA Bundle configmap value and propagate it as a annotation to the
+	// deployment's pods to trigger a new rollout when the CA is rotated.
+	dep.Spec.Template.Annotations["monitoring.openshift.io/kubelet-serving-ca-bundle-hash"] = hashStringMap(kubeletCABundle.Data)
+
+	// Hash the TLS secret and propagate it as a annotation to the
+	// deployment's pods to trigger a new rollout when the TLS certificate/key
+	// are rotated.
+	dep.Spec.Template.Annotations["monitoring.openshift.io/serving-ca-secret-hash"] = hashByteMap(tlsSecret.Data)
+
+	// Hash the metrics client cert and propagate it as a annotation to the
+	// deployment's pods to trigger a new rollout when the metrics client cert
+	// is rotated.
+	dep.Spec.Template.Annotations["monitoring.openshift.io/metrics-client-cert-hash"] = hashByteMap(metricsClientCert.Data)
 
 	config := f.config.ClusterMonitoringConfiguration.MetricsServerConfig
 	if config == nil {
@@ -2912,7 +2926,7 @@ func (f *Factory) MonitoringPluginDeployment(tlsSecret *v1.Secret) (*appsv1.Depl
 	// Hash the TLS secret and propagate it as an annotation to the
 	// deployment's pods to trigger a new rollout when the TLS certificate/key
 	// are rotated.
-	d.Spec.Template.Annotations["monitoring.openshift.io/hash"] = hashSecretData(tlsSecret)
+	d.Spec.Template.Annotations["monitoring.openshift.io/hash"] = hashByteMap(tlsSecret.Data)
 
 	cfg := f.config.ClusterMonitoringConfiguration.MonitoringPluginConfig
 	if cfg == nil {
@@ -3709,14 +3723,21 @@ func containerNameEquals(name string) func(corev1.Container) bool {
 	}
 }
 
-// hashSecretData hashes the secret's data and returns the hash value as a string.
-func hashSecretData(s *v1.Secret) string {
+func hashByteMap(s map[string][]byte) string {
 	h := fnv.New64()
 	// The data's keys need to be sorted in a predictable order to always
 	// produce the same hash.
-	for _, k := range sets.StringKeySet[[]byte](s.Data).List() {
-		h.Write(s.Data[k])
+	for _, k := range sets.StringKeySet[[]byte](s).List() {
+		h.Write(s[k])
 	}
 
 	return strconv.FormatUint(h.Sum64(), 32)
+}
+
+func hashStringMap(m map[string]string) string {
+	byteMap := make(map[string][]byte, len(m))
+	for k, v := range m {
+		byteMap[k] = []byte(v)
+	}
+	return hashByteMap(byteMap)
 }
--- a/pkg/tasks/metricsserver.go
+++ b/pkg/tasks/metricsserver.go
@@ -6,6 +6,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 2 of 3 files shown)


pkg/tasks/metricsserver.go
```diff
@@ -104,9 +105,18 @@ func (t *MetricsServerTask) create(ctx context.Context) error {
...
 		}
-	}
-	{
-		dep, err := t.factory.MetricsServerDeployment()
+
+		kubeletCABundleConfigMap, err := t.client.WaitForConfigMapByNsName(
+			ctx,
+			types.NamespacedName{
+				Namespace: s.Namespace,
+				Name:      "kubelet-serving-ca-bundle",
```
machine424, line 113: maybe use `PrometheusK8sKubeletServingCABundle` + `WaitForConfigMap` to avoid hardcoding.

pkg/manifests/manifests.go
```diff
@@ -3720,3 +3724,16 @@ func hashSecretData(s *v1.Secret) string {
 
 	return strconv.FormatUint(h.Sum64(), 32)
 }
+
+// hashConfigMapData hashes the configmap's data and returns the hash value as a string.
```
machine424, line 3728: maybe make the above `hashSecretData` take a `map[string][]byte` (and change its name) to make it work for secrets and configmaps

pkg/manifests/manifests.go
```diff
@@ -2140,6 +2140,10 @@ func (f *Factory) MetricsServerDeployment() (*appsv1.Deployment, error) {
 	containers[idx].Args = f.setTLSSecurityConfiguration(podSpec.Containers[0].Args,
 		MetricsServerTLSCipherSuitesFlag, MetricsServerTLSMinTLSVersionFlag)
 
+	// Hash the Kubelet Serving CA Bundle configmap value and propagate it as an annotation to the
+	// deployment's pods to trigger a new rollout when the CA is rotated.
+	dep.Spec.Template.Annotations["monitoring.openshift.io/hash"] = hashConfigMapData(kubeletCABundle)
```
machine424, line 2145: I see we also use `monitoring.openshift.io/hash` as a label sometimes https://github.com/search?q=repo%3Aopenshift%2Fcluster-monitoring-operator%20monitoring.openshift.io%2Fhash&type=code, maybe we should unify this?
cc @simonpasquier

machine424 (commented):
Thanks JP
Did you check if we should do the same for `metrics-client-certs` and `metrics-server-tls`?

pkg/manifests/manifests.go
slashpai: This annotation has a special meaning IIUC based on which deployment will be rollout https://github.com/openshift/library-go/blob/964bcb3f545c24f15294fd8ba914529bf4fe8c4d/pkg/certs/cert-inspection/certgraphanalysis/collection_options.go#L86

simonpasquier: Good catch! I'm guilty introducing a different approach using an annotation instead of a label in #2233. Since this change isn't published yet, I'm fine switching to a label instead.

pkg/manifests/manifests.go
simonpasquier: sounds good to me. Something like?

````
func hashByteMap(map[string][]byte) string {
// same as hashSecretData()
}

func hashStringMap(m map[string]string) string {
  byteMap := make(map[string][]byte, len(m)
  for k, v := range m {
    byteMap[k] = []byte(v)
  }
  return hashByteMap(byteMap)
}

**Review by simonpasquier — COMMENTED** [review]
In addition to the kubelet CA, do we need to account also for the client cert rotation?

**Comment by slashpai** [comment]
> In addition to the kubelet CA, do we need to account also for the client cert rotation?

I think we would need to. I just added the labels for serving ca and metrics-client cert as well

File: pkg/manifests/manifests.go [hunk 4]
```diff
@@ -2140,6 +2140,20 @@ func (f *Factory) MetricsServerDeployment() (*appsv1.Deployment, error) {
 	containers[idx].Args = f.setTLSSecurityConfiguration(podSpec.Containers[0].Args,
 		MetricsServerTLSCipherSuitesFlag, MetricsServerTLSMinTLSVersionFlag)
 
+	// Hash the Kubelet Serving CA Bundle configmap value and propagate it as a label to the
```
machine424, line 2143: I'm afraid the values will be overridden and only `hashByteMap(metricsClientCert.Data)` will be applied.
we should either:
- Group the 3 hashes and put them under the same label, I don't think `monitoring.openshift.io/hash` is the suitable name as it may mean the hash of the resource itself (Pod in this case) which may be confusing. Maybe we could use something like `monitoring.openshift.io/datastore-hash` or something.
- have 3 different labels `monitoring.openshift.io/CONFIGMAP-OR-SECRET-NAME-hash` for each configmap/secret.

WDYT?

If we move away from `monitoring.openshift.io/hash`, we could use annotations, in the Helm world, we use sth like that, they're called checksums https://github.com/search?q=repo%3Abitnami%2Fcharts%20checksum&type=code maybe we could use that term instead of hash to align with Helm.

pkg/manifests/manifests.go
machine424: Maybe we could still use an annotation after all :) https://github.com/openshift/cluster-monitoring-operator/pull/2242#discussion_r1464544438 wdyt?

pkg/tasks/metricsserver.go
```diff
@@ -104,9 +105,34 @@ func (t *MetricsServerTask) create(ctx context.Context) error {
...
+
+		kscm, err := t.client.GetConfigmap(ctx, "openshift-config-managed", "kubelet-serving-ca")
+		if err != nil {
+			return fmt.Errorf("openshift-config-managed/kubelet-serving-ca: %w", err)
+		}
+
+		cacm, err := t.factory.PrometheusK8sKubeletServingCABundle(kscm.Data)
+		if err != nil {
+			return fmt.Errorf("initializing kubelet serving CA Bundle ConfigMap failed: %w", err)
+		}
```
machine424, line 117: We could just use `PrometheusK8sKubeletServingCABundle` with empty data as we're only interested in namespace/name to avoid hardcoding them and to signal that we rely on that resource as well.

```suggestion

               // Only interested in its name and to declare dependency on the asset.
		cacm, err := t.factory.PrometheusK8sKubeletServingCABundle([]string{})
		if err != nil {
			return fmt.Errorf("initializing an empty kubelet serving CA Bundle ConfigMap failed: %w", err)
		}
		then WaitForConfigMap(cacm)
```

pkg/tasks/metricsserver.go
```diff
@@ -104,9 +105,34 @@ func (t *MetricsServerTask) create(ctx context.Context) error {
...
+			types.NamespacedName{
+				Namespace: s.Namespace,
+				Name:      s.Annotations["service.beta.openshift.io/serving-cert-secret-name"],
+			},
+		)
+		if err != nil {
+			return err
+		}
+
+		metricsClientSecret, err := t.factory.MetricsClientCerts()
```
machine424, line 130: maybe we should wait for this as well.

pkg/manifests/manifests.go
slashpai: Thats a good catch, I think we should have separate label/annotation in that case. 
For telemeter also we use annotation https://github.com/openshift/cluster-monitoring-operator/blob/70998442501d8a2d0d35dbb493a623161b3c56a3/pkg/manifests/manifests.go#L3121

slashpai:
@machine424 @simonpasquier addressed comments, could you please review again?

slashpai:
Adding placeholder for backporting once PR is merged

/cherry-pick release-4.15

pkg/manifests/manifests.go
```diff
@@ -2912,7 +2926,7 @@ func (f *Factory) MonitoringPluginDeployment(tlsSecret *v1.Secret) (*appsv1.Depl
 	// Hash the TLS secret and propagate it as an annotation to the
 	// deployment's pods to trigger a new rollout when the TLS certificate/key
 	// are rotated.
-	d.Spec.Template.Annotations["monitoring.openshift.io/hash"] = hashSecretData(tlsSecret)
+	d.Spec.Template.Annotations["monitoring.openshift.io/hash"] = hashByteMap(tlsSecret.Data)
```
machine424, line 2929:
```suggestion
	d.Spec.Template.Annotations["monitoring.openshift.io/monitoring-plugin-cert-hash"] = hashByteMap(tlsSecret.Data)
```

nit: if you want to include this change in here (otherwise I can open a PR for it)

normally CMO should handle this change well: https://github.com/openshift/cluster-monitoring-operator/blob/60312aa5cb3d27d9c4134c55a1f4d0116bf8e71d/pkg/client/client.go#L1972-L1991

pkg/tasks/metricsserver.go
```diff
@@ -104,9 +105,51 @@ func (t *MetricsServerTask) create(ctx context.Context) error {
...
+		if err != nil {
+			return fmt.Errorf("initializing metrics-client-cert failed: %w", err)
+		}
+
+		mcs, err := t.client.WaitForSecretByNsName(
```
machine424, line 141: nit: we could use `WaitForSecret(metricsClientSecret)` directly I think

pkg/manifests/manifests.go
machine424: I'll take care of this

PR #2244, opened 2024-01-25
OCPBUGS-27289: followup of https://github.com/openshift/cluster-monitoring-operator/pull/2242

machine424:
chore: adjust monitoring plugin cert hash annotation name to avoid confusion.

chore: adjust code for metrics server certs rotation and add test

followup of https://github.com/openshift/cluster-monitoring-operator/pull/2242

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -2926,7 +2926,7 @@ func (f *Factory) MonitoringPluginDeployment(tlsSecret *v1.Secret) (*appsv1.Depl
 	// Hash the TLS secret and propagate it as an annotation to the
 	// deployment's pods to trigger a new rollout when the TLS certificate/key
 	// are rotated.
-	d.Spec.Template.Annotations["monitoring.openshift.io/hash"] = hashByteMap(tlsSecret.Data)
+	d.Spec.Template.Annotations["monitoring.openshift.io/cert-hash"] = hashByteMap(tlsSecret.Data)
 
 	cfg := f.config.ClusterMonitoringConfiguration.MonitoringPluginConfig
 	if cfg == nil {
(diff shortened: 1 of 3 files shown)


pkg/manifests/manifests.go
```diff
@@ -2926,7 +2926,7 @@ func (f *Factory) MonitoringPluginDeployment(tlsSecret *v1.Secret) (*appsv1.Depl
 	// Hash the TLS secret and propagate it as an annotation to the
 	// deployment's pods to trigger a new rollout when the TLS certificate/key
 	// are rotated.
-	d.Spec.Template.Annotations["monitoring.openshift.io/hash"] = hashByteMap(tlsSecret.Data)
+	d.Spec.Template.Annotations["monitoring.openshift.io/monitoring-plugin-cert-hash"] = hashByteMap(tlsSecret.Data)
```
machine424, line 2929: CMO should handle this change well: https://github.com/openshift/cluster-monitoring-operator/blob/60312aa5cb3d27d9c4134c55a1f4d0116bf8e71d/pkg/client/client.go#L1972-L1991

simonpasquier: nit-picking but `monitoring-plugin-cert-hash` is a bit verbose since the annotation is set on the monitoring-plugin deployment. `cert-hash` would be more adequate.

machine424: was trying to stick to `RESOURCE-NAME-hash`, but sure, let's save some bytes.

pkg/manifests/manifests_test.go
```diff
@@ -2759,8 +2759,8 @@ metricsServer:
 			Namespace: "openshift-monitoring",
 		},
 		Data: map[string][]byte{
-			"tls.crt": []byte("foo"),
-			"tls.key": []byte("bar"),
+			"tls.crt": []byte("bar"),
```
slashpai, line 2762: Just wondering was this swap intentional? 🙂

simonpasquier (commented):
/lgtm

PR #2247, opened 2024-01-29
MON-3699: chore: merge OmitFromDoc with HideFromDoc introduced in https://github.com/https://github.com/openshift/cluster-monitoring-operator/pull/2210 to hide fields from the doc

machine424:
…b.com/openshift/cluster-monitoring-operator/pull/2210 to hide fields from the doc

### DIFF
--- a/hack/docgen/format/asciidocs/api.go
+++ b/hack/docgen/format/asciidocs/api.go
@@ -133,9 +133,6 @@ func PrintAPIDocs(args []string) {
 			moduleContent += "|===\n"
 			moduleContent += "| Property | Type | Description \n"
 			for _, f := range t.Fields {
-				if strings.HasPrefix(fmt.Sprint(f.Description()), "OmitFromDoc") {
-					continue
-				}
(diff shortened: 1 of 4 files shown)


machine424:
Where could we doc this to avoid a recurrence?

simonpasquier:
/lgtm

> Where could we doc this to avoid a recurrence?

It's already in the documentation (https://github.com/openshift/cluster-monitoring-operator/tree/master/hack/docgen#ignoring-fields-from-the-doc) but I suppose that we don't have enough exposure with the tooling to remember it.

machine424:
> It's already in the documentation (https://github.com/openshift/cluster-monitoring-operator/tree/master/hack/docgen#ignoring-fields-from-the-doc) but I suppose that we don't have enough exposure with the tooling to remember it.

Indeed, I'll need to open my eyes next time :)


/retitle MON-3699: chore: merge OmitFromDoc with HideFromDoc introduced in https://github.com/https://github.com/openshift/cluster-monitoring-operator/pull/2210 to hide fields from the doc

PR #2250, opened 2024-01-31
OCPBUGS-28246: fix: set externalURL in UWM Prometheus

jan--f:
This sets the correct source URL in UWM alert notifications.


simonpasquier (commented):
/lgtm

just to be clear, it only affects PrometheusRules objects with the `openshift.io/prometheus-rule-evaluation-scope: leaf-prometheus` label?

PR #2251, opened 2024-01-31
MON-3697: use `maximumStartupDurationSeconds` instead of container patch

simonpasquier:
The latest version of the Prometheus CRD allows to customize the overall startup probe duration. This commit makes use of it instead of patching the Prometheus container definition at runtime.

### DIFF
--- a/jsonnet/components/prometheus.libsonnet
+++ b/jsonnet/components/prometheus.libsonnet
@@ -382,6 +382,10 @@ function(params)
             value: '15ms',
           },
         ],
+        // Increase the startup probe timeout to 1h from 15m to avoid restart
+        // failures when the WAL replay takes a long time.
+        // See https://issues.redhat.com/browse/OCPBUGS-4168 for details.
+        maximumStartupDurationSeconds: 3600,
         containers: [
           {
             name: 'prometheus-proxy',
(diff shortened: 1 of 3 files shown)


jsonnet/components/prometheus.libsonnet
```diff
@@ -382,6 +382,10 @@ function(params)
             value: '15ms',
           },
         ],
+        // Increase the startup probe timeout to 1h from 15m to avoid restart
+        // failures when the WAL replay takes a long time.
+        // See https://issues.redhat.com/browse/OCPBUGS-4168 for details.
+        maximumStartupDurationSeconds: 3600,
```
machine424, line 388: According to https://github.com/prometheus-operator/prometheus-operator/blob/e0b26a167be886fa2a8cce24109bb5599dce6728/pkg/prometheus/statefulset.go#L504

We'll end up with a `PeriodSecond` of 60s, this maybe make prometheus take 45s longer to be ready. Is it ok?

simonpasquier: That's correct. I feel like it's an acceptable trade-off. The (indirect) upside is that it will leave more time for the newly started Prometheus pod to scrape targets so it minimizes the risk of metric gaps.

PR #2252, opened 2024-01-31
fix: update generate_versions.go

slashpai:
Follow-up of https://github.com/rhobs/syncbot/pull/72

* [ x] No user facing changes, so no entry in CHANGELOG was needed.

### DIFF
--- a/hack/go/generate_versions.go
+++ b/hack/go/generate_versions.go
@@ -17,11 +17,11 @@ import (
 )
 
 const (
-	mainBranch        = "master"
-	metricsServerRepo = "openshift/kubernetes-metrics-server"
-	versionFile       = "../../jsonnet/versions.yaml"
+	mainBranch          = "master"
+	metricsServerRepo   = "openshift/kubernetes-metrics-server"
+	versionFile         = "../../jsonnet/versions.yaml"
 	versionNotFound     = "N/A"
-	OCPVersionHeader   = " OCP Version"
+	OCPVersionHeader    = " OCP Version"
 	depsVersionsFile    = "../../Documentation/deps-versions.md"
 	versionFileComments = `---
 # This file is meant to be managed by hack/go/generate_versions.go script
@@ -146,7 +146,7 @@ func updateDepsVersionsFile(fileP string, components Components) error {
 		}
 		rows = append(rows, row)
 		err = releaseVersion.IncrementMinor()
-		if err != nil{
+		if err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
/lgtm

hack/go/generate_versions.go
```diff
@@ -181,7 +181,7 @@ func getVersion(repo, ref string) (string, error) {
 	baseURL := fmt.Sprintf("https://raw.githubusercontent.com/%s/%s", repo, ref)
 	link := fmt.Sprintf("%s/VERSION", baseURL)
 	if repo == metricsServerRepo {
-		link = fmt.Sprintf("%s/manifests/release/kustomization.yaml", baseURL)
+		link = fmt.Sprintf("%s/manifests/components/release/kustomization.yaml", baseURL)
```
machine424, line 184: I'm afraid we'll have to branch as before `v0.7.0`, the file was in fact at `/manifests/release/kustomization.yaml`: https://github.com/kubernetes-sigs/metrics-server/blob/v0.6.4/manifests/release/kustomization.yaml

we started using v0.7.0 in 4.16, so we can branch on that.
Or fallback to the other path on a 404, as you want :)

machine424: I didn't check, but maybe we could add a VERSION file in our downstream?
syncbot should know the tag, we could just ask to put it there...

rexagod: Ah! [That](https://github.com/openshift/cluster-monitoring-operator/pull/2252/files#r1474205497) makes so much sense, I missed this! This will definitely break jobs on `<4.16` PRs, true. [This](https://github.com/openshift/cluster-monitoring-operator/pull/2252/files#r1474211363) sounds like a good long-term approach, but would adding a conditional for checking both paths suffice here?

machine424: Yes let's branch to unblock the CI, we'll try the other approach later.

rexagod: Just to confirm, if https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-release-4.16.yaml and https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-monitoring-operator/openshift-cluster-monitoring-operator-release-4.15.yaml are sourced from the respective branches instead of `master`, wouldn't that automatically fix things for us?

rexagod: (since this commit won't go into `4.15` as it's out, and only `4.16` will be up-to-date with `master`?)

machine424: But with the code from `master` we fetch versions for `4.14` and `4.17`, so it should be compatible with both.

rexagod: Right, I was confused by that. Could you elaborate on the branching approach, since we already have branches till `4.17` on both. I'd also like to reiterate the idea of having a conditional that checks both `/manifests/release/kustomization.yaml` and `/manifests/components/release/kustomization.yaml` for a `200`, and goes with that, which would fix things on both fronts. We can drop the older path once the CMO versions using them correspond to EOL releases. WDYT?

rexagod:
JFYI This needs `make versions`.

hack/go/generate_versions.go
machine424: Sorry, I wasn't explicit, by branching I mean "conditional branching": adding a condition on the version

PR #2254, opened 2024-02-05
OCPBUGS-18326: add Console cap annotation to dashboards

simonpasquier:
This commit adds the "capability.openshift.io/name: Console" annotation to the dashboard configmaps.

It also renames the existing configmaps from "grafana-xxx" to "xxx" because otherwise CVO would enable the Console capability on all clusters, even those where the capability was initially disabled [1]. The CVO team has a feature request [2] to improve the situation but for now, we have no other option than creating the configmaps with a different name and deleting the old configmaps (anyway the "grafana-" prefix is misleading since OCP doesn't ship any Grafana instance by default).

[1] https://github.com/openshift/enhancements/blob/master/enhancements/installer/component-selection.md#how-to-implement-a-new-capability
[2] https://issues.redhat.com/browse/OTA-1039

### DIFF
--- a/jsonnet/components/dashboards.libsonnet
+++ b/jsonnet/components/dashboards.libsonnet
@@ -144,26 +144,58 @@ function(params)
     consoleDashboardDefinitions: {
       apiVersion: 'v1',
       kind: 'ConfigMapList',
-      items: std.filterMap(
-        // etcd dashboard is deployed by cluster-etcd-operator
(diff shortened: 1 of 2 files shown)


jan--f:
Nice, thank you! Love the commit message :heart: 

/lgtm

machine424:
Thanks
/lgtm

PR #2258, opened 2024-02-08
MON-3706: chore: poll immediately in the e2e tests

simonpasquier:
It should speed up the test run if we check the condition immediately instead of waiting for 5 seconds.

### DIFF
--- a/test/e2e/alertmanager_test.go
+++ b/test/e2e/alertmanager_test.go
@@ -171,8 +171,6 @@ func testAlertmanagerReady(t *testing.T, name, ns string, validator ...validator
 }
 
 func testAlertmanagerTenancyAPI(t *testing.T, host string) {
-	t.Helper()
-
 	ctx := context.Background()
 	const testNs = "tenancy-api-e2e-test"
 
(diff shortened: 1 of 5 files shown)


machine424:
/lgtm
Thought about doing the same for the client's one https://github.com/openshift/cluster-monitoring-operator/blob/dbcfed9c91a37adbc1e2eb5af5e09172d701cdff/pkg/client/client.go#L2037
but I was worried about getting more failure logs + making unnecessary requests. (especially for delete ones that are known to take time.)
unlike tests, It' not a problem if the sync takes an extra 5 seconds (especially that the tasks run in //).

simonpasquier:
@machine424 100% agreed: for CMO tasks, waiting an extra 5s is often not a big deal.

simonpasquier:
The latest run took `5561s` (https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2188/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-operator/1757316599150284800/artifacts/e2e-agnostic-operator/test/build-log.txt). I've looked at another run (https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2188/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic-operator/1757316599150284800/artifacts/e2e-agnostic-operator/test/build-log.txt) and it took `6855s`.

test/e2e/alertmanager_test.go
```diff
@@ -171,8 +171,6 @@ func testAlertmanagerReady(t *testing.T, name, ns string, validator ...validator
 }
 
 func testAlertmanagerTenancyAPI(t *testing.T, host string) {
-	t.Helper()
```
machine424, line 174: I assume we'll get better context without it?

test/e2e/alertmanager_test.go
```diff
@@ -242,44 +240,53 @@ func testAlertmanagerTenancyAPI(t *testing.T, host string) {
 		now.Add(time.Hour).Format(time.RFC3339),
 	))
 
-	assertDo := func(user string, expectedCode int, do func() (*http.Response, error)) []byte {
-		t.Helper()
```
machine424, line 246: here as well

machine424 (commented):
/lgtm
just have a couple of questions about `t.Helper`

machine424:
/hold
unhold whenever you want.

test/e2e/alertmanager_test.go
simonpasquier: Correct.  Otherwise it's hard to know which step failed exactly.

PR #2261, opened 2024-02-14
MON-3380: re-add kube-rbac-proxy for /metrics and /federate

simonpasquier:
Until https://github.com/brancz/kube-rbac-proxy/issues/146 is implemented, we need a dedicated kube-rbac-proxy for static authz. Otherwise it will always do a SubjectAccessReview for requests authenticated using bearer tokens.


jsonnet/components/prometheus.libsonnet
```diff
@@ -424,11 +422,6 @@ function(params)
                 mountPath: '/etc/tls/private',
                 name: 'secret-' + prometheusTLSSecret,
               },
-              {
```
danielmellado, line 427: I assume no need for tls client here as it'd be done by the rbac proxy?

PR #2264, opened 2024-02-15
OCPBUGS-29531: Add include.release.openshift.io/hypershift label

danielmellado:

### DIFF
--- a/jsonnet/components/cluster-monitoring-operator.libsonnet
+++ b/jsonnet/components/cluster-monitoring-operator.libsonnet
@@ -150,6 +150,7 @@ function(params) {
       name: cfg.name,
       annotations: {
         'include.release.openshift.io/ibm-cloud-managed': 'true',
+        'include.release.openshift.io/hypershift': 'true',
         'include.release.openshift.io/self-managed-high-availability': 'true',
         'include.release.openshift.io/single-node-developer': 'true',
       },
@@ -266,6 +267,7 @@ function(params) {
       name: 'cluster-monitoring-operator-namespaced',
       annotations: {
         'include.release.openshift.io/ibm-cloud-managed': 'true',
+        'include.release.openshift.io/hypershift': 'true',
         'include.release.openshift.io/self-managed-high-availability': 'true',
         'include.release.openshift.io/single-node-developer': 'true',
       },
@@ -317,6 +319,7 @@ function(params) {
       namespace: cfg.namespace,
       annotations: {
         'include.release.openshift.io/ibm-cloud-managed': 'true',
+        'include.release.openshift.io/hypershift': 'true',
         'include.release.openshift.io/self-managed-high-availability': 'true',
         'include.release.openshift.io/single-node-developer': 'true',
       },
--- a/Documentation/data-collection.md
+++ b/Documentation/data-collection.md
@@ -1065,6 +1065,7 @@ metadata:
   name: telemetry-config
   namespace: openshift-monitoring
   annotations:
+    include.release.openshift.io/hypershift: "true"
     include.release.openshift.io/ibm-cloud-managed: "true"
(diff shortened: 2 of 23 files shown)


marioferh (commented):
/lgtm

danielmellado:
@bburt-rh mind taking a look at this for the docs-approved label? thanks!

ashwindasr:
/hold

@danielmellado Looks like the annotation is defined in other places in the code base as well. Can you check this [diff](https://github.com/danielmellado/cluster-monitoring-operator/compare/danielmellado:add_hypershift_annotation...openshift-art-build-bot:art-8431-test?expand=1) and pull this in as well if possible?

danielmellado:
Hi @ashwindasr I've identified a few more places where this was used. Regarding the diff I just see the same files I had, not sure if you wanted to show something else! I can't just pull it in as the manifests are generated from the jsonnet files and they would be overridden. Let me know!

ashwindasr:
Thanks @danielmellado !

As long as `include.release.openshift.io/hypershift': 'true'` follows all instances of `'include.release.openshift.io/ibm-cloud-managed': 'true'` we should be good!

jsonnet/components/cluster-monitoring-operator.libsonnet
```diff
@@ -150,6 +150,7 @@ function(params) {
       name: cfg.name,
       annotations: {
         'include.release.openshift.io/ibm-cloud-managed': 'true',
+        'include.release.openshift.io/hypershift': 'true',
```
machine424, line 153: I'd be great if we're able to identify those CVO resources in `add-annotations.libsonnet` and inject the needed annotations there as well.

machine424:
I don't know the effect of the annotation, but
/lgtm

machine424:
Note that not all files from `manifests/` were annotated https://github.com/openshift/cluster-monitoring-operator/tree/master/manifests
/hold
please unhold if it's ok.

danielmellado:
Hm, true, i.e. https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster-monitoring-operator_04-config.yaml#L1060 @simonpasquier would it be ok just to update those manually?

jsonnet/components/cluster-monitoring-operator.libsonnet
danielmellado: Agree, I'll tackle the libsonnet cleanup on a different PR to unblock this one. Some files are not generated at all by the jsonnet and that's why they're left ignored. It'd be good to consolidate this.

danielmellado:
@ashwindasr the version gate is expected to fail given the current openshift release status. So let's just merge this as soon as the other gates pass again (they were passing before your retest xD) Thanks!

danielmellado:
CI Issues seem totally unrelated ({Operator degraded (MissingDependency): machine-api not ready  Operator degraded (MissingDependency): machine-api not ready}
) so let's check if we're lucky this time

danielmellado:
@simonpasquier I'm a bit lost here, this was previously passing and tbh I feel that the issues are unrelated to CMO :\

danielmellado:
there was an issue with CI, let's double check now

danielmellado:
this passed... not sure what's going on with the CI, it doesn't even get to deploy the cluster.

danielmellado:
Adding @jparrill and @csrwng, mind taking a look at this CI runs? not sure about the hypershift annotation and I'm aware about the MachineConfigPool recent CI outage but besides that IPI install doesn't even get to finish. Thanks!

danielmellado:
note: Agreed with the ART team to skip a specific ROKS annotation for this effort which was making IPI install phase fail

marioferh (commented):
/lgtm

PR #2267, opened 2024-02-19
OCPBUGS-28246: fix Thanos ruler alert generator url

jan--f:

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -3249,7 +3249,6 @@ func (f *Factory) ThanosRulerRBACProxyMetricsSecret() (*v1.Secret, error) {
 }
 
 func (f *Factory) ThanosRulerCustomResource(
-	queryURL string,
 	trustedCA *v1.ConfigMap,
 	grpcTLS *v1.Secret,
 	alertmanagerConfig *v1.Secret,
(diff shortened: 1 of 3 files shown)


pkg/manifests/manifests.go
```diff
@@ -3340,8 +3339,11 @@ func (f *Factory) ThanosRulerCustomResource(
 	f.mountThanosRulerAlertmanagerSecrets(t)
 	f.injectThanosRulerAlertmanagerDigest(t, alertmanagerConfig)
 
-	if queryURL != "" {
-		t.Spec.AlertQueryURL = queryURL
+	if f.consoleConfig != nil && f.consoleConfig.Status.ConsoleURL != "" {
```
machine424, line 3342: Maybe we could refactor this under an util, we repeat the same logic in other places as well https://github.com/search?q=repo%3Aopenshift%2Fcluster-monitoring-operator%20consoleConfig.Status.ConsoleUR&type=code

machine424: Will `/monitoring` be accessible/useful for non-admins as well?

(Simon mentioned sth similar at https://github.com/openshift/cluster-monitoring-operator/blob/1878d8205445a8b3eaf594b48976c3fa3956f173/pkg/manifests/manifests.go#L480-L481)

jan--f: > Maybe we could refactor this under an util, we repeat the same logic in other places as well https://github.com/search?q=repo%3Aopenshift%2Fcluster-monitoring-operator%20consoleConfig.Status.ConsoleUR&type=code

I thought about it, but given that we add the console URL different types and fields with different names I decided not to. Happy to be convinced with a good approach though...me failing to come up with something clever means very little.

machine424: (cc @simonpasquier, if you know that https://github.com/openshift/cluster-monitoring-operator/blob/1878d8205445a8b3eaf594b48976c3fa3956f173/pkg/manifests/manifests.go#L480-L481 can be done now, otherwise I can dig in.)

jan--f: You are right of course. Added a commit accordingly.

PR #2269, opened 2024-02-21
OCPBUGS-30257: Making sure proxy settings are correctly forwarded in the generated remote write configs

Nikokolas3270:
Fix some golang mistake in which the `remoteWrite` config objects were copied instead of being passed by reference.
**This may impact the customers as the remote write calls may now be proxied:**
- We may want to fix this issue first: https://github.com/prometheus-operator/prometheus-operator/issues/6301
- Also it would be nice to have the Prometheus Operator and CMO forward the `noProxy` option on top of the `proxyUrl`; indeed Prometheus is supporting that option: https://github.com/prometheus/prometheus/blob/main/docs/configuration/configuration.md#oauth2

c.c. @simonpasquier 

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1428,7 +1428,9 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, trustedCABundleCM *v1.Config
 		p.Spec.RemoteWrite = addRemoteWriteConfigs(clusterID, p.Spec.RemoteWrite, f.config.ClusterMonitoringConfiguration.PrometheusK8sConfig.RemoteWrite...)
 	}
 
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
Thanks! Ideally we'd need a unit test to avoid any regression.

simonpasquier:
/retitle OCPBUGS-30257: Making sure proxy settings are correctly forwarded in the generated remote write configs

cloned the original ticket and moved to OCPBUGS so it can be backported if needed.

simonpasquier:
/hold

we need test case.

jan--f:
I added a test in #2281 and confirmed this change fixes it. @simonpasquier Are you ok to go ahead with this?

Nikokolas3270:
@simonpasquier : added a test case.

PR #2272, opened 2024-02-28
MON-3706: chore: simplify GetServiceAccountToken()

simonpasquier:

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -33,6 +33,8 @@ import (
 	routev1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	"github.com/openshift/cluster-monitoring-operator/pkg/client"
 	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	"k8s.io/utils/ptr"
 
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1beta1"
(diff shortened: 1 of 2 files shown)


test/e2e/framework/framework.go
```diff
@@ -306,26 +308,35 @@ func (f *Framework) CreateServiceAccount(namespace, serviceAccount string) (Clea
...
-	var token string
-	err := Poll(5*time.Second, time.Minute, func() error {
-		secrets, err := f.KubeClient.CoreV1().Secrets(namespace).List(ctx, metav1.ListOptions{})
+	var (
+		ctx             = context.Background()
+		token           string
+		tokenExpiration = time.Duration(time.Hour * 12)
+		expirationTime  = metav1.NewTime(time.Now().Add(tokenExpiration))
+	)
+	err := Poll(time.Second, time.Minute, func() error {
```
danielmellado, line 317: Just curious about the change in the poll interval, any reason about the change in the poll interval? that's just a nit but I'm wondering if this would be more poll intensive.

test/e2e/framework/framework.go
```diff
@@ -306,26 +308,35 @@ func (f *Framework) CreateServiceAccount(namespace, serviceAccount string) (Clea
...
-			if _, dockerToken := secret.Annotations["openshift.io/create-dockercfg-secrets"]; dockerToken {
-				continue
-			}
-			if strings.Contains(secret.Name, fmt.Sprintf("%s-token-", name)) {
-				token = string(secret.Data["token"])
-				return nil
-			}
+
+		if tokenReq.Status.ExpirationTimestamp.Before(&expirationTime) {
+			return fmt.Errorf("expiration too short: %v < %v", tokenReq.Status.ExpirationTimestamp, expirationTime)
```
danielmellado, line 333: total nit, I'd be explicit about that the expiration refers to the token.

test/e2e/user_workload_monitoring_test.go
```diff
@@ -323,7 +323,7 @@ func assertMetricsForMonitoringComponents(t *testing.T) {
 	} {
 		t.Run(service, func(t *testing.T) {
 			f.ThanosQuerierClient.WaitForQueryReturn(
-				t, 10*time.Minute, fmt.Sprintf(`count(up{service="%s",namespace="openshift-user-workload-monitoring"} == 1)`, service),
+				t, time.Minute, fmt.Sprintf(`count(up{service="%s",namespace="openshift-user-workload-monitoring"} == 1)`, service),
```
danielmellado, line 326: I assume that this is just an optimization to make tests faster.

test/e2e/framework/framework.go
simonpasquier: The previous approach was to extract the token from the SA secret and it could take time from Kubernetes to populate the secret. The new approach gets a token from the API which should succeed from the first try in the vast majority of runs.

danielmellado: Thanks for the clarifications ;)

PR #2273, opened 2024-03-06
OCPBUGS-30294: Add aggregate-to-view to cluster-monitoring-view

danielmellado:
Since OpenShift 4.15.0, AlertManager requires the user to have an RBAC
to the subresource of 'prometheuses/api'. The ClusterRole
'cluster-monitoring-view' is not configured to use aggregate-to-view [1]

This commit adds the rbac.authorization.k8s.io/aggregate-to-view: 'true'
label to the cluster-monitoring-view ClusterRole.

---
[1] https://kubernetes.io/docs/reference/access-authn-authz/rbac/#default-roles-and-role-bindings

### DIFF
--- a/assets/cluster-monitoring-operator/cluster-role-view.yaml
+++ b/assets/cluster-monitoring-operator/cluster-role-view.yaml
@@ -4,6 +4,7 @@ metadata:
   labels:
     app.kubernetes.io/managed-by: cluster-monitoring-operator
     app.kubernetes.io/part-of: openshift-monitoring
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
   name: cluster-monitoring-view
 rules:
 - apiGroups:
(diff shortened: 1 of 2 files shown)


marioferh (commented):
/lgtm

machine424:
I don't see any yaml changes and the generate job is passing, maybe the change has no effect or it's already applied.
/hold

simonpasquier:
/hold

While the change might solve the reported issue, we need to review the implications..

juzhao:
@simonpasquier @machine424 @danielmellado please see my comment in bug, edit clusterrole cluster-monitoring-view to add label: rbac.authorization.k8s.io/aggregate-to-view: "true", login console with the user, could view Alerts/ALerting rules, but there is warning on page
```
Error loading silences from Alertmanager. Some of the alerts below may actually be silenced.
Forbidden 
```
also Forbidden error for Silences tab, can not view silences

PR #2275, opened 2024-03-06
MON-3789: add a golangci-lint-fix makefile target to fix golangci-lint errros when possible

machine424:
…hen possible

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,7 +195,11 @@ go-fmt:
 
 .PHONY: golangci-lint
 golangci-lint: $(GOLANGCI_LINT_BIN)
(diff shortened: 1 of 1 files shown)


danielmellado (commented):
/lgtm althought we could use a small note on the dev docs

machine424:
> /lgtm althought we could use a small note on the dev docs

I don't see any section about linting, I could take care of that but in another PR.

simonpasquier (commented):
/lgtm

machine424:
/retitle MON-3789: add a golangci-lint-fix makefile target to fix golangci-lint errros when possible

PR #2277, opened 2024-03-07
OBSDOCS-883: change docs to correct retention time uwm prometheus

marioferh:
Documentation for User Workload Monitoring implies that default retention time is 15d, when it is actually 24h in practice


slashpai (approved):
/lgtm

danielmellado:
@marioferh mind checking @machine424 comment about the hold? I'd like to see if we get this in soon.

PR #2280, opened 2024-03-12
MON-3748: Enable audit logs by default for metrics-server

slashpai:
Enabling audit logs helps troubleshoot issues in Metrics Server. The details of the audit level can be tweaked by selecting an audit profile which corresponds to audit Log Level.
See: https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#audit-policy

The following profiles are provided:
  - Metadata:  The default audit profile
  - None
  - Request
  - RequestResponse

User can pick any of the profile by modifying the CMO configmap as follows

```
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    metricsServer:
      audit:
        profile: Request
```

**Adapted the change as in Prometheus Adapter**

### DIFF
--- a/Documentation/openshiftdocs/modules/metricsserverconfig.adoc
+++ b/Documentation/openshiftdocs/modules/metricsserverconfig.adoc
@@ -18,6 +18,8 @@ Appears in: link:clustermonitoringconfiguration.adoc[ClusterMonitoringConfigurat
 [options="header"]
 |===
 | Property | Type | Description 
+|audit|*Audit|Defines the audit configuration used by the Metrics Server instance. Possible profile values are: `metadata`, `request`, `requestresponse`, and `none`. The default value is `metadata`.
+
 |nodeSelector|map[string]string|Defines the nodes on which the pods are scheduled.
 
 |tolerations|[]v1.Toleration|Defines tolerations for the pods.
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -203,6 +203,7 @@ var (
 	MetricsServerClusterRoleBinding              = "metrics-server/cluster-role-binding.yaml"
 	MetricsServerClusterRoleBindingAuthDelegator = "metrics-server/cluster-role-binding-auth-delegator.yaml"
 	MetricsServerRoleBindingAuthReader           = "metrics-server/role-binding-auth-reader.yaml"
+	MetricsServerConfigMapAuditPolicy            = "metrics-server/configmap-audit-profiles.yaml"
 	MetricsServerDeployment                      = "metrics-server/deployment.yaml"
 	MetricsServerService                         = "metrics-server/service.yaml"
 	MetricsServerServiceMonitor                  = "metrics-server/service-monitor.yaml"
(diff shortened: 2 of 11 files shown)


pkg/manifests/manifests.go
```diff
@@ -2098,6 +2103,19 @@ func (f *Factory) MetricsServerDeployment(kubeletCABundle *v1.ConfigMap, tlsSecr
...
 	}
 
+	if err := validateAuditProfile(config.Audit.Profile); err != nil {
+		return nil, err
+	}
+
+	profile := strings.ToLower(string(config.Audit.Profile))
+	containers[idx].Args = append(containers[idx].Args,
+		fmt.Sprintf("--audit-policy-file=/etc/audit/%s-profile.yaml", profile),
+		"--audit-log-path=/var/log/metrics-server/audit.log",
```
machine424, line 2113: I'd send them directly to stdout for more visibility, also to avoid reserving 500MB of disk (for a Deployment it could be considered a lot).
And if they'll not be enabled all the time (my other proposition),  they'll not harm the logging stack.

Documentation/openshiftdocs/modules/metricsserverconfig.adoc
```diff
@@ -18,6 +18,8 @@ Appears in: link:clustermonitoringconfiguration.adoc[ClusterMonitoringConfigurat
 [options="header"]
 |===
 | Property | Type | Description 
+|audit|*Audit|Defines the audit configuration used by the Metrics Server instance. Possible profile values are: `metadata`, `request`, `requestresponse`, and `none`. The default value is `metadata`.
```
machine424, line 21: I'd just stuck with one profile that will help with debugging for simplicity.
And I suggest disabling the audit logs by default and only propose a flag to enable them.
We could even enable them when debugging is enabled (if we already have a flag for that).

machine424 (commented):
Cool, I thought the api-server audit logs were already covering requests to metrics server.

Documentation/openshiftdocs/modules/metricsserverconfig.adoc
danielmellado: +1 on not enabling this by default. It would make sense that a user wouldn't like to have debug mode enabled all the time so we should offer some option, bonus point on matching with debugging is enabled.

machine424:
(I'd wait until we're sure those audit logs are useful (helped during a debug or sth) before allowing to enable them (maybe debug logs are sufficient), but your call, as this was already accepted in the enhancement proposal :))

pkg/manifests/manifests_test.go
```diff
@@ -2797,6 +2797,135 @@ metricsServer:
...
+			}
+
+			adapterArgs := d.Spec.Template.Spec.Containers[0].Args
+			auditArgs := []string{}
+			for _, arg := range adapterArgs {
+				if strings.HasPrefix(arg, "--audit-") {
+					auditArgs = append(auditArgs, arg)
+				}
+			}
+			assertDeepEqual(t, test.args, auditArgs,
```
machine424, line 2923: nit: we could use `require.Contains` (trying to advertise `testify` on every PR I review :))

machine424 (commented):
After discussing this, we agreed to proceed in the same manner as what has already been implemented for the prometheus-adapter.
/lgtm

pkg/manifests/manifests.go
slashpai: Since we will be including it in must-gather we will be keeping it in storage

PR #2281, opened 2024-03-12
manifest: test proxy setting propagates to remote write

jan--f:

### DIFF
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -54,13 +54,17 @@ func defaultInfrastructureReader() InfrastructureReader {
 	return &fakeInfrastructureReader{highlyAvailableInfrastructure: true, hostedControlPlane: false}
 }
 
-type fakeProxyReader struct{}
+type fakeProxyReader struct {
+	httpProxy  string
+	httpsProxy string
+	noProxy    string
+}
 
-func (f *fakeProxyReader) HTTPProxy() string { return "" }
+func (f *fakeProxyReader) HTTPProxy() string { return f.httpProxy }
 
-func (f *fakeProxyReader) HTTPSProxy() string { return "" }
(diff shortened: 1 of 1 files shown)


jan--f:
needs #2269 to succeed.

pkg/manifests/manifests_test.go
```diff
@@ -1299,6 +1303,34 @@ func TestPrometheusK8sRemoteWriteURLs(t *testing.T) {
...
+	p, err := f.PrometheusK8s(
+		&v1.Secret{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+		&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "foo"}},
+		nil,
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if p.Spec.RemoteWrite[0].URL != "https://test.remotewrite.com/api/write" {
```
machine424, line 1326: nit: maybe we could use `testify/require`

pkg/manifests/manifests_test.go
```diff
@@ -1299,6 +1303,34 @@ func TestPrometheusK8sRemoteWriteURLs(t *testing.T) {
...
+		t.Errorf("want remote write URL https://test.remotewrite.com/api/write, got %v", p.Spec.RemoteWrite[0].URL)
+	}
+	if p.Spec.RemoteWrite[0].ProxyURL != "http://test.proxy.com" {
```
machine424, line 1329: nit: maybe we could add a check to `TestPrometheusK8sConfiguration` that no proxy is set by default (who knows).

machine424 (commented):
/lgtm
/hold
for some nits, don't hesitate to unhold.

jan--f:
/close
test was addressed in #2269

PR #2283, opened 2024-03-12
MON-3752: expose metric-denylist for KSM

rexagod:
_Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>_

<details>
<summary>CMO Config (with denied metrics)</summary>

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    kubeStateMetrics:
      metricDenylist:
        - ^kube_.+_created$
        - ^kube_.+_annotations$
```

</details>

<details>
<summary>KSM Logs</summary>

```
I0312 21:31:50.753546       1 wrapper.go:120] "Starting kube-state-metrics"
W0312 21:31:50.753856       1 client_config.go:618] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0312 21:31:50.754060       1 server.go:192] "Used default resources"
I0312 21:31:50.754100       1 types.go:184] "Using all namespaces"
I0312 21:31:50.754164       1 server.go:225] "Metric allow-denylisting" allowDenyStatus="Excluding the following lists that were on denylist: ^kube_.+_annotations$, ^kube_.+_created$"
W0312 21:31:50.754199       1 client_config.go:618] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0312 21:31:50.754825       1 utils.go:70] "Tested communication with server"
I0312 21:31:50.771593       1 utils.go:75] "Run with Kubernetes cluster version" major="1" minor="28" gitVersion="v1.28.7+6e2789b" gitTreeState="clean" gitCommit="dfd36a72760e09d4971f8b49ed335f5522dab8af" platform="linux/amd64"
I0312 21:31:50.771662       1 utils.go:76] "Communication with server successful"
I0312 21:31:50.772149       1 server.go:347] "Started metrics server" metricsServerAddress="127.0.0.1:8081"
I0312 21:31:50.772315       1 metrics_handler.go:99] "Autosharding disabled"
I0312 21:31:50.774043       1 server.go:336] "Started kube-state-metrics self metrics server" telemetryAddress="127.0.0.1:8082"
I0312 21:31:50.776631       1 server.go:72] levelinfomsgListening onaddress127.0.0.1:8082
I0312 21:31:50.776699       1 server.go:72] levelinfomsgTLS is disabled.http2falseaddress127.0.0.1:8082
I0312 21:31:50.778155       1 builder.go:271] "Active resources" activeStoreNames="certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,ingresses,jobs,leases,limitranges,mutatingwebhookconfigurations,namespaces,networkpolicies,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses,validatingwebhookconfigurations,volumeattachments"
I0312 21:31:50.778438       1 server.go:72] levelinfomsgListening onaddress127.0.0.1:8081
I0312 21:31:50.778514       1 server.go:72] levelinfomsgTLS is disabled.http2falseaddress127.0.0.1:8081                                    
```

</details>

***


rexagod:
The [description](https://github.com/openshift/cluster-monitoring-operator/pull/2283#issue-2182704971) shows the current behavior, which is to replace the deny-list with the one is specified by the user (CMO configuration deny-list takes precedence over the default one).

However, owing to this, users could very well find themselves (a) denying certain default metrics that various alerts and dashboards depend upon, or (b) enabling certain default metrics that could cause cardinality issues, `^kube_.+_annotations$`, for instance.

jan--f:
Adding high cardinality metrics is a concern, but I'm more concerned about users removing metrics that we rely on in dashboards and alerts.
The feature request around this is specifically to add previously denied metrics back. Giving users full reign over the denylist strikes me as quite dangerous, as they now easily render the stack mostly useless.

pkg/manifests/manifests.go
```diff
@@ -771,6 +773,89 @@ func (f *Factory) KubeStateMetricsDeployment() (*appsv1.Deployment, error) {
...
+				Path:   "/metrics",
+			}
+			tr := &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			}
+			client := &http.Client{Transport: tr}
+			resp, err = client.Get(u.String())
+			if err == nil {
+				break
+			} // Unauthorized.
```
rexagod, line 815: Todo: fix this.

pkg/tasks/kubestatemetrics.go
```diff
@@ -95,6 +95,21 @@ func (t *KubeStateMetricsTask) Run(ctx context.Context) error {
 		return fmt.Errorf("reconciling kube-state-metrics Deployment failed: %w", err)
 	}
 
+	prometheusK8sTokenSecret, err := t.client.GetTokenSecret(ctx, dep.Namespace, "prometheus-k8s")
```
rexagod, line 98: I've used the `prometheus-k8s-token-...` secret to query the `/metrics` endpoint, since that was [allowed by KRP](https://github.com/openshift/cluster-monitoring-operator/blob/6795f509e77cce6d24e5a3e371a432ca22e1a8e7/assets/kube-state-metrics/kube-rbac-proxy-secret.yaml#L17).

pkg/manifests/manifests.go
```diff
@@ -772,6 +774,94 @@ func (f *Factory) KubeStateMetricsDeployment() (*appsv1.Deployment, error) {
...
+	var p int32
+	for _, port := range service.Spec.Ports {
+		if port.Name == "https-main" {
+			p = port.Port
+			break
+		}
+	}
+
+	// Query the endpoint.
+	t := time.NewTimer(5 * time.Second)
```
danielmellado, line 796: I'm usually not a huge fan of hardcoding times within the code, but I see the point. How about setting a var for this in any case? I know it's only used there but it'd be cool from a readability standpoint and for a latter change if needed.

jan--f: To add to this, is a Timer the right tool here? Wouldn't the worst case here be a hot loop creating http requests for 5 seconds? 
Wouldn't a Poll variant with timeout and maybe backoff make sense here?

pkg/manifests/manifests.go
```diff
@@ -772,6 +774,94 @@ func (f *Factory) KubeStateMetricsDeployment() (*appsv1.Deployment, error) {
...
+		case <-t.C:
+			return nil, fmt.Errorf("timed out waiting to fetch kube-state-metrics /metrics endpoint: %w", err)
+		default:
+			u := &url.URL{
+				Scheme: "https",
+				Host:   fmt.Sprintf("%s.%s.svc:%d", service.GetName(), service.GetNamespace(), p),
+				Path:   "/metrics",
+			}
+			client := &http.Client{
+				Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}},
```
jan--f, line 811: Is it too complicated here to find the right TLS artifacts?
The `metrics-client-certs` secret should also have mTLS artifacts, this code wouldn't have to retrieve a bearer token for auth I think. Though there might be some fault tolerance implications when using mTLS only. @simonpasquier wdyt?

rexagod: > Is it too complicated here to find the right TLS artifacts?

I missed incorporating this, but I think that should be doable. I can use `metrics-client-certs` for the TLS certificates, and perhaps `metrics-client-ca` for the CA. I believe we'll also need to redeploy KSM if these dependent assets are rotated (same as [`metrics-server`](https://github.com/openshift/cluster-monitoring-operator/blob/bccc6b7292136f9b14dee42fc78d0b788739754d/pkg/manifests/manifests.go#L2161-L2173)).

pkg/manifests/manifests.go
```diff
@@ -772,6 +774,94 @@ func (f *Factory) KubeStateMetricsDeployment() (*appsv1.Deployment, error) {
 	return d, nil
 }
 
+// KubeStateMetricsDenylistBoundsCheck ensures that the user is:
+// * able to enable metrics that are denied by default, and,
+// * unable to disable metrics that are enabled by default.
+func (f *Factory) KubeStateMetricsDenylistBoundsCheck(deployment *appsv1.Deployment, service *v1.Service, secret *v1.Secret) (*appsv1.Deployment, error) {
```
machine424, line 780: Clever idea :) but I have some concerns:

- Making an operator depend on one of its operands doesn't seem that common and natural to me. Here, for example, if CMO cannot reach out to KSM's` /metrics`, it'll abort its task and not try to "help it". This could be fixed, but maybe there will be other deadlocks like this.
- Does KSM always expose all the metrics that it can expose? I mean, does it initialize all the metrics? If not, `/metrics` will not really be a source of truth: A user could drop a not-there-yet metric.
- The code seems a little bit hacky as we're re-implementing some parts of a scraper, and I'm concerned about its maintainability.

I don't have a better alternative but I see that for the majority of metrics in the current `--metric-denylist` we specify the exact metrics. How about expanding the ones with `xxx_.+_yyy` and just allowing users to select from that list? But maybe it'll be harder to maintain. (we can still have tests with `xxx_.+_yyy` to be sure we take new metrics we want to drop into account)

Maybe there is also something to be done in connection with collection profiles...

rexagod: > Making an operator depend on one of its operands doesn't seem that common and natural to me. Here, for example, if CMO cannot reach out to KSM's /metrics, it'll abort its task and not try to "help it". This could be fixed, but maybe there will be other deadlocks like this.

Ah, I believe CMO itself won't lean on KSM anymore than it did before this patch. A bounds-check for an intrinsic property (`--metric-denylist`) of the operand is nothing more than its own behavior being handled by CMO, same as any other non-repairable (by choice or by trial) error it'd exhibit (for instance, an invalid CRS config). We are failing KSM reconciliation and throwing that error without making any assumptions or implicit conversions to be explicit towards the user on CMO's resolution of their supplied deny-list. PLMK if I missed something.

> Does KSM always expose all the metrics that it can expose? I mean, does it initialize all the metrics? If not, /metrics will not really be a source of truth: A user could drop a not-there-yet metric.

Since help-texts will always be present in the exposition data, so [we should be safe](https://github.com/openshift/cluster-monitoring-operator/pull/2283/files#diff-1f9d2d18d4cabd8262e585b932330183f7cb9b544ac9f27c8bcb3ef9c8d14376R840-R841).

> [...] re-implementing some parts of a scraper [...]

Do we already have helpers for this in CMO? The single-scrape logic essentially crafts a request and interprets it, while pinging the `/metrics` endpoint with the necessary auth permissions. It seemed like a straightforward implementation, but feel free to drop any suggestions (in addition to Jan's comment above, as I've made those changes locally but I want to solve the auth issue before I push again) that may abstract this and reduce maintainability costs.

machine424: > Ah, I believe CMO itself won't lean on KSM anymore than it did before this patch. A bounds-check for an intrinsic property (--metric-denylist) of the operand is nothing more than its own behavior being handled by CMO, same as any other non-repairable (by choice or by trial) error it'd exhibit (for instance, an invalid CRS config). We are failing KSM reconciliation and throwing that error without making any assumptions or implicit conversions to be explicit towards the user on CMO's resolution of their supplied deny-list. PLMK if I missed something.

I see, but in this case, it seems to me that if KSM itself is broken (or CMO thinks it’s broken), CMO will not sync KSM, which isn’t the behavior we expect from an operator. For example, if one makes an unintentional change to the KSM Deployment that broke it, CMO will abandon it instead of fixing it as expected.
CMO being unable to proceed because of some third-party dependency that CMO cannot even control/adjust (CR, a resource managed by another operator e.g.) is okay, but CMO not being able to fix an operand because the broken operand isn’t responding to CMO, well, that’s a circular dependency :)
Again, this is just an example, maybe there are others.

> Since help-texts will always be present in the exposition data, so [we should be safe](https://github.com/openshift/cluster-monitoring-operator/pull/2283/files#diff-1f9d2d18d4cabd8262e585b932330183f7cb9b544ac9f27c8bcb3ef9c8d14376R840-R841).

Ok, then KSM initializes all the metrics, (even the empty ones), good to know.

> Do we already have helpers for this in CMO? The single-scrape logic essentially crafts a request and interprets it, while pinging the /metrics endpoint with the necessary auth permissions. It seemed like a straightforward implementation, but feel free to drop any suggestions (in addition to Jan's comment above, as I've made those changes locally but I want to solve the auth issue before I push again) that may abstract this and reduce maintainability costs.

I don't think there are any existing utils for this (the only similar ones I see are used for tests), it'll be a new "trait/responsibility" that I'm not sure we should be adding to CMO. But the alternative I'm proposing (to expand the`xxx_.+_yyy`) isn't without its flaws.
Of course, if we agree to merge this, maybe we should consider moving that new logic from `manifests.go` and adding some tests.

pkg/manifests/manifests.go
rexagod: Done, PTAL.

pkg/manifests/manifests.go
rexagod: As a side note, I believe we should drop the following from the TLS cipher suite (for all components).

```
W0416 13:58:20.519009       1 secure_serving.go:69] Use of insecure cipher 'TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256' detected.
W0416 13:58:20.519041       1 secure_serving.go:69] Use of insecure cipher TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256' detected.
```

pkg/manifests/manifests.go
rexagod: IMHO this boils down to a question of trading-off between:

* failing on an invalid deny-list, albeit at the expense of degrading a healthy KSM instance, or,
* subtly emitting or logging the issue, and relying on the editor to diagnose the issue.

I guess I'm okay with wherever the team lands on this.

> Ok, then KSM initializes all the metrics, (even the empty ones), good to know.

Ah, sorry, I believe I should've been more explicit. So KSM has a [list of default resources](https://github.com/kubernetes/kube-state-metrics/blob/93e79fc9d8741c5e605fdab0e56700fdc320c7f1/pkg/options/resource.go#L27) that it builds metric stores for from the get-go, but these have to be present in the cluster in order for the informers to trigger the metric family population. If so, these are initialized, and I believe this set should be a safe assumption for all OpenShift variants (I've tested this on `hypershift`, but not all of them individually). LMK if you think otherwise.

pkg/manifests/manifests.go
jan--f: Mind raising a bug for us to address this?

pkg/manifests/manifests.go
jan--f: Thanks Ayoub for raising this, I think its well worth thinking carefully about the failure scenarios here.
Afaiu Ayoubs concern is that an erroneous user supplied deny list will cause unrelated tasks to fail. In the current iteration this is true. If `KubeStateMetricsDenylistBoundsCheck` returns an error the KSM PrometheusRule and ServiceMonitor will not be reconciled. Also the config sharing task (in its own task group) will not run. 

Maybe the application of the user supplied deny list should run in its own task in the last task group. If that fails at least the remaining stack is already rolled out completely.

machine424: Thanks for this rubberducking session :)

Actually, I thought that `CreateOrUpdateDeployment` call was moved after `KubeStateMetricsDenylistBoundsCheck`. That's why I was talking about a deadlock: "If CMO cannot reachKSM /metrics because a user broke the `Deployment` resource on the cluster with a non-denylist related change, CMO will not update the `Deployment` to try to fix KSM...". However, in truth, we call `CreateOrUpdateDeployment` both before and after. My bad.

Still, as Jan mentioned, `KubeStateMetricsDenylistBoundsCheck` may prevent the following `PrometheusRule/ServiceMonitor` related operations. But if `KubeStateMetricsDenylistBoundsCheck` **can only** return an error in case of an erroneous user-supplied denylist, I think it's okay to fail the task. I don't believe we validate at 100% all user inputs elsewhere. (If a user really wants to break a component, they'd be able to do so).

If the "expand the `xxx_.+_yyy` metrics and let them choose from them" approach isn't appealing, why couldn't we just pass the input as is? As we do for the net interfaces, e.g.: https://github.com/openshift/cluster-monitoring-operator/blob/4b8086df25735168099a7f18ab78deaa30b5c73d/pkg/manifests/config.go#L306-L320. I'm not sure, but maybe excluding the wrong interface can already "break" some stuff.

Besides the ones for alerts, dashboards, etc., which metrics can we not afford to lose? Couldn't we define them explicitly as part of a "minimal collection profile" and ensure they're never excluded? (This slightly aligns with the "expand the `xxx_.+_yyy` metrics" approach). This way, we only have one method for defining required/minimal metrics: collection profiles.

(Sorry, if this was discussed elsewhere, please point me to that)

test/e2e/kube_state_metrics_test.go
```diff
@@ -65,3 +66,115 @@ func TestKSMMetricsSuppression(t *testing.T) {
...
+
+func TestKSMDenylistBoundsCheck(t *testing.T) {
+
+	// Update CMO config to include an invalid deny-list.
+	cfg, err := getOrCreateCMOConfig(t)
+	if err != nil {
+		t.Errorf("failed to get cluster-monitoring-config configmap: %v", err)
+	}
+	cfgCopy := cfg.DeepCopy()
+	invalidMetricExpression := "kube_lease_owner"
```
juzhao, line 78: not sure why `kube_lease_owner` is invalid, from api
`| metricDenylist | []string | Comma-separated list of metrics not to be enabled. This list comprises exact metric names and/or regex patterns. |`
seems it's valid, or it should be `^kube_lease_owner$`?

Tai-RedHat:
Seem there is no logs feedback after I modified KSM deploy.
I tried delete `kube_*_created,` or delete the whole denylist. Then I check the deploy logs, no new updates.
`oc -n openshift-monitoring logs deploy/kube-state-metrics`

pkg/manifests/manifests.go
rexagod: > However, in truth, we call CreateOrUpdateDeployment both before and after. My bad.

I'm not sure I understand. KSM's task will still error out if a non-empty invalid deny-list is supplied.

> Besides the ones for alerts, dashboards, etc., which metrics can we not afford to lose? Couldn't we define them explicitly as part of a "minimal collection profile" and ensure they're never excluded? (This slightly aligns with the "expand the xxx_.+_yyy metrics" approach). This way, we only have one method for defining required/minimal metrics: collection profiles.

The problem with doing that, is multi-staged and multi-faceted. Note that some bullets below may go outside of the scope of your question, but I've included them for the sake of arriving on a consensus.

- We cannot exhaustively _lasso_ all metrics that critical OpenShift components depend upon at **any** given time, unless we have [a set of `targetSelector`s](https://github.com/rexagod/cpv/tree/main#extraction) that allows us to do that.
- [From the point-of-view of the users being able to customize denied metrics, and CPs being a deny-list alternative] If CMO was to "inject" the user's deny-list regex expressions into the profile (or give them any control over it), this will allow users to "break" CMO's "minimal" commitment, rendering the full and minimal profiles similar, etc.
- CPs that the CMO is responsible for are not meant to be dynamic (as in, configurable-on-the-fly). Using the deny-list option in the configuration, users have reasonable control over their **generated metrics**, without CMO reconciling them back. Additionally, CPs will choke out a lot of high-cardinality metrics **after they have been generated**, rather than not generating them at all (deny-list).
- We don't have a concept of per-component collection profiles yet, but hypothetically, if we had a `custom` collection profile for custom use cases (disabling or enabling a set of metrics that don't fit into the `minimal` profile's characteristics), that will **disallow users to apply other profiles** at the same time. This means that if they want a set of metrics denied, but at the same time have a **cluster-wide** `minimal` profile (or any other profile that's different from the currently applied (`custom`) profile that has denied metrics at the user's discretion) that they want to apply, they won't be able to do so. We will lose the ability to decouple (a) "denying" a set of metrics from (b) "defining" a set of metrics that establish various cluster behaviors.
- `drop`ing metrics in the KSM `minimal` service monitor will replace a KSM behavior, that it already has explicit support for, with a CMO responsibility entirely. IMHO choosing between extending an already existing behavior that a component exhibits (as we do stack-wide for a lot of them) versus not using that at all and replacing that with a CMO-specific trait is not ideal, given that we already rely on deny-listing a set of default metrics in the KSM deployment. It only makes sense to extend it as compared to hooking that to one of KSM's profiles (`minimal`, `custom`, etc.).

> But if KubeStateMetricsDenylistBoundsCheck can only return an error in case of an erroneous user-supplied denylist, I think it's okay to fail the task.

IIUYC we'd still need to modify the deployment after the bounds check, to make the valid user deny-list come into effect.

> [...] why couldn't we just pass the input as is?

Regexes are supported in the `kubeStateMetrics.metricDenylist`, if that's what you meant here.

***

> Maybe the application of the user supplied deny list should run in its own task in the last task group. If that fails at least the remaining stack is already rolled out completely.

ACK, I'll send a patch.

rexagod: > Besides the ones for alerts, dashboards, etc., which metrics can we not afford to lose? Couldn't we define them explicitly as part of a "minimal collection profile" and ensure they're never excluded? (This slightly aligns with the "expand the xxx_.+_yyy metrics" approach). This way, we only have one method for defining required/minimal metrics: collection profiles.

The problem with doing that, is multi-staged and multi-faceted. Note that some bullets below may go outside of the scope of your question, but I've included them for the sake of arriving on a consensus.

- We cannot exhaustively _lasso_ all metrics that critical OpenShift components depend upon at **any** given time, unless we have [a set of `targetSelector`s](https://github.com/rexagod/cpv/tree/main#extraction) that allows us to do that.
- [From the point-of-view of the users being able to customize denied metrics, and CPs being a deny-list alternative] If CMO was to "inject" the user's deny-list regex expressions into the profile (or give them any control over it), this will allow users to "break" CMO's "minimal" commitment, rendering the full and minimal profiles similar, etc.
- CPs that the CMO is responsible for are not meant to be dynamic (as in, configurable-on-the-fly). Using the deny-list option in the configuration, users have reasonable control over their **generated metrics**, without CMO reconciling them back. Additionally, CPs will choke out a lot of high-cardinality metrics **after they have been generated**, rather than not generating them at all (deny-list).
- We don't have a concept of per-component collection profiles yet, but hypothetically, if we had a `custom` collection profile for custom use cases (disabling or enabling a set of metrics that don't fit into the `minimal` profile's characteristics), that will **disallow users to apply other profiles** at the same time. This means that if they want a set of metrics denied, but at the same time have a **cluster-wide** `minimal` profile (or any other profile that's different from the currently applied (`custom`) profile that has denied metrics at the user's discretion) that they want to apply, they won't be able to do so. We will lose the ability to decouple (a) "denying" a set of metrics from (b) "defining" a set of metrics that establish various cluster behaviors.
- `drop`ing metrics in the KSM `minimal` service monitor will replace a KSM behavior, that it already has explicit support for, with a CMO responsibility entirely. IMHO choosing between extending an already existing behavior that a component exhibits (as we do stack-wide for a lot of them) versus not using that at all and replacing that with a CMO-specific trait is not ideal, given that we already rely on deny-listing a set of default metrics in the KSM deployment. It only makes sense to extend it as compared to hooking that to one of KSM's profiles (`minimal`, `custom`, etc.).

> But if KubeStateMetricsDenylistBoundsCheck can only return an error in case of an erroneous user-supplied denylist, I think it's okay to fail the task.

IIUYC we'd still need to modify the deployment after the bounds check, to make the valid user deny-list come into effect.

> [...] why couldn't we just pass the input as is?

Regexes are supported in the `kubeStateMetrics.metricDenylist`, if that's what you meant here.

***

> Maybe the application of the user supplied deny list should run in its own task in the last task group. If that fails at least the remaining stack is already rolled out completely.

ACK, I'll send a patch.

test/e2e/kube_state_metrics_test.go
rexagod: Hey, @juzhao! The metric is invalid in the context of the metric deny-list test bounds. This basically implies that this metric cannot be disabled by the user (by specifying it in `metricDenylist`) due to the fact that it is not a part of the [default set of metrics that are deny-listed](https://github.com/openshift/cluster-monitoring-operator/blob/937046a650f32bdca4666cf6923feb956ed40d82/assets/kube-state-metrics/deployment.yaml#L38-L55), which is the only valid pool of valid metrics that can be denied by the user using this option, whose value will always be a subset of this pool.

rexagod:
@Tai-RedHat PTAL at my comment here: https://github.com/openshift/cluster-monitoring-operator/pull/2283#discussion_r1571542539, which should help provide more context around this patch.

test/e2e/kube_state_metrics_test.go
rexagod: > seems it's valid since it can without regex patterns , or it should be ^kube_lease_owner$?

I noticed earlier today that the KSM deny-list compatibility took a bit of a hit for no reason in this patch, in the sense that the regexes specified by the user were earlier being matched against the payload lines, rather than a 1-1 against the metrics themselves.

So, earlier, the regexes still worked, but boundaries would cause unexpected results: `^foo$` would not match against a payload line that contained a `foo` metric, since the line did not start and end with just `foo`. To resolve this and maintain complete KSM compliance I have pushed a commit so users can specify expressions exactly as they would do in KSM's deny-list.

TL;DR `^kube_lease_owner$` is the correct way to specify an expression that matches against the `kube_lease_owner` metric **only**, although this will error since its not in the default deny-list set.

juzhao: thanks,
`[#2283](https://github.com/openshift/cluster-monitoring-operator/pull/2283) Expose KSM's `--metric-denylist` in `cluster-monitoring-config` ConfigMap, and allow users to enable metrics that are "denied" by default.`
I did not realized the `default` means [default set of metrics that are deny-listed](https://github.com/openshift/cluster-monitoring-operator/blob/937046a650f32bdca4666cf6923feb956ed40d82/assets/kube-state-metrics/deployment.yaml#L38-L55), the valid value for `metricDenylist` should be a subset of it, otherwise it's invalid

Tai-RedHat:
this PR needs to rebase
```
launch 4.16,openshift/cluster-monitoring-operator#2283 gcp
pull request openshift/cluster-monitoring-operator#2283 needs to be rebased to branch master
````

**Comment by Tai-RedHat** [comment]
PR tested with cluster-bot
/label qe-approved

BTW, I found that if I give an error config map like:
```
% oc apply -f -<<EOF
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    kubeStateMetrics:
      metricDenylist:
        - 123
EOF
```
the operator will not degrade, is this expected?  
```
% oc get co/monitoring
NAME         VERSION                                                AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
monitoring   4.16.0-0.test-2024-04-23-015105-ci-ln-fdn20ht-latest   True        False         False      45m
```
I'm asking this because when I give a invalid value such as `^kube_lease_owner$`, the operator will degrated
```
% oc get co/monitoring
NAME         VERSION                                                AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
monitoring   4.16.0-0.test-2024-04-23-015105-ci-ln-fdn20ht-latest   False       True          True       29s     UpdatingPeripherals: verifying kube-state-metrics deny-list bounds failed: cannot deny a metric enabled by default: kube_lease_owner
```

**Comment by juzhao** [comment]
> PR tested with cluster-bot /label qe-approved
> 
> BTW, I found that if I give an error config map like:
> 
> ```
> % oc apply -f -<<EOF
> apiVersion: v1
> kind: ConfigMap
> metadata:
>   name: cluster-monitoring-config
>   namespace: openshift-monitoring
> data:
>   config.yaml: |
>     kubeStateMetrics:
>       metricDenylist:
>         - 123
> EOF

does your cluster have `123` metrics which exposed by KSM? if no, monitoring is not degrade, this is expected

**Comment by rexagod** [comment]
> BTW, I found that if I give an error config map like:
% oc apply -f -<<EOF
apiVersion: v1
kind: ConfigMap
metadata:
data:
  config.yaml: |
    kubeStateMetrics:
      metricDenylist:
        - 123
EOF
the operator will not degrade, is this expected?

Yep, it is, since 123 is not within the set of metrics enabled by default, so its okay to disable that.

> I'm asking this because when I give a invalid value such as ^kube_lease_owner$, the operator will degrated.

Unlike 123, `kube_lease_owner` is a metric that will be enabled by default, so disabling that will cause the operator to become degraded.

File: pkg/client/client.go [hunk 7]
```diff
@@ -612,6 +612,22 @@ func (c *Client) GetSecret(ctx context.Context, namespace, name string) (*v1.Sec
...
+
+	for _, secret := range secrets.Items {
+		if strings.HasPrefix(secret.GetName(), prefix+"-token-") {
+			return &secret, nil
+		}
+	}
+
+	return nil, fmt.Errorf("no secret found with the prefix \"%s-token\" in namespace %s", prefix, namespace)
+}
+
```
jan--f, line 630: This seems unused now.

pkg/manifests/manifests.go
```diff
@@ -759,6 +765,114 @@ func (f *Factory) KubeStateMetricsDeployment() (*appsv1.Deployment, error) {
...
+	// Fetch the `/metrics` port.
+	var p int32
+	for _, port := range service.Spec.Ports {
+		if port.Name == "https-main" {
+			p = port.Port
+			break
+		}
+	}
+
+	// Query the endpoint.
```
jan--f, line 786: Maybe the query part should be refactored into a client method? `KubeStateMetricsDenylistBoundsCheck` is a rather long function and retrieving the metrics endpoint response seems like a good candidate to be split out into its own function?

pkg/manifests/types.go
```diff
@@ -146,6 +146,9 @@ type DedicatedServiceMonitors struct {
 // The `KubeStateMetricsConfig` resource defines settings for the
 // `kube-state-metrics` agent.
 type KubeStateMetricsConfig struct {
+	// Comma-separated list of metrics not to be enabled.
+	// This list comprises exact metric names and/or regex patterns.
+	MetricDenylist []string `json:"metricDenylist,omitempty"`
```
jan--f, line 151: Since this will end up in our official docs and the behavior can be somewhat surprising, I think this doc string should be a bit more extensive.

pkg/operator/operator.go
```diff
@@ -774,6 +774,12 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 				newTaskSpec("ConfigurationSharing", tasks.NewConfigSharingTask(o.client, factory, config)),
 			},
 		),
+		// Peripheral tasks are volatile and run after critical tasks to ensure a stable stack rollout.
+		tasks.NewTaskGroup(
+			[]*tasks.TaskSpec{
+				newTaskSpec("Peripherals", tasks.NewPeripheralsTask(o.client, factory)),
+			},
+		),
```
jan--f, line 782: Not sure we need a new TaskGroup here. Since all tasks in a TaskGroup will run concurrently, the new task can be added in the existing last task group.
Also the new task should be focused in what it does, check the user provided metrics-deny list and apply it. If we get more tasks like that, they will get their own task implementation. How about calling the task something like `UserKSMDenyListTask`?

rexagod: ACK, I had this in a separate task group in order for all peripherals to execute sequentially after the last one, but we don't need to do that since running that concurrently in the shared config task should achieve the same effect. I've made the change and renamed the file.

rexagod:
Cluster bot is taking longer than expected, so I've pushed the commit here to let the CI save us some time. I'll check locally too once I have one.

pkg/manifests/manifests.go
machine424: Thanks for sharing that.
I wasn't suggesting dynamically changing CPs, but using the minimal CP as the source of truth: make `KubeStateMetricsDenylistBoundsCheck` rely on it instead of the `/metrics` output. If we had a minimal CP for KSM, CMO could only accept `kubeStateMetrics.metricDenylist` that doesn't drop metrics from the minimal CP.
Again, I'm just thinking out loud, maybe this needs more discussion later or somewhere else.

> Regexes are supported in the kubeStateMetrics.metricDenylist, if that's what you meant here.

Actually, I was suggesting to pass the user input to KSM as is, without any validation in CMO, like we do for the node-exporter net interfaces I shared above.

> ACK, thank you for clearing this up. I'll send a patch that should establish common ground.

Maybe we can just run that **at the end** of the KSM task (if it's possible)? That way, we're sure it's not blocking anything. This will save us the overhead of adding another task and all the conflicts and races that may occur if the two tasks run concurrently.

pkg/operator/operator.go
```diff
@@ -768,10 +768,13 @@ func (o *Operator) sync(ctx context.Context, key string) error {
...
 				newUWMTaskSpec("ThanosRuler", tasks.NewThanosRulerUserWorkloadTask(o.client, factory, config)),
 			}),
-		// The shared configmap depends on resources being created by the previous tasks hence run it last.
 		tasks.NewTaskGroup(
 			[]*tasks.TaskSpec{
+				// The shared configmap depends on resources being created by the previous tasks.
 				newTaskSpec("ConfigurationSharing", tasks.NewConfigSharingTask(o.client, factory, config)),
+				// NewUserKubeStateMetricsDenyListTask can degrade the operator if an invalid metrics-denylist is
+				// provided by the user. It is run after critical tasks to ensure a stable stack rollout.
+				newTaskSpec("Peripherals", tasks.NewUserKubeStateMetricsDenyListTask(o.client, factory)),
```
jan--f, line 777:
```suggestion
				newTaskSpec("UserKubeStateMetricsDenyList", tasks.NewUserKubeStateMetricsDenyListTask(o.client, factory)),
```

pkg/manifests/types.go
```diff
@@ -146,6 +146,14 @@ type DedicatedServiceMonitors struct {
...
 // `kube-state-metrics` agent.
 type KubeStateMetricsConfig struct {
+	// Comma-separated list of metrics not to be enabled. This list comprises
+	// exact metric names and/or regex patterns. CMO has a default deny-list
+	// that forms the overall scope of the set of metrics that are allowed to
+	// be enabled. However, metrics that are not in the default deny-list
+	// cannot be disabled by the user, since various OpenShift components rely
+	// on them. Doing so will cause the operator to go into a degraded state,
+	// until a valid (or empty) list is provided by the user.
+	MetricDenylist []string `json:"metricDenylist,omitempty"`
```
jan--f, line 156: @eromanova97 @bburt-rh Mind having a look at the doc string proposal here.

jan--f (commented):
One minor nit and a request for docs to check out the api doc string.

pkg/manifests/manifests.go
jan--f: > I wasn't suggesting dynamically changing CPs, but using the minimal CP as the source of truth: make KubeStateMetricsDenylistBoundsCheck rely on it instead of the /metrics output. If we had a minimal CP for KSM, CMO could only accept kubeStateMetrics.metricDenylist that doesn't drop metrics from the minimal CP.
Again, I'm just thinking out loud, maybe this needs more discussion later or somewhere else.

This is definitely worth exploring. I'd suggest through we wait to collection profiles are GA to pursue this.

> Maybe we can just run that at the end of the KSM task (if it's possible)? That way, we're sure it's not blocking anything. This will save us the overhead of adding another task and all the conflicts and races that may occur if the two tasks run concurrently.

This would still mean that the last task group wouldn't run (config sharing task). I think running it in the last task group is the best bang for the buck.

rexagod:
Squashed all `fixup!`s.

rexagod:
Getting a `WARN: cannot link 'github.com/thanos-io/thanos/mixin' to '/Users/rexagod/repositories/work/cluster-monitoring-operator/jsonnet/vendor/mixin', because package 'github.com/etcd-io/etcd/contrib/mixin' already uses that name. The absolute import still works` on `make update`.

Documentation/api.md
```diff
@@ -172,6 +172,7 @@ The `KubeStateMetricsConfig` resource defines settings for the `kube-state-metri
 
 | Property | Type | Description |
 | -------- | ---- | ----------- |
+| metricDenylist | []string | Comma-separated list of metrics not to be enabled. This list comprises exact metric names and/or regex patterns. CMO has a default deny-list that forms the overall scope of the set of metrics that are allowed to be enabled. However, metrics that are not in the default deny-list cannot be disabled by the user, since various OpenShift components rely on them. Doing so will cause the operator to go into a degraded state, until a valid (or empty) list is provided by the user. |
```
eromanova97, line 175:
```suggestion
| metricDenylist | []string | A comma-separated list of metrics that are disabled by default. This list comprises exact metric names and/or regex patterns. You can enable the metrics from the CMO default deny-list. However, you cannot disable metrics that are not in the default deny-list, because various OpenShift components rely on them. Doing so causes the Operator to go into a degraded state, until a valid (or empty) list is provided by the user. |
```

eromanova97 (commented):
Suggesting some small changes to make it a little easier to read.

Documentation/api.md
rexagod: Done, PTAL.

eromanova97: LGTM

Tai-RedHat:
re-test PR with cluster-bot, LGTM

pkg/manifests/manifests.go
rexagod: It seems I missed the updates on this thread.

> I wasn't suggesting dynamically changing CPs, but using the minimal CP as the source of truth: make KubeStateMetricsDenylistBoundsCheck rely on it instead of the /metrics output. If we had a minimal CP for KSM, CMO could only accept kubeStateMetrics.metricDenylist that doesn't drop metrics from the minimal CP.
Again, I'm just thinking out loud, maybe this needs more discussion later or somewhere else.

That is doable if the metric expressions within KSM's minimal service monitor are not regexes themselves (at the moment, these are deterministic values surrounded by `|`, which can be used to statically validate the user-defined list). However,
- if they are, we'll need to either (a) switch to minimal profile every time we want to validate the denylist (to compile the regex, same problem we solve here by querying `/metrics`), or (b) have a complex regex implementation that verifies if a particular regex is a subset of the other, and additionally,
- this will also significantly increase the number of metrics that can be denied by the user in a non-minimal environment, which other components may depend upon (the number of metrics in KSM's `minimal` SM `keep` is much lesser than the metrics not in the default deny-list, giving the user much more control).

I guess this boils down to the fact that making KSM's `minimal` CP the source of truth for all non-`minimal` CP environments can be guaranteed to safeguard all metrics that will ever be relied on by any internal component. Note that while CP may be moving to GA, components will still need time to adapt to the `minimal` environment as there is no such concept being enforced currently (this is exactly what [CPV](https://github.com/rexagod/cpv) aims to help developers with).

> Actually, I was suggesting to pass the user input to KSM as is, without any validation in CMO, like we do for the node-exporter net interfaces I shared above.

By "validation", I'm assuming you mean the default-denylist subset-check that we do here, in which case, not doing so would give the user way too much control and break the "flexibility-stability" bounds this PR rests on. Not sure I follow the idea here, but feel free to correct me in case I misunderstood.

rexagod:
/hold

Until ongoing discussions are done with.

pkg/manifests/manifests.go
rexagod: Friendly ping. @machine424 PLMK if you'd prefer for me to schedule a call to discuss this thread, since reaching a consensus here in time may help us get this in 4.16. I deferred scheduling a call till now as I thought the discussion may end here and not need a call at all, but PLMK if you think otherwise, and if that would help accelerate things. Thanks!

machine424: Thanks Pranshu for your answer.
(I don't think we need to switch to `minimal`, as CMO always knows about the SM manifest as it's in its assets/, we could also use the current profile instead of the minimal one)
As I mentioned, I'm not trying/aiming to block this PR, even though I don’t 100% agree with the approach ;) If the other colleague are okay with it, let's merge it.
Let's resume the profiles related discussion once they're GA or during their transition to GA.

test/e2e/utils.go
```diff
@@ -157,3 +162,28 @@ func getSecurityContextRestrictedProfile() *v1.SecurityContext {
 		},
 	}
 }
+
+func getOrCreateCMOConfig(t *testing.T) (*v1.ConfigMap, error) {
```
machine424, line 166: I didn't have a look at all the recent changes, but this drew my attention.
are the existing `BuildCMOConfigMap` and `MustCreateOrUpdateConfigMap` not sufficient for this?

pkg/manifests/manifests.go
rexagod: > I don't think we need to switch to minimal, as CMO always knows about the SM manifest as it's in its assets/, we could also use the current profile instead of the minimal one.

It's not possible to statically compile regexes (PTAL at my first point [above](https://github.com/openshift/cluster-monitoring-operator/pull/2283#discussion_r1586906524)), or atleast without a lot of computational (and maintenance) complexity.

***

> As I mentioned, I'm not trying/aiming to block this PR, even though I don’t 100% agree with the approach ;) If the other colleague are okay with it, let's merge it.

I'm not sure I see the cause for the incorporation of collection profiles here at the moment. Nonetheless, your review has been constructive and critical to this patch, and not at all blocking IMHO. I'd much rather defer this PR to after 4.16, than rush it right now against your acumen. Besides, we have little to gain from merging this PR in 4.16, as there are no such commitments that would tie us to doing so.

rexagod: I'll setup a call later this week to continue this discussion. That should make it easier to arrive at a consensus. :)

test/e2e/utils.go
rexagod: Right, we can use `BuildCMOConfigMap` to construct the ConfigMap [here](https://github.com/openshift/cluster-monitoring-operator/pull/2283/files#diff-2a8a052d42916d916acab3eed2d52f7039647cc165e3086304d8c4e3e3b800bcR172-R180). 👍🏼

pkg/manifests/manifests.go
rexagod: Not used anymore.

PR #2286, opened 2024-03-13
MON-3792: fix metrics-server path for /test versions

rexagod:
Owing to the older path being referenced in the prow workflow, we saw consistent failure for /test versions job, this fixes that.

***


hack/go/generate_versions.go
```diff
@@ -199,20 +204,50 @@ func getVersion(repo, ref string) (string, error) {
...
+	var (
+		version string
+		err     error
+	)
+	versionChan := make(chan string, 1)
+	errChan := make(chan error, 1)
+	ctx, cancel := context.WithCancel(context.Background())
+
+	for _, path := range paths {
+		go func(path string, cancel context.CancelFunc) {
```
machine424, line 218: No need for concurrency here I think, no problem if it takes an extra second.
I'd run it sequentially to keep the code simple/more readable.
/lgtm
so we can move forward, but I'll
/hold
please don't hesitate to unhold.

rexagod: Removed.

Documentation/deps-versions.md
```diff
@@ -1,10 +0,0 @@
-|  OCP Version |                                   alertmanager                                   |                              kubeRbacProxy                               |                              kubeStateMetrics                               |                              kubernetesMetricsServer                              |                             monitoringPlugin                              |                             nodeExporter                              |                              promLabelProxy                              |                             prometheus                              |                                prometheusAdapter                                |                              prometheusOperator                              |                             thanos                              |
```
machine424, line 1: I'm afraid, it generates an empty table now

machine424:
Thanks
/lgtm

PR #2290, opened 2024-03-20
OCPBUGS-31920: go.mod: bump openshift/api

jan--f:
Changes in openshift/api now require us to
`import _ "github.com/openshift/api/monitoring/v1/zz_generated.crd-manifests" in order to get the generated CRD manifests in vendor/.


jan--f:
I came across this on other business and thought we might as well get ahead of things.
I'm mostly wondering about the import location. Not sure if adding this to the `main` package has any unintended consequences.

PR #2292, opened 2024-03-27
MON-3850: Lint CMO tests

rexagod:
Adds an explicit `vet` target and also integrates it into the `verify` flow to catch any such violations within the source code.

***

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,6 +1,6 @@
 run:
   timeout: 5m
-  tests: false
+  tests: true
   skip-dirs:
     - vendor
     - test
@@ -28,6 +28,31 @@ linters:
     - wastedassign
     - whitespace
     - gci
+
+issues:
+  exclude-rules:
+    - path: _test\.go
+      linters:
+        # The list of linters below is exhaustive with respect to the linters defined above, and as such, and linter
+        # that needs to be run on test files should be dropped from here.
+        - depguard
+        - errcheck
+        - errorlint
+        - bodyclose
+        - exportloopref
+        - gosimple
+        - ineffassign
+        - misspell
+        - nolintlint
+        - nosprintfhostport
+        - staticcheck
+        - tenv
+        - typecheck
+        - unconvert
+        - unused
+        - wastedassign
+        - whitespace
+        - gci
 
 linters-settings:
   depguard:
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -2237,7 +2237,8 @@ func TestPollUntil(t *testing.T) {
 	require.ErrorContains(t, err, "context deadline exceeded")
 
 	// the parent context times out before poll times out.
-	parentCtx1, _ := context.WithTimeout(context.Background(), 10*time.Millisecond)
+	parentCtx1, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
+	defer cancel()
 	err = Poll(parentCtx1, func(ctx context.Context) (bool, error) {
 		// This also ensures that when the parent context is Done, the condition context is cancelled as well.
 		select {


simonpasquier:
Hmm I can't get `go vet ./...` to fail on my machine :-/

rexagod:
Strange, are you on `go1.21`? I can reproduce this in a container if that's better.

simonpasquier (commented):
/lgtm

my bad, I suppose that I didn't revert the fix in pkg/client/client_test.go when checking.

.golangci.test.yaml
```diff
@@ -0,0 +1,11 @@
+run:
```
machine424, line 1: How about expressing this is `.golangci.yaml` itself? I think we can enable linters on tests and then exclude the ones we don't want/want later under [issues](https://golangci-lint.run/usage/configuration/#issues-configuration)

rexagod: My only concern here is that by switching to `issues:` from `run:`, we will potentially lose out on all options that the latter field offers over the former.

rexagod: I've pushed a commit with the proposed changes. However, I'm not sure if there's an "include-only" approach under `issues:` that would allow us to not duplicate the entire list across `run:` and `issues:`. I tried the following with no luck.

```go

issues:
  exclude-rules:
	# Run some linter only for test files by excluding its issues for everything else.    	- path-except: _test\.go
      linters:
       - govet
```

Also, not that we use presets, but `issues:` prevents us from using those as well. I haven't resolved all the lint issues yet, but I'll hop on those once this thread resolves.

machine424: I, personally, have no problem with the disabled linters for tests being enumerated explicitly.
Sure, nothing is set in stone, we can always change the config to adapt to the future needs.
You don't have to fix all the linters errors for tests immediately, you can just start with govet if you want.
Gradually, we can work on reducing that list of "disabled linters for tests", although there might be some that we may never wish to activate for tests...

rexagod: - `tests` is `false`: All `_test.go` lint errors are ignored (eg., `pkg/client/client_test.go`).
- `tests` is `true`: All non-test [linters](https://github.com/openshift/cluster-monitoring-operator/pull/2292/files#diff-9917ddc9f1c3304218f7269265b746d997c5c0615478177b5fceecd33ef47cb5R51) are used for tests, which is not intended.

Unless I'm missing something here, I believe `.golangci.test.yaml` is the only way to go here. WDYT?

machine424: I still think we can do it using the same `.golangci.yaml`
Just like what you did:
- set `tests` to `false`
- disable all the linters except `govet` for test files:
```

Comment (machine424) [reply .golangci.test.yaml]: I've just tested it and I got:
```
pkg/client/client_test.go:2241:13: lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (govet)
	parentCtx, _ := context.WithTimeout(context.Background(), 10*time.Millisecond)
```

File: .golangci.yaml [hunk 2]
```diff
@@ -28,6 +27,7 @@ linters:
     - wastedassign
     - whitespace
     - gci
+    - govet
```
machine424, line 30: `govet` is already enabled in line `17`

.golangci.yaml
```diff
@@ -44,3 +44,28 @@ linters-settings:
       - standard
       - default
       - prefix(github.com/openshift/cluster-monitoring-operator)
+
+issues:
+  exclude-rules:
+    - path: _test\.go
+      linters:
+      # Only enable the following linters for test files (commented out):
```
machine424, line 52: we can remove these comments I think, the list of linters above can be the source of truth.

Makefile
```diff
@@ -195,11 +195,11 @@ go-fmt:
 
 .PHONY: golangci-lint
 golangci-lint: $(GOLANGCI_LINT_BIN)
-	$(GOLANGCI_LINT_BIN) run --verbose --print-resources-usage
+	$(GOLANGCI_LINT_BIN) run -c .golangci.yaml --verbose --print-resources-usage
```
machine424, line 198: is this needed?

rexagod:
I think I pushed pre-maturely here (owing to testing out different configurations). I'll try https://github.com/openshift/cluster-monitoring-operator/pull/2292#discussion_r1621036854 on my local and update the PR.

.golangci.test.yaml
rexagod: I force-pushed the expected set of changes based on your review below. However, skipping the `test` directory was something I intentionally left out in the previous commit (https://github.com/openshift/cluster-monitoring-operator/pull/2292/commits/6b2fafcbc1e345c2773586dbf4d8c29962e9c0ef#diff-99f59e8337d93933947245ba0c7be00c7b35470b0e47f6dd13986f77150775e5R4-R5), in order to encompass non-`_test.go` files in the `test` directory.

```yaml
skip-dirs:
    - vendor
    - tests
```

Including this back in, I see none of the aforementioned `test/!.*_test.go` files are checked by the linter.

Also, I'm sorry to raise this again but I still don't see any point in maintaining two identical lists (since `issues.exclude-rules:` works in an inverse way as we've seen, as opposed to just defining the linters we need) in the linter configuration, as this may very well lead to consistency issues that will go unnoticed (unless we draft a script of sorts to check that, which feels redundant).

PS. I reckon the issue above can be fixed by tweaking the `path:` to `test/` or something similar (although this didn't work for me so far, and still only returns `_test.go` files). Nonetheless, the time we've spent here to add a basic functionality is indicative of the effort someone will need to go through to customize this in the future, if `issues:`'s narrow featureset supports that.

rexagod:
Posting this here explicitly so GitHub doesn't skip past that: https://github.com/openshift/cluster-monitoring-operator/pull/2292#discussion_r1621732073.

.golangci.test.yaml
machine424: I see the `golangci-lint` job is failing on the govet linter for that test file now: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2292/pull-ci-openshift-cluster-monitoring-operator-master-golangci-lint/1796402383471775744

Can you think of any scenarios that might lead to unnoticed consistency issues? If we add a new linter in the future, we'll include it in the first list and run golang-lint. If it fails on tests, we can either fix the tests if we want to keep it enabled or add it to the second list to disable it. If the linter doesn't fail on tests, that's good and means we're checking the tests as well. If the tests change in the future in a way that breaks the linter, we will always have the option to disable the linter for them.

With versioning (git) and reviews, we'll track every change to that file and ensure that the disabling or enabling of a linter for tests is justified.

And we'll also try to `Gradually, we can work on reducing that list of "disabled linters for tests", although there might be some that we may never wish to activate for tests...`

rexagod: > I see the golangci-lint job is failing on the govet linter for that test file now: [prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2292/pull-ci-openshift-cluster-monitoring-operator-master-golangci-lint/1796402383471775744](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2292/pull-ci-openshift-cluster-monitoring-operator-master-golangci-lint/1796402383471775744)

The problem is `test/!.*_test.go` files not being covered by the configuration.

> Can you think of any scenarios that might lead to unnoticed consistency issues? [...] With versioning (git) and reviews, we'll track every change to that file and ensure that the disabling or enabling of a linter for tests is justified.

The committer needs to update two lists now. This is only verified by the manual review process that is subject to human oversight.

> If it fails on tests, we can either fix the tests if we want to keep it enabled or add it to the second list to disable it. If the linter doesn't fail on tests, that's good and means we're checking the tests as well. If the tests change in the future in a way that breaks the linter, we will always have the option to disable the linter for them.

Irrespective of linter pass or fail status, we would've added a linter to the tests without knowing about that. Imagine if a couple of linters have no problem with the tests and got in as a result of human oversight, until they do. At that point, we will be alarmed by the CI and either have to drop them or fix all the lint issues. Pragmatically, neither of these should happen, and the code, as well as its toolchain configuration, should always be in a deterministic state at all times. The CI is already fragile enough for us to include unchecked linters to the list.

machine424: I see but I maintain my previous stance on this:

> If we add a new linter in the future, we'll include it in the first list and run golang-lint. If it fails on tests, we can either fix the tests if we want to keep it enabled or add it to the second list to disable it. If the linter doesn't fail on tests, that's good and means we're checking the tests as well. If the tests change in the future in a way that breaks the linter, we will always have the option to disable the linter for them.

Additionally, as far as I understand, determinism implies that the linter should yield the same results for the same test code. Of course, if the tests change, we should expect the linter results to change as well. This doesn't only apply to tests, the same principle could apply to the rest of the codebase.

Nothing is irreversible or set in stone, we always have the option to make changes in the future. I'd prefer to keep things simple and as similar as possible to what others do. I've never seen multiple `.golangci.yaml` files and I don't believe our use case is different from others.

rexagod: Let me digress a bit. WRT to the configuration being proposed, how do we enable specific linters that are only intended for tests, and not the codebase?

rexagod: `path-except` didn't seem to invert `path`'s behavior in the way we expect, so I've pushed the suggestion before that.

rexagod:
(bump)

machine424:
/lgtm
Thanks for this.

PR #2293, opened 2024-03-28
MON-3749: enable request headers flags for metrics server

slashpai:

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -2050,7 +2050,7 @@ func (f *Factory) MetricsServerRoleBindingAuthReader() (*rbacv1.RoleBinding, err
 	return f.NewRoleBinding(f.assets.MustNewAssetSlice(MetricsServerRoleBindingAuthReader))
 }
 
-func (f *Factory) MetricsServerDeployment(kubeletCABundle *v1.ConfigMap, tlsSecret, metricsClientCert *v1.Secret) (*appsv1.Deployment, error) {
+func (f *Factory) MetricsServerDeployment(kubeletCABundle *v1.ConfigMap, tlsSecret, metricsClientCert *v1.Secret, requestheader map[string]string) (*appsv1.Deployment, error) {
 	dep, err := f.NewDeployment(f.assets.MustNewAssetSlice(MetricsServerDeployment))
 	if err != nil {
 		return nil, err
@@ -2086,6 +2086,26 @@ func (f *Factory) MetricsServerDeployment(kubeletCABundle *v1.ConfigMap, tlsSecr
 	if config == nil {
 		return dep, nil
 	}
+
+	r := newErrMapReader(requestheader)
+
+	var (
+		requestheaderAllowedNames       = strings.Join(r.slice("requestheader-allowed-names"), ",")
+		requestheaderExtraHeadersPrefix = strings.Join(r.slice("requestheader-extra-headers-prefix"), ",")
+		requestheaderGroupHeaders       = strings.Join(r.slice("requestheader-group-headers"), ",")
+		requestheaderUsernameHeaders    = strings.Join(r.slice("requestheader-username-headers"), ",")
+	)
+
+	if r.Error() != nil {
+		return nil, fmt.Errorf("value not found in extension api server authentication configmap: %w", r.err)
+	}
+
+	containers[idx].Args = append(containers[idx].Args,
+		"--requestheader-allowed-names="+requestheaderAllowedNames,
+		"--requestheader-extra-headers-prefix="+requestheaderExtraHeadersPrefix,
+		"--requestheader-group-headers="+requestheaderGroupHeaders,
+		"--requestheader-username-headers="+requestheaderUsernameHeaders,
+	)
 
 	if err := validateAuditProfile(config.Audit.Profile); err != nil {
 		return nil, err
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -2050,7 +2050,7 @@ func (f *Factory) MetricsServerRoleBindingAuthReader() (*rbacv1.RoleBinding, err
 	return f.NewRoleBinding(f.assets.MustNewAssetSlice(MetricsServerRoleBindingAuthReader))
 }
 
-func (f *Factory) MetricsServerDeployment(kubeletCABundle *v1.ConfigMap, tlsSecret, metricsClientCert *v1.Secret, requestheader map[string]string) (*appsv1.Deployment, error) {
+func (f *Factory) MetricsServerDeployment(apiAuthSecretName string, kubeletCABundle *v1.ConfigMap, servingCASecret, metricsClientCert *v1.Secret, requestheader map[string]string) (*appsv1.Deployment, error) {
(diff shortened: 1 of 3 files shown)


slashpai:
will update adding `requestheader-client-ca-file` as well

danielmellado:
Small nit, can we squash (unless squash and merge is enabled in CMO repo) these commits before merging this?

slashpai:
> Small nit, can we squash (unless squash and merge is enabled in CMO repo) these commits before merging this?

ya sure I wanted to verify if updated changes works needed to keep separate commit for easy update. First commit is already verified working. I can squash commit once I test the change locally :)

pkg/manifests/manifests.go
```diff
@@ -2087,6 +2087,28 @@ func (f *Factory) MetricsServerDeployment(kubeletCABundle *v1.ConfigMap, tlsSecr
 		return dep, nil
 	}
 
+	r := newErrMapReader(requestheader)
```
danielmellado, line 2090: from the definition, errMapReader would expose the first occurred error in errMapReader.err. Does this mean that if there are several errors we will just miss others than the first one? just a nit, but wanted to double check.

pkg/manifests/manifests.go
```diff
@@ -2050,7 +2050,7 @@ func (f *Factory) MetricsServerRoleBindingAuthReader() (*rbacv1.RoleBinding, err
 	return f.NewRoleBinding(f.assets.MustNewAssetSlice(MetricsServerRoleBindingAuthReader))
 }
 
-func (f *Factory) MetricsServerDeployment(kubeletCABundle *v1.ConfigMap, tlsSecret, metricsClientCert *v1.Secret) (*appsv1.Deployment, error) {
+func (f *Factory) MetricsServerDeployment(apiAuthSecretName string, kubeletCABundle *v1.ConfigMap, servingCASecret, metricsClientCert *v1.Secret, requestheader map[string]string) (*appsv1.Deployment, error) {
```
danielmellado, line 2053: apiAuthSecretName seems to be unused within the function. Am I missing something?

slashpai: mm ya I had a volume mount for that fixing it :)

slashpai: updated

pkg/manifests/manifests.go
slashpai: I am not sure why doc string is put that way, we use same for adapter as well. 
IIUC it means if first error occurs itself it will stop execution and won't check for more errors?

pkg/manifests/manifests_test.go
```diff
@@ -2814,7 +2814,13 @@ metricsServer:
...
 		},
 	}
-	d, err := f.MetricsServerDeployment(kubeletCABundle, servingCASecret, metricsClientSecret)
+	apiAuthConfigMapData := map[string]string{
+		"requestheader-allowed-names":        "",
+		"requestheader-extra-headers-prefix": "",
+		"requestheader-group-headers":        "",
+		"requestheader-username-headers":     "",
+	}
+	d, err := f.MetricsServerDeployment("foo", kubeletCABundle, servingCASecret, metricsClientSecret, apiAuthConfigMapData)
```
danielmellado, line 2823: if MetricsServerDeployment apiAuthSecretName param gor removed, it should be also removed on the tests.

pkg/manifests/manifests_test.go
```diff
@@ -2969,8 +2975,13 @@ metricsServer:
...
 			}
-
-			d, err := f.MetricsServerDeployment(kubeletCABundle, servingCASecret, metricsClientSecret)
+			apiAuthConfigMapData := map[string]string{
+				"requestheader-allowed-names":        "",
+				"requestheader-extra-headers-prefix": "",
+				"requestheader-group-headers":        "",
+				"requestheader-username-headers":     "",
+			}
+			d, err := f.MetricsServerDeployment("foo", kubeletCABundle, servingCASecret, metricsClientSecret, apiAuthConfigMapData)
```
danielmellado, line 2984: same here ;)

pkg/manifests/manifests_test.go
slashpai: no that was a need a needed parameter, I fixed the code in MetricsServerDeployment() :)

danielmellado: oh, I didn't get to see the fix there, thanks!

slashpai:
/hold
To clarify a query about mount path

pkg/manifests/manifests.go
```diff
@@ -2119,6 +2160,52 @@ func (f *Factory) MetricsServerDeployment(kubeletCABundle *v1.ConfigMap, tlsSecr
...
+			Labels: map[string]string{
+				"monitoring.openshift.io/name": "metrics-server",
+				"monitoring.openshift.io/hash": hash,
+			},
+		},
+		Data: map[string][]byte{
+			"client-ca-file":               []byte(clientCA),
+			"requestheader-client-ca-file": []byte(requestheaderClientCA),
+			"tls.crt":                      []byte(tlsCA),
+			"tls.key":                      []byte(tlsKey),
```
machine424, line 2204: I don't see where we mount/set these two.

pkg/tasks/metricsserver.go
```diff
@@ -155,7 +155,32 @@ func (t *MetricsServerTask) create(ctx context.Context) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to wait for kube-system/extension-apiserver-authentication configmap: %w", err)
+		}
+
+		secret, err := t.factory.MetricsServerSecret(tlsSecret, apiAuthConfigmap)
+		if err != nil {
+			return fmt.Errorf("failed to create metrics-server secret: %w", err)
+		}
+
+		err = t.deleteOldMetricsServerSecrets(secret.Labels["monitoring.openshift.io/hash"])
```
machine424, line 173: Not related to the PR: Couldn't this be replaced by a hash on the Deployment itself as we do for the certs?
So we have only one way of "reloading the Deployment if the secret changes".
(I'd create a ticket for that)
(cc @simonpasquier in case you know that this logic is needed for other things than reloading)

machine424 (commented):
If you agree with my comments JP, I'd be happy to address them myself as the PR is already merged.
Hopefully metrics-server would learn to do this itself: https://github.com/kubernetes-sigs/metrics-server/issues/1432

pkg/tasks/metricsserver.go
slashpai: I wasn't sure either why this was there for prometheus-adapter as well. Happy to move to similar logic we have for other secrets if this is not needed

rexagod: +1 on doing that since we [already use the other logic within MS](https://github.com/openshift/cluster-monitoring-operator/blob/f0dea72dd9e387f0239410a3d17a5fb2ca4a6507/pkg/manifests/manifests.go#L2057-L2069).

pkg/manifests/manifests.go
rexagod: https://github.com/openshift/cluster-monitoring-operator/blob/900ffda2f64b256c739679a90555e896033dca7c/assets/metrics-server/deployment.yaml#L85-L86

PR #2294, opened 2024-04-02
MON-3700: replace OAuth proxy for Thanos Ruler

simonpasquier:


Tai-RedHat:
PR tested with cluster-bot
/label qe-approved

assets/thanos-ruler/service-account.yaml
```diff
@@ -1,8 +1,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  annotations:
-    serviceaccounts.openshift.io/oauth-redirectreference.thanos-ruler: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"thanos-ruler"}}'
+  annotations: {}
```
machine424, line 4: I'm afraid we'll need to do it in two steps https://github.com/openshift/cluster-monitoring-operator/blob/e3ea72fdc49d14443583418002e25b136f69d692/Documentation/development.md?plain=1#L52
(just realized that section is in the wrong place, I'll move, tell me if you have any preference.)
(I didn't test/check this, ignore if you did)

pkg/tasks/thanos_ruler_user_workload.go
```diff
@@ -441,6 +432,11 @@ func (t *ThanosRulerUserWorkloadTask) destroy(ctx context.Context) error {
 		return fmt.Errorf("deleting ThanosRuler object failed: %w", err)
 	}
 
+	err = t.client.DeleteSecret(ctx, acs)
+	if err != nil {
+		return fmt.Errorf("creating Thanos Ruler alertmanager config Secret failed: %w", err)
```
machine424, line 437:
```suggestion
		return fmt.Errorf("deleting Thanos Ruler alertmanager config Secret failed: %w", err)
```

pkg/tasks/thanos_ruler_user_workload.go
```diff
@@ -291,6 +278,35 @@ func (t *ThanosRulerUserWorkloadTask) create(ctx context.Context) error {
 		}
 	}
 
+	// TODO(simonpasquier): remove this step after OCP 4.16 is released.
```
machine424, line 281: If it's not already the case, it'll be great to have a ticket to keep track of thee TODOs.

assets/thanos-ruler/thanos-ruler.yaml
```diff
@@ -76,8 +60,10 @@ spec:
     volumeMounts:
     - mountPath: /etc/tls/private
       name: secret-thanos-ruler-tls
-    - mountPath: /etc/proxy/secrets
-      name: secret-thanos-ruler-oauth-cookie
+      readOnly: true
+    - mountPath: /etc/kube-rbac-proxy
```
machine424, line 64: I don't know if there is a mechanism to "reload" the proxies if such secrets change without the deployments being changed. (or if we need to pair them, it's not as critical as the certs as we'll be making the changes thus we'll test them, but we may overlook things)

pkg/tasks/thanos_ruler_user_workload.go
```diff
@@ -174,22 +177,6 @@ func (t *ThanosRulerUserWorkloadTask) create(ctx context.Context) error {
 	}
 
 	{
-		// Create trusted CA bundle ConfigMap.
-		trustedCA, err := t.factory.ThanosRulerTrustedCABundle()
```
machine424, line 178: I think we can get rid of `ThanosRulerTrustedCABundle()` and its deps.

machine424 (changes requested):
Thanks for the cleaning

assets/thanos-ruler/thanos-ruler.yaml
machine424: (not related to this PR, we can discuss/address this later)

simonpasquier: kube-rbac-proxy should live-reload certificates

pkg/tasks/thanos_ruler_user_workload.go
simonpasquier: I'll create one

assets/thanos-ruler/thanos-ruler.yaml
machine424: yep, I was talking about the proxy config itself.
Whether changing `thanos-user-workload-kube-rbac-proxy-web` in the future requires a manual reloading of the proxy or if we should add a hash or sth as we do for some certs for others components.

assets/thanos-ruler/service-account.yaml
simonpasquier: thanks, updated!

machine424: I think `-` should be a suffix :)
I'll add an example as well. I think`kubectl label` uses the same syntax.

pkg/tasks/thanos_ruler_user_workload.go
simonpasquier: https://issues.redhat.com/browse/MON-3801

assets/thanos-ruler/thanos-ruler.yaml
machine424: NP, I'll create a ticket for that.

machine424:
Thanks.
/lgtm

PR #2298, opened 2024-04-08
OCPBUGS-28650: fix generation of telemeter token hash

simonpasquier:

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -3076,8 +3076,9 @@ func (f *Factory) TelemeterClientDeployment(proxyCABundleCM *v1.ConfigMap, s *v1
 
 	// Set annotation on deployment to trigger redeployments
 	if s != nil {
-		hash := sha256.New()
-		d.Spec.Template.Annotations["telemeter-token-hash"] = string(hash.Sum(s.Data["token"]))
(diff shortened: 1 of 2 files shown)


pkg/manifests/manifests.go
```diff
@@ -3076,8 +3076,9 @@ func (f *Factory) TelemeterClientDeployment(proxyCABundleCM *v1.ConfigMap, s *v1
 
 	// Set annotation on deployment to trigger redeployments
 	if s != nil {
-		hash := sha256.New()
-		d.Spec.Template.Annotations["telemeter-token-hash"] = string(hash.Sum(s.Data["token"]))
+		h := fnv.New64()
```
danielmellado, line 3079: Can we have some quick note about the motivation for the change betweeh sha256 and the fnv hash? Thanks!

simonpasquier: added a description in the commit message.

rexagod (approved):
IIUC even though we could `strconv.ParseUint(<telemeter-token-hash>, 32, 64)` to get back the generated hash (`h.Sum64()`), that shouldn't matter as it'd be different from the actual token hash in `pull-secret` since the hashing algorithm being used here is different.

PR #2299, opened 2024-04-08
MON-3801: remove oauth-redirectreference annotations

simonpasquier:


machine424:
Good catch!
I'll need to check all the other PRs that may have deleted annotations/labels for 4.16 before https://github.com/openshift/cluster-monitoring-operator/pull/2226.

PR #2306, opened 2024-04-09
[release-4.12] OCPBUGS-28766: fix generation of telemeter token hash

simonpasquier:
It also changes the hash algorithm from sha256 to fnv to be consistent with the rest of code and to signal that the use of the function is non-cryptographic.


Tai-RedHat:
PR tested with cluster-bot, telemeter token has been encrypted
/label cherry-pick-approved
/label qe-approved

PR #2307, opened 2024-04-09
MON-3802: implement cross-namespace rules for UWM

simonpasquier:
This change introduces a way to deploy user-defined rules which are not
    scoped to their namespace of origin.
    
To enable the feature, a user-defined monitoring admin needs to
    configure at least one namespace in the UWM ConfigMap:
    
````
    kind: ConfigMap
    apiVersion: v1
    metadata:
    data:
      config.yaml: |-
        namespacesWithoutLabelEnforcement: [ user-monitoring-shared ]
````
    
For all `PrometheusRule` objects defined in the `user-monitoring-shared`
    namespace, Prometheus and Thanos Ruler evaluate the PromQL expressions
    without enforcing the namespace label of origin. It makes it possible to
    have generic rules that get applied to all (or a subset of) the user
    projects instead of having individual rule objects in each user project.
    
 The capability is enabled by default but a cluster admin can decide to
    disable it in the CMO ConfigMap:
    

````
    kind: ConfigMap
    apiVersion: v1
    metadata:
    data:
      config.yaml: |-
        userWorkloadEnabled: true
        userWorkload:
         rulesWithoutLabelEnforcementAllowed: false
````

For example, a user-defined admin can create a single rule that fires
    when a user namespace doesn't enforce the Restricted pod security
    policy.
    
````
    kind: PrometheusRule
    metadata:
      name: security
      namespace: user-monitoring-shared
    spec:
      groups:
        - name: pod-security-policy
          rules:
            - alert: "NamespaceNotEnforcingRestrictedPolicy"
              expr: kube_namespace_labels{namespace!~"(openshift|kube).*|default",label_pod_security_kubernetes_io_enforce!="restricted"}
              for: 5m
              annotations:
                summary: "Restricted policy not enforced"
                description: "Namespace {{ $labels.namespace }} doesn't enforce the Restricted pod security policy."
              labels:
                severity: warning
````

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,39 +112,32 @@ This targets needs `docker` to be installed on host and was not tested with othe
 
 ## Testing
 
-Supposing $KUBECONFIG is set to the config file of a Kubernetes cluster running the codes to test. We can run all tests by using this command `make test`.
+Assuming that the `KUBECONFIG` environment variable is set to the config file of an OpenShift cluster against which the tests will run, you can execute all the tests running `make test`.
 
-The testing consist of 3 aspects:
-- unit tests, can be run separately by `make test-unit`.
-- Prometheus rule tests, can be run separately by `make test-rules`.
-- end-to-end tests, can be run separately by `make test-e2e`.
+Testing entails 3 types of tests:
+- Unit tests which can be run separately with `make test-unit`.
+- Prometheus rule tests which can be run separately with `make test-rules`.
+- End-to-end tests which can be run separately with `make test-e2e`.
 
-If we need to run a specific test case of the E2E test, we can use the following command.
+If you need to run a specific unit test (for example `TestHashSecret`), you can use the following command:
+
 ```bash
-go test -v -timeout=120m ./test/e2e/ --kubeconfig $KUBECONFIG -run TestBodySizeLimit
-```
-Attention that we have to pass a valid $KUBECONFIG explicitly `--kubeconfig $KUBECONFIG` even if ~/.kube/config exists.
-
-To run a specific test case of unit tests, we can use the command `go test -v $PACKAGE_DIR -run $TEST_FUNC_NAME`.
-The `$PACKAGE_DIR` is where the source files of a Go package lives. The `$TEST_FUNC_NAME` is the test function whose name always starts with "Test" (regex pattern `TEST\w+`).
-For example, we have a test function `TestImageParsing` in package `manifests`.The source file `./pkg/manifests/image_test.go` and other source code files of this package lives in `./pkg/manifests`. So we can use the following command to run the test function `TestImageParsing` in package `manifests`.
-```bash
-go test -v ./pkg/manifests -run TestImageParsing
+go test -v ./pkg/... -run TestHashSecret
 ```
 
+If you need to run a specific test case from the E2E test suite (for instance `TestBodySizeLimit`), you can use the following command:
 
-## Coding Style
-
-cluster-monitoring-operator projects written in Go follow a set of style guidelines that we've documented [here](https://github.com/coreos/docs/tree/master/golang).
-Please follow them when working on your contributions.
+```bash
+go test -v -timeout=120m -run TestBodySizeLimit ./test/e2e/ --kubeconfig $KUBECONFIG
+```
+Note: you have to set the `KUBECONFIG` environment variable even if `~/.kube/config` exists.
 
 ## Format of Pull Requests
-We are making heavy use of bots and integrations.
-In order for those to work properly, your Pull Request should match the following structure:
 
-```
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

**Comment by simonpasquier** [comment]
current status:
* Need to fix the e2e tests when cross-namespace rules are disabled.
* I found an issue with the dev console: the alerts generated from cross-namespace rules aren't visible because the console queries for the rules with the exact namespace value.

**Comment by jan--f** [comment]
As mentioned in the sync earlier, I would like to make sure we make it very clear to the user what we mean by `cross-namespace`. Iiuc this means user namespace as well as system namespaces. Since this currently talks about cross namespace in the context of UWM, I think we should make it explicit that system namespaces can be queried as well for  such rules (if that is actually the case).
I'm sure @bburt-rh would know best how to phrase that.

**Comment by Tai-RedHat** [comment]
PR tested with cluster-bot, other user-defined namespace could trigger the alert in `user-monitoring-shared` namespace.
test case: https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-75384
/label qe-approved

**Comment by Tai-RedHat** [comment]
If I set two or even more user-monitoring-shared namespaces, and then create same prometheusrules to each namespaces. This will result in repeated alerts. Should we remind or restrict users in this situation?
`config.yaml: 'namespacesWithoutLabelEnforcement: [ns1, ns2]'`

Screenshot 2024-08-13 at 12 55 49

**Comment by machine424** [comment]
I've just realized this is assigned to me, I'll take a look.

File: pkg/manifests/types.go [hunk 1]
```diff
@@ -532,6 +547,19 @@ type UserWorkloadConfiguration struct {
 	// Defines the settings for the Thanos Ruler component in user workload
 	// monitoring.
 	ThanosRuler *ThanosRulerConfig `json:"thanosRuler,omitempty"`
+
+	// Defines the list of namespaces for which Prometheus and Thanos Ruler in
+	// user-defined monitoring don't enforce the `namespace` label value in
+	// `PrometheusRule` objects.
+	//
+	// It allows to define recording and alerting rules that can query across
+	// multiple projects instead of deploying identical `PrometheusRule`
```
machine424, line 556:
```suggestion
	// multiple projects (including default projects) instead of deploying identical `PrometheusRule`
```

openshift-xxx etc. I think that's how we call them.

machine424: As Jan mentioned, maybe we should be more explicit in the docs about the fact that enabling this breaks/bypasses the tenancy model as any user that can read metrics for a namespace in NamespacesWithoutLabelEnforcement and write PrometheusRules in that namespace, can now get access to any metric in any namespace.

(maybe it's "safer" to have RulesWithoutLabelEnforcementAllowed disabled by default.)

CONTRIBUTING.md
```diff
@@ -112,39 +112,32 @@ This targets needs `docker` to be installed on host and was not tested with othe
 
 ## Testing
 
-Supposing $KUBECONFIG is set to the config file of a Kubernetes cluster running the codes to test. We can run all tests by using this command `make test`.
+Assuming that the `KUBECONFIG` environment variable is set to the config file of an OpenShift cluster against which the tests will run, you can execute all the tests running `make test`.
```
machine424, line 115: Thanks for this.

machine424 (commented):
Glad to see this taking shape, well done!
I just have some prepositions.

pkg/manifests/types.go
simonpasquier: > any user that can read metrics for a namespace in NamespacesWithoutLabelEnforcement and write PrometheusRules in that namespace, can now get access to any metric in any namespace.

hmm not quite possible IIUC. But I agree that it should be clear from the docs that you don't want regular users to have PrometheusRule write access to the `NamespacesWithoutLabelEnforcement` namespace(s).

simonpasquier: sure but from my POV, it's by design: when you trust a user/service account to create non-namespaced rules, you're effectively providing them access to all metrics.

simonpasquier:
From https://github.com/openshift/cluster-monitoring-operator/pull/2307#discussion_r1734441777

> (maybe it's "safer" to have RulesWithoutLabelEnforcementAllowed disabled by default.)

My initial intention was to avoid friction in adopting this feature but I'm also ok making it opt-in as it's less surprising for platform admins. We can also keep it opt-in for a few releases and then turn it on by default.

@jan--f WDYT?

pkg/manifests/types.go
machine424: (I agree with you, just suggested that to be more explicit in the doc)

machine424: how about the suggested addition of `(including default projects)`? , I'd even go with `can query across ALL projects`

machine424:
/lgtm
You'll make many users happy with this.

PR #2308, opened 2024-04-10
MON-3701: remove references to OAuth proxy

simonpasquier:

### DIFF
--- a/test/e2e/alertmanager_test.go
+++ b/test/e2e/alertmanager_test.go
@@ -482,8 +482,8 @@ func TestAlertmanagerDataReplication(t *testing.T) {
 	}
 }
 
-// The Alertmanager API should be protected by the OAuth proxy.
-func TestAlertmanagerOAuthProxy(t *testing.T) {
+// The Alertmanager API should be protected by authentication/authorization.
+func TestAlertmanagerAPI(t *testing.T) {
 	err := framework.Poll(5*time.Second, 5*time.Minute, func() error {
 		body, err := f.AlertmanagerClient.GetAlertmanagerAlerts(
 			"filter", `alertname="Watchdog"`,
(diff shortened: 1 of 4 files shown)


marioferh (commented):
/lgtm

test/e2e/alertmanager_test.go
```diff
@@ -482,8 +482,8 @@ func TestAlertmanagerDataReplication(t *testing.T) {
 	}
 }
 
-// The Alertmanager API should be protected by the OAuth proxy.
-func TestAlertmanagerOAuthProxy(t *testing.T) {
+// The Alertmanager API should be protected by authentication/authorization.
+func TestAlertmanagerAPI(t *testing.T) {
```
machine424, line 486: I don't see how the test is checking the auth

machine424 (commented):
Thanks.
I see some other references that we can get rid of as well:
- https://github.com/openshift/cluster-monitoring-operator/blob/7f498b49343756681f34d0d75013f193cfdc8a66/jsonnet/components/prometheus-user-workload.libsonnet#L57
- https://github.com/openshift/cluster-monitoring-operator/blob/7f498b49343756681f34d0d75013f193cfdc8a66/examples/user-guides/configuring-cluster-monitoring/custom-image-config.yaml#L10 (I'm not even sure if we still consume that)
- https://github.com/openshift/cluster-monitoring-operator/blob/7f498b49343756681f34d0d75013f193cfdc8a66/test/e2e/thanos_querier_test.go#L75 (I don't see what that test is checking, maybe we should improve the utils)

test/e2e/alertmanager_test.go
machine424: (not related to the PR)

machine424: cc @simonpasquier if you missed this (I forgot o hold)
I could take care of it if you want.

simonpasquier:
@machine424 

> I see some other references that we can get rid of as well:

Yes I plan to submit separate PRs to smooth the reviews and CI debugging.

> I don't see how the test is checking the auth

It's not very explicit but the test SA has the necessary RBAC to query the API. Testing that unauthorized users are rejected would be good indeed (we have it for tenancy services but for not admin-like services).

PR #2310, opened 2024-04-10
MON-3701: clean-up injection of trusted CA bundle for main Alertmanager

simonpasquier:

### DIFF
--- a/pkg/tasks/alertmanager.go
+++ b/pkg/tasks/alertmanager.go
@@ -185,17 +185,12 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 			return fmt.Errorf("initializing Alertmanager CA bundle ConfigMap failed: %w", err)
 		}
 
-		cbs := &caBundleSyncer{
-			client:  t.client,
-			factory: t.factory,
-			prefix:  "alertmanager",
-		}
-		trustedCA, err = cbs.syncTrustedCABundle(ctx, trustedCA)
+		err = t.client.CreateOrUpdateConfigMap(ctx, trustedCA)
 		if err != nil {
-			return fmt.Errorf("syncing Alertmanager trusted CA bundle ConfigMap failed: %w", err)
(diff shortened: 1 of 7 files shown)


simonpasquier:
/hold
want to test a small variation

simonpasquier (commented):
/hold cancel

pkg/tasks/alertmanager.go
```diff
@@ -185,17 +185,12 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 			return fmt.Errorf("initializing Alertmanager CA bundle ConfigMap failed: %w", err)
 		}
 
-		cbs := &caBundleSyncer{
-			client:  t.client,
-			factory: t.factory,
-			prefix:  "alertmanager",
-		}
-		trustedCA, err = cbs.syncTrustedCABundle(ctx, trustedCA)
```
machine424, line 193: Cool that we're simplifying code like this.
I had a question about a similar practice (generating secrets with hashes...) here https://github.com/openshift/cluster-monitoring-operator/pull/2293#discussion_r1557101087
and it'd be great if we can get rid of the logic there as well and only use the "secret/config hash on Deployment's template" approach as simpler.

---

I have a question about the root CA hot-reloading though, I know KRP supports that, but I think it only concerns `--tls-cert-file` and `--tls-private-key-file`, I think root CA from `/etc/pki/ca-trust/extracted/pem/` etc. are managed by Go itself (in the case of KRP) and according to https://github.com/golang/go/issues/41888 they're not reloaded.

simonpasquier: kube-rbac-proxy doesn't rely on the cluster CA bundle. Here is the CA bundle is mounted into the Alertmanager container because it may be configured to send notifications to 3rd party services that use certificates generated by the cluster CA.

machine424: Ah, ok. I don't know if alertmanager supports hot-reloading then.

machine424:
/lgtm
please unhold when it's ok for you.

PR #2313, opened 2024-04-10
OCPBUGS-32055: backport ols metric allowlist to 4.15

bparees:


bparees:
manually labeling to valid bug as there is/was/will be no 4.16+ bug since the fix merged to main prior to branching for those releases.  (and this isn't really a bug)

juzhao:
@bparees I am afraid you also need to add these metrics to teletmer-server side, did not see these metrics in
https://github.com/rhobs/configuration/blob/main/configuration/telemeter/metrics.json
then you won't be able to query these metrics from telemeter server

bparees:
> @bparees I am afraid you also need to add these metrics to teletmer-server side, did not see these metrics in
https://github.com/rhobs/configuration/blob/main/configuration/telemeter/metrics.json
then you won't be able to query these metrics from telemeter server

thanks i'll do that next but i don't think it precludes merging this backport, does it?

juzhao:
> thanks i'll do that next but i don't think it precludes merging this backport, does it?
no, it won't prevent merging this PR, only prevent searching these metrics from telemeter server

bparees:
@simonpasquier @jan--f ptal

PR #2315, opened 2024-04-11
MON-3814: Follow-up #2293

slashpai:
Follow-up from https://github.com/openshift/cluster-monitoring-operator/pull/2293

### DIFF
--- a/assets/metrics-server/deployment.yaml
+++ b/assets/metrics-server/deployment.yaml
@@ -52,6 +52,7 @@ spec:
         - --tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
         - --shutdown-send-retry-after=true
         - --shutdown-delay-duration=150s
+        - --disable-http2-serving=true
         image: registry.k8s.io/metrics-server/metrics-server:v0.7.2
         imagePullPolicy: IfNotPresent
         livenessProbe:
--- a/jsonnet/components/metrics-server.libsonnet
+++ b/jsonnet/components/metrics-server.libsonnet
@@ -202,6 +202,7 @@ function(params) {
                 '--shutdown-send-retry-after=true',
                 // wait long enough for the readiness probe's failure threshold to be breached
                 '--shutdown-delay-duration=150s',
+                '--disable-http2-serving=true',
               ],
               imagePullPolicy: 'IfNotPresent',
               livenessProbe: {


pkg/manifests/manifests.go
```diff
@@ -2106,6 +2106,8 @@ func (f *Factory) MetricsServerDeployment(apiAuthSecretName string, kubeletCABun
 		"--requestheader-extra-headers-prefix="+requestheaderExtraHeadersPrefix,
 		"--requestheader-group-headers="+requestheaderGroupHeaders,
 		"--requestheader-username-headers="+requestheaderUsernameHeaders,
+		"--tls-cert-file=/etc/client-ca-bundle/tls.crt",
```
machine424, line 2109: No test was failing when using `/etc/tls/private/xxx`, do they have the same content? Is testing this difficult?

slashpai: no failure I just moved it from jsonnet to here 
https://github.com/openshift/cluster-monitoring-operator/pull/2293/files#r1557090512

slashpai: https://github.com/openshift/cluster-monitoring-operator/blob/889bb2ce19cd1018888f506dd8c401cca163cfc8/jsonnet/components/metrics-server.libsonnet#L120 (metrics-server-tls) is been mounted at /etc/tls/private (https://github.com/openshift/cluster-monitoring-operator/blob/889bb2ce19cd1018888f506dd8c401cca163cfc8/jsonnet/components/metrics-server.libsonnet#L248)

In the updated approach as well it uses the same secret https://github.com/openshift/cluster-monitoring-operator/pull/2293/files#diff-0bbbc0b6e7ee4c4efe218030b64c2ef604fda58d92e51e52c206b6c5c0a4a5a3R158

slashpai: I updated the PR to remove unused in volume mounts https://github.com/openshift/cluster-monitoring-operator/pull/2293/files#r1557090512

pkg/manifests/manifests.go
```diff
@@ -2162,16 +2162,14 @@ func (f *Factory) MetricsServerSecret(tlsSecret *v1.Secret, apiAuthConfigmap *v1
 	var (
 		clientCA              = r.value("client-ca-file")
 		requestheaderClientCA = r.value("requestheader-client-ca-file")
-		tlsCA                 = r.value("tls.crt")
-		tlsKey                = r.value("tls.key")
 	)
```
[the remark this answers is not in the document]
rexagod: True, and I believe the following can be dropped:

```go

		tlsSecret, err := t.client.WaitForSecretByNsName(ctx, types.NamespacedName{Namespace: t.namespace, Name: "metrics-server-tls"})
		if err != nil {
			return fmt.Errorf("failed to wait for metrics-server-tls secret: %w", err)
		}

```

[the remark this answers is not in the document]
machine424: (i just realized we didin't need that wait. as we already waited for that secret here https://github.com/openshift/cluster-monitoring-operator/blob/65ee64665d9553c70f9cbfae5c2794d327c4fe32/pkg/tasks/metricsserver.go#L122-L128)

slashpai:
Closing in favour of https://github.com/openshift/cluster-monitoring-operator/pull/2329

machine424:
I'll
/reopen
this, it'll be nice to have this cleaning in 4.17

machine424:
I think we can merge this once `tlsSecret` is dropped as an argument to `MetricsServerSecret`.

PR #2316, opened 2024-04-11
OCPBUGS-31411: add runbook link for KubeAggregatedAPIErrors alert

slashpai:


juzhao:
@slashpai see comment in OCPBUGS-31411, @jan--f recommends to delete https://github.com/openshift/runbooks/blob/master/alerts/AggregatedAPIErrors.md, do we have a PR for it?

PR #2317, opened 2024-04-11
WIP DO NOT MERGE [only testing]: Test Metrics Server Feature Gate Default

slashpai:
/hold

This PR is only meant to test if switching Metrics Server to default breaks test
original PR to update api is https://github.com/openshift/api/pull/1851. For testing purpose I replaced to my fork

may be this will also need to update other libs (CCO) I am not sure but creating a WIP PR to see

cc: @simonpasquier


slashpai:
I think we need change to cluster config operator as well to get the change since it's still not turned to default even after bumping api to fork

PR #2323, opened 2024-04-16
MON-3701: clean-up injection of trusted CA bundle for k8s Prometheus

simonpasquier:

### DIFF
--- a/jsonnet/components/prometheus.libsonnet
+++ b/jsonnet/components/prometheus.libsonnet
@@ -565,6 +565,24 @@ function(params)
           },
           {
             name: 'prometheus',
+            volumeMounts+: [
+              {
+                name: $.trustedCaBundle.metadata.name,
+                mountPath: '/etc/pki/ca-trust/extracted/pem/',
+              },
+            ],
+          },
+        ],
+        volumes+: [
+          {
+            name: $.trustedCaBundle.metadata.name,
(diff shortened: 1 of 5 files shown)


jsonnet/components/prometheus.libsonnet
```diff
@@ -565,6 +565,24 @@ function(params)
...
+          },
+        ],
+        volumes+: [
+          {
+            name: $.trustedCaBundle.metadata.name,
+            configMap: {
+              name: $.trustedCaBundle.metadata.name,
+              items: [{
+                key: 'ca-bundle.crt',
+                path: 'tls-ca-bundle.pem',
```
rexagod, line 583: (suggestion) It might be worth putting [`src/crypto/x509/root_linux.go#L15`](https://cs.opensource.google/go/go/+/master:src/crypto/x509/root_linux.go;l=15?q=%2Fpki&ss=go%2Fgo) in a comment here, but nothing that would block this PR.

rexagod (approved):

jsonnet/components/prometheus.libsonnet
simonpasquier: I'd rather do it once for all components if we feel that it's important.

simonpasquier:
@rexagod can you `/lgtm` explicitly please?

PR #2326, opened 2024-04-19
MON-3707: Add ipsec state metric into telemetry

JoshSalomon:
Add the metric openshift:openshift_network_operator_ipsec_state:sum to telemetry
This metric captures ipsec state of the cluster (Disabled, External or Full) and whether the state was set by the legacy API (OCP 4.14 or before) or the new API (OCP 4.15+)


simonpasquier (commented):
Looking at the Prometheus dump from the upgrade job, I see a bump in the metric value

image

IIUC the goal of the metric is to report 1 for any active combination of mode/legacy API. The following expression would be more suited:

````
group by(mode,is_legacy_api) (openshift_network_operator_ipsec_state{namespace=~"openshift-network-operator"})
````

image

Finally the metric name could be improved IMHO: `openshift:openshift_network_operator_ipsec_state:sum` stutters, I'd suggest `cluster:openshift_network_operator_ipsec_state:info`.

JoshSalomon:
> 
> IIUC the goal of the metric is to report 1 for any active combination of mode/legacy API. The following expression would be more suited:
> 
> ```
> group by(mode,is_legacy_api) (openshift_network_operator_ipsec_state{namespace=~"openshift-network-operator"})
> ```
> 
At any point in time, there is only one active combination (the code resets the gauge and then calls `ipsecStateGauge.WithLabelValues` once, so I don't think there is a big difference here.

> 
> Finally the metric name could be improved IMHO: `openshift:openshift_network_operator_ipsec_state:sum` stutters, I'd suggest `cluster:openshift_network_operator_ipsec_state:info`.

I agree, (with both comments) - but is it critical, the CNO PR took a long time to approve, and we need this in 4.16 - are these comments really critical to put this feature in risk for 4.16?

rexagod:
> At any point in time, there is only one active combination (the code resets the gauge and then calls `ipsecStateGauge.WithLabelValues` once), so I don't think there is a big difference here.

If the underlying code changes, a `group by` will ensure the value safely remains a boolean, and won't be prone to going out of bounds.

Additionally, wrt the [source](https://github.com/openshift/cluster-network-operator/blob/a5909c5af10683639fbbdbc822bc0b13c98520d5/bindata/cluster-network-operator/prometheus.yaml#L16):

- https://github.com/openshift/cluster-network-operator/blob/a5909c5af10683639fbbdbc822bc0b13c98520d5/bindata/cluster-network-operator/prometheus.yaml#L16: I see a `=~`, are multiple namespaces expected?

- https://github.com/openshift/cluster-network-operator/blob/a5909c5af10683639fbbdbc822bc0b13c98520d5/bindata/cluster-network-operator/prometheus.yaml#L18: `s/sum/info`, otherwise this won't work.

rexagod:
PS. The tests are failing due to [internal changes](https://github.com/openshift/cluster-monitoring-operator/pull/2153#issuecomment-2068769453), and should be fixed soon.

zshi-redhat:
> > At any point in time, there is only one active combination (the code resets the gauge and then calls `ipsecStateGauge.WithLabelValues` once), so I don't think there is a big difference here.
> 
> If the underlying code changes, a `group by` will ensure the value safely remains a boolean, and won't be prone to going out of bounds.
> 
> This is obvious but just to be explicit, you'll also need to [`s/sum/info`](https://github.com/openshift/cluster-network-operator/blob/a5909c5af10683639fbbdbc822bc0b13c98520d5/bindata/cluster-network-operator/prometheus.yaml#L18) in the source, otherwise this won't work. I'm not sure if there's a PR up for that in CNO.

PR is up in CNO: https://github.com/openshift/cluster-network-operator/pull/2346
it got lgtmed by networking team. @rexagod  would you mind double check?

JoshSalomon:
the SDN PR https://github.com/openshift/cluster-network-operator/pull/2346 was merged

PR #2328, opened 2024-04-20
MON-3839: test: add skip tests for prometheus adapter tests

slashpai:

### DIFF
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -488,6 +488,7 @@ func TestTelemeterClientSecret(t *testing.T) {
 }
 
 func TestClusterMonitorK8sPromAdapterConfig(t *testing.T) {
+	skipPrometheusAdapterTests(t)
 	const (
 		deploymentName = "prometheus-adapter"
 	)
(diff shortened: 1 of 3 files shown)


slashpai:
ok I need to remove prometheus-adapter from metrics map when feature gate is enabled

```
prometheus_test.go:53: context deadline exceeded: error getting first value from response body "{\"status\":\"success\",\"data\":{\"resultType\":\"vector\",\"result\":[],\"analysis\":{}}}\n" for query "count(up{service=\"prometheus-adapter\",namespace=\"openshift-monitoring\"} == 1)": expected body to contain single timeseries but got 0
--- FAIL: TestPrometheusMetrics (645.17s)
    --- PASS: TestPrometheusMetrics/metrics-server (5.01s)
    --- FAIL: TestPrometheusMetrics/prometheus-adapter (600.02s)
```

jan--f:
Ahh right, `WaitForQueryReturn` can't deal with no result.

slashpai:
@jan--f please review again updated

PR #2330, opened 2024-04-23
NO-JIRA: Increase `checkAlertmanagerAPIVerbs` timeout

rexagod:


jan--f:
Next time, feel free to raise a bug for things like this.

PR #2332, opened 2024-04-25
OBSDOCS-883: Fix the wrong retention time comment for user-defined project monitoring

eromanova97:
Issue: [OBSDOCS-883](https://issues.redhat.com/browse/OBSDOCS-883)

This is to fix the wrong comment about retention time for user-defined project monitoring.


machine424:
note that we already have a similar PR: https://github.com/openshift/cluster-monitoring-operator/pull/2277

eromanova97:
> note that we already have a similar PR: #2277

Thank you for letting me know, I missed that one. I will close this PR then.

PR #2333, opened 2024-04-27
MON-3846: create separate kubelet-serving-ca cm for metrics-server

slashpai:
This is to avoid similar situation like https://issues.redhat.com/browse/OCPBUGS-32510 so that metrics-server
doesn't need to wait for prometheus task to create the configmap

cc: @machine424 @simonpasquier @jan--f

### DIFF
--- a/pkg/tasks/metricsserver.go
+++ b/pkg/tasks/metricsserver.go
@@ -106,17 +106,19 @@ func (t *MetricsServerTask) create(ctx context.Context) error {
 			return fmt.Errorf("reconciling MetricsServer Service failed: %w", err)
 		}
 
-		cacm, err := t.factory.PrometheusK8sKubeletServingCABundle(map[string]string{})
+		kscm, err := t.client.GetConfigmap(ctx, "openshift-config-managed", "kubelet-serving-ca")
+		if err != nil {
+			return fmt.Errorf("openshift-config-managed/kubelet-serving-ca: %w", err)
+		}
+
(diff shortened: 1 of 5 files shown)


slashpai:
cc @machine424

slashpai:
/hold
let's revisit this later after temporary fix #2334

slashpai:
@machine424 what do you think about proceeding with this or do we want to wait?

pkg/tasks/metricsserver.go
```diff
@@ -106,17 +106,19 @@ func (t *MetricsServerTask) create(ctx context.Context) error {
 			return fmt.Errorf("reconciling MetricsServer Service failed: %w", err)
 		}
 
-		cacm, err := t.factory.PrometheusK8sKubeletServingCABundle(map[string]string{})
+		kscm, err := t.client.GetConfigmap(ctx, "openshift-config-managed", "kubelet-serving-ca")
```
machine424, line 109: Because the bundle isn't nominative (like the client certificate), the optimal setup would be to have a unique configmap created by a task from the first stage (like `NewMetricsClientCATask` or something similar).
However, for now, let's just add some comments here + around https://github.com/openshift/cluster-monitoring-operator/blob/a83f250fbf7efd2863a6faab471242322e563796/pkg/tasks/prometheus.go#L70 and around the two jsonnet resources, warning about the "duplication", to help avoid divergence between the two setups.

PR #2335, opened 2024-05-02
AUTH-482: set required-scc for openshift workloads

liouk:

This PR sets the required SCC explicitly on each workload in `openshift-*` namespaces. The SCC chosen is the one that the pods are getting admitted with, so no change expected there. This is to protect the pods from getting admitted with a different custom SCC than the one intended.

This PR
- updates `manifests/`
- updates any workloads defined in `jsonnet/`
- generates new manifests in `assets/` based on jsonnet changes


simonpasquier (commented):
/lgtm

PR #2341, opened 2024-05-08
Remove job:kube_pv:count

weirdwiz:
This commit removes job:kube_pv:count from telemetry as it's been deprecated and no longer contains any value


machine424:
Thanks for this.
Do we have a guarantee that no `<4.12` cluster is sending this metric anymore?
IIUC, there is no versioning on telemeter server side: https://github.com/rhobs/configuration/blob/d2353fb83f94d23a173f56b93a5329e5f85c9e13/Makefile#L110-L115 (server's allow list uses CMO's maqter)
If we remove this metric here, we may prevent those old clusters from sending it.
I think, if it's just a clean-up change, it's safer to keep things as they are.
Maybe we could mention that next to these "deprecated" comments.

Of course, if you DO want this metric to disappear, that's another discussion.
cc @jan--f @simonpasquier

simonpasquier:
+1 to what Ayoub said. Dropping the metric from this list means that it will be rejected on the telemeter server side. It might be better to document the metric as being deprecated and remove it once it's not available in any supported OCP version.
Anyway thanks for the clean-up!

PR #2347, opened 2024-05-14
MON-3867: generate CRD's and api

marioferh:
First PR to implement CRD Based CMO:

Add controller-gen, manifests and generate-crd to makefile
Add pkg/apis/cmo/v1
Add Generate CRD
Add example CustomResource


config/examples/clustermonitoringoperator.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: cmo.example.com/v1
```
marioferh, line 1: I need the correct name for this, some ideas?

juzhao: how about `monitoring.coreos.com/v1`

```
$  oc explain Prometheus
GROUP:      monitoring.coreos.com
KIND:       Prometheus
VERSION:    v1
...
```

marioferh: thanks, I don't know if correct group is that or cmo should go in another one

marioferh:
moved to https://github.com/openshift/api/

PR #2350, opened 2024-05-15
WIP: MON-3868: Crd based watch events

marioferh:


marioferh:
change to go 1.22, otherwise code-generator does not work. We can wait until bump to go 1.22 or try to fix code generator with 1.21

PR #1289, opened 2021-07-19
Bug 1982778: jsonnet: thanosquery: Use HTTP probes as opposed to exec

PhilipGough:
This is cherry pick of #1277 which backports to 4.8.z

* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

PR #2352, opened 2024-05-16
OCPBUGS-33620: define *probes for KSM

rexagod:
Will set a target version post-4.16 branching.

### DIFF
--- a/assets/kube-state-metrics/deployment.yaml
+++ b/assets/kube-state-metrics/deployment.yaml
@@ -59,7 +59,7 @@ spec:
         image: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.13.0
         livenessProbe:
           httpGet:
-            path: livez
+            path: /livez
             port: main
             scheme: HTTPS
         name: kube-state-metrics
@@ -70,7 +70,7 @@ spec:
           name: self
         readinessProbe:
           httpGet:
-            path: metrics
+            path: /readyz
             port: self
             scheme: HTTPS
         resources:
@@ -124,7 +124,7 @@ spec:
         - --tls-private-key-file=/etc/tls/private/tls.key
         - --client-ca-file=/etc/tls/client/client-ca.crt
         - --config-file=/etc/kube-rbac-policy/config.yaml
-        - --ignore-paths=/metrics
+        - --ignore-paths=/readyz
         image: quay.io/brancz/kube-rbac-proxy:v0.17.1
         name: kube-rbac-proxy-self
         ports:
--- a/assets/kube-state-metrics/deployment.yaml
+++ b/assets/kube-state-metrics/deployment.yaml
@@ -57,7 +57,22 @@ spec:
           ^kube_pod_completion_time$,
           ^kube_pod_status_scheduled$
         image: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.13.0
+        livenessProbe:
+          httpGet:
+            path: livez
+            port: main
+            scheme: HTTPS
         name: kube-state-metrics
+        ports:
+        - containerPort: 8443
+          name: main
+        - containerPort: 9443
+          name: self
+        readinessProbe:
+          httpGet:
+            path: metrics
+            port: self
+            scheme: HTTPS
         resources:
           requests:
             cpu: 2m
@@ -79,6 +94,7 @@ spec:
         - --tls-private-key-file=/etc/tls/private/tls.key
         - --client-ca-file=/etc/tls/client/client-ca.crt
         - --config-file=/etc/kube-rbac-policy/config.yaml
+        - --ignore-paths=/livez
         image: quay.io/brancz/kube-rbac-proxy:v0.17.1
         name: kube-rbac-proxy-main
         ports:
@@ -108,6 +124,7 @@ spec:
         - --tls-private-key-file=/etc/tls/private/tls.key
         - --client-ca-file=/etc/tls/client/client-ca.crt
         - --config-file=/etc/kube-rbac-policy/config.yaml
+        - --ignore-paths=/metrics
         image: quay.io/brancz/kube-rbac-proxy:v0.17.1
         name: kube-rbac-proxy-self
         ports:
--- a/jsonnet/components/kube-state-metrics.libsonnet
+++ b/jsonnet/components/kube-state-metrics.libsonnet
@@ -208,6 +208,7 @@ function(params)
                         '--tls-private-key-file=/etc/tls/private/tls.key',
                         '--client-ca-file=/etc/tls/client/client-ca.crt',
                         '--config-file=/etc/kube-rbac-policy/config.yaml',
+                        '--ignore-paths=' + std.join(',', if std.endsWith(c.name, '-self') then ['/metrics'] else ['/livez']),
                       ],
                       volumeMounts: [
                         {
@@ -266,6 +267,36 @@ function(params)
                           readOnly: true,
                         },
                       ],
+                      local mainPort = 8443,
+                      local mainPortName = 'main',
+                      local selfPort = 9443,
+                      local selfPortName = 'self',
+                      ports::: [
+                        {
+                          containerPort: mainPort,
+                          name: mainPortName,
+                        },
+                        {
+                          containerPort: selfPort,
+                          name: selfPortName,
+                        },
+                      ],
+                      local livenessProbePath = 'livez',
+                      local readinessProbePath = 'metrics',
+                      livenessProbe::: {
+                        httpGet: {
+                          path: livenessProbePath,
+                          port: mainPortName,
+                          scheme: 'HTTPS',
+                        },
+                      },
+                      readinessProbe::: {
+                        httpGet: {
+                          path: readinessProbePath,
+                          port: selfPortName,
+                          scheme: 'HTTPS',
+                        },
+                      },
                     },
                 super.containers,
               ),
--- a/jsonnet/components/kube-state-metrics.libsonnet
+++ b/jsonnet/components/kube-state-metrics.libsonnet
@@ -208,7 +208,7 @@ function(params)
                         '--tls-private-key-file=/etc/tls/private/tls.key',
                         '--client-ca-file=/etc/tls/client/client-ca.crt',
                         '--config-file=/etc/kube-rbac-policy/config.yaml',
-                        '--ignore-paths=' + std.join(',', if std.endsWith(c.name, '-self') then ['/metrics'] else ['/livez']),
+                        '--ignore-paths=' + std.join(',', if std.endsWith(c.name, '-self') then ['/readyz'] else ['/livez']),
                       ],
                       volumeMounts: [
                         {
@@ -281,8 +281,8 @@ function(params)
                           name: selfPortName,
                         },
                       ],
-                      local livenessProbePath = 'livez',
-                      local readinessProbePath = 'metrics',
+                      local livenessProbePath = '/livez',
+                      local readinessProbePath = '/readyz',
                       livenessProbe::: {
                         httpGet: {
                           path: livenessProbePath,


jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -266,6 +266,38 @@ function(params)
...
+                          containerPort: metricsPort,
+                          name: 'metrics',
+                        },
+                        {
+                          containerPort: selfPort,
+                          name: 'self',
+                        },
+                      ],
+                      local livenessProbePath = 'healthz',
+                      local readinessProbePath = '',
```
rexagod, line 282: https://github.com/kubernetes/kube-state-metrics/blob/6de105ebbe0eeb5d97d9e6adf1cc83314434e8cc/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet#L200

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -266,6 +266,38 @@ function(params)
                           readOnly: true,
                         },
                       ],
+                      local metricsPort = 8081,
+                      local selfPort = 8082,
+                      ports::: [
```
rexagod, line 271: `:::` since [kube-prometheus hides these fields](https://github.com/prometheus-operator/kube-prometheus/blob/main/jsonnet/kube-prometheus/components/kube-state-metrics.libsonnet#L164-L166).

simonpasquier (commented):
Exec probes have lots of edge cases and I'd rather stay away from them.
Also are we convinced that a liveness probe would have caught the network issue?

rexagod:
There were a couple of asks in the original ticket.

- Being able to "figure out" network outages.
- Being able to restart the `kube-state-metrics` container if that happens.

I believe for the second case, exec probes may be the only way since kubelet pings are blocked off by KRP. In other cases, we've used [HTTP-based probes](https://github.com/openshift/cluster-monitoring-operator/pull/1289/commits/7b50ff1548c91415a5e3595e21fbf5ac444577ca) for the outer KRP instance, but a non-`200` status there will end up restarting the KRP instance, and not the proxied container.

For the first case, I believe we should keep the endpoints here as they are defined upstream, but make sure that `/healthz` (in upstream KSM) considers network outages before sending a `200` (querying a known resource type for metrics, for instance, or maybe a more suited endpoint). This way, liveness will restart the pod if there's an outage, while readiness will query `:8082` to determine if the pod's good for traffic (so no changes there).

IMHO I agree, and I'd prefer HTTP-based approach over this given the edge cases, but this seems to check all the asks.

simonpasquier:
> For the first case, I believe we should keep the endpoints here as they are defined upstream, but make sure that /healthz (in upstream KSM) considers network outages before sending a 200 (querying a known resource type for metrics, for instance, or maybe a more suited endpoint). This way, liveness will restart the pod if there's an outage, while readiness will query :8082 to determine if the pod's good for traffic (so no changes there).

https://github.com/kubernetes/kube-state-metrics/blob/a4ddfe6ed901c80b96e1dfab00a2df1820925b53/pkg/app/server.go#L405-L409

AFAICT /healthz always return 200 OK. I'd be on the fence adding checks against Kubernetes API to the /healthz endpoint: strictly speaking kube-state-metrics is still alive even if it can't reach the API. But it's an upstream discussion anyway.

rexagod:
Right, and to make this a non-breaking change, would it make sense to add outage checks to a `/livez` endpoint instead? If so, I can pitch this upstream.

simonpasquier:
Checking against what metrics-server does:
* /livez reports whether the process is dead-locked or not.
* /healthz reports whether all dependencies are operational.
* /readyz reports that the process is ready to serve HTTP requests.

Even to avoid a breaking change, I'm not sure that you want to swap the semantics of /livez and /healthz?

rexagod:
Hmm, there seems to be a similar pattern (eg., `/healthz` being influenced by dependencies) across components that are coupled with API-server ([`controller-manager`](https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/healthcheck/proxier_health.go#L193)) or aggregate-API-server (`metrics-server`).

But for out-of-tree non-`apiserver`-dependent projects like [`usage-metrics-collector`](https://github.com/kubernetes-sigs/usage-metrics-collector/blob/7a27bc576308daf15dcfecca259c76bb492438b8/cmd/metrics-prometheus-collector/cmd/cmd.go#L210), the `/healthz` seems to be in line with KSM, and seems to choose [`controller-runtime`](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.18.4/pkg/healthz/healthz.go#L187)'s style over `apiserver`'s.

I believe this pattern would warrant for including outage checks within `/livez`, while making it responsible for looking out for any non-healing symptoms (i.e., outages) which would demand a restart.

WDYT?

rexagod:
Friendly ping.

PS. I might be wrong here, but I wanted to confirm my deduction with you to devise my approach to resolve this upstream.

assets/kube-state-metrics/deployment.yaml
```diff
@@ -57,7 +57,24 @@ spec:
           ^kube_pod_completion_time$,
           ^kube_pod_status_scheduled$
         image: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.12.0
+        livenessProbe:
+          exec:
+            command:
+            - sh
+            - -c
+            - if [ -x "$(command -v curl)" ]; then exec curl -s -I -f http://localhost:8081/livez; elif [ -x "$(command -v wget)" ]; then exec wget --quiet --tries=1 --spider http://localhost:8081/livez; else exit 1; fi
```
simonpasquier, line 65: /livez returns 200 OK currently because the HTTP server returns the home URL. I'd rather have the /livez endpoint implemented before having the change here.

simonpasquier (commented):
As I said above, I've been burnt too many times by exec probes. Can we configure kube-rbac-proxy to disable authn/authz for probes?

assets/kube-state-metrics/deployment.yaml
rexagod: I've raised a PR for `livez` here: https://github.com/kubernetes/kube-state-metrics/pull/2418.

rexagod:
> As I said above, I've been burnt too many times by exec probes. Can we configure kube-rbac-proxy to disable authn/authz for probes?

ACK, I'll make the changes and drop the ones here.

rexagod:
> Can we configure kube-rbac-proxy to disable authn/authz for probes?

We can, but to the best of my understanding, not in a way that satisfies the [second point above](https://github.com/openshift/cluster-monitoring-operator/pull/2352#issuecomment-2146883283). While it is possible to define probes on the KRP containers to get an idea about the proxied application's health (just as we do for [Thanos Querier](https://github.com/openshift/cluster-monitoring-operator/blob/7a814977559b542c59bdf628b067a5c55dd49f0e/assets/thanos-querier/deployment.yaml#L100-L109)), these won't help Kubelet to actually restart the container (but instead the proxy).

I further verified this by making the following changes and deploying KRP's [`--ignore-path` test assets](https://github.com/brancz/kube-rbac-proxy/tree/master/test/e2e/ignorepaths).

```diff
diff --git a/test/e2e/ignorepaths/deployment.yaml b/test/e2e/ignorepaths/deployment.yaml
index 8e1f545b..cbc6bfee 100644
--- a/test/e2e/ignorepaths/deployment.yaml
+++ b/test/e2e/ignorepaths/deployment.yaml
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: kube-rbac-proxy
       containers:
         - name: kube-rbac-proxy
-          image: quay.io/brancz/kube-rbac-proxy:local
+          image: quay.io/brancz/kube-rbac-proxy:v0.18.0
           args:
             - "--secure-listen-address=0.0.0.0:8443"
             - "--upstream=http://127.0.0.1:8081/"
@@ -30,3 +30,7 @@ spec:
           image: quay.io/brancz/prometheus-example-app:v0.1.0
           args:
             - "--bind=127.0.0.1:8081"
+          livenessProbe:
+            httpGet:
+              port: 8081
+              scheme: HTTPS
```

assets/kube-state-metrics/deployment.yaml
```diff
@@ -57,7 +57,22 @@ spec:
...
+          httpGet:
+            path: livez
+            port: 8443
+            scheme: HTTPS
         name: kube-state-metrics
+        ports:
+        - containerPort: 8443
+          name: metrics
+        - containerPort: 9443
+          name: self
```
rexagod, line 70: (point probes for the KSM container to it's respective proxied ports, this helps us actually restart the container on `livenessProbe` failure)

assets/kube-state-metrics/deployment.yaml
```diff
@@ -57,7 +57,22 @@ spec:
           ^kube_pod_completion_time$,
           ^kube_pod_status_scheduled$
         image: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.12.0
+        livenessProbe:
+          httpGet:
+            path: livez
+            port: 8443
+            scheme: HTTPS
```
rexagod, line 64: (point `livenessProbe` for the KSM container to it's respective proxied port, this helps us actually restart the container on a probe failure)

rexagod:
https://github.com/kubernetes/kube-state-metrics/pull/2418#event-13278813804 is in, and will be included in https://github.com/kubernetes/kube-state-metrics/pull/2419 (v2.13.0).

rexagod:
Pinging @simonpasquier for one last review here to make sure everything looks good before we release KSMv2.13.0, so all that's left after that is merging this. :)

assets/kube-state-metrics/deployment.yaml
```diff
@@ -108,6 +124,7 @@ spec:
         - --tls-private-key-file=/etc/tls/private/tls.key
         - --client-ca-file=/etc/tls/client/client-ca.crt
         - --config-file=/etc/kube-rbac-policy/config.yaml
+        - --ignore-paths=/metrics
```
simonpasquier, line 127: not sure to understand why we ignore /metrics here?

assets/kube-state-metrics/deployment.yaml
```diff
@@ -57,7 +57,22 @@ spec:
           ^kube_pod_completion_time$,
           ^kube_pod_status_scheduled$
         image: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.12.0
+        livenessProbe:
+          httpGet:
+            path: livez
```
simonpasquier, line 62: should be /livez?

simonpasquier (commented):
I don't understand how the e2e tests pass given that the /livez endpoint doesn't exist.

assets/kube-state-metrics/deployment.yaml
rexagod: Right, even though we've always used telemetry metrics to check for readiness upstream, nonetheless, exposing that information is susceptible to malicious intent. I'll patch upstream to restrict that information from being exposed and add a dedicated `/readyz`.

assets/kube-state-metrics/deployment.yaml
rexagod: May be [omitted](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/url/url.go;l=842), I'll patch this up though for it to be consistent with other paths.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sample-deployment
  namespace: default
  labels:
    app: sample-app
spec:
  replicas: 3
  selector:
    matchLabels:
      app: sample-app
  template:
    metadata:
      labels:
        app: sample-app
    spec:
		...
        - containerPort: 80
        livenessProbe:
          httpGet:
            path: healthfoo
            port: 80
			...
```
```
Containers:
  sample-container:
    Container ID:   cri-o://736a33ba57ec594200107098015e4d0fd38c03d24e46e65b359bd83e43172095
    Image:          nginx:latest
    Image ID:       docker.io/library/nginx@sha256:67682bda769fae1ccf5183192b8daf37b64cae99c6c3302650f6f8bf5f0f95df
    Port:           80/TCP
    Host Port:      0/TCP
    State:          Running
      Started:      Wed, 03 Jul 2024 23:08:25 +0530
    Ready:          False
    Restart Count:  0
    Liveness:       http-get http://:80/healthfoo delay=15s timeout=2s period=5s #success=1 #failure=3
    Readiness:      http-get http://:80/readyz delay=5s timeout=2s period=5s #success=1 #failure=3
```

rexagod:
That is indeed strange, none of the KSM e2e tests should've passed. I'll take a look, maybe something's up with the custom/default parameters.

assets/kube-state-metrics/deployment.yaml
rexagod: On second thought, it'd be better to not rely on telemetry endpoints at all and base all probes off of the actual KSM resource metrics server.

rexagod: Opened https://github.com/kubernetes/kube-state-metrics/pull/2442.

juzhao:
@rexagod do we have other code change for this PR?

rexagod:
@juzhao Since https://github.com/openshift/cluster-monitoring-operator/pull/2352/files#r1664732654 is in, I'll update the PR. :)

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -208,7 +208,7 @@ function(params)
                         '--tls-private-key-file=/etc/tls/private/tls.key',
                         '--client-ca-file=/etc/tls/client/client-ca.crt',
                         '--config-file=/etc/kube-rbac-policy/config.yaml',
-                        '--ignore-paths=' + std.join(',', if std.endsWith(c.name, '-self') then ['/metrics'] else ['/livez']),
+                        '--ignore-paths=' + std.join(',', if std.endsWith(c.name, '-self') then ['/readyz'] else ['/livez']),
```
rexagod, line 211: Upstream `libsonnet`: https://github.com/kubernetes/kube-state-metrics/pull/2442/files#diff-cbebc0f980963cca7e4f8c36c69571676835a7d16f296123376a588b204c5c99.

rexagod:
`v2.13` is released, this can be merged now.

simonpasquier (commented):
shouldn't we propagate the probes enhancement to kube-prometheus first?

rexagod:
Currently, `kube-prometheus` hides probe specifications. I believe modifying them would break the expectations laid forth here: https://github.com/prometheus-operator/kube-prometheus/pull/436#issuecomment-594433918.

simonpasquier:
Unfortunately https://github.com/prometheus-operator/kube-prometheus/pull/436 lacks context about the issue being addressed. If CMO can deploy kube-state-metrics with liveness/readiness probes, I see no reason why it wouldn't be beneficial for kube-prometheus too. We shoould be able to find a way which is backward compatible (e.g. probes disabled by default but they can be opted in).

rexagod:
Opened a PR here: https://github.com/prometheus-operator/kube-prometheus/pull/2505.

PR #2359, opened 2024-05-22
MON-3801: clean-up unused OAuth proxy references

simonpasquier:
After the 4.16-release branch being cut, we have code related to OAuth proxy that can be removed now.


machine424:
/lgtm
thanks

PR #2360, opened 2024-05-22
MON-3701: clean-up tests

simonpasquier:


machine424:
/lgtm
Thanks.
There is also https://github.com/openshift/cluster-monitoring-operator/blob/db006a94385df720f1351b2dae636b53e40a271c/hack/ocp-images.sh#L103, but if don't use that script anymore, we can just remove it.
/hold
please unhold when you want.

PR #2362, opened 2024-05-22
MON-3701: clean-up injection of CA bundle for user Prometheus

simonpasquier:

### DIFF
--- a/assets/prometheus-user-workload/prometheus.yaml
+++ b/assets/prometheus-user-workload/prometheus.yaml
@@ -167,6 +167,9 @@ spec:
       name: secret-grpc-tls
   - name: prometheus
     terminationMessagePolicy: FallbackToLogsOnError
+    volumeMounts:
+    - mountPath: /etc/pki/ca-trust/extracted/pem/
+      name: prometheus-user-workload-trusted-ca-bundle
   enableFeatures:
   - extra-scrape-metrics
   enforcedNamespaceLabel: namespace
@@ -262,3 +265,10 @@ spec:
         memory: 100Mi
     version: 0.35.0
   version: 2.52.0
+  volumes:
+  - configMap:
+      items:
+      - key: ca-bundle.crt
+        path: tls-ca-bundle.pem
+      name: prometheus-user-workload-trusted-ca-bundle
+    name: prometheus-user-workload-trusted-ca-bundle
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1620,7 +1620,7 @@ func (f *Factory) PrometheusUserWorkloadAdditionalAlertManagerConfigsSecret() (*
 	}, nil
 }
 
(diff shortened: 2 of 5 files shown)


assets/prometheus-user-workload/prometheus.yaml
```diff
@@ -262,3 +265,10 @@ spec:
         memory: 100Mi
     version: 0.35.0
   version: 2.52.0
+  volumes:
+  - configMap:
+      items:
+      - key: ca-bundle.crt
```
machine424, line 271: I don't know if kube makes sure that the key value isn't empty before starting the Pod and if the value can be empty during a transient/temporary period...
(but even in that case, maybe, the Pod will converge)

pkg/manifests/manifests_test.go
```diff
@@ -1765,6 +1763,19 @@ func TestPrometheusUserWorkloadConfiguration(t *testing.T) {
 	if !reflect.DeepEqual(*f.config.UserWorkloadConfiguration.Prometheus.Resources, p.Spec.Resources) {
 		t.Fatal("Prometheus UWM resources not configured correctly")
 	}
+
+	for _, container := range p.Spec.Containers {
+		switch container.Name {
```
machine424, line 1768: nit: maybe we can raise in `default:` so we have a safe guard.

machine424 (commented):
Thanks.

pkg/tasks/prometheus_user_workload.go
```diff
@@ -318,6 +312,12 @@ func (t *PrometheusUserWorkloadTask) create(ctx context.Context) error {
 		}
 	}
 
+	// TODO(simonpasquier): remove this step after OCP 4.17 is released.
```
machine424, line 315: I created https://issues.redhat.com/browse/MON-3880 for those

assets/prometheus-user-workload/prometheus.yaml
simonpasquier: no Kubernetes wouldn't check that but yes it should converge eventually.

pkg/manifests/manifests.go
```diff
@@ -1751,13 +1741,7 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret, trustedCABundleCM *
 				PeriodSeconds:    15,
 				FailureThreshold: 240,
 			}
-			// Support CA bundles for Prometheus UWM.
-			if trustedCABundleVolumeName != "" {
-				p.Spec.Containers[i].VolumeMounts = append(
-					p.Spec.Containers[i].VolumeMounts,
-					trustedCABundleVolumeMount(trustedCABundleVolumeName),
```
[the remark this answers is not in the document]
simonpasquier: No it's used for telemeter

pkg/manifests/manifests_test.go
simonpasquier: changed a bit the test to make it more explicit indeed.

PR #2367, opened 2024-05-23
MON-3884: Send metric acm_managed_cluster_worker_cores via Telemetry

elgnay:
The related issue: https://issues.redhat.com/browse/MON-3884


Documentation/data-collection.md
```diff
@@ -703,11 +703,16 @@ data:
...
-    # owners: (https://github.com/open-cluster-management, @open-cluster-management/cluster-lifecycle-admin)
+    # owners: (https://github.com/orgs/stolostron/teams/server-foundation, @acm-server-foundation)
     #
     # acm_managed_cluster_info provides Subscription watch and other information for the managed clusters for an ACM Hub cluster.
     - '{__name__="acm_managed_cluster_info"}'
     #
+    # owners: (https://github.com/orgs/stolostron/teams/server-foundation, @acm-server-foundation)
+    #
+    # acm_managed_cluster_worker_cores:sum tracks the number of CPU cores on the worker nodes of the ACM managed clusters.
+    - '{__name__="acm_managed_cluster_worker_cores:sum"}'
```
juzhao, line 714: 1. what's the prometheus expr for recording rule `acm_managed_cluster_worker_cores:sum`, how the metric is calculated
2. https://github.com/rhobs/configuration/blob/main/configuration/telemeter/metrics.json did not see `acm_managed_cluster_worker_cores:sum` added

elgnay: `acm_managed_cluster_worker_cores:sum` is a recording rule belongs to ACM. That will be created once ACM 2.11+ is installed on an OCP cluster. Here is the definition of this recording rule: https://github.com/stolostron/backplane-operator/pull/718.

Should I add `acm_managed_cluster_worker_cores:sum` into https://github.com/rhobs/configuration/blob/main/configuration/telemeter/metrics.json?

juzhao: > Should I add `acm_managed_cluster_worker_cores:sum` into https://github.com/rhobs/configuration/blob/main/configuration/telemeter/metrics.json?

yes, should add the metric to rhobs repo, otherwise, you can not find the metrics from telemeter server even you can find it in your local cluster, see `Synchronize the Telemeter server’s configuration` part of https://rhobs-handbook.netlify.app/products/openshiftmonitoring/telemetry.md/

juzhao:
ci/prow/versions job failed for thanos version
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2367/pull-ci-openshift-cluster-monitoring-operator-master-versions/1796013340560986112
```
 --- a/jsonnet/versions.yaml
+++ b/jsonnet/versions.yaml
@@ -25,4 +25,4 @@ versions:
     prometheus: 2.52.0
     prometheusAdapter: 0.11.2
     prometheusOperator: 0.73.2
-    thanos: 0.35.0
+    thanos: 0.35.1
make: *** [Makefile:177: check-versions] Error 1
{"component":"entrypoint","error":"wrapped process failed: exit status 2","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:84","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2024-05-30T03:03:17Z"} 
```
@machine424 @marioferh , since ci/prow/versions job is not a required job, I think we could ignore the failure, or do you think we'd better fix it?

Documentation/data-collection.md
simonpasquier: Why using sum() and not max()?
IIUC the source metric comes from https://github.com/stolostron/clusterlifecycle-state-metrics/ and though sum and max should return the same result, max() is a better aggregation method, just in case Prometheus scrapes 2 instances by accident.

elgnay: That makes sense. I'll update the recording rule on the ACM side and then this PR.

elgnay: The update on ACM side is done and this PR has been updated as well.

simonpasquier:
@juzhao 

> since ci/prow/versions job is not a required job, I think we could ignore the failure, or do you think we'd better fix it?

Correct, we don't need to worry about it. It is only there to remind us that we need to merge PRs like https://github.com/openshift/cluster-monitoring-operator/pull/2376.

PR #2369, opened 2024-05-23
MON-3885: Crd based cmo codegen

marioferh:


marioferh:
Add vendoring to hack tools, need it for updatea-gen.sh

hack/update-codegen.sh
```diff
@@ -0,0 +1,40 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+
+(cd hack/tools/; go mod vendor)
```
marioferh, line 23: script needs vendoring but not sure if this is best way

PR #2380, opened 2024-06-13
MON-3894: revert(monitoring-plugin): undo nginx caching tweaks as the issue was…

machine424:
… resolved at its root cause.

Revert "prevent plugin entry assets from caching"

This reverts commit c6652ade52bd56bb5f4c4af10cfd28b6b350eccd.

Revert "Move config to jsonnet code and regenerate"

This reverts commit eb6a40c4ea30304c03eb4709d7e82ece7b98bd1a.

### DIFF
--- a/assets/monitoring-plugin/config-map.yaml
+++ b/assets/monitoring-plugin/config-map.yaml
@@ -12,20 +12,6 @@ data:
         ssl_certificate     /var/cert/tls.crt;
         ssl_certificate_key /var/cert/tls.key;
         root                /usr/share/nginx/html;
-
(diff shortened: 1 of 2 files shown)


simonpasquier:
/lgtm
/hold

I guess that we don't have other ways to verify the non-regression than checking manually?

machine424:
> I guess that we don't have other ways to verify the non-regression than checking manually?

I'm afraid we don't have any tests for this, I wanted to make the change at the beginning of 4.17 dev and wait for feedback.

I'll see with the console folks for the tests, there are others we can add as well.

PR #2386, opened 2024-06-19
WIP pkg/manifests: use prometheus-adapter config when metrics-server config not defined

slashpai:
Per https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/replace-prometheus-adapter-with-metrics-server.md#design-details we wanted to use prometheus adapter config for metrics server. This change is to handle that. We would want to backport this to 4.16 and remove it in 4.17


simonpasquier (commented):
Instead of doing it field by field, I'd rather check if the metrics server config is undefined/empty and only in this case, use the existing prometheus-adapter config.

I might also move the logic to Config.applyDefaults()

slashpai:
Closing this as this seems to be error prone, will revisit later if required

PR #2392, opened 2024-06-25
MON-3800: jsonnet: update prometheus dep to fetch the PrometheusKuber…

machine424:
…netesListWatchFailures alert added upstream in

https://github.com/prometheus/prometheus/pull/14308

The metric was added in prometheus v2.52.0, thus available in the current downstream version.

---

Also this:

jsonnet: Pin https://github.com/kubernetes-monitoring/kubernetes-mixin.git dep version

as newer versions are breaking.

This is a temporary solution to avoid blocking the other deps updates

See https://issues.redhat.com/browse/MON-3837

### DIFF
--- a/assets/prometheus-k8s/prometheus-rule.yaml
+++ b/assets/prometheus-k8s/prometheus-rule.yaml
@@ -36,6 +36,15 @@ spec:
       for: 20m
       labels:
         severity: warning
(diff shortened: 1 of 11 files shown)


simonpasquier (commented):
/lgtm

PR #2395, opened 2024-06-26
OCPBUGS-17113: Lazy updates for Prometheus

rexagod:
Incorporate https://github.com/openshift/library-go/pull/1575 downstream.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ GOPROXY?=http://proxy.golang.org
 export GO111MODULE
 export GOPROXY
 
-# go pakages for unit tests, excluding e2e tests
+# go packages for unit tests, excluding e2e tests
 PKGS=$(shell go list ./... | grep -v /test/e2e)
 GOLANG_FILES:=$(shell find . -name \*.go -print)
 # NOTE: grep -v %.yaml is needed  because "%s-policy.yaml" is used
@@ -248,6 +248,10 @@ check-runbooks:
 ###########
 # Testing #
 ###########
+
+.PHONY: bench
+bench:
+	go test -run NONE -bench=. -benchmem $(PKGS)
 
 .PHONY: test
 test: test-unit test-rules test-e2e
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -50,6 +50,7 @@ import (
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -58,6 +59,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/metadata"
 	"k8s.io/client-go/rest"
@@ -81,6 +83,7 @@ type Client struct {
 	userWorkloadNamespace string
 
 	kclient     kubernetes.Interface
+	dclient     dynamic.Interface
 	mdataclient metadata.Interface
 	osmclient   openshiftmonitoringclientset.Interface
 	oscclient   openshiftconfigclientset.Interface
@@ -105,6 +108,14 @@ func NewForConfig(cfg *rest.Config, version string, namespace, userWorkloadNames
 			return nil, fmt.Errorf("creating kubernetes clientset client: %w", err)
 		}
 		client.kclient = kclient
+	}
+
+	if client.dclient == nil {
+		dclient, err := dynamic.NewForConfig(cfg)
+		if err != nil {
+			return nil, fmt.Errorf("creating dynamic clientset client: %w", err)
+		}
+		client.dclient = dclient
 	}
 
 	if client.eclient == nil {
@@ -200,6 +211,12 @@ type Option = func(*Client)
 func KubernetesClient(kclient kubernetes.Interface) Option {
 	return func(c *Client) {
 		c.kclient = kclient
+	}
+}
+
+func DynamicClient(dclient *dynamic.DynamicClient) Option {
+	return func(c *Client) {
+		c.dclient = dclient
 	}
 }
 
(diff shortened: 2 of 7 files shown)


pkg/client/client.go
```diff
@@ -632,29 +649,106 @@ func (c *Client) GetAlertingRule(ctx context.Context, namespace, name string) (*
...
+	if err != nil {
+		return &didUpdate, fmt.Errorf("updating Prometheus object failed: %w", err)
+	}
 
-	required.ResourceVersion = existing.ResourceVersion
-	_, err = pclient.Update(ctx, required, metav1.UpdateOptions{})
+	return &didUpdate, nil
+}
+
+func prometheusDefaultingFunc(unstructuredPrometheus *unstructured.Unstructured) {
```
machine424, line 708: I thought that after our discussion here https://github.com/openshift/library-go/pull/1575#discussion_r1634427896, we "agreed" to give the "cache approach" a try for now, as it's simpler (less code/less maintenance), and with the cache, IIUC, I don't think we need to care about the defaults (if the wanted config is the one stored in the cache which mean the one deployed, skip the update).
Any reasons behind the change?

pkg/client/client_test.go
```diff
@@ -1787,34 +1800,276 @@ func TestCreateOrUpdatePrometheus(t *testing.T) {
...
+			Name:      "test-prometheus",
+			Namespace: "test-namespace",
+		},
+	}
+	prometheusGVR := initialPrometheus.GroupVersionKind().GroupVersion().WithResource("prometheuses")
+
+	c := &Client{}
+	c.dclient = dynamicFake.NewSimpleDynamicClient(runtime.NewScheme(), initialPrometheus.DeepCopy())
+	c.eventRecorder = events.NewInMemoryRecorder("create-or-update-prometheus-test")
+	c.resourceCache = resourceapply.NewResourceCache()
```
machine424, line 1848: Note that the code seems to be using the "comparison approach" (I don't see any cache initialized there), but here we're testing the "cache approach", am I missing sth?

pkg/client/client_test.go
```diff
@@ -1787,34 +1800,276 @@ func TestCreateOrUpdatePrometheus(t *testing.T) {
 				},
 			}
 
+			s := runtime.NewScheme()
+			_ = monv1.AddToScheme(s)
```
machine424, line 1804: these may be worth a comment

pkg/client/client.go
```diff
@@ -632,29 +649,106 @@ func (c *Client) GetAlertingRule(ctx context.Context, namespace, name string) (*
...
+			return nil, fmt.Errorf("creating Prometheus object failed: %w", err)
 		}
-		return nil
+		return ptr.To(true), nil
 	}
 	if err != nil {
-		return fmt.Errorf("retrieving Prometheus object failed: %w", err)
+		return nil, fmt.Errorf("retrieving Prometheus object failed: %w", err)
 	}
+	unstructuredRequiredPrometheusMetadataLabels := unstructuredRequiredPrometheus.GetLabels()
```
machine424, line 683: these may be worth a comment.
Just to be sure I understand, is this done to avoid https://github.com/openshift/cluster-monitoring-operator/blob/5341b80d4868fb1a330b14c8beff4c3ebba01b69/Documentation/development.md?plain=1#L52?

pkg/client/client.go
```diff
@@ -632,29 +649,106 @@ func (c *Client) GetAlertingRule(ctx context.Context, namespace, name string) (*
 	return c.osmclient.MonitoringV1().AlertingRules(namespace).Get(ctx, name, metav1.GetOptions{})
 }
 
-func (c *Client) CreateOrUpdatePrometheus(ctx context.Context, p *monv1.Prometheus) error {
-	pclient := c.mclient.MonitoringV1().Prometheuses(p.GetNamespace())
-	existing, err := pclient.Get(ctx, p.GetName(), metav1.GetOptions{})
+func (c *Client) CreateOrUpdatePrometheus(ctx context.Context, structuredRequiredPrometheus *monv1.Prometheus) (*bool, error) {
```
machine424, line 652: why do we need to return a pointer instead of just a bool here?
I don't think tests should rely on this, I think we should retrieve whether the object was updated from the API/client.

machine424 (commented):
Thanks!
I left some comments.

pkg/client/client.go
rexagod: Referring to the same thread: https://github.com/openshift/library-go/pull/1575#discussion_r1650646219, we do this in addition to the cache approach since that doesn't cache `spec` fields, in case of a metadata cache miss (same as what Trevor's patch targets in https://github.com/openshift/cluster-monitoring-operator/commit/ef5f348a939419818f9da862ff554af8853c9738#diff-2895f30116602d8fe6b0545c186d4f81247fa4eaab049dab174c6c91c0036e08R639-R667).

pkg/client/client.go
rexagod: I looked at the usages and there doesn't seem to be a need for three states. Fixed! 👍🏼

pkg/client/client.go
```diff
@@ -632,29 +649,106 @@ func (c *Client) GetAlertingRule(ctx context.Context, namespace, name string) (*
...
+	unstructuredExistingPrometheus, err := c.dclient.
+		Resource(prometheusGVR).
+		Namespace(structuredRequiredPrometheus.GetNamespace()).
+		Get(ctx, structuredRequiredPrometheus.GetName(), metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		_, err := pclient.Create(ctx, p, metav1.CreateOptions{})
+		_, err = c.dclient.
+			Resource(prometheusGVR).
+			Namespace(structuredRequiredPrometheus.GetNamespace()).
+			Create(ctx, unstructuredRequiredPrometheus, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
rexagod: You mean in the call that happens below? CMIIW But shouldn't we generally fail fast and return early?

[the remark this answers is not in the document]
rexagod: I see. I tried to [preserve the earlier behavior](https://github.com/openshift/cluster-monitoring-operator/pull/2395/files#diff-2dad64e28ce5e2eddf0adf303052d38e3b5b4fcba6b32fc4ca26d5e4d6b8fa4aR2519), albeit at the cost of no cache updates during creation, but this makes sense. Done!

pkg/client/client.go
rexagod: I looked at the usages and there doesn't seem to be a need for three states. Fixed! 👍🏼

pkg/client/client.go
rexagod: That's done to preserve the past behavior: PTAL https://github.com/rexagod/cluster-monitoring-operator/commit/529b00520a5ab9534d93169bccf4249c9dcd99a0#diff-2895f30116602d8fe6b0545c186d4f81247fa4eaab049dab174c6c91c0036e08R301.

pkg/client/client_test.go
rexagod: Both of these are in use (PTAL my comment https://github.com/openshift/cluster-monitoring-operator/pull/2395/files#r1696698968). Here, we initialize the cache, and down below we use the defaulting approach to reject default `spec` updates (`shouldUpdate` is `false` for [this case](https://github.com/openshift/cluster-monitoring-operator/pull/2395/files#diff-2dad64e28ce5e2eddf0adf303052d38e3b5b4fcba6b32fc4ca26d5e4d6b8fa4aR1861)).

pkg/client/client_test.go
```diff
@@ -1870,13 +1869,15 @@ func TestCreateOrUpdatePrometheus(t *testing.T) {
...
 			},
+			shouldUpdate: true,
 		},
 		{
 			description: "apply no value to the existing prometheus",
 			prometheus: &monv1.Prometheus{
 				TypeMeta:   initialPrometheus.TypeMeta,
 				ObjectMeta: initialPrometheus.ObjectMeta,
 			},
+			shouldUpdate: true,
```
rexagod, line 1880: With the no defaulting approach, there won't be a difference between default and non-default applications, as both will trigger updates as compared to just the latter earlier, at the benefit of not maintaining hardcoded default values in CMO. LMK if I should revert this or if this sounds good.

rexagod, line 1880: With the no defaulting approach, there won't be a difference between default and non-default applications, as both will trigger updates as compared to just the latter earlier, at the benefit of not maintaining hardcoded default values in CMO. LMK if I should revert this or if this sounds good.

Makefile
```diff
@@ -249,17 +249,21 @@ check-runbooks:
...
 .PHONY: test-unit
 test-unit:
-	go test -race -short $(PKGS) -count=1
+	go test -run ^Test -race -short $(PKGS) -count=1
 
 .PHONY: test-e2e
 test-e2e: KUBECONFIG?=$(HOME)/.kube/config
 test-e2e:
-	go test -v -timeout=150m ./test/e2e/ --kubeconfig $(KUBECONFIG)
+	go test -run ^Test -v -timeout=150m ./test/e2e/ --kubeconfig $(KUBECONFIG)
```
jan--f, line 266: Are the `-run` parameters really needed? I thought `go test` already only runs functions that start with `Test` and `go run -bench .` runs the `Benchmark` prefixed tests.

pkg/client/client.go
```diff
@@ -632,29 +649,70 @@ func (c *Client) GetAlertingRule(ctx context.Context, namespace, name string) (*
...
+		c.eventRecorder,
+		unstructuredRequiredPrometheus,
+		c.resourceCache,
+		prometheusGVR,
+		nil,
+		nil,
+	)
 	if err != nil {
-		return fmt.Errorf("updating Prometheus object failed: %w", err)
+		return didUpdate, fmt.Errorf("updating Prometheus object failed: %w", err)
```
jan--f, line 712: I find the interface a bit confusing. The bool is ignored in all call sites except for tests. But the tests only test `shouldUpdate=true and err != nil`. Its not clear what that scenario actually means and why `CreateOrUpdatePrometheus` needs to add this to its signature.
We should at least have some clarifying comments.

rexagod: ACK, I believe the additional `bool` in the upstream signature is an artifact of a previous oversight which created scenarios where [an error could be returned with a true `didUpdate` value](https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/monitoring.go#L94). I'll drop the `bool` from the signature here since it contributes little, and also refactor upstream to do away with the anti-pattern.

Makefile
rexagod: I verified that the tests and benches can indeed be run using `go test` and `go test -bench=.`, not sure why I assumed otherwise. Sorry for the confusion here, reverted.

pkg/client/client.go
rexagod: Added a comment to clarify the matrix.

rexagod:
/test e2e-aws-ovn-single-node

Flaking.

rexagod:
Blocked by https://github.com/openshift/library-go/pull/1823.

rexagod:
- [ ] Refactor the `resourceapply` signature as stated https://github.com/openshift/cluster-monitoring-operator/pull/2395#discussion_r1767017238 to drop the `bool` param.

pkg/client/client.go
rexagod: Resolved over Slack and https://github.com/openshift/library-go/pull/1833#discussion_r1931753875.

rexagod:
```
{  1 events happened too frequently

event happened 39 times, something is wrong: namespace/openshift-monitoring deployment/cluster-monitoring-operator hmsg/10ad659236 - reason/PrometheusUpdated Updated Prometheus.monitoring.coreos.com/k8s -n openshift-monitoring because it changed (09:44:45Z) result=reject }
```

Looking into it.

rexagod:
Plausible.

https://github.com/openshift/cluster-monitoring-operator/pull/2395/files#diff-2dad64e28ce5e2eddf0adf303052d38e3b5b4fcba6b32fc4ca26d5e4d6b8fa4aR2051-R2070

rexagod:
Oof, this is pretty ancient. I'll try to pick this up next sprint 🤞🏼 


/remove-lifecycle stale

PR #2409, opened 2024-07-09
MON-3900: Remove prometheus-adapter code

slashpai:
pkg: remove prometheus-adapter code


pkg/manifests/types.go
```diff
@@ -104,7 +104,7 @@ type AlertmanagerMainConfig struct {
 }
 
 // The `K8sPrometheusAdapter` resource defines settings for the Prometheus Adapter component.
-// This is deprecated and will be removed in a future version.
+// This is deprecated config, setting this has no effect and will be removed in a future version.
```
slashpai, line 107: Keeping this to check in 4.18 release if any user defined config exists in cmo's configmap and then block upgrades till removed

slashpai:
cc @simonpasquier @jan--f @machine424

pkg/manifests/types.go
simonpasquier: Can you add a TODO comment? It doesn't have to be 4.18 "exactly" but setting a target is good.

slashpai:
@simonpasquier could you please review again?

jan--f:
Looks good! Happy to merge when CI and Simon are happy :)

slashpai:
I will rebase once https://github.com/openshift/cluster-monitoring-operator/pull/2414 is merged to fix generate issue

madeelrh:
Why was it removed? and what it is being replaced with? It is not mentioned in this PR.

slashpai:
This PR was part of clean up. Its been replaced with metrics-server for resource metrics API 
You can find the details in https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/replace-prometheus-adapter-with-metrics-server.md

PR #2412, opened 2024-07-16
OU-450: Migrate Monitoring Plugin Deployment from nginx to Go server

PeterYurkovich:
This PR looks to update CMO to format the monitoring-plugin kubernetes objects such it will work with the new go backend for monitoring-plugin. This PR should not be merged until its associated [PR](https://github.com/openshift/monitoring-plugin/pull/128) in the monitoring-plugin repo has been merged. 

The primary change in this PR is the removal of the nginx configuration which is no longer needed as the go backend will serve the web server.


PeterYurkovich:
/restest

Best,

Peter Yurkovich

Tai-RedHat:
faild to launch with cluster-bot: [logs](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/1821729833231585280) [logs2](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/1821750107280773120)
```
level=error msg=Cluster operator monitoring Available is False with UpdatingConsolePluginComponentsFailed: UpdatingConsolePluginComponents: reconciling Console Plugin Deployment failed: updating Deployment object failed: waiting for DeploymentRollout of openshift-monitoring/monitoring-plugin: context deadline exceeded: got 2 unavailable replicas
```

Tai-RedHat:
re-test with cluster-bot, monitoring-plugin works well with no other regression issues.
/label qe-approved

Tai-RedHat:
BTW, can I get a brief explanation of why we deleted the monitor-plugin configmap?

jgbernalp:
> BTW, can I get a brief explanation of why we deleted the monitor-plugin configmap?

We are migrating the backend to Go, so the config map holding the old nginx configuration is not needed anymore.

pkg/tasks/monitoring_plugin.go
```diff
@@ -71,17 +71,6 @@ func (t *MonitoringPluginTask) Run(ctx context.Context) error {
 		}
 	}
 
-	{ // config map
-		cm, err := t.factory.MonitoringPluginConfigMap()
-		if err != nil {
-			return fmt.Errorf("initializing Console Plugin ConfigMap failed: %w", err)
-		}
-
-		if err = t.client.CreateOrUpdateConfigMap(ctx, cm); err != nil {
```
simonpasquier, line 80: we would need to delete the configmap (4.18) then remove the deletion in 4.19.

PeterYurkovich: I updated the PR to hardcode deleting the expected config map in the most recent commit

machine424:
Thanks for this.

Great to see that we’re working on unifying the plugin backends. I had raised this with the console folkshttps://redhat-internal.slack.com/archives/C6A3NV5J9/p1700750927381839 but never pushed it. I assume the next step would be to unify the Go backends now (https://github.com/openshift/monitoring-plugin/tree/main/pkg  and https://github.com/openshift/logging-view-plugin/tree/main/pkg/server etc.) into a common repository?

I will run some of the IPv6 and dual-stack payloads here as it took us some time and tweaks to make nginx work there:

/payload-job periodic-ci-openshift-release-master-nightly-4.18-e2e-metal-ipi-ovn-dualstack periodic-ci-openshift-release-master-nightly-4.18-e2e-metal-ipi-ovn-ipv6

Also, looking at some of the backend logs, I can see:

```
time="2024-08-22T21:20:59Z" level=info msg="enabled features: []\n" module=main
time="2024-08-22T21:20:59Z" level=warning msg="cannot read config file, serving plugin with default configuration, tried /etc/plugin/config.yaml" error="open /etc/plugin/config.yaml: no such file or directory" module=server
time="2024-08-22T21:20:59Z" level=error msg="cannot read base manifest file" error="open opt/app-root/web/dist/plugin-manifest.json: no such file or directory" module=manifest
time="2024-08-22T21:20:59Z" level=info msg="listening on https://:9443" module=server
```
I don’t know if that’s normal. If it is, could we hide or omit the error?

Note that our e2e tests regarding the plugin aren’t complete. Maybe we’re letting something slip. I don’t think we check that the backend is reachable e.g. (I’ll try to add one in CMO, even though the more realistic test should involve the console being able to reach the backend).

Also, adding a readiness probe to the Deployment could be helpful.

assets/monitoring-plugin/deployment.yaml
```diff
@@ -45,17 +45,12 @@ spec:
...
-          sed "s/LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME/$LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME/g" /etc/nginx/nginx.conf > /tmp/nginx.conf
-          exec nginx -c /tmp/nginx.conf -g 'daemon off;'
+      - args:
+        - -static-path=/opt/app-root/web/dist
+        - -cert=/var/cert/tls.crt
+        - -key=/var/cert/tls.key
+        command:
+        - /opt/app-root/plugin-backend
         env:
         - name: POD_IP
```
machine424, line 55: `POD_IP` was only used above, we can delete it now.

machine424:
here is the PR to add that "readiness/reachability" check https://github.com/openshift/cluster-monitoring-operator/pull/2444

assets/monitoring-plugin/deployment.yaml
PeterYurkovich: I removed the POD_IP from the `deployment.yaml` and the `monitoring-plugin.libsonnet` in the most recent commit

PeterYurkovich:
> here is the PR to add that "readiness/reachability" check https://github.com/openshift/cluster-monitoring-operator/pull/2444, please tell me if that's sufficient.

That readiness check seems sufficient to me. I wouldn't expect or want CMO to check for much other than valid JSON as changes upstream shouldn't impact the downstream distribution.

> Great to see that we’re working on unifying the plugin backends. I had raised this with the console folks [redhat-internal.slack.com/archives/C6A3NV5J9/p1700750927381839](https://redhat-internal.slack.com/archives/C6A3NV5J9/p1700750927381839) but never pushed it. I assume the next step would be to unify the Go backends now ([openshift/monitoring-plugin@main/pkg](https://github.com/openshift/monitoring-plugin/tree/main/pkg?rgh-link-date=2024-08-27T08%3A15%3A53Z) and [openshift/logging-view-plugin@main/pkg/server](https://github.com/openshift/logging-view-plugin/tree/main/pkg/server?rgh-link-date=2024-08-27T08%3A15%3A53Z) etc.) into a common repository?

The primary reasons for us making this change are:

1) Standardization across all our plugins ([1](https://github.com/openshift/monitoring-plugin/tree/main/pkg), [2](https://github.com/openshift/logging-view-plugin/tree/main/pkg/server), [3](https://github.com/openshift/console-dashboards-plugin/tree/main/pkg), [4](https://github.com/openshift/distributed-tracing-console-plugin/tree/main/pkg), [5](https://github.com/openshift/troubleshooting-panel-console-plugin/tree/main/pkg))
2) Greater control over the server in preparation for changes being made server-side to the `package-manifest.json` as well as feature flags to support an [Alerting UI in ACM](https://issues.redhat.com/browse/OU-396)

I think that bringing all the backends together in a single repository could be helpful, but it isn't in our short term goals that I'm aware of.

> I don’t know if that’s normal. If it is, could we hide or omit the error?

That error seems out of place, as the go server should be able to locate the `plugin-manifest.json`. I'll take a look at it

PeterYurkovich:
The CI is a but upset right now due to some pod taint, I'll rerun in a few hours

machine424:
Thanks for the context.

I see the server exposes a `/health` route
https://github.com/openshift/monitoring-plugin/blob/eb2a1bd822dc7a6f490276121bd654c5b000127d/pkg/server.go#L82 what do you think about adding a readinessProbe to the Deployment based on that like we do here https://github.com/openshift/cluster-monitoring-operator/blob/f7e92e869c43fa0455d656dcfc89045b60e5baa1/assets/admission-webhook/deployment.yaml#L60-L64? (no need to set up certs and stuff as kubelet will ignore that).

That will help us ensure real minimal readiness (the server starts up and can serve a route) of the plugin on all the flavors (ipv4, ipv6, dual stack etc.) that `origin` tests run on + https://github.com/openshift/cluster-monitoring-operator/pull/2444 will help us ensure that `/plugin-manifest.json` as well is accessible on the flavors that CMO e2e tests run on.

PeterYurkovich:
The latest commits add a readiness probe as requested and fix the error that was being shown. An invalid default value was fixed upstream ([fix](https://github.com/openshift/monitoring-plugin/pull/174)) and this PR now also passes in that value.

machine424:
Good news, thanks again for this.

(I can see the server is logging every call to `/health`, I think we can hide those, it's not a blocker for this PR)

I'll re-reun those ipv6 and dual stack payloads just in case:

/payload-job periodic-ci-openshift-release-master-nightly-4.18-e2e-metal-ipi-ovn-dualstack periodic-ci-openshift-release-master-nightly-4.18-e2e-metal-ipi-ovn-ipv6

But this
/lgtm
I'll 
/hold
just in case @simonpasquier has anything to add.

machine424:
going once, going twice...
/unhold
/retitle OU-450: Migrate Monitoring Plugin Deployment from nginx to Go server

Thanks again @PeterYurkovich for this.
(Tell me if I should create a follow up ticket to reduce the verbosity on /health)

PeterYurkovich:
I'll create the ticket in our board. Most of the backend code is shared currently, so I'll need to be done across all of our plugins, and too make sure that our deployments have a similar readiness/liveliness probe.

Thanks for working through this

PR #2405, opened 2024-07-03
WIP (review not needed): better errors and logging

machine424:
* [ ] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #2417, opened 2024-07-24
OCPBUGS-17506: Make error messages clearer on poll timeouts

simonpasquier:

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -913,13 +913,22 @@ func (c *Client) DeleteSecret(ctx context.Context, s *v1.Secret) error {
 	return err
 }
 
+func wrapWithResource(kind string, name types.NamespacedName, err error) error {
+	return wrapWithResourcef(kind, name, "%w", err)
+}
+
+func wrapWithResourcef(kind string, name types.NamespacedName, format string, a ...any) error {
+	return fmt.Errorf("%s %q: %w", kind, name.String(), fmt.Errorf(format, a...))
+}
+
 // validatePrometheusResource is a helper method for ValidatePrometheus.
+// It returns true only when the Prometheus resource is available and reconciled.
 // NOTE: this function is refactored out of wait.Poll for testing
 func (c Client) validatePrometheusResource(ctx context.Context, prom types.NamespacedName) (bool, []error) {
 	p, err := c.mclient.MonitoringV1().Prometheuses(prom.Namespace).Get(ctx, prom.Name, metav1.GetOptions{})
 	if err != nil {
-		// failing to get Prometheus -> Degraded: Unknown & Unavailable: Unknown
-		klog.V(4).Info("validate prometheus failed to get prometheus: ", err)
+		err = wrapWithResourcef("Prometheus", prom, "failed to get: %w", err)
+		// Report Degraded=Unknown and Unavailable=Unknown if the API request failed.
 		return false, []error{
 			NewUnknownAvailabiltyError(err.Error()),
 			NewUnknownDegradedError(err.Error()),
@@ -928,8 +937,8 @@ func (c Client) validatePrometheusResource(ctx context.Context, prom types.Names
 
 	avail, err := getMonitoringCondition(p.Status.Conditions, monv1.Available)
 	if err != nil {
-		err = fmt.Errorf("prometheus: %w", err)
-		// failing to get Prometheus.Status.Condtion -> Degraded: Unknown & Unavailable: Unknown
+		// Report Degraded=Unknown and Unavailable=Unknown if the condition can't be found.
+		err = wrapWithResource("Prometheus", prom, err)
 		return false, []error{
 			NewUnknownAvailabiltyError(err.Error()),
 			NewUnknownDegradedError(err.Error()),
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -929,54 +930,54 @@ func (c Client) validatePrometheusResource(ctx context.Context, prom types.Names
 	avail, err := getMonitoringCondition(p.Status.Conditions, monv1.Available)
 	if err != nil {
 		err = fmt.Errorf("prometheus: %w", err)
```
jan--f, line 932:
```suggestion
		err = fmt.Errorf("get prometheus status: %w", err)
```

pkg/client/client.go
```diff
@@ -929,54 +930,54 @@ func (c Client) validatePrometheusResource(ctx context.Context, prom types.Names
...
 			NewUnknownDegradedError(err.Error()),
 		}
 	}
 
 	if avail.Status == monv1.ConditionTrue {
-		// Prometheus is Available; check reconciled Condition as well
+		// Check the Reconciled condition.
 		reconciled, err := getMonitoringCondition(p.Status.Conditions, monv1.Reconciled)
 		if err != nil {
 			err = fmt.Errorf("prometheus: %w", err)
```
jan--f, line 944:
```suggestion
			err = fmt.Errorf("get prometheus status condition: %w", err)
```

pkg/client/client.go
```diff
@@ -1843,17 +1854,38 @@ func WithLastError(e *error) func(o *pollOptions) {
...
+	var (
+		conditionErr error
+		done         bool
+	)
+	deadlineCtx, deadlineCancel := context.WithTimeout(ctx, opts.timeout)
+	defer deadlineCancel()
+	if err := wait.PollUntilContextCancel(deadlineCtx, opts.interval, false, func(context.Context) (bool, error) {
+		// Don't pass the deadline context to the condition function to avoid
+		// errors when calling the API server.
+		done, conditionErr = condition(ctx)
```
jan--f, line 1874: I'm not sure I fully understand. Iiuc the goal here is to have `func Poll` be in charge of the timeout/deadline for the polling, instead of sharing the deadline ctx with the `condition()`?
If that is right, shouldn't we pass a completely new context to `condition()`? With the current implementation a caller of `Poll()` can still pass in a ctx with a timeout and our new `deadlineCtx` can only ever timeout earlier, but no later the the passed ctx.

simonpasquier: Yes, the idea is that the condition function should complete even if the poll context terminates during its execution.

In my first version, I created a new context but then I realize that the parent context is a cancel-only context which is done only when the program terminates (SIGTERM) so it feels better to stop the condition function ASAP in this case.

jan--f: Oh I see now. `PollUntilContextTimeout` sets the timeout on `ctx` before passing it to `condition`. Thanks

pkg/client/client.go
```diff
@@ -914,12 +914,13 @@ func (c *Client) DeleteSecret(ctx context.Context, s *v1.Secret) error {
...
 
 // validatePrometheusResource is a helper method for ValidatePrometheus.
+// It returns true only when the Prometheus resource is available and reconciled.
 // NOTE: this function is refactored out of wait.Poll for testing
 func (c Client) validatePrometheusResource(ctx context.Context, prom types.NamespacedName) (bool, []error) {
 	p, err := c.mclient.MonitoringV1().Prometheuses(prom.Namespace).Get(ctx, prom.Name, metav1.GetOptions{})
 	if err != nil {
-		// failing to get Prometheus -> Degraded: Unknown & Unavailable: Unknown
-		klog.V(4).Info("validate prometheus failed to get prometheus: ", err)
+		err = fmt.Errorf("get prometheus status: %w", err)
```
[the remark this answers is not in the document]
jan--f: Just realized this is similar to https://github.com/openshift/cluster-monitoring-operator/pull/2405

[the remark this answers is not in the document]
simonpasquier: :facepalm:

pkg/client/client.go
simonpasquier: I've removed the prefix since the error is unlikely to happen in real life (e.g. it'd be a bug in the code or in the Prometheus operator not populating the conditions properly).

simonpasquier:
Tested with a live cluster after triggering deployment failures.

Before this PR

```
UpdatingTelemeterClient: reconciling Telemeter client Deployment failed: updating Deployment object failed: waiting for DeploymentRollout of openshift-monitoring/telemeter-client: context deadline exceeded,
UpdatingPrometheus: client rate limiter Wait returned an error: context deadline exceeded
```

After this PR

```
UpdatingTelemeterClient: reconciling Telemeter client Deployment failed: updating Deployment object failed: waiting for DeploymentRollout of openshift-monitoring/telemeter-client: context deadline exceeded: got 1 unavailable replicas,
UpdatingPrometheus: StatefulSetNotFound: shard 0: statefulset openshift-monitoring/prometheus-k8s not found
```

pkg/client/client.go
```diff
@@ -928,55 +929,53 @@ func (c Client) validatePrometheusResource(ctx context.Context, prom types.Names
 
 	avail, err := getMonitoringCondition(p.Status.Conditions, monv1.Available)
 	if err != nil {
-		err = fmt.Errorf("prometheus: %w", err)
-		// failing to get Prometheus.Status.Condtion -> Degraded: Unknown & Unavailable: Unknown
+		// Report Degraded=Unknown and Unavailable=Unknown if the condition can't be found.
```
jan--f, line 932: /hold
Sorry...did you mean to remove the `Errorf` here?

pkg/client/client.go
```diff
@@ -1843,17 +1852,38 @@ func WithLastError(e *error) func(o *pollOptions) {
...
+		timeout:  5 * time.Minute,
+		interval: time.Second,
+	}
 	for _, o := range options {
 		o(&opts)
 	}
 
-	if err := wait.PollUntilContextTimeout(ctx, opts.interval, opts.timeout, false, condition); err != nil {
+	var (
+		conditionErr error
```
machine424, line 1864: why do we need this `conditionErr`?
A unit test would be great to show when `Poll` couldn't catch such error.

pkg/client/client.go
machine424: I'm leaning towards adding more context, sth like
`"Prometheus CR %s: %w", prom.String()...` to make sure users will go take a look at the right CR status for more details.
I was also testing the inclusion of the CR reason/message directly to the error message https://github.com/openshift/cluster-monitoring-operator/pull/2405/files

machine424 (commented):
Thanks!
Here are the unit tests I told you about https://github.com/openshift/cluster-monitoring-operator/compare/master...machine424:cluster-monitoring-operator:poll-tests?expand=1, pick the ones that seem interesting to you.

pkg/client/client.go
simonpasquier: yes the removal was intentional.  The returned error is already informative (it says "failed to find condition ...").
I'm updating the patch to always add the resource identifier.

pkg/client/client.go
simonpasquier: we need it at L1881. Eventually I think that we can rid of the WithLastError() construct too.

pkg/client/client.go
```diff
@@ -1843,17 +1852,38 @@ func WithLastError(e *error) func(o *pollOptions) {
...
+		done         bool
+	)
+	deadlineCtx, deadlineCancel := context.WithTimeout(ctx, opts.timeout)
+	defer deadlineCancel()
+	if err := wait.PollUntilContextCancel(deadlineCtx, opts.interval, false, func(context.Context) (bool, error) {
```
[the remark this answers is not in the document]
simonpasquier: No you want the wait function to honor the deadline context which is why we don't use PollUntilContextTimeout() here.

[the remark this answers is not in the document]
machine424: but we can still force the context used in the condition using the wrapper, for me, the existing is equivalent to:
```
if err := wait.PollUntilContextTimeout(ctx, ..., func(_ context.Context) (bool, error) {
  return condition(ctx) //  ignore PollUntilContextTimeout's context `_` and use the parent context `ctx`
 })
```

pkg/client/client.go
simonpasquier: IMHO a simpler contract is to declare that the function is retried is retried until no error is returned. From what I can tell, there's zero occurrences where the condition function returns an error directly (e.g. it's always `return false, nil` or `return true, nil`).

simonpasquier:
/hold cancel
@machine424 I've included the unit tests that you've mentioned earlier :)

pkg/client/client.go
machine424: Yep, but I still think that catching the `condition` error (via `conditionErr`) should be done in a separate PR. AFAIU, this doesn't help with anything regarding OCPBUGS-17506, but I might be missing something. A unit test showing the current error-catching logic's shortcomings would be helpful.

We can leave this for another PR and also see if we can completely remove the `lastError` logic.

/lgtm 
Thanks for this!

/hold
Please unhold if/when needed.

PR #2422, opened 2024-07-29
OCPBUGS-35095: unpin `kubernetes-mixin`

rexagod:
Unpin `kubernetes-mixin` to rely on the version pulled in from `kube-prometheus`, and update the dependencies as a result. Also, update the `version` manifest as well, as that changed during the tenure of this PR.


simonpasquier:
@rexagod need to run "make generate"?

rexagod:
@simonpasquier It seems updating `k-mixin` dependency to the corresponding upstream commit breaks the dashboards. Earlier I used to `jb update <dependency>@<commit>` and `generate` the manifests, but I was afraid that we'll hit a point where it won't be possible anymore due to the recent refactor.

rexagod:
I can push for a branch upstream that ports post-refactor commits while making use of the earlier Grafana v11 dashboards. Or we might want to refactor our jsonnets internally.

rexagod:
Unblocked, needs to be re-implemented using the newly introduced method.

jsonnet/jsonnetfile.json
```diff
@@ -58,14 +58,13 @@
       "version": "main"
     },
     {
-      "name": "kubernetes-mixin is pinned because newer versions are breaking MON-3837",
```
rexagod, line 61: cc @simonpasquier @machine424 

Since https://github.com/openshift/cluster-monitoring-operator/pull/2482 removed CMO's dependency on `k/k-mixin` for dashboards, I believe we can unpin and sync with upstream to update the recording rules that depend on the same.

rexagod: > A follow-up PR will unpin the kubernetes-monitoring/kubernetes-mixin version.

Ah, nvm I just noticed Simon's comment there.

rexagod:
Good for reviews now. 🙂

jsonnet/jsonnetfile.json
simonpasquier: I think that we can remove the whole entry for kubernetes-monitoring/kubernetes-mixin and just depend on the version pulled by kube-prometheus.

assets/control-plane/prometheus-rule.yaml
```diff
@@ -336,7 +336,7 @@ spec:
         description: The kubernetes apiserver has terminated {{ $value | humanizePercentage }} of its incoming requests.
         summary: The kubernetes apiserver has terminated {{ $value | humanizePercentage }} of its incoming requests.
       expr: |
-        sum(rate(apiserver_request_terminations_total{job="apiserver"}[10m]))  / (  sum(rate(apiserver_request_total{job="apiserver"}[10m])) + sum(rate(apiserver_request_terminations_total{job="apiserver"}[10m])) ) > 0.20
+        sum by(cluster) (rate(apiserver_request_terminations_total{job="apiserver"}[10m])) / ( sum by(cluster) (rate(apiserver_request_total{job="apiserver"}[10m])) + sum by(cluster) (rate(apiserver_request_terminations_total{job="apiserver"}[10m])) ) > 0.20
```
simonpasquier, line 339: unrelated to this PR but I'd recommend to remove the `KubeAPITerminatedRequests` alerting rule since we're not owners for the API server. And same goes with `KubeAPIDown`.

rexagod: ACK, I'll open up a new PR.

machine424:
The PR title is a bit misleading :) I initially thought it would be easy to review, but I had to spend more time understanding what the PR is actually doing. The PR introduces three main changes: updating jsonnet dependencies, updating operands versions, and removing a pin. This makes it harder to review.
/lgtm
Thanks for this.
It would be great if you could add more details about what the PR does in the description and commits.

jan--f:
+1 to Ayoubs comment, lets update the PR title and perhaps description. Otherwise
/lgtm

jan--f:
Feel free to unhold when ready

rexagod:
Updated the description and squashed the commits (and updated the commit message). Needs another `/lgtm`.

/unhold
/jira refresh

machine424:
Thanks again
/lgtm

rexagod:
/jira backport release-4.18,release-4.17,release-4.16,release-4.15,release-4.14

PR #2424, opened 2024-07-31
MON-3961: inject proxy env variables in Alertmanager

simonpasquier:


assets/alertmanager/alertmanager.yaml
```diff
@@ -27,6 +27,15 @@ spec:
         topologyKey: kubernetes.io/hostname
   automountServiceAccountToken: true
   containers:
+  - env:
+    - name: HTTP_PROXY
```
danielmellado, line 31: just a tiny minor thing, I was wondering if we should also pass lower cased versions of these envs as well (http_proxy, https_proxy, no_proxy)

PR #2425, opened 2024-07-31
MON-3900: follow-up: clean up operator logic and some tests now that metrics-server is the default Metrics API backend

machine424:
Came across this while working on https://github.com/openshift/cluster-monitoring-operator/pull/2423

/cc @slashpai @jan--f 

/hold
let's wait for https://github.com/openshift/cluster-monitoring-operator/pull/2423 to get merged first.


slashpai:
Thank you. I was going to have this as follow up PR. We should be able to remove feature gate as well from api

jan--f:
Thanks!
/lgtm

PR #2428, opened 2024-08-02
Allow scrapping of OpenStack load-balancer-id

mandre:
This PR is made of two separate changes.

The first one remove `kube_*_annotations` from the deny list, as it is no longer needed since https://github.com/kubernetes/kube-state-metrics/pull/2145 and prevents the collection of `kube_*_annotations`.

The second change effectively enables scrapping of OpenStack load-balancer-id.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 4.17
 
 - [#2409](https://github.com/openshift/cluster-monitoring-operator/issues/2409) Remove prometheus-adapter code from CMO
+- [#2428](https://github.com/openshift/cluster-monitoring-operator/issues/2428) Allow scrapping of OpenStack load-balancer-id
 
 ## 4.16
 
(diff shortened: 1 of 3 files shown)


jan--f:
I'm generally in favor of this, though I think we have yet another setting to gatekeep that should really be user accessible, similar to https://issues.redhat.com/browse/MON-2809.

Iiuc we also no longer need to deny `kube_secret_labels`?

mandre:
> I'm generally in favor of this, though I think we have yet another setting to gatekeep that should really be user accessible, similar to https://issues.redhat.com/browse/MON-2809.
> 
> Iiuc we also no longer need to deny `kube_secret_labels`?

Yep, that's also my understanding I can do that in a separate PR if you'd like.

mandre:
Closing in favor of https://github.com/openshift/cluster-monitoring-operator/pull/2430. We want to use a more dynamic approach for collecting the needed annotation thanks to CRS.
/close

PR #2430, opened 2024-08-02
Remove `kube_*_annotations` and `kube_*_labels` from deny list

mandre:
Since https://github.com/kubernetes/kube-state-metrics/pull/2145, kube-state-metrics does not collect `kube_*_annotations` or `kube_*_labels` metrics by default. It's no longer necessary to add them to the metrics' deny list.

Removing `kube_*_annotations` from the deny list allows us to enable scrapping of annotation metrics via the `--metric-annotations-allowlist` option.

Additionally, we were missing a comma, which might have been a problem in the comma-separated list of arguments.

### DIFF
--- a/assets/kube-state-metrics/deployment.yaml
+++ b/assets/kube-state-metrics/deployment.yaml
@@ -38,8 +38,6 @@ spec:
         - --telemetry-port=8082
         - |
           --metric-denylist=
-          ^kube_secret_labels$,
-          ^kube_.+_annotations$
           ^kube_customresource_.+_annotations_info$,
           ^kube_customresource_.+_labels_info$,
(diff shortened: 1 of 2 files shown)


assets/kube-state-metrics/deployment.yaml
```diff
@@ -38,8 +38,6 @@ spec:
         - --telemetry-port=8082
         - |
           --metric-denylist=
-          ^kube_secret_labels$,
-          ^kube_.+_annotations$
```
machine424, line 42: cc @rexagod, apparently the comma was omitted during https://github.com/openshift/cluster-monitoring-operator/commit/bc1174a6f7746950a11bc405bc825a132c4957a7, not sure if this was intentional.

rexagod: Definitely not intentional, missed adding a comma there. I believe the other upstream patch went in before causing this to not blow up. Glad to see this was caught here!

PR #2441, opened 2024-08-20
OCPBUGS-38289: fix(prometheus): avoid passing the cluster-wide proxy config to a remote endpoint when noProxy will prevent the proxuying anyway

machine424:
Also

docs: hide PrometheusK8sConfig.RemoteWrite.ProxyURL from docs until its usage is clarified

### DIFF
--- a/Documentation/openshiftdocs/modules/remotewritespec.adoc
+++ b/Documentation/openshiftdocs/modules/remotewritespec.adoc
@@ -35,7 +35,7 @@ link:prometheusrestrictedconfig.adoc[PrometheusRestrictedConfig]
 
 |oauth2|*monv1.OAuth2|Defines OAuth2 authentication settings for the remote write endpoint.
 
-|proxyUrl|string|Defines an optional proxy URL.
+|proxyUrl|string|Defines an optional proxy URL. It is superseded by the cluster-wide proxy, if enabled.
 
 |queueConfig|*monv1.QueueConfig|Allows tuning configuration for remote write queue parameters.
 
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -37,6 +37,7 @@ import (
 	securityv1 "github.com/openshift/api/security/v1"
 	"github.com/openshift/library-go/pkg/crypto"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"golang.org/x/net/http/httpproxy"
 	yaml2 "gopkg.in/yaml.v2"
 	admissionv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
@@ -1424,12 +1425,41 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret) 
 	for k := range p.Spec.RemoteWrite {
 		rw := &p.Spec.RemoteWrite[k]
 
+		var tmpRWProxyURL *string
 		if f.proxy.HTTPProxy() != "" {
-			rw.ProxyURL = ptr.To(f.proxy.HTTPProxy())
+			tmpRWProxyURL = ptr.To(f.proxy.HTTPProxy())
 		}
 		if f.proxy.HTTPSProxy() != "" {
-			rw.ProxyURL = ptr.To(f.proxy.HTTPSProxy())
+			tmpRWProxyURL = ptr.To(f.proxy.HTTPSProxy())
 		}
+
+		// Fixes: https://issues.redhat.com/browse/OCPBUGS-38289
+		// The code is condensed here to simplify backporting.
+		// This is a temporary workaround because the NoProxy field isn't exported on the Prometheus CRD in older versions.
+		// If the cluster proxy or remoteWrite config change, this code will re-run, ensuring no stale state.
+		// The httpproxy package used here is the same as used by Prometheus, which helps with consistency.
+
+		// proxy URL is not to be set/overwritten in here, nothing to do.
+		if tmpRWProxyURL == nil {
+			continue
+		}
+		// Suppose that CMO has set the proxy URL.
+		proxyConfig := &httpproxy.Config{
+			HTTPProxy:  *tmpRWProxyURL,
+			HTTPSProxy: *tmpRWProxyURL,
+			NoProxy:    f.proxy.NoProxy(),
+		}
+		proxyFunc := proxyConfig.ProxyFunc()
+		// Ignore errors here.
+		u, _ := url.Parse(rw.URL)
+		// Check if the endpoint URL will be using that proxy URL.
+		proxyURL, err := proxyFunc(u)
+		if err == nil && proxyURL == nil {
+			// Proxing will not be used, no need to set it.
+			continue
+		}
+		// The proxy will be used, set it.
+		rw.ProxyURL = tmpRWProxyURL
 	}
 
 	if f.config.Images.Thanos != "" {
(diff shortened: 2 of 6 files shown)


pkg/manifests/types.go
```diff
@@ -757,7 +757,7 @@ type RemoteWriteSpec struct {
 	Name string `json:"name,omitempty"`
 	// Defines OAuth2 authentication settings for the remote write endpoint.
 	OAuth2 *monv1.OAuth2 `json:"oauth2,omitempty"`
-	// Defines an optional proxy URL.
```
machine424, line 760: This should be reflected in the docs as well.
(Tell me if we should keep this to another PR)

pkg/manifests/manifests.go
```diff
@@ -1424,6 +1425,29 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret)
...
+		proxyFunc := proxyConfig.ProxyFunc()
+		// Ignore errors here.
+		u, _ := url.Parse(rw.URL)
+		proxyURL, err := proxyFunc(u)
+		if err == nil && proxyURL == nil {
+			// Proxing will not be used, no need to set it.
+			continue
+		}
+
 		if f.proxy.HTTPProxy() != "" {
```
simonpasquier, line 1451: hmm shouldn't we remove this code and use `proxyURL` instead?

pkg/manifests/manifests.go
```diff
@@ -1424,6 +1425,29 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret)
...
+		}
+		// Check if the endpoint URL needs proxying.
+		proxyConfig := &httpproxy.Config{
+			HTTPProxy:  f.proxy.HTTPProxy(),
+			HTTPSProxy: f.proxy.HTTPSProxy(),
+			NoProxy:    f.proxy.NoProxy(),
+		}
+		proxyFunc := proxyConfig.ProxyFunc()
+		// Ignore errors here.
+		u, _ := url.Parse(rw.URL)
```
simonpasquier, line 1444: An improvement for later (and it can wait for the CMO config CRD) but we could detect invalid URLs and fail early. I suppose that right now, a not parseable URL would fail the prometheus operator reconciliation.

pkg/manifests/manifests.go
```diff
@@ -1424,6 +1425,29 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret)
 	for k := range p.Spec.RemoteWrite {
 		rw := &p.Spec.RemoteWrite[k]
 
+		// Fixes: https://issues.redhat.com/browse/OCPBUGS-38289
+		// The code is condensed here to simplify backporting.
+		// This is a temporary workaround because the NoProxy field isn't exported on the Prometheus CRD in older versions.
+		// If the cluster proxy or remoteWrite config change, this code will re-run, ensuring no stale state.
+		// The httpproxy package used here is the same as used by Prometheus, which helps with consistency.
+		if f.proxy.HTTPSProxy() == "" && f.proxy.HTTPProxy() == "" {
```
simonpasquier, line 1433: (nit) we could as well avoid the early continue and always call ProxyFunc()?

pkg/manifests/types.go
simonpasquier: I see several options:
1. keep the current behavior (though it's awkward), if proxyURL is set then use this value, expect if a global HTTP proxy is configured. And it should be explicit in the documentation.
2. Document that the field is deprecated and ignore it.
3. Remove the field (not desired because it could break upgrades).

We could mix 1 and 2: keep the current behavior for a couple of releases and warn users (via the ClusterOperator upgrade condition) that they need to remove it before upgrading to version 4.x (tbd).

pkg/manifests/manifests.go
machine424: Yes, we just need to make sure they use the same definition of a valid URL.
I thought that it would be blocked by the validation webhook, but I'm not sure.

pkg/manifests/manifests.go
machine424: I adjust the code to make it simulate what would happen if CMO set/overwrites the proxy URL here.

pkg/manifests/manifests.go
machine424: I wanted the added logic to be the least intrusive possible.
I adjusted it

pkg/manifests/types.go
machine424: I still think this is a bug, config fields explicitly provided by users should always have precedence, otherwise it makes to sense to keep them exposed.

Let's go with `1.` for now and see.

simonpasquier: I agree that it's a bug *but* changing the behavior would be a breaking change for existing users.

pkg/manifests/manifests.go
simonpasquier: which validation webhook?

pkg/manifests/manifests.go
```diff
@@ -1424,12 +1425,41 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret)
...
+		// Ignore errors here.
+		u, _ := url.Parse(rw.URL)
+		// Check if the endpoint URL will be using that proxy URL.
+		proxyURL, err := proxyFunc(u)
+		if err == nil && proxyURL == nil {
+			// Proxing will not be used, no need to set it.
+			continue
+		}
+		// The proxy will be used, set it.
+		rw.ProxyURL = tmpRWProxyURL
```
simonpasquier, line 1462:
```suggestion
		rw.ProxyURL = proxyURL.String()
```

pkg/manifests/types.go
machine424: We can always start warning about the current "unusual" behavior and fix that in a future `4.x.0` making sure it's mentioned in the release notes.

(This "unusual" behavior was introduced in 4.16, so it can also be considered as a breaking change for someone upgrading from 4.15 that used to configure the proxyURL via the configmap because they don't want to go through the cluster-wide one.)

But let's continue this discussion somewhere else :)

pkg/manifests/manifests.go
machine424: prom-operator's, but maybe Prometheus CRs do not go through it.

pkg/manifests/manifests.go
machine424: I think it's easier to understand the logic with `tmpRWProxyURL`
(if `noProxy` wouldn't change anything, we fallback to the old behaviour: CMO itself setting `rw.ProxyURL`)

Documentation/openshiftdocs/modules/remotewritespec.adoc
```diff
@@ -35,7 +35,7 @@ link:prometheusrestrictedconfig.adoc[PrometheusRestrictedConfig]
 
 |oauth2|*monv1.OAuth2|Defines OAuth2 authentication settings for the remote write endpoint.
 
-|proxyUrl|string|Defines an optional proxy URL.
+|proxyUrl|string|Defines an optional proxy URL. It is superseded by the cluster-wide proxy if enabled.
```
machine424, line 38: cc @eromanova97, whether the wording is ok for you.

pkg/manifests/manifests.go
simonpasquier: yeah, the validation webhook only works against PrometheusRule and AlertmanagerConfig objects.

pkg/manifests/manifests.go
simonpasquier: fair enough :)

simonpasquier (commented):
nit but commit ec95ac1bf012e0a31a52cc48965ede061455778c has a typo ("proxuying" > "proxying")

simonpasquier:
otherwise lgtm!

Documentation/openshiftdocs/modules/remotewritespec.adoc
eromanova97:
```suggestion
|proxyUrl|string|Defines an optional proxy URL. It is superseded by the cluster-wide proxy, if enabled.
```
Just a very small suggestion :)

simonpasquier (commented):
/lgtm

PR #2442, opened 2024-08-20
OCPBUGS-38780: update links to OpenShift documentation

iNecas:
The original links were not working anymore. Also, in some places, it hardcoded specific version. Changed to use `latest` instead.


pkg/operator/operator.go
```diff
@@ -1064,7 +1064,7 @@ func (o *Operator) Config(ctx context.Context, key string) (*manifests.Config, e
 // version can be computed, the link will point to the documentation for that
 // version, else it will point to latest documentation.
 func (o Operator) storageNotConfiguredMessage() string {
-	const docURL = "https://docs.openshift.com/container-platform/%s/monitoring/configuring-the-monitoring-stack.html"
+	const docURL = "https://docs.openshift.com/container-platform/%s/observability/monitoring/configuring-the-monitoring-stack.html"
```
simonpasquier, line 1067: this alone would require an OCPBUGS ticket and backport to 4.16.z (it looks correct for 4.15 and below).

simonpasquier (commented):
thanks!

simonpasquier (commented):
/lgtm
/approve

PR #2444, opened 2024-08-27
MON-3990: make TestClusterMonitorConsolePlugin check that one of the pods can serve /plugin-manifest.json

machine424:
… pods can serve /plugin-manifest.json

This is all the more important since the plugin Pods do not have readiness probes.

A more realistic test would ensure the console can reach the plusgins, that can be added in origin/

### DIFF
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -16,14 +16,19 @@ package e2e
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
 	"testing"
 	"time"
 
+	"github.com/Jeffail/gabs"
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-monitoring-operator/pkg/client"
 	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
+	"github.com/stretchr/testify/require"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


test/e2e/config_test.go
```diff
@@ -727,6 +731,40 @@ func TestUserWorkloadMonitorThanosRulerConfig(t *testing.T) {
...
+			return err
+		}
+		defer resp.Body.Close()
+
+		b, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("fail to read response body: %w", err)
+		}
+
+		if resp.StatusCode != http.StatusOK {
```
simonpasquier, line 760: (nit) I would check the status code first and then verify that the response body length is > 0 (or even valid JSON?).

test/e2e/config_test.go
```diff
@@ -727,6 +732,49 @@ func TestUserWorkloadMonitorThanosRulerConfig(t *testing.T) {
...
+			return fmt.Errorf("expected status %d, got %d (%q)", resp.StatusCode, http.StatusOK, framework.ClampMax(b))
+		}
+
+		res, err := gabs.ParseJSON(b)
+		if err != nil {
+			return err
+		}
+		name, ok := res.Path("name").Data().(string)
+		if !ok || name != pluginName {
+			t.Errorf("expected plugin name to be %s, got `%s`", pluginName, name)
```
simonpasquier, line 771: (nit) if !ok then name is empty? Also shouldn't we return an err rather than calling t.Errorf()?

machine424: Good catch for the `t.Errorf`
> (nit) if !ok then name is empty

That's why I'm putting it between ``

machine424: `t.Errorf` still makes the test fail, but it should be cleaner now.

machine424:
/retitle MON-3990: make TestClusterMonitorConsolePlugin check that one of the pods can serve /plugin-manifest.json

machine424:
Note that a readinessProbe will as well be added for the new Go backend https://github.com/openshift/cluster-monitoring-operator/pull/2412#issuecomment-2315085438

test/e2e/config_test.go
```diff
@@ -727,6 +732,49 @@ func TestUserWorkloadMonitorThanosRulerConfig(t *testing.T) {
...
+			return fmt.Errorf("expected status %d, got %d (%q)", resp.StatusCode, http.StatusOK, framework.ClampMax(b))
+		}
+
+		res, err := gabs.ParseJSON(b)
+		if err != nil {
+			return err
+		}
+		name, ok := res.Path("name").Data().(string)
+		if !ok || name != pluginName {
+			return fmt.Errorf("expected plugin name to be %s, got `%s`", pluginName, name)
```
simonpasquier, line 771: (nit++)
```suggestion
			return fmt.Errorf("expected plugin name to be %q, got %q", pluginName, name)
```

machine424: noooooo

test/e2e/config_test.go
```diff
@@ -727,6 +732,48 @@ func TestUserWorkloadMonitorThanosRulerConfig(t *testing.T) {
...
+			return fmt.Errorf("expected status %d, got %d (%q)", resp.StatusCode, http.StatusOK, framework.ClampMax(b))
+		}
+
+		res, err := gabs.ParseJSON(b)
+		if err != nil {
+			return err
+		}
+		name, ok := res.Path("name").Data().(string)
+		if !ok || name != pluginName {
+			return fmt.Errorf("expected plugin name to be %q, got `%q`", pluginName, name)
```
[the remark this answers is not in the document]
machine424: noooooo

simonpasquier (commented):
/lgtm

finally :smile:

PR #2452, opened 2024-09-02
OCPBUGS-39126: disable user-defined monitoring per object

simonpasquier:
Before this change, a project owner could only disable user-defined monitoring per namespace/project (typically to prevent the `PrometheusOperatorRejectedResources` alert from firing).

To provide greater flexibility, it is now possible to exclude individual objects (e.g. `ServiceMonitor`, `PodMonitor` and `PrometheusRule`) by adding the `openshift.io/user-monitoring="true"` label to them.

### DIFF
--- a/assets/prometheus-user-workload/prometheus.yaml
+++ b/assets/prometheus-user-workload/prometheus.yaml
@@ -204,7 +204,12 @@ spec:
       operator: NotIn
       values:
       - "false"
-  podMonitorSelector: {}
+  podMonitorSelector:
+    matchExpressions:
+    - key: openshift.io/user-monitoring
+      operator: NotIn
+      values:
+      - "false"
   priorityClassName: openshift-user-critical
   probeNamespaceSelector:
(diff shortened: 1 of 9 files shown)


simonpasquier:
/hold
need to update the e2e tests

simonpasquier:
I've tested the PR to see if it could help with issues like 

After deploying the compliance operator and otel operator + enabling user-defined monitoring, I see the PrometheusOperatorRejectedResources active (as expected):

image

image

After labeling the 2 service monitor objects with `openshift.io/user-monitoring="false"`, the alert went away:

image

machine424:
/lgtm
But I’m concerned this might discourage users from addressing the root cause of the issue: deploying operators in the wrong ns.
Also, this could create confusion (I don't know if we want to document it or not) as now looking at the ns labels isn't enough to be sure a resource would be considered.

nit: I think you meant `openshift.io/user-monitoring="false"` in the PR desc.

PR #2453, opened 2024-09-02
MON-3991: set noProxy for Prometheus remote-write via the CR.

machine424:
this is a 'permanent' fix for https://issues.redhat.com/browse/OCPBUGS-38289

### DIFF
--- a/Documentation/openshiftdocs/modules/remotewritespec.adoc
+++ b/Documentation/openshiftdocs/modules/remotewritespec.adoc
@@ -35,7 +35,7 @@ link:prometheusrestrictedconfig.adoc[PrometheusRestrictedConfig]
 
 |oauth2|*monv1.OAuth2|Defines OAuth2 authentication settings for the remote write endpoint.
 
-|proxyUrl|string|Defines an optional proxy URL. It is superseded by the cluster-wide proxy when it is enabled, prioritizing the HTTPS proxy.
+|proxyUrl|string|Defines an optional proxy URL. If the cluster-wide proxy is enabled, it replaces the proxyUrl setting. The cluster-wide proxy supports both HTTP and HTTPS proxies, with HTTPS taking precedence.
 
 |queueConfig|*monv1.QueueConfig|Allows tuning configuration for remote write queue parameters.
 
--- a/Documentation/openshiftdocs/modules/remotewritespec.adoc
+++ b/Documentation/openshiftdocs/modules/remotewritespec.adoc
@@ -35,7 +35,7 @@ link:prometheusrestrictedconfig.adoc[PrometheusRestrictedConfig]
 
 |oauth2|*monv1.OAuth2|Defines OAuth2 authentication settings for the remote write endpoint.
 
-|proxyUrl|string|Defines an optional proxy URL. It is superseded by the cluster-wide proxy, if enabled.
+|proxyUrl|string|Defines an optional proxy URL. It is superseded by the cluster-wide proxy when it is enabled, prioritizing the HTTPS proxy.
 
 |queueConfig|*monv1.QueueConfig|Allows tuning configuration for remote write queue parameters.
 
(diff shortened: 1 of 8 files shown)


simonpasquier (commented):
Shouldn't we rather set `rw.ProxyFromEnvironment = true` and inject the proxy variables? This way Prometheus would select the correct proxy based on HTTP/HTTPS.

machine424:
> Shouldn't we rather set `rw.ProxyFromEnvironment = true` and inject the proxy variables? This way Prometheus would select the correct proxy based on HTTP/HTTPS.

TBH, give the latest regression, I don't want to touch/change the logic: CMO gives priority to HTTPS proxy and set it. And no body is complaining about that for now.

Plus, given all the possible places in Prometheus where proxying can be set (scraping, remote wr, read etc) and even though prometheus operator isn't exposing all of that for now, I think it's better to be explicit (set the proxy in the CR) especially that users may need to have a different proxy for each part.

simonpasquier:
Fair enough but in this case, CMO should look at the scheme of the remote-write URL to decide whether it needs to set proxy_url to HTTProxy() or HTTPSProxy().

machine424:
> Fair enough but in this case, CMO should look at the scheme of the remote-write URL to decide whether it needs to set proxy_url to HTTProxy() or HTTPSProxy().

Actually that's the behavior I'm avoiding to change: maybe some users are expecting/relying on the fact that the httpsproxy is always prioritized... That's why I think it's safer to keep it like this until someone complains about it.
(We could mention that here https://github.com/openshift/cluster-monitoring-operator/blob/c5668e364b2dfd25a3e0277c75491e0a18da9b90/Documentation/api.md?plain=1#L525)

simonpasquier:
> maybe some users are expecting/relying on the fact that the httpsproxy is always prioritized

Hmm I'm not sure that it would make sense but indeed it can be tackled with another JIRA ticket.

cc @jan--f for another view point :)

jan--f:
Afaiu http proxies work very differently from https proxies and generally one can't assume the url scheme corresponds to the the proxy to be used. E.g. sending https to an http proxy is a valid setup.
Not changing behavior seems like a good strategy here, maybe with a log message to make this behavior more visible if anyone cares to look?

Documentation/openshiftdocs/modules/remotewritespec.adoc
```diff
@@ -35,7 +35,7 @@ link:prometheusrestrictedconfig.adoc[PrometheusRestrictedConfig]
 
 |oauth2|*monv1.OAuth2|Defines OAuth2 authentication settings for the remote write endpoint.
 
-|proxyUrl|string|Defines an optional proxy URL. It is superseded by the cluster-wide proxy, if enabled.
+|proxyUrl|string|Defines an optional proxy URL. It is superseded by the cluster-wide proxy when it is enabled, prioritizing the HTTPS proxy.
```
machine424, line 38: cc @eromanova97, if there is any better wording for this.
(when cluster-wide proxy is enabled, it can have an HTTP and an HTTPS proxy and we make HTTPS take precedence)

machine424:
> Not changing behavior seems like a good strategy here, maybe with a log message to make this behavior more visible if anyone cares to look?

I added a mention of this into the docs, if it's ok for you. We can even backport it to the docs of previous versions.

Documentation/openshiftdocs/modules/remotewritespec.adoc
eromanova97, line 38: Hello, I like the change you are suggesting, I believe it is just missing the information about the replacement there. So I would suggest something like:

```suggestion
|proxyUrl|string|Defines an optional proxy URL. If the cluster-wide proxy is enabled, it replaces the proxyUrl setting. The cluster-wide proxy supports both HTTP and HTTPS proxies, with HTTPS taking precedence.
```
I hope this is correct. I was not sure if the "it" in your suggested text refers to the cluster-wide proxy or the proxyURL property :sweat_smile:

machine424: much clearer, thanks.

PR #2459, opened 2024-09-06
Clarify service.yaml comment about role bindings.

alanconway:


simonpasquier:
The change would need to go to jsonnet/components/thanos-querier.jsonnet. @eromanova97 has updated the ref docs with the correct information in https://github.com/openshift/openshift-docs/pull/80127

PR #2464, opened 2024-09-11
MON-3934: merge TestAlertManagerHasAdditionalAlertRelabelConfigs into…

machine424:
… TestMultinamespacePrometheusRule to ensure the existence of an alert with the labels the check is looking for.

move client.DeleteIfExists into framework.

clean TestMultinamespacePrometheusRule and fix the namespace.

* [ ] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #2465, opened 2024-09-11
MON-3934: move port forward setup inside retry loop for more resiliency

machine424:
avoid declaring new variables for cases as tests are not run in // or because we're already using 1.22 https://go.dev/blog/loopvar-preview

---

only the last commit is to be reviewed, the other one is part of https://github.com/openshift/cluster-monitoring-operator/pull/2464, building on it to avoid conflicts.


machine424:
/hold
needs https://github.com/openshift/cluster-monitoring-operator/pull/2464

PR #2470, opened 2024-09-13
[release-4.17] OCPBUGS-41908: filter alerts sent to Telemeter

simonpasquier:
This change ensures that only alerts with a valid severity value are sent to Telemeter.

Valid severities are defined in
https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md


Documentation/sample-metrics.md
```diff
@@ -13,7 +13,11 @@ return the full set of metrics that the Telemeter client captures:
 
 [embedmd]:# (telemetry/telemeter_query txt)
 ```txt
+<<<<<<< HEAD
```
machine424, line 16: oops

PR #2475, opened 2024-09-16
OCPBUGS-41158: add pod-metrics-reader cluster role

simonpasquier:
The new `pod-metrics-reader` cluster role allows GET *and* POST requests to the namespace-aware Thanos Querier API endpoint.


jan--f:
/lgtm
Nice, thanks!

PR #2476, opened 2024-09-16
OCPBUGS-42067: Reapply "chore: poll immediately in the e2e tests"

simonpasquier:
This reverts commit 10160fd3fefb3791121ad8f5b8bf043938085884.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -655,7 +655,7 @@ func (f *Framework) ForwardPort(t *testing.T, ns, svc string, port int) (string,
 func Poll(interval, timeout time.Duration, f func() error) error {
 	var lastErr error
 
-	err := wait.PollUntilContextTimeout(context.Background(), interval, timeout, false, func(context.Context) (bool, error) {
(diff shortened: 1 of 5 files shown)


test/e2e/framework/framework.go
```diff
@@ -655,7 +655,7 @@ func (f *Framework) ForwardPort(t *testing.T, ns, svc string, port int) (string,
 func Poll(interval, timeout time.Duration, f func() error) error {
 	var lastErr error
 
-	err := wait.PollUntilContextTimeout(context.Background(), interval, timeout, false, func(context.Context) (bool, error) {
```
machine424, line 658: All of the runs I checked were timing out, I think this is the change that may make a difference. Do you think we need to have the rest as well? (as I'm not seeing any flaky tests)

machine424:
Discussed offline, let's give this a try, thanks.
/lgtm

simonpasquier:
From the e2e-agnostic-operator job run

```
Step phase test succeeded after 1h33m48s
````

PR #2477, opened 2024-09-17
Update ConsolePlugin to include specs for i18n.loadType : Preload. 

zhuje:
- Update ConsolePlugin to include specs for `i18n.loadType : Preload`. 
- Allows 'monitoring-plugin' to be found when listed in `window.SERVER_FLAGS.i18nNamespaces` when inspecting in a web console. 
- Allows monitoring-plugin > console-extension.json to use local translations for tabs under Observe  > Alerting, Metrics, Targets, and Dashboards.
- Related to this [PR openshift/monitoring#184](https://github.com/openshift/monitoring-plugin/pull/184), which consolidates the translations of the monitoring plugin to only use translation files from the monitoring plugin (i18n translation files named 'plugin__monitoring-plugin'). Previously translations were used from both the openshift/console (i18n translation files named 'public') and openshift/monitoring-plugin.
 


assets/monitoring-plugin/console-plugin.yaml
```diff
@@ -16,3 +16,5 @@ spec:
       port: 9443
     type: Service
   displayName: monitoring-plugin
+  i18n:
```
machine424, line 19: these are generated via jsonnet, you'd need to add that here https://github.com/openshift/cluster-monitoring-operator/blob/df5e2abc418929264edc59193bc3641b811d461b/jsonnet/components/monitoring-plugin.libsonnet#L75, then `make build-jsonnet`

Thanks for this.

machine424:
/lgtm
thanks.

PR #2479, opened 2024-09-19
NO-JIRA: clean-up generated dashboards

simonpasquier:
This change also removes the dependency on the etcd mixin since everything is now managed by the etcd operator.


slashpai (approved):
/lgtm

marioferh (commented):
/lgtm

PR #2491, opened 2024-10-03
MON-4042: new label for acm_console_page_count metric

zlayne:

### DIFF
--- a/Documentation/data-collection.md
+++ b/Documentation/data-collection.md
@@ -4,7 +4,8 @@ Red Hat values our customers' experience and privacy. It is important to us that
 
 OpenShift 4 clusters send anonymized telemetry back to Red Hat about the following attributes. The telemetry is gathered by referencing your cluster ID and pull secret:
 
-[embedmd]:# (../manifests/0000_50_cluster-monitoring-operator_04-config.yaml)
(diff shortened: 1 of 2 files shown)


Documentation/data-collection.md
```diff
@@ -758,7 +758,7 @@ data:
     # owners: (https://github.com/orgs/stolostron/teams/search-admin, @acm-observability-search)
     #
     # acm_console_page_count:sum counts the total number of visits for each page in ACM console.
-    - '{__name__="acm_console_page_count:sum", page=~"overview-classic|overview-fleet|search|search-details|clusters|application|governance"}'
+    - '{__name__="acm_console_page_count:sum", page=~"overview-fleet|search|search-details|clusters|virtual-machines|application|governance"}'
```
simonpasquier, line 761: you probably want to keep the deprecated `overview-classic` label because otherwise it might be removed from the Telemeter server allow-list (need @saswatamcode @philipgough confirmation here). But a comment + TODO would be useful.

zlayne: @simonpasquier thanks for the comment. If we remove the `overview-classic` label here will we lose history? If not then we are okay not to receive any new data.

PR #2492, opened 2024-10-04
NO-JIRA: use scrape class to avoid TLS config repetitions

simonpasquier:

### DIFF
--- a/assets/alertmanager-user-workload/service-monitor.yaml
+++ b/assets/alertmanager-user-workload/service-monitor.yaml
@@ -17,11 +17,9 @@ spec:
     port: metrics
     scheme: https
     tlsConfig:
-      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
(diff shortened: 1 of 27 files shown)


slashpai (commented):
/retest-required

simonpasquier:
@slashpai the CI failure looks legit

simonpasquier:
@jan--f this is a small improvement to reduce coupling between the service monitors and the platform Prometheus: instead of hardcoding file paths in all service monitors, we can refer to the scrape class name defined in the Prometheus object. Eventually we can modify other core operators to do the same.

jan--f:
Sorry, lost track of this. Feel free to mark this critical if this should make it into 4.18. I assume though it can wait?
/lgtm

simonpasquier:
> I assume though it can wait?

Totally.

cc @frzifus, if/when this merges, you won't be able to reuse the service monitors without supporting scrape classes in the opentelemetry operator (or a similar mechanism).

PR #2494, opened 2024-10-10
OCPBUGS-43690: chore: introduce a config parsing check via a strict unmarshaller for…

machine424:
… the monitoring and set CMO to Upgradeable=false if errors are found

The strict unmarshaller is currently in advisory mode. Setting CMO to Upgradeable=false ensures that configurations meet strict unmarshaller checks before it replacesthe current relaxed one.


test/e2e/config_test.go
```diff
@@ -49,14 +48,25 @@ func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
...
 		},
 	}
 	f.MustCreateOrUpdateConfigMap(t, cm)
 
+	t.Log("asserting that CMO only goes Upgradeable=false as unknown fields are only caught by the strict unmarshaller")
+	f.AssertOperatorCondition(configv1.OperatorUpgradeable, configv1.ConditionFalse)(t)
+	f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionFalse)(t)
+	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionTrue)(t)
+
+	// Even when an invalid configuration is caught by both unmarshallers, the operator is still set
```
machine424, line 61: FYI

pkg/manifests/config.go
```diff
@@ -46,8 +45,25 @@ const (
...
+// InvalidConfigWarning is a disguised error.
+type InvalidConfigWarning error
+
+type UpgradeBlockingInvalidConfigWarning struct {
+	ConfigMap string
+	Err       error
+}
+
+func (e *UpgradeBlockingInvalidConfigWarning) Error() string {
+	return fmt.Sprintf("configuration within the %s ConfigMap is invalid and should be adjusted to make the operator Upgradeable again: %s", e.ConfigMap, e.Err)
```
simonpasquier, line 62:
```suggestion
	return fmt.Sprintf("configuration in the %q ConfigMap is invalid and should be fixed: %s", e.ConfigMap, e.Err)
```

pkg/operator/operator.go
```diff
@@ -951,78 +961,80 @@ func (o *Operator) loadConsoleConfig(ctx context.Context) (*configv1.Console, er
...
+		return nil, warning, fmt.Errorf("the Cluster Monitoring ConfigMap could not be parsed: %w", err)
 	}
 
-	return cParsed, nil
+	return cParsed, warning, nil
 }
 
-func (o *Operator) Config(ctx context.Context, key string) (*manifests.Config, error) {
-	c, err := o.loadConfig(key)
+func (o *Operator) Config(ctx context.Context, key string) (*manifests.Config, manifests.InvalidConfigWarning, error) {
```
simonpasquier, line 1018: should we return a slice of warnings? I feel that it would be easier to reason about.
```suggestion
func (o *Operator) Config(ctx context.Context, key string) (*manifests.Config, []manifests.InvalidConfigWarning, error) {
```

machine424: why not.

machine424:
cc @simonpasquier 
I think we're good now.

pkg/operator/operator.go
```diff
@@ -1066,15 +1107,9 @@ func (o Operator) storageNotConfiguredMessage() string {
...
 	// if cluster version cannot be obtained due to any failure, point to the
 	// latest documentation
-	cv, err := o.client.GetClusterVersion(context.Background(), "version")
-	if err != nil {
-		klog.Warningf("failed to find the cluster version: %s", err)
-		return latestDocMsg
-	}
-
-	v, err := semver.Make(cv.Status.Desired.Version)
+	v, err := o.getClusterVersion()
```
simonpasquier, line 1110: I would argue that CMO should pull the cluster version only once at startup and save it in the Operator struct but :shrug:

pkg/operator/operator.go
```diff
@@ -745,12 +745,28 @@ func newUWMTaskSpec(targetName string, task tasks.Task) *tasks.TaskSpec {
...
+	if err != nil {
+		klog.Errorf("error occurred while setting Upgradeable status: %v", err)
+	}
+}
+
 func (o *Operator) sync(ctx context.Context, key string) error {
-	config, err := o.Config(ctx, key)
+	config, warnings, err := o.Config(ctx, key)
+
+	invalidConfigReason := "InvalidConfiguration"
```
simonpasquier, line 758: hmm shouldn't we try to have different reasons for platform and user-defined configs to help managed service SREs? Like we did for tasks.

pkg/operator/operator.go
machine424: I admit I just re-used the existing API supposing it works as expected (as it kind of already handles the same use case) but I agree we can do better. I reverted the reference linking as it'd be too much for this PR, also it was just a nice to have I wanted to add.
We could always add it back in another PR.

pkg/operator/operator.go
machine424: for now, we've only done that for 4.18 https://issues.redhat.com/browse/OCPBUGS-33863, we could backport that, I can take care or that in another PR.

machine424:
cc @simonpasquier if it's ok for you now.

juzhao:
tested with PR, seems the issue is not fixed for 4.17, see comments in https://issues.redhat.com/browse/OCPBUGS-43690

machine424:
@simonpasquier I think the `backport-risk-assessed` comment was written but never published :)

pkg/operator/operator.go
simonpasquier: sounds good

simonpasquier (commented):
/lgtm

PR #2503, opened 2024-10-21
MON-4057: Expose `scrapeInterval` setting for UWM Prometheus

slashpai:

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -484,6 +484,7 @@ The `PrometheusRestrictedConfig` resource defines the settings for the Prometheu
 
 | Property | Type | Description |
 | -------- | ---- | ----------- |
+| scrapeInterval | string | Configures the default interval between consecutive scrapes in case the `ServiceMonitor` or `PodMonitor` resource does not specify any value. The interval must be set between 5 seconds and 5 minutes. The value can be expressed in: seconds (for example `30s`.), minutes (for example `1m`.) or a mix of minutes and seconds (for example `1m30s`.). The default value is `30s`. |
 | additionalAlertmanagerConfigs | [][AdditionalAlertmanagerConfig](#additionalalertmanagerconfig) | Configures additional Alertmanager instances that receive alerts from the Prometheus component. By default, no additional Alertmanager instances are configured. |
 | enforcedLabelLimit | *uint64 | Specifies a per-scrape limit on the number of labels accepted for a sample. If the number of labels exceeds this limit after metric relabeling, the entire scrape is treated as failed. The default value is `0`, which means that no limit is set. |
 | enforcedLabelNameLengthLimit | *uint64 | Specifies a per-scrape limit on the length of a label name for a sample. If the length of a label name exceeds this limit after metric relabeling, the entire scrape is treated as failed. The default value is `0`, which means that no limit is set. |
--- a/Documentation/openshiftdocs/modules/prometheusrestrictedconfig.adoc
+++ b/Documentation/openshiftdocs/modules/prometheusrestrictedconfig.adoc
@@ -18,6 +18,8 @@ Appears in: link:userworkloadconfiguration.adoc[UserWorkloadConfiguration]
 [options="header"]
 |===
 | Property | Type | Description 
+|scrapeInterval|string|Configures the default interval between consecutive scrapes in case the `ServiceMonitor` or `PodMonitor` resource does not specify any value. The interval must be set between 5 seconds and 5 minutes. The value can be expressed in: seconds (for example `30s`.), minutes (for example `1m`.) or a mix of minutes and seconds (for example `1m30s`.). The default value is `30s`.
+
 |additionalAlertmanagerConfigs|[]link:additionalalertmanagerconfig.adoc[AdditionalAlertmanagerConfig]|Configures additional Alertmanager instances that receive alerts from the Prometheus component. By default, no additional Alertmanager instances are configured.
 
 |enforcedLabelLimit|*uint64|Specifies a per-scrape limit on the number of labels accepted for a sample. If the number of labels exceeds this limit after metric relabeling, the entire scrape is treated as failed. The default value is `0`, which means that no limit is set.
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/alecthomas/units"
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/prometheus/common/model"
 	v1 "k8s.io/api/core/v1"
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
@@ -124,6 +125,35 @@ func (c Config) GetThanosRulerAlertmanagerConfigs() []AdditionalAlertmanagerConf
 	}
 
 	return alertmanagerConfigs
+}
+
+func scrapeIntervalLimits() (model.Duration, model.Duration) {
+	lowerLimit, _ := model.ParseDuration("5s")
+	upperLimit, _ := model.ParseDuration("5m")
+	return lowerLimit, upperLimit
+}
+
+func (u *UserWorkloadConfiguration) check() error {
+	if u == nil {
+		return nil
+	}
+
+	if u.Prometheus == nil || u.Prometheus.ScrapeInterval == "" {
+		return nil
+	}
+
+	scrapeInterval, err := model.ParseDuration(u.Prometheus.ScrapeInterval)
+
+	if err != nil {
+		return fmt.Errorf("invalid scrape interval value: %w", err)
+	}
+
+	allowedLowerLimit, allowedUpperLimit := scrapeIntervalLimits()
+
+	if (scrapeInterval < allowedLowerLimit) || (scrapeInterval > allowedUpperLimit) {
+		return fmt.Errorf("scrape interval value %q outside of the allowed range [%q, %q]", u.Prometheus.ScrapeInterval, allowedLowerLimit, allowedUpperLimit)
+	}
+	return nil
 }
 
 type Images struct {
@@ -541,6 +571,11 @@ func NewUserConfigFromString(content string) (*UserWorkloadConfiguration, error)
 	}
 
 	u.applyDefaults()
+
+	if err := u.check(); err != nil {
+		return nil, err
+	}
+
 	return u, nil
 }
 
--- a/pkg/manifests/config_test.go
+++ b/pkg/manifests/config_test.go
@@ -484,6 +484,64 @@ func TestLoadEnforcedBodySizeLimit(t *testing.T) {
 	}
 }
 
+func TestScrapeIntervalUWMPreCheck(t *testing.T) {
+	for _, tc := range []struct {
+		name          string
+		uwmconfig     string
+		expectedError bool
+	}{
+		{
+			name:          "default",
+			uwmconfig:     "",
+			expectedError: false,
+		},
+		{
+			name: "scrapeInterval valid within limits",
+			uwmconfig: `prometheus:
+  scrapeInterval: 15s
+  `,
+			expectedError: false,
+		},
+		{
+			name: "scrapeInterval valid within limits-mix-of-minutes-seconds",
+			uwmconfig: `prometheus:
+  scrapeInterval: 1m30s
+  `,
+			expectedError: false,
+		},
+		{
+			name: "scrapeInterval < allowed lower limit",
+			uwmconfig: `prometheus:
+  scrapeInterval: 2s
+  `,
+			expectedError: true,
+		},
+		{
+			name: "scrapeInterval > allowed upper limit",
+			uwmconfig: `prometheus:
+  scrapeInterval: 10m
+  `,
+			expectedError: true,
+		},
+		{
+			name: "incorrect scrape interval value",
+			uwmconfig: `prometheus:
+  scrapeInterval: 1234www
+  `,
+			expectedError: true,
+		},
+	} {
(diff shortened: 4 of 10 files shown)


pkg/manifests/types.go
```diff
@@ -601,6 +601,9 @@ type AlertmanagerUserWorkloadConfig struct {
 // The `PrometheusRestrictedConfig` resource defines the settings for the
 // Prometheus component that monitors user-defined projects.
 type PrometheusRestrictedConfig struct {
+	// Configures Interval between consecutive scrapes.
```
simonpasquier, line 604:
```suggestion
	// Configures the default interval between consecutive scrapes in case the ServiceMonitor or PodMonitor resource doesn't specify any value.
	// Allowed values are between 10s and 300s (5m).
```

simonpasquier (commented):
I suggest to validate the input value to be within acceptable boundaries (e.g. not too low and not too high).

simonpasquier:
/cc @eromanova97 

if you want to check the updated documentation page.

Documentation/openshiftdocs/modules/prometheusrestrictedconfig.adoc
```diff
@@ -18,6 +18,8 @@ Appears in: link:userworkloadconfiguration.adoc[UserWorkloadConfiguration]
 [options="header"]
 |===
 | Property | Type | Description 
+|scrapeInterval|string|Configures Interval between consecutive scrapes. The default value is `30s`.
```
eromanova97, line 21:
```suggestion
|scrapeInterval|string|Configures interval between consecutive scrapes. The default value is `30s`.
```

pkg/manifests/types.go
eromanova97: I would just add to this suggestion to replace `doesn't` for `does not`. Thanks!

eromanova97:
Added very small suggestions, thank you!

slashpai:
addressed review comments
@simonpasquier @eromanova97 PTAL

pkg/manifests/config.go
```diff
@@ -126,6 +127,27 @@ func (c Config) GetThanosRulerAlertmanagerConfigs() []AdditionalAlertmanagerConf
 	return alertmanagerConfigs
 }
 
+func (c Config) HasInvalidScrapeIntervalDuration() bool {
```
simonpasquier, line 130: It'd be better to return an error that would be wrapped by the caller with `ErrUserWorkloadValidation`.
```suggestion
func (c Config) checkUserWorkloadPrometheusConfig() error {
```

pkg/manifests/config.go
```diff
@@ -126,6 +127,27 @@ func (c Config) GetThanosRulerAlertmanagerConfigs() []AdditionalAlertmanagerConf
...
+	}
+
+	if c.UserWorkloadConfiguration.Prometheus == nil || c.UserWorkloadConfiguration.Prometheus.ScrapeInterval == "" {
+		return false
+	}
+
+	scrapeInterval, err := model.ParseDuration(c.UserWorkloadConfiguration.Prometheus.ScrapeInterval)
+
+	if err != nil {
+		return true
```
simonpasquier, line 142:
```suggestion
		return fmt.Errorf("invalid scrape interval value: %w", err)
```

pkg/manifests/config.go
```diff
@@ -126,6 +127,27 @@ func (c Config) GetThanosRulerAlertmanagerConfigs() []AdditionalAlertmanagerConf
...
+	}
+
+	allowedLowerLimit, _ := model.ParseDuration("5s")
+	allowedUpperLimit, _ := model.ParseDuration("5m")
+
+	return (scrapeInterval < allowedLowerLimit) || (scrapeInterval > allowedUpperLimit)
```
simonpasquier, line 148: It's better user experience to report why the value isn't valid.
```suggestion
	if (scrapeInterval < allowedLowerLimit) || (scrapeInterval > allowedUpperLimit) {
		return fmt.Errorf("scrape interval value %q outside of the allowed range [%q, %q]", c.UserWorkloadConfiguration.Prometheus.ScrapeInterval, allowedLowerLimit, allowedUpperLimit)
```

pkg/manifests/config.go
```diff
@@ -126,6 +127,27 @@ func (c Config) GetThanosRulerAlertmanagerConfigs() []AdditionalAlertmanagerConf
...
+		return false
+	}
+
+	scrapeInterval, err := model.ParseDuration(c.UserWorkloadConfiguration.Prometheus.ScrapeInterval)
+
+	if err != nil {
+		return true
+	}
+
+	allowedLowerLimit, _ := model.ParseDuration("5s")
```
simonpasquier, line 145: (nit) can we declare constants?

pkg/manifests/types.go
```diff
@@ -601,6 +601,9 @@ type AlertmanagerUserWorkloadConfig struct {
 // The `PrometheusRestrictedConfig` resource defines the settings for the
 // Prometheus component that monitors user-defined projects.
 type PrometheusRestrictedConfig struct {
+	// Configures the default interval between consecutive scrapes in case the ServiceMonitor or PodMonitor resource does not specify any value.
```
simonpasquier, line 604:
```suggestion
	// Configures the default interval between consecutive scrapes in case the ServiceMonitor or PodMonitor resource does not specify any value.
	// The interval needs to be greater than or equal to 5 seconds and less than or equal to 5 minutes. 
	// The value can be expressed as:
	// * seconds, for example `30s`.
	// * minutees, for example `1m`.
	// * a mix of minutes and seconds, for example `1m30s`.
```

pkg/manifests/config.go
```diff
@@ -436,6 +458,9 @@ func (c *Config) LoadEnforcedBodySizeLimit(pcr PodCapacityReader, ctx context.Co
 }
 
 func (c *Config) Precheck() error {
+	if c.HasInvalidScrapeIntervalDuration() {
+		return fmt.Errorf("%w: scrapeInterval specified should be between 5s and 5m", ErrUserWorkloadValidation)
```
machine424, line 462: nit: if we could avoid having the limits hardcoded here (we can just remove them) + the error may be not related to limits (it may be a parsing error), we could build the errors in `HasInvalidScrapeIntervalDuration` and just forward them here.

pkg/manifests/manifests.go
```diff
@@ -329,7 +329,8 @@ var (
 )
 
 var (
-	ErrConfigValidation = fmt.Errorf("invalid value for config")
+	ErrConfigValidation       = fmt.Errorf("invalid value for config")
+	ErrUserWorkloadValidation = fmt.Errorf("invalid value for user-workload config")
```
machine424, line 333: I don't think we need `ErrUserWorkloadValidation` in here, maybe I'm missing sth, but we don't have any check on it (not used) in the caller code.

(not related to the PR: same for `ErrConfigValidation` we only check it in the tests, I think we can get rid of it.)

slashpai:
Fixing one more thing

pkg/manifests/manifests.go
slashpai: I removed it, I will takecare of ErrConfigValidation in separate PR :)

pkg/manifests/config.go
slashpai: We can't create constants for multi valued IIUC. I changed to a method for better readability

slashpai:
This is good for review again

Documentation/api.md
```diff
@@ -484,6 +484,7 @@ The `PrometheusRestrictedConfig` resource defines the settings for the Prometheu
 
 | Property | Type | Description |
 | -------- | ---- | ----------- |
+| scrapeInterval | string | Configures the default interval between consecutive scrapes in case the ServiceMonitor or PodMonitor resource does not specify any value. The interval needs to be greater than or equal to 5 seconds and less than or equal to 5 minutes. The value can be expressed as: * seconds, for example `30s`. * minutees, for example `1m`. * a mix of minutes and seconds, for example `1m30s` The default value is `30s`. |
```
simonpasquier, line 487: The formatting is a bit off.
```suggestion
| scrapeInterval | string | Configures the default interval between consecutive scrapes in case the ServiceMonitor or PodMonitor resource does not specify any value. The interval needs to be greater than or equal to 5 seconds and less than or equal to 5 minutes. The value can be expressed as seconds (for example `30s`.), minutes (for example `1m`.) or a mix of minutes and seconds (for example `1m30s`). The default value is `30s`. |
```

pkg/manifests/config_test.go
```diff
@@ -318,6 +318,68 @@ func TestLoadEnforcedBodySizeLimit(t *testing.T) {
...
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			c := NewDefaultConfig()
+			uwc, err := NewUserConfigFromString(tc.uwmconfig)
+			require.NoError(t, err)
+			c.UserWorkloadConfiguration = uwc
+			err = c.UserWorkloadPrecheck()
+			if err != nil && tc.expectedError {
+				return
+			}
```
simonpasquier, line 376:
```suggestion
			if tc.expectedError {
				require.Error(t, err)
				return
			}
```

pkg/manifests/config.go
```diff
@@ -465,6 +495,14 @@ func (c *Config) Precheck() error {
 	return nil
 }
 
+func (c *Config) UserWorkloadPrecheck() error {
```
simonpasquier, line 498: can't we move this directly to NewUserConfigFromString()?

Documentation/api.md
eromanova97, line 487:
```suggestion
| scrapeInterval | string | Configures the default interval between consecutive scrapes in case the `ServiceMonitor` or `PodMonitor` resource does not specify any value. The interval must be set between 5 seconds and 5 minutes. The value can be expressed in seconds, for example `30s`. * minutees, for example `1m`. * a mix of minutes and seconds, for example `1m30s` The default value is `30s`. |
```

I added some additional suggestions, plus I agree with Simon's suggestions.

Documentation/openshiftdocs/modules/prometheusrestrictedconfig.adoc
```diff
@@ -18,6 +18,8 @@ Appears in: link:userworkloadconfiguration.adoc[UserWorkloadConfiguration]
 [options="header"]
 |===
 | Property | Type | Description 
+|scrapeInterval|string|Configures the default interval between consecutive scrapes in case the ServiceMonitor or PodMonitor resource does not specify any value. The interval needs to be greater than or equal to 5 seconds and less than or equal to 5 minutes. The value can be expressed as: * seconds, for example `30s`. * minutees, for example `1m`. * a mix of minutes and seconds, for example `1m30s` The default value is `30s`.
```
eromanova97, line 21: Same suggestions here.

slashpai:
@simonpasquier @eromanova97 updated with changes

pkg/manifests/config.go
```diff
@@ -126,6 +127,35 @@ func (c Config) GetThanosRulerAlertmanagerConfigs() []AdditionalAlertmanagerConf
 	return alertmanagerConfigs
 }
 
+func scrapeIntervalLimits() (model.Duration, model.Duration) {
+	lowerLimit, _ := model.ParseDuration("5s")
+	upperLimit, _ := model.ParseDuration("5m")
+	return lowerLimit, upperLimit
+}
+
+func (u *UserWorkloadConfiguration) checkPrometheusConfig() error {
```
simonpasquier, line 136:
```suggestion
func (u *UserWorkloadConfiguration) check() error {
```

pkg/manifests/config.go
```diff
@@ -466,6 +496,14 @@ func (c *Config) Precheck() error {
...
 }
 
+func (u *UserWorkloadConfiguration) Precheck() error {
+	if err := u.checkPrometheusConfig(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
```
simonpasquier, line 506:
```suggestion
```

pkg/manifests/config.go
```diff
@@ -541,6 +579,11 @@ func NewUserConfigFromString(content string) (*UserWorkloadConfiguration, error)
 	}
 
 	u.applyDefaults()
+
+	if err := u.Precheck(); err != nil {
```
simonpasquier, line 583:
```suggestion
	if err := u.check(); err != nil {
```

pkg/manifests/config_test.go
```diff
@@ -484,6 +484,66 @@ func TestLoadEnforcedBodySizeLimit(t *testing.T) {
...
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			c := NewDefaultConfig()
+			uwc, err := NewUserConfigFromString(tc.uwmconfig)
+			if tc.expectedError {
+				require.Error(t, err)
+				return
+			}
+			c.UserWorkloadConfiguration = uwc
```
simonpasquier, line 541:
```suggestion
```

pkg/manifests/config_test.go
```diff
@@ -484,6 +484,66 @@ func TestLoadEnforcedBodySizeLimit(t *testing.T) {
...
+		{
+			name: "incorrect scrape interval value",
+			uwmconfig: `prometheus:
+  scrapeInterval: 1234www
+  `,
+			expectedError: true,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			c := NewDefaultConfig()
```
simonpasquier, line 535:
```suggestion
```

slashpai:
@simonpasquier addressed comments PTAL

slashpai:
cc @juzhao

juzhao:
tested with PR, no issue
/label qe-approved

simonpasquier (commented):
/lgtm

PR #2515, opened 2024-11-07
MON-4078: report vendor_model:node_accelerator_cards:sum to Telemetry

simonpasquier:


simonpasquier:
/hold
needs https://github.com/openshift/node_exporter/pull/154 and explicit approval

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -632,6 +632,15 @@ data:
...
     - '{__name__="profile:cluster_monitoring_operator_collection_profile:max"}'
     #
+    # owners: (@openshift/openshift-team-monitoring)
+    #
+    # vendor_model:node_accelerator_cards:sum reports the total number of accelerator cards
+    # in the cluster per vendor and model.
+    # Possible label values are:
+    #   vendor: NVIDIA
+    #   model: 100, RTX_A6000, RTX_4090, A40, V100
+    - '{__name__="vendor_model:node_accelerator_cards:sum",vendor=~"NVIDIA",model=~"A100|RTX_A6000|RTX_4090|A40|V100"}'
```
yevgeny-shnaidman, line 642: do we need to specify vendor and model?  every vendor_model:node_accelerator_cards:sum metric should be sent to telemetry, no?

simonpasquier: This is the recommended approach to control exactly what's being sent to telemetry. As I mentioned in https://github.com/openshift/node_exporter/pull/154#discussion_r1831267742, the mapping to vendors and models should eventually be stored in the CMO repository so we can ensure that the allow-list is kept in sync.

yevgeny-shnaidman: Lets say we will have 20 models for NVIDIA, 20 for AMD and 10 for QUALCOMM: does it mean that we will have to specify all the 50 models in this line, in addition to the vendors?  Can't we maybe just specify the vendors without models? The original metrics (node_accelerator_card_info) is produced by the code that already verifies the vendors and the models.

jan--f: Just to expand on the ast question: Having to adjust this when more label values appear is the feature here. We want to guard our telemetry pipeline. The later change might look like you suggest or change to a regex of some sort. Point is we want to be aware when more series are shipped to telemetry

PR #2523, opened 2024-11-19
OCPBUGS-38655: take cluster proxy into account for UWM Prometheus rem…

machine424:
…ote write config


jan--f:
retest

PR #2524, opened 2024-11-22
OCPBUGS-35726: remove certificate hash annotation for monitoring-plugin

simonpasquier:
With https://github.com/openshift/monitoring-plugin/pull/264, the monitoring plugin binary reloads the certificate and key files from disk automatically.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -2804,7 +2804,7 @@ func (f *Factory) MonitoringPlugin() (*consolev1.ConsolePlugin, error) {
 	return f.NewConsolePlugin(f.assets.MustNewAssetSlice(MonitoringPlugin))
 }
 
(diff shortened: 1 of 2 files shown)


simonpasquier:
/hold

this needs https://github.com/openshift/monitoring-plugin/pull/264 first

simonpasquier:
/retest-required
/hold cancel
The monitoring-plugin PR has merged.

jan--f:
The last failed test isn't required, looks like an issue with the test itself.
/lgtm

machine424:
Glad to see this happen, thanks Simon!
another
/lgtm
(note that due to the [unfamous](https://github.com/openshift/cluster-monitoring-operator/blob/4d9bb6ae9c512211e071846ba9fbc42cf13d8402/Documentation/development.md?plain=1#L52) logic/limitation in library-go, the label will not disappear, but I don't think that's a problem.)

PR #2526, opened 2024-11-25
OCPBUGS-44971: Add new metrics for OpenShift logging telemetry

xperimental:
Backport of #2512 to `release-4.17`. Updated with fixes from #2528.

* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

PR #2528, opened 2024-11-27
OCPBUGS-44724: Use explicit name label for logging telemetry metrics

xperimental:
Updates the metrics queries for the logging telemetry to explicitly use the `__name__` label. Needs to be backported to 4.14 - 4.17. See also #2526 (backport to 4.17) and #2529 (this change on 4.19)


xperimental:
If I read the prow status correctly then it is still waiting for the `cherry-pick-approved` label.

PR #2529, opened 2024-11-28
OCPBUGS-45179: Use explicit name label for logging telemetry metrics

xperimental:
Updates the metrics queries for the logging telemetry to explicitly use the `__name__` label.

See also #2526 (backport to 4.18) and #2528 (backport of original change to 4.17 including this fix).


marioferh (commented):
/lgtm

PR #2530, opened 2024-11-30
MGMT-19498: Adding accelerators configuration file to the node-exporter

yevgeny-shnaidman:
1) adding ConfigMap that contains the yaml structure describing
   monitored accelerators per vendor/model
2) mapping accelerators ConfigMap into the node-exporter pods 3) adding env variable that contains the exact path for the mapped
   ConfigMap

### DIFF
--- a/assets/node-exporter/daemonset.yaml
+++ b/assets/node-exporter/daemonset.yaml
@@ -79,6 +79,9 @@ spec:
         - mountPath: /var/node_exporter/textfile
           name: node-exporter-textfile
           readOnly: true
+        - mountPath: /var/node_exporter/accelerators_collector_config
+          name: node-exporter-accelerators-collector-config
+          readOnly: true
         workingDir: /var/node_exporter/textfile
       - args:
         - --secure-listen-address=[$(IP)]:9100
@@ -182,6 +185,12 @@ spec:
       - name: node-exporter-kube-rbac-proxy-config
         secret:
           secretName: node-exporter-kube-rbac-proxy-config
+      - configMap:
+          items:
+          - key: config.yaml
+            path: config.yaml
+          name: node-exporter-accelerators-collector-config
+        name: node-exporter-accelerators-collector-config
   updateStrategy:
     rollingUpdate:
       maxUnavailable: 10%
--- a/jsonnet/components/node-exporter.libsonnet
+++ b/jsonnet/components/node-exporter.libsonnet
@@ -3,6 +3,26 @@ local textfileVolumeName = 'node-exporter-textfile';
 local tlsVolumeName = 'node-exporter-tls';
 local wtmpPath = '/var/log/wtmp';
 local wtmpVolumeName = 'node-exporter-wtmp';
+local configDir = '/var/node_exporter/accelerators_collector_config';
+local configVolumeName = 'node-exporter-accelerators-collector-config';
+local acceleratorsConfigFileName = 'config.yaml';
+local acceleratorsConfigMapName = 'node-exporter-accelerators-collector-config';
+local acceleratorsConfigData = [
+  {
+    vendorName: 'NVIDIA',
+    vendorID: '0x10de',
+    models: [
+      { pciID: '0x20b5', pciName: 'A100' },
+      { pciID: '0x2230', pciName: 'RTX_A6000' },
+      { pciID: '0x2717', pciName: 'RTX_4090' },
+      { pciID: '0x2235', pciName: 'A40' },
+      { pciID: '0x1df5', pciName: 'V100' },
+      { pciID: '0x20f1', pciName: 'A100 40G' },
+      { pciID: '0x1ff2', pciName: 'T400 4GB' },
+      { pciID: '0x1eb8', pciName: 'Tesla T4' },
+    ],
+  },
+];
 
 local nodeExporter = import 'github.com/prometheus-operator/kube-prometheus/jsonnet/kube-prometheus/components/node-exporter.libsonnet';
 local generateSecret = import '../utils/generate-secret.libsonnet';
@@ -269,11 +289,18 @@ function(params)
                           exec /bin/node_exporter "$0" "$@"
                         |||,
                       ],
-                      volumeMounts+: [{
-                        mountPath: textfileDir,
-                        name: textfileVolumeName,
-                        readOnly: true,
-                      }],
+                      volumeMounts+: [
+                        {
+                          mountPath: textfileDir,
+                          name: textfileVolumeName,
+                          readOnly: true,
+                        },
+                        {
+                          mountPath: configDir,
+                          name: configVolumeName,
+                          readOnly: true,
+                        },
+                      ],
                       workingDir: textfileDir,
                       resources+: {
                         requests+: {
@@ -324,6 +351,18 @@ function(params)
                   secretName: 'node-exporter-kube-rbac-proxy-config',
                 },
               },
+              {
+                name: configVolumeName,
+                configMap: {
+                  name: acceleratorsConfigMapName,
+                  items: [
+                    {
+                      key: acceleratorsConfigFileName,
+                      path: acceleratorsConfigFileName,
+                    },
+                  ],
+                },
+              },
             ],
             securityContext: {},
             priorityClassName: 'system-cluster-critical',
@@ -348,6 +387,18 @@ function(params)
             }],
           },
         ],
+      },
+    },
+
+    acceleratorsCollectorConfigmap: {
+      apiVersion: 'v1',
+      kind: 'ConfigMap',
+      metadata: {
+        name: acceleratorsConfigMapName,
+        namespace: cfg.namespace,
+      },
+      data: {
+        [acceleratorsConfigFileName]: std.manifestYamlDoc(acceleratorsConfigData),
       },
     },
   }
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -138,6 +138,7 @@ var (
 	NodeExporterMinimalServiceMonitor      = "node-exporter/minimal-service-monitor.yaml"
 	NodeExporterPrometheusRule             = "node-exporter/prometheus-rule.yaml"
 	NodeExporterKubeRbacProxySecret        = "node-exporter/kube-rbac-proxy-secret.yaml"
+	NodeExporterAcceleratorsConfigMap      = "node-exporter/accelerators-collector-configmap.yaml"
 
 	PrometheusK8sClusterRoleBinding               = "prometheus-k8s/cluster-role-binding.yaml"
 	PrometheusK8sRoleBindingConfig                = "prometheus-k8s/role-binding-config.yaml"
@@ -1031,6 +1032,10 @@ func (f *Factory) NodeExporterPrometheusRule() (*monv1.PrometheusRule, error) {
 
 func (f *Factory) NodeExporterRBACProxySecret() (*v1.Secret, error) {
 	return f.NewSecret(f.assets.MustNewAssetSlice(NodeExporterKubeRbacProxySecret))
+}
+
+func (f *Factory) NodeExporterAcceleratorsCollectorConfigMap() (*v1.ConfigMap, error) {
+	return f.NewConfigMap(f.assets.MustNewAssetSlice(NodeExporterAcceleratorsConfigMap))
 }
 
 func (f *Factory) PrometheusK8sClusterRoleBinding() (*rbacv1.ClusterRoleBinding, error) {
--- a/pkg/tasks/nodeexporter.go
+++ b/pkg/tasks/nodeexporter.go
@@ -94,6 +94,15 @@ func (t *NodeExporterTask) Run(ctx context.Context) error {
 		return fmt.Errorf("reconciling node-exporter Service failed: %w", err)
 	}
 
+	cm, err := t.factory.NodeExporterAcceleratorsCollectorConfigMap()
+	if err != nil {
+		return fmt.Errorf("initializing node-exporter accelerators collector ConfigMap failed: %w", err)
+	}
+	err = t.client.CreateOrUpdateConfigMap(ctx, cm)
+	if err != nil {
+		return fmt.Errorf("reconciling node-exporter accelerators collector ConfigMap failed: %w", err)
+	}
+
 	ds, err := t.factory.NodeExporterDaemonSet()
 	if err != nil {
 		return fmt.Errorf("initializing node-exporter DaemonSet failed: %w", err)
--- a/assets/node-exporter/accelerators-collector-configmap.yaml
+++ b/assets/node-exporter/accelerators-collector-configmap.yaml
@@ -1,0 +1,29 @@
+apiVersion: v1
+data:
+  config.yaml: |-
+    - "models":
+      - "pciID": "0x20b5"
+        "pciName": "A100"
+      - "pciID": "0x2230"
+        "pciName": "RTX_A6000"
+      - "pciID": "0x2717"
+        "pciName": "RTX_4090"
+      - "pciID": "0x2235"
+        "pciName": "A40"
+      - "pciID": "0x1df5"
+        "pciName": "V100"
+      - "pciID": "0x20f1"
+        "pciName": "A100 40G"
+      - "pciID": "0x1ff2"
+        "pciName": "T400 4GB"
+      - "pciID": "0x1eb8"
+        "pciName": "Tesla T4"
+      "vendorID": "0x10de"
+      "vendorName": "NVIDIA"
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/managed-by: cluster-monitoring-operator
+    app.kubernetes.io/part-of: openshift-monitoring
+  name: node-exporter-accelerators-collector-config
+  namespace: openshift-monitoring


assets/node-exporter/accelerators-configmap.yaml
```diff
@@ -0,0 +1,29 @@
...
+        - pciID: 0x2235
+          pciName: A40
+        - pciID: 0x1df5
+          pciName: V100
+        - pciID: 0x20f1
+          pciName: A100 40G
+        - pciID: 0x1ff2
+          pciName: T400 4GB
+        - pciID: 0x1eb8
+          pciName: Tesla T4
```
romfreiman, line 22: why not to add here more ids?

yevgeny-shnaidman: i want the code change to be approved first. Once that is done, then i will add the rest of the accelerators

jsonnet/components/node-exporter.libsonnet
```diff
@@ -3,6 +3,31 @@ local textfileVolumeName = 'node-exporter-textfile';
 local tlsVolumeName = 'node-exporter-tls';
 local wtmpPath = '/var/log/wtmp';
 local wtmpVolumeName = 'node-exporter-wtmp';
+local configDir = '/var/node_exporter/config';
+local configVolumeName = 'node-exporter-config';
+local acceleratorsConfigFileName = 'accelerators-config.yaml';
+local acceleratorsConfigMapName = 'accelerators-configmap';
+local acceleratorsConfigData = |||
```
simonpasquier, line 10: you can use `std.manifestYamlDoc()` to convert a valid jsonnet object to its YAML representation. It will avoid any indentation issue.
https://jsonnet.org/ref/stdlib.html

jsonnet/components/node-exporter.libsonnet
```diff
@@ -290,6 +322,10 @@ function(params)
                           name: 'DBUS_SYSTEM_BUS_ADDRESS',
                           value: 'unix:path=/host/root/var/run/dbus/system_bus_socket',
                         },
+                        {
+                          name: 'ACCELERATORS_CONFIG_FILE',
```
simonpasquier, line 326: I don't think that we need to make it configurable given that it's a pure downstream feature: the accelerator collector config file location can be hardcoded in node_exporter and mounted by CMO at the expected path.

pkg/manifests/manifests.go
```diff
@@ -138,6 +138,7 @@ var (
 	NodeExporterMinimalServiceMonitor      = "node-exporter/minimal-service-monitor.yaml"
 	NodeExporterPrometheusRule             = "node-exporter/prometheus-rule.yaml"
 	NodeExporterKubeRbacProxySecret        = "node-exporter/kube-rbac-proxy-secret.yaml"
+	NodeExporterConfigMap                  = "node-exporter/accelerators-configmap.yaml"
```
jan--f, line 141:
```suggestion
	NodeExporterAcceleratorsConfigMap                  = "node-exporter/accelerators-configmap.yaml"
```
Maybe a more descriptive name?

jsonnet/components/node-exporter.libsonnet
yevgeny-shnaidman: Using the environment variable means that we determine the mounting path only at CMO, and there is no need to synchronize between CMO and node-exporter. IMHO in case it will be changed in the future for any reason, using ENV will allow us to avoid synchronization bugs

pkg/manifests/manifests.go
yevgeny-shnaidman: Fixed

jsonnet/components/node-exporter.libsonnet
yevgeny-shnaidman: Fixed

yevgeny-shnaidman:
@jan--f @simonpasquier , fixed and answered the remarks, please take a look again

jsonnet/components/node-exporter.libsonnet
machine424: I agree with Simon, hardcoding it would be better as we control the whole chain, and in case of a sync issue, we should be expecting the collector and the tests to yell.
(But if you think this is still needed I'd go for a cli arg for consistency, like `--collector.textfile.directory=/var/node_exporter/textfile` e.g.)

jsonnet/versions.yaml
```diff
@@ -15,7 +15,7 @@ repos:
     thanos: openshift/thanos
 versions:
     alertmanager: 0.27.0
-    kubeRbacProxy: 0.18.1
+    kubeRbacProxy: 0.18.2
```
machine424, line 18: could you revert this? as it's not related to the PR. (in case the PR is reverted we don't want this to get reverted e.g.)
We'll take case of it in another PR. `versions` job isn't required you can keep it failing.

jsonnet/components/node-exporter.libsonnet
```diff
@@ -3,6 +3,26 @@ local textfileVolumeName = 'node-exporter-textfile';
 local tlsVolumeName = 'node-exporter-tls';
 local wtmpPath = '/var/log/wtmp';
 local wtmpVolumeName = 'node-exporter-wtmp';
+local configDir = '/var/node_exporter/config';
+local configVolumeName = 'node-exporter-config';
+local acceleratorsConfigFileName = 'accelerators-config.yaml';
+local acceleratorsConfigMapName = 'accelerators-configmap';
```
machine424, line 9: to reduce the risk of any collision and to provide as much context as possible I'd go with `node-exporter-accelerators-collector-config` (this is also consistent with other naming)

jsonnet/components/node-exporter.libsonnet
```diff
@@ -324,6 +355,18 @@ function(params)
                   secretName: 'node-exporter-kube-rbac-proxy-config',
                 },
               },
+              {
+                name: configVolumeName,
+                configMap: {
+                  name: acceleratorsConfigMapName,
+                  items: [
+                    {
+                      key: 'accelerators-config.yaml',
```
machine424, line 364:
```suggestion
                      key: 'accelerators-collector-config.yaml',
```

pkg/manifests/manifests.go
```diff
@@ -138,6 +138,7 @@ var (
 	NodeExporterMinimalServiceMonitor      = "node-exporter/minimal-service-monitor.yaml"
 	NodeExporterPrometheusRule             = "node-exporter/prometheus-rule.yaml"
 	NodeExporterKubeRbacProxySecret        = "node-exporter/kube-rbac-proxy-secret.yaml"
+	NodeExporterAcceleratorsConfigMap      = "node-exporter/accelerators-configmap.yaml"
```
machine424, line 141:
```suggestion
	NodeExporterAcceleratorsConfigMap      = "node-exporter/accelerators-collector-configmap.yaml"
```

pkg/tasks/nodeexporter.go
```diff
@@ -94,6 +94,15 @@ func (t *NodeExporterTask) Run(ctx context.Context) error {
 		return fmt.Errorf("reconciling node-exporter Service failed: %w", err)
 	}
 
+	cm, err := t.factory.NodeExporterConfigMap()
+	if err != nil {
+		return fmt.Errorf("initializing node-exporter ConfigMap failed: %w", err)
```
machine424, line 99:
```suggestion
		return fmt.Errorf("initializing node-exporter accelerators collector ConfigMap failed: %w", err)
```

pkg/tasks/nodeexporter.go
```diff
@@ -94,6 +94,15 @@ func (t *NodeExporterTask) Run(ctx context.Context) error {
...
+	}
+	err = t.client.CreateOrUpdateConfigMap(ctx, cm)
+	if err != nil {
+		return fmt.Errorf("reconciling node-exporter ConfigMap failed: %w", err)
```
machine424, line 103:
```suggestion
		return fmt.Errorf("reconciling node-exporter accelerators collector ConfigMap failed: %w", err)
```

pkg/tasks/nodeexporter.go
```diff
@@ -94,6 +94,15 @@ func (t *NodeExporterTask) Run(ctx context.Context) error {
 		return fmt.Errorf("reconciling node-exporter Service failed: %w", err)
 	}
 
+	cm, err := t.factory.NodeExporterConfigMap()
```
machine424, line 97:
```suggestion
	cm, err := t.factory.NodeExporterAcceleratorsCollectorConfigMap()
```

machine424 (changes requested):
passing through
left some naming suggestions.

jsonnet/versions.yaml
jan--f: No need to worry about the `versions` ci job. It just informative for us and not a required check.

yevgeny-shnaidman: Fxied

jsonnet/components/node-exporter.libsonnet
yevgeny-shnaidman: Fixed

jsonnet/components/node-exporter.libsonnet
yevgeny-shnaidman: Fixed

pkg/manifests/manifests.go
yevgeny-shnaidman: Fixed

pkg/tasks/nodeexporter.go
yevgeny-shnaidman: Fixed

pkg/tasks/nodeexporter.go
yevgeny-shnaidman: Fixed

pkg/tasks/nodeexporter.go
yevgeny-shnaidman: Fixed

jsonnet/components/node-exporter.libsonnet
yevgeny-shnaidman: i cannot add the cli arg in the current commit, since it will crush the node-exporter ( the collector.accelerators.configmpa-path flag is not defined there yet. So  in this PR i will just remove the EVN variable, and then add the flag in the node-exporter PR, and then will add another commit in this repo for defining the cli args

assets/node-exporter/daemonset.yaml
```diff
@@ -79,6 +79,9 @@ spec:
         - mountPath: /var/node_exporter/textfile
           name: node-exporter-textfile
           readOnly: true
+        - mountPath: /var/node_exporter/config
```
machine424, line 82:
```suggestion
        - mountPath: /var/node_exporter/accelerators_collector_config
```

assets/node-exporter/daemonset.yaml
```diff
@@ -182,6 +185,12 @@ spec:
       - name: node-exporter-kube-rbac-proxy-config
         secret:
           secretName: node-exporter-kube-rbac-proxy-config
+      - configMap:
+          items:
+          - key: accelerators-collector-config.yaml
+            path: accelerators-collector-config.yaml
```
machine424, line 191:
```suggestion
          - key: config.yaml
            path: config.yaml
```
(given the change, to the folder name's, above)

assets/node-exporter/daemonset.yaml
```diff
@@ -79,6 +79,9 @@ spec:
         - mountPath: /var/node_exporter/textfile
           name: node-exporter-textfile
           readOnly: true
+        - mountPath: /var/node_exporter/config
+          name: node-exporter-config
```
machine424, line 83:
```suggestion
          name: node-exporter-accelerators-collector-config
```

assets/node-exporter/daemonset.yaml
```diff
@@ -182,6 +185,12 @@ spec:
       - name: node-exporter-kube-rbac-proxy-config
         secret:
           secretName: node-exporter-kube-rbac-proxy-config
+      - configMap:
+          items:
+          - key: accelerators-collector-config.yaml
+            path: accelerators-collector-config.yaml
+          name: node-exporter-accelerators-collector-config
+        name: node-exporter-config
```
machine424, line 193:
```suggestion
        name: node-exporter-accelerators-collector-config
```

jsonnet/components/node-exporter.libsonnet
```diff
@@ -350,4 +389,16 @@ function(params)
...
+
+    acceleratorsCollectorConfigmap: {
+      apiVersion: 'v1',
+      kind: 'ConfigMap',
+      metadata: {
+        name: acceleratorsConfigMapName,
+        namespace: cfg.namespace,
+      },
+      data: {
+        'accelerators-collector-config.yaml': std.manifestYamlDoc(acceleratorsConfigData),
```
machine424, line 401: we could use `acceleratorsConfigFileName` here I think.

jsonnet/components/node-exporter.libsonnet
```diff
@@ -3,6 +3,26 @@ local textfileVolumeName = 'node-exporter-textfile';
 local tlsVolumeName = 'node-exporter-tls';
 local wtmpPath = '/var/log/wtmp';
 local wtmpVolumeName = 'node-exporter-wtmp';
+local configDir = '/var/node_exporter/config';
+local configVolumeName = 'node-exporter-config';
+local acceleratorsConfigFileName = 'accelerators-collector-config.yaml';
```
machine424, line 8:
```suggestion
local acceleratorsConfigFileName = 'config.yaml';
```
see other change suggestions.

Documentation/deps-versions.md
```diff
@@ -1,7 +1,7 @@
...
+| release-4.20 | [0.27.0](https://github.com/openshift/prometheus-alertmanager/blob/release-4.20) | [0.18.2](https://github.com/openshift/kube-rbac-proxy/blob/release-4.20) | [2.13.0](https://github.com/openshift/kube-state-metrics/blob/release-4.20) | [0.7.2](https://github.com/openshift/kubernetes-metrics-server/blob/release-4.20) | [1.0.0](https://github.com/openshift/monitoring-plugin/blob/release-4.20) | [1.8.2](https://github.com/openshift/node_exporter/blob/release-4.20) | [0.11.0](https://github.com/openshift/prom-label-proxy/blob/release-4.20) | [2.55.1](https://github.com/openshift/prometheus/blob/release-4.20) | [0.78.2](https://github.com/openshift/prometheus-operator/blob/release-4.20) | [0.36.1](https://github.com/openshift/thanos/blob/release-4.20) |
```
machine424, line 3: you could revert this change as well.

machine424 (commented):
I had some final naming suggestions. I understand this may seem nitpicky, but I think they'll be helpful to anyone who is not very familiar with node-exporter or the downstream accelerators collector, and who will need to interact with them by providing clear and consistent naming.

This
/lgtm
Thanks for working on this.
/hold
In case the other colleagues have anything to add. 
I understand if you don't want/have time for the suggestions, I can take care of it in another PR.

The failures in okd-scos-e2e-aws-ovn don't seem to be related to the changes. Apparently, the test is flaky (which explains is why it isn't required).

jsonnet/components/node-exporter.libsonnet
yevgeny-shnaidman: fixed

jsonnet/components/node-exporter.libsonnet
yevgeny-shnaidman: in case it is a key, jsonnet does not substitute the variable for its value, but uses the name of the variable directly as key name

yevgeny-shnaidman: fixed

assets/node-exporter/daemonset.yaml
yevgeny-shnaidman: fixed

assets/node-exporter/daemonset.yaml
yevgeny-shnaidman: fixed

assets/node-exporter/daemonset.yaml
yevgeny-shnaidman: fixed

assets/node-exporter/daemonset.yaml
yevgeny-shnaidman: fixed

machine424:
/lgtm
thanks again.

yevgeny-shnaidman:
> Thanks, this lgtm overall. I think there are two things we still need (please correct me if I'm missing anything):
> 
> 1. The change for a default config map location in [Add configmap support to the accelerator's collector node_exporter#155](https://github.com/openshift/node_exporter/pull/155)
> 2. And perhaps a test confirming that we remain consistent with the default location? For example create a invalid config map at the default location and expect an error, i.e. fail the test if no error is thrown? For inspiration see https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/node_exporter_test.go.
If i understand correctly , both those issues are relevant for the node exporter PR:
1) default config map location can be added  once the PR node-exporter repo is pushed.
2) The same as above: once we also push the PR for node-exporter , then we can add e2e test

machine424:
As I mentioned earlier, no need to retrigger `okd-scos-e2e-aws-ovn`, it's flaky but it's not blocking the PR as it's not required.

The PR is blocked because of [tide](https://prow.ci.openshift.org/pr?query=is%3Apr+repo%3Aopenshift%2Fcluster-monitoring-operator+author%3Ayevgeny-shnaidman+head%3Ayevgeny%2Fadding-configmap); the PR should be attached to a valid jira ticket https://docs.ci.openshift.org/docs/architecture/jira/#non-bug-jira-references

and the /hold should be removed, cc @jan--f if https://github.com/openshift/cluster-monitoring-operator/pull/2530#issuecomment-2529304530 is ok for you.

yevgeny-shnaidman:
/retitle MGMT-19498: Adding accelerators configuration file to the node-exporter

PR #2535, opened 2024-12-10
NI-JIRA: update go deps

machine424:
/hold

let's wait after acknowledge-critical-fixes-only 


test/e2e/framework/prometheus.go
```diff
@@ -147,7 +147,7 @@ func (f Framework) PrometheusConfigFromSecret(t *testing.T, namespace, secretNam
 	require.NoError(t, err)
 	unzippedData, err := io.ReadAll(reader)
 	require.NoError(t, err)
-	prometheusConfig, err := promConfig.Load(string(unzippedData), false, log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr)))
+	prometheusConfig, err := promConfig.Load(string(unzippedData), promslog.NewNopLogger())
```
rexagod, line 150: Can we use [New()](https://pkg.go.dev/github.com/prometheus/common/promslog#New) here instead? That should preserve earlier behavior of logging to `Stderr`.

machine424: I added a comment about this that apparently disappeared:
"I don't think that's worth the complexity, we can always add it in the future if needed for debugging"

machine424: But, it's actually not complex, I'll take care of that.

machine424:
well the tests require more adjustments apparently (prom 3 related), turning this into a draft.

slashpai:
Do we want to update the prometheus module to v3 now? or may be after updated in openshift/prometheus?

machine424:
> Do we want to update the prometheus module to v3 now? or may be after updated in openshift/prometheus?

I'll move all the required changes for pom3 (if any) to another PR and keep this one for the other deps upgrades (waiting for new versions of some openshift api deps, thus it's a draft/on hold)

jan--f:
I assume this is no longer needed.

/close

PR #2538, opened 2024-12-16
NO-JIRA: Bump prometheus-operator to v0.79.1

slashpai:


slashpai:
/hold
We would need a patch release 0.79.1 https://github.com/prometheus-operator/prometheus-operator/pull/7197

PR #2539, opened 2024-12-17
OCPBUGS-45895: Reconcile pre-existing Secrets

rexagod:
Until now, CMO did not reconcile existing secrets, even if their data changed. This changes that behavior.

### DIFF
--- a/pkg/tasks/alertmanager.go
+++ b/pkg/tasks/alertmanager.go
@@ -111,7 +111,7 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 		return fmt.Errorf("initializing Alertmanager RBAC proxy Secret failed: %w", err)
 	}
 
-	err = t.client.CreateIfNotExistSecret(ctx, rs)
+	err = t.client.CreateOrUpdateSecret(ctx, rs)
 	if err != nil {
 		return fmt.Errorf("creating Alertmanager RBAC proxy Secret failed: %w", err)
 	}
@@ -121,7 +121,7 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 		return fmt.Errorf("initializing Alertmanager RBAC proxy metric Secret failed: %w", err)
 	}
 
-	err = t.client.CreateIfNotExistSecret(ctx, rsm)
+	err = t.client.CreateOrUpdateSecret(ctx, rsm)
 	if err != nil {
 		return fmt.Errorf("creating Alertmanager RBAC proxy metric Secret failed: %w", err)
 	}
@@ -173,7 +173,7 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 		return fmt.Errorf("initializing Alertmanager proxy web Secret failed: %w", err)
 	}
 
-	err = t.client.CreateIfNotExistSecret(ctx, ps)
+	err = t.client.CreateOrUpdateSecret(ctx, ps)
 	if err != nil {
 		return fmt.Errorf("creating Alertmanager proxy web Secret failed: %w", err)
 	}
--- a/test/e2e/reconcile_objects_test.go
+++ b/test/e2e/reconcile_objects_test.go
@@ -14,6 +14,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 )
 
 // TestSecretsReconciliation tests whether the secrets created by the operator are reconciled correctly. These include:
@@ -53,14 +54,12 @@ func TestSecretsReconciliation(t *testing.T) {
 		},
 	}
 
-	restoreSecrets := func(secrets []types.NamespacedName) {
-		for _, secret := range secrets {
+	// Restore unsynced secrets to their original state after the test.
+	// Synced secrets don't need restoration since the operator reconciles those.
+	t.Cleanup(func() {
+		for _, secret := range unsyncedSecrets {
 			gotSecret, err := f.KubeClient.CoreV1().Secrets(secret.Namespace).Get(context.Background(), secret.Name, metav1.GetOptions{})
-			if err != nil {
-				// Secret may have been rotated (e.g. hash-suffixed TLS secrets).
-				t.Logf("skipping restore of %s: %v", secret.String(), err)
-				continue
-			}
+			require.NoError(t, err)
 			data := gotSecret.Data
 			for k, v := range data {
 				if isSecretDataJSONEncoded(gotSecret) {
@@ -78,70 +77,70 @@ func TestSecretsReconciliation(t *testing.T) {
 			_, err = f.KubeClient.CoreV1().Secrets(secret.Namespace).Update(context.Background(), gotSecret, metav1.UpdateOptions{})
 			require.NoError(t, err)
 		}
-	}
+	})
 
-	var syncedSecrets []types.NamespacedName
 	// Fetch all secrets since we are responsible for the whole namespace.
+	var syncedSecrets []types.NamespacedName
 	secretsNS, err := f.KubeClient.CoreV1().Secrets(f.Ns).List(context.Background(), metav1.ListOptions{})
 	require.NoError(t, err)
-
 	secretsUWMNS, err := f.KubeClient.CoreV1().Secrets(f.UserWorkloadMonitoringNs).List(context.Background(), metav1.ListOptions{})
 	require.NoError(t, err)
 
 	for _, secret := range append(secretsNS.Items, secretsUWMNS.Items...) {
-		secretNamespacedName := types.NamespacedName{
+		nn := types.NamespacedName{
 			Name:      secret.Name,
 			Namespace: secret.Namespace,
 		}
-		if slices.Contains(unsyncedSecrets, secretNamespacedName) {
+		if slices.Contains(unsyncedSecrets, nn) {
 			continue
 		}
-		syncedSecrets = append(syncedSecrets, secretNamespacedName)
+		// Skip TLS secrets since they have hash suffixes and get rotated (replaced, not updated in-place).
+		if strings.Contains(secret.Name, "tls") {
+			continue
+		}
+		syncedSecrets = append(syncedSecrets, nn)
 	}
 	require.NotEmpty(t, syncedSecrets)
 
 	secrets := append(syncedSecrets, unsyncedSecrets...)
 
-	t.Cleanup(func() { restoreSecrets(secrets) })
-
 	// Update the aforementioned secrets' data.
-	var mutatedSecrets []types.NamespacedName
 	for _, secret := range secrets {
-		var gotSecret *v1.Secret
-		gotSecret, err = f.KubeClient.CoreV1().Secrets(secret.Namespace).Get(context.Background(), secret.Name, metav1.GetOptions{})
-		if err != nil {
-			// Secret may have been rotated (e.g. hash-suffixed TLS secrets).
-			t.Logf("skipping mutation of %s: %v", secret.String(), err)
-			continue
-		}
-		data := gotSecret.Data
-		for k, v := range data {
-			if isSecretDataJSONEncoded(gotSecret) {
-				var jsonData map[string]interface{}
-				err = json.Unmarshal(v, &jsonData)
-				require.NoError(t, err)
-				jsonData[t.Name()] = t.Name()
-				v, err = json.Marshal(jsonData)
-				require.NoError(t, err)
-				data[k] = v
-			} else {
-				data[k] = []byte(t.Name() + string(v))
+		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			gotSecret, err := f.KubeClient.CoreV1().Secrets(secret.Namespace).Get(context.Background(), secret.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
 			}
-			break
-		}
-
-		_, err = f.KubeClient.CoreV1().Secrets(secret.Namespace).Update(context.Background(), gotSecret, metav1.UpdateOptions{})
+			data := gotSecret.Data
+			for k, v := range data {
+				if isSecretDataJSONEncoded(gotSecret) {
+					var jsonData map[string]interface{}
+					if err := json.Unmarshal(v, &jsonData); err != nil {
+						return err
+					}
+					jsonData[t.Name()] = t.Name()
+					v, err = json.Marshal(jsonData)
+					if err != nil {
+						return err
+					}
+					data[k] = v
+				} else {
+					data[k] = []byte(t.Name() + string(v))
+				}
+				break
+			}
+			_, err = f.KubeClient.CoreV1().Secrets(secret.Namespace).Update(context.Background(), gotSecret, metav1.UpdateOptions{})
+			return err
+		})
 		require.NoError(t, err)
-		mutatedSecrets = append(mutatedSecrets, secret)
 	}
 
 	// Check for reconciliation of secrets.
-	for _, secret := range mutatedSecrets {
-		// Check if the secrets were reconciled as expected.
+	for _, secret := range secrets {
+		// Synced secrets should be reconciled, i.e., the test prefix must be removed.
 		if slices.Contains(syncedSecrets, secret) {
-			err = framework.Poll(10*time.Second, 5*time.Minute, func() error {
-				var updatedSecret *v1.Secret
-				updatedSecret, err = f.KubeClient.CoreV1().Secrets(secret.Namespace).Get(context.Background(), secret.Name, metav1.GetOptions{})
+			err := framework.Poll(10*time.Second, 5*time.Minute, func() error {
+				updatedSecret, err := f.KubeClient.CoreV1().Secrets(secret.Namespace).Get(context.Background(), secret.Name, metav1.GetOptions{})
 				if err != nil {
 					return err
 				}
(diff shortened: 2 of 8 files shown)


pkg/tasks/alertmanager.go
```diff
@@ -77,7 +77,7 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
 		return fmt.Errorf("initializing Alertmanager configuration Secret failed: %w", err)
 	}
 
-	err = t.client.CreateIfNotExistSecret(ctx, s)
+	err = t.client.CreateOrUpdateSecret(ctx, s)
```
machine424, line 80: Good catch for the rbac proxy secrets, we should for sure update them when needed. Not sure about how we can easily have a regression test for this.

For some secrets like the one containing alertmanager config, CMO shouldn't override it as it may be customized by users https://docs.openshift.com/container-platform/4.16/observability/monitoring/managing-alerts.html#configuring-notifications-for-default-platform-alerts_managing-alerts

rexagod: Yeah, that's a tricky case, since if we make updates to that secret between releases, and upgraded CMO version won't reconcile it.

rexagod: I'll revert the change for those cases.

machine424: we can go through them one by one ;)
But AFAIK, users aren't meant to change the rbac proxy secrets.

rexagod: Refactored to exclude `alertmanager` configuration secrets for UWM and in-cluster environments, as users may want to modify those.

machine424: I have an idea of an e2e test for this:
- keep a list of secretsNotToSync (with the alertmanager config secret etc, the first test run will help us build it)
- loop over all the secrets in the two monitoring namespaces (if we can get all the operands enabled it'd be great) if not in secretsNotToSync, alter them (add a `foo: bar` data entry or sth)
- trigger a CMO sync and make sure the added entry gets deleted from all secrets.

We can extend this to configmaps, RBAC etc. in the future.
If we can have it for secrets as part of this PR it'd be great.

rexagod: Sounds good, patched.

rexagod:
I'm currently having some problems with `make versions generate` on MacOS.

Documentation/deps-versions.md
```diff
@@ -1,7 +1,7 @@
...
+| release-4.20 | [0.27.0](https://github.com/openshift/prometheus-alertmanager/blob/release-4.20) | [0.18.2](https://github.com/openshift/kube-rbac-proxy/blob/release-4.20) | [2.13.0](https://github.com/openshift/kube-state-metrics/blob/release-4.20) | [0.7.2](https://github.com/openshift/kubernetes-metrics-server/blob/release-4.20) | [1.0.0](https://github.com/openshift/monitoring-plugin/blob/release-4.20) | [1.8.2](https://github.com/openshift/node_exporter/blob/release-4.20) | [0.11.0](https://github.com/openshift/prom-label-proxy/blob/release-4.20) | [2.55.1](https://github.com/openshift/prometheus/blob/release-4.20) | [0.79.2](https://github.com/openshift/prometheus-operator/blob/release-4.20) | [0.37.2](https://github.com/openshift/thanos/blob/release-4.20) |
```
machine424, line 3: If you could rebase, those changes are already on `main`

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+func TestSecretsReconciliation(t *testing.T) {
+	// List of secrets that should not be synced during operator's reconciliation.
+	var (
+		namespaceMonitoring                 = f.Ns
+		namespaceUserWorkloadMonitoring     = f.UserWorkloadMonitoringNs
+		extractNamespacedNameFromFileSecret = extractNamespacedNameFromFile[*v1.Secret]
+	)
+	unsyncedSecrets := []types.NamespacedName{
+		extractNamespacedNameFromFileSecret(t, manifests.AlertmanagerConfig),
```
machine424, line 43: I think it's simpler/more readable/less error prone is we just hardcode the names here

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,185 @@
...
+		extractNamespacedNameFromFileSecret(t, manifests.AlertmanagerUserWorkloadSecret),
+	}
+	syncedSecrets := []types.NamespacedName{
```
machine424, line 46: As I see it, we can do without this (to avoid the maintenance burden also)

We can do with:
list all secrets in the namespace, ensure there are at least 5 (we know there are more than 5 secrets) or `len(unsyncedSecrets)` or sth, to make sure the test is really doing sth.

Then loop over all the secrets in there and alter them.

Then loop over all of them, if they're not in `unsyncedSecrets`, check the suffix was removed.

Then at the end, loop over `unsyncedSecrets` and make sure the prefix is still there. (this way we can say we waited for all the syncs to get triggered and those secrets weren't touched)

We can have subtests or sth to run the test on both namespaces or you can consider the secrets frpm both of them in one shot

machine424 (commented):
thanks for the test, some suggestions.

test/e2e/reconcile_objects_test.go
rexagod: Nice idea, let me patch this.

rexagod: Thinking a bit more on this, let's say if a `synced` secret unintentionally becomes `unsynced`, we'd be only able to catch it randomly?

machine424: if a `synced` secret becomes `unsynced` the step where we'd check that the suffix is removed will fail.

test/e2e/reconcile_objects_test.go
machine424: we don't add new secrets regularly, it's always better to minimize risks that the test itself is broken.
Having hardcoded names is more readable (no need to go refer to the manifest to get the name of a secret) and we don't have to worry that the "fetch secret code" is broken (I'm not saying it's broken now :))

We can always change in the future if we start adding/removing secrets regularly in the future.

rexagod:
@machine424 Made the requested changes, PTAL :)

rexagod:
`okd-scos-e2e-aws-ovn` errors are unrelated, so this is good for reviews.

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,168 @@
...
+			Name:      "alertmanager-user-workload",
+			Namespace: f.UserWorkloadMonitoringNs,
+		},
+	}
+	cleanup := func() {
+		// Restore all unsynced secrets to their original state.
+		for _, secret := range unsyncedSecrets {
+			gotSecret, err := f.KubeClient.CoreV1().Secrets(secret.Namespace).Get(context.Background(), secret.Name, metav1.GetOptions{})
+			if err != nil {
+				if errors.IsNotFound(err) {
```
machine424, line 58: I think we should be made aware of this; whether `unsyncedSecrets` would need to be updated or the test deletes the secret by mistakes in both cases we need to surface that

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,168 @@
...
+			_, err = f.KubeClient.CoreV1().Secrets(secret.Namespace).Update(context.Background(), gotSecret, metav1.UpdateOptions{})
+			require.NoError(t, err)
+		}
+	}
+	defer cleanup()
+
+	// Prepare synced secrets.
+	var syncedSecrets []types.NamespacedName
+	secretsNS, err := f.KubeClient.CoreV1().Secrets(f.Ns).List(context.Background(), metav1.ListOptions{
+		LabelSelector: "app.kubernetes.io/managed-by=cluster-monitoring-operator",
```
machine424, line 80: because the two namespaces are our responsibility, let's just fetch all; if another test leaves a secret around we need to make it clean it, if another operator doesn't sync a secret that it should, we want to identify that and reach out to them

Documentation/deps-versions.md
```diff
@@ -1,8 +1,8 @@
...
+| release-4.20 | [0.28.0](https://github.com/openshift/prometheus-alertmanager/blob/release-4.20) | [0.19.0](https://github.com/openshift/kube-rbac-proxy/blob/release-4.20) | [2.15.0](https://github.com/openshift/kube-state-metrics/blob/release-4.20) | [0.7.2](https://github.com/openshift/kubernetes-metrics-server/blob/release-4.20) | [1.0.0](https://github.com/openshift/monitoring-plugin/blob/release-4.20) | [1.8.2](https://github.com/openshift/node_exporter/blob/release-4.20) | [0.11.0](https://github.com/openshift/prom-label-proxy/blob/release-4.20) | [3.0.1](https://github.com/openshift/prometheus/blob/release-4.20)  | [0.80.1](https://github.com/openshift/prometheus-operator/blob/release-4.20) | [0.37.2](https://github.com/openshift/thanos/blob/release-4.20) |
```
machine424, line 3: if we could have those versions bumps in another PR, that'd be great

machine424 (commented):
Thanks again, we're almost ready to merge this, just some final suggestions

rexagod:
🤦🏼 So I just noticed I had a patch on local for this that I never pushed, I did that now, and I think some of the review comments are addressed by it. I'll go over them again to make sure I act on them.

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+func TestSecretsReconciliation(t *testing.T) {
+	// Create assets under both scenarios for us to work with.
+	setupUserWorkloadAssetsWithTeardownHook(t, f)
+	userWorkloadConfigMap := f.BuildUserWorkloadConfigMap(t, `alertmanager:
+  enabled: true
+`)
+	f.MustCreateOrUpdateConfigMap(t, userWorkloadConfigMap)
+	defer f.MustDeleteConfigMap(t, userWorkloadConfigMap)
+	for _, secret := range []types.NamespacedName{
```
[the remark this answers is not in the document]
rexagod: Used a `require.NotEmpty(t, syncedSecrets)` right after we fill `syncedSecrets` to ensure it's not empty.

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+		syncedSecrets = append(syncedSecrets, types.NamespacedName{
+			Name:      secret.Name,
+			Namespace: secret.Namespace,
+		})
+	}
+	require.NotEmpty(t, syncedSecrets)
+
+	// Update the aforementioned secrets' data.
+	for _, secret := range append(syncedSecrets, unsyncedSecrets...) {
```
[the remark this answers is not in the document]
rexagod: Not sure what this means, could you please elaborate a bit? There's no `secrets` IIUC but only `syncedSecrets` and `unsyncedSecrets`.

test/e2e/reconcile_objects_test.go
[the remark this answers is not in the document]
machine424: then I don't see why this `f.AssertSecretExists` is needed.

test/e2e/reconcile_objects_test.go
machine424: a log can be missed, let's just remove this special treatment for `IsNotFound` and just fail

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"time"
+
+	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestSecretsReconciliation(t *testing.T) {
```
machine424, line 18: let's add a short comment about the overall logic in the test

test/e2e/reconcile_objects_test.go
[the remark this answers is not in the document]
machine424: I meant we can only have `secrets` with all secrets and loop over them only once and then use the `slices.Contains`  check to see whether we want to `// Check if the secrets were reconciled as expected.` or `// Check if the secrets were reconciled unexpectedly.`. so we only have one check loop.
(this is a nit, I just find it more readable that way)

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,151 @@
...
+	// Create assets under both scenarios for us to work with.
+	setupUserWorkloadAssetsWithTeardownHook(t, f)
+	userWorkloadConfigMap := f.BuildUserWorkloadConfigMap(t, `alertmanager:
```
machine424, line 21: good catch, I think this config should make CMO produce all/most of the secrets.

rexagod:
Need to test this out on a cluster, will request a review once that's green.

rexagod:
This needs to be updated a bit, will request a review once I clean this up.

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// TestSecretsReconciliation tests whether the secrets created by the operator are reconciled correctly. These include:
+// * unsynced secrets: secrets that are deployed by, but not synced by the operator, and,
+// * synced secrets: secrets that are deployed by, and should be synced by the operator.
+
+// TODO: Exclude all secrets that are initially empty and populated by other operators.
```
machine424: I don't think we should be excluding any, same reasoning than https://github.com/openshift/cluster-monitoring-operator/pull/2539#discussion_r2022599773, the two monitoring namespaces are our responsibility, if any other operator doesn't sync its operands, we want to be aware of that

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,183 @@
...
+			Name:      "thanos-ruler-user-workload-web-config",
+			Namespace: f.UserWorkloadMonitoringNs,
+		},
+	}
+
+	// Restore all unsynced secrets to their original state.
+	cleanup := func() {
+		for _, secret := range unsyncedSecrets {
+			gotSecret, err := f.KubeClient.CoreV1().Secrets(secret.Namespace).Get(context.Background(), secret.Name, metav1.GetOptions{})
+			if errors.IsNotFound(err) {
```
machine424: How could this happen? I don't think we should skip here, that may mean the list above is no longer up to date and we should update it.

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,183 @@
...
+			require.NoError(t, err)
+			data := gotSecret.Data
+			for k, v := range data {
+				data[k] = []byte(strings.TrimPrefix(string(v), t.Name()))
+			}
+			_, err = f.KubeClient.CoreV1().Secrets(secret.Namespace).Update(context.Background(), gotSecret, metav1.UpdateOptions{})
+			require.NoError(t, err)
+		}
+	}
+	defer cleanup()
```
machine424: niy: we could use `t.Cleanup()`

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+	var syncedSecrets []types.NamespacedName
+	secretsNS, err := f.KubeClient.CoreV1().Secrets(f.Ns).List(context.Background(), metav1.ListOptions{
+		// Intentionally commented out as we want to fetch all secrets.
+		// LabelSelector: "app.kubernetes.io/managed-by=cluster-monitoring-operator",
```
machine424: nit: we can remove this or replace it with a `we're responsible for the whole namespace`

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,187 @@
...
+	})
+	require.NoError(t, err)
+
+	secretsUWMNS, err := f.KubeClient.CoreV1().Secrets(f.UserWorkloadMonitoringNs).List(context.Background(), metav1.ListOptions{
+		// Intentionally commented out as we want to fetch all secrets.
+		// LabelSelector: "app.kubernetes.io/managed-by=cluster-monitoring-operator",
```
machine424: nit: we can remove this or replace it with a `we're responsible for the whole namespace`

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+		syncedSecrets = append(syncedSecrets, secretNamespacedName)
+	}
+	require.NotEmpty(t, syncedSecrets)
+
+	secrets := append(syncedSecrets, unsyncedSecrets...)
+
+	var filteredSecrets []types.NamespacedName
+	for _, secret := range secrets {
+		if strings.Contains(secret.Name, "tls") {
```
machine424: mmm any reason for ignoring those?

machine424 (commented):
I realized I never pushed the comments

test/e2e/reconcile_objects_test.go
```diff
@@ -0,0 +1,170 @@
...
+				require.NoError(t, err)
+				jsonData[t.Name()] = t.Name()
+				v, err = json.Marshal(jsonData)
+				require.NoError(t, err)
+				data[k] = v
+			} else {
+				data[k] = []byte(t.Name() + string(v))
+			}
+			break
+		}
```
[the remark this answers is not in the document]
rexagod: I believe this is addressed now: https://github.com/openshift/cluster-monitoring-operator/pull/2539/changes#diff-940ff24a237421831783d2b59569062d3a03444707cc4fe8b990c99a6f3d09d9R156

PR #2540, opened 2024-12-17
MGMT-19498:Fixing the configmap data fields

yevgeny-shnaidman:
Renaming the pciName field to modelName


machine424:
/lgtm
/hold
until the testwith tests in https://github.com/openshift/node_exporter/pull/155 pass

machine424:
(this needs to be attached to a jira ticket or ...)

PR #2542, opened 2024-12-18
NO-JIRA: Bump prometheus-operator to 0.79.2

slashpai:


slashpai:
cc @machine424 since you reviewed 0.79.1 we had another patch release, thought to keep it up todated even though no change in CRD in this patch

machine424:
/lgtm
thanks

PR #2543, opened 2024-12-19
NO-JIRA: fix TestPrometheusRemoteWrite to align the image used for th…

machine424:
…e remote write receiver with k8s'

instead of the operator's upstream default image

This broke once prometheus-operator's default image is a prom 3 see https://github.com/openshift/prometheus-operator/pull/321#issuecomment-2552954671: prometheus 3 no longer understand the `remote-write-receiver` feature flag.

also moved from the feature gate to the spec field as it'll be required for prom 3

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -123,7 +123,7 @@ type remoteWriteTest struct {
 func TestPrometheusRemoteWrite(t *testing.T) {
 	ctx := context.Background()
 
-	// Get the Prometheus image to use, use the same than k8s's.
+	// Use the same image than k8s' for the remote write target.
 	k8sProm, err := f.MonitoringClient.Prometheuses(f.Ns).Get(ctx, "k8s", metav1.GetOptions{})
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


slashpai:
Good catch 🎉

slashpai (approved):
/lgtm

marioferh (commented):
/lgtm

machine424:
moved from the feature gate to the spec field as it'll be required for prom 3

test/e2e/prometheus_test.go
```diff
@@ -123,6 +123,13 @@ type remoteWriteTest struct {
 func TestPrometheusRemoteWrite(t *testing.T) {
 	ctx := context.Background()
 
+	// Get the Prometheus image to use, use the same than k8s's.
```
slashpai, line 126: nit: I am not sure if I understand the comment correctly

machine424: adjusted it.

slashpai (commented):
/lgtm

machine424:
/retitle MON-4116: fix TestPrometheusRemoteWrite to align the image used for the remote write receiver with k8s'

PR #2545, opened 2024-12-23
MON-3960: test: enable back TestTLSSecurityProfileConfiguration and m…

machine424:
…ake it only check the default TLSProfiles as changing the profiles is disruptive (nodes rollout), other profiles/cases are now checked in new unit tests

### DIFF
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -4953,3 +4953,82 @@ xx_omitted_before_deploy__test_file_name:foo.yaml`,
 		})
 	}
 }
+
+func TestSetTLSSecurityConfiguration(t *testing.T) {
+	tests := []struct {
+		name               string
+		apiServerConfig    *APIServerConfig
+		initialArgs        []string
+		tlsCipherSuitesArg string
+		tlsMinVersionArg   string
+		finalArgs          []string
+	}{
+		{
+			name:               "Empty APIServerConfig",
+			apiServerConfig:    &APIServerConfig{},
+			initialArgs:        []string{"--foo=bar"},
+			tlsCipherSuitesArg: "--tls-cipher-suites=",
+			tlsMinVersionArg:   "--tls-min-version=",
+			finalArgs: []string{
+				"--foo=bar",
+				"--tls-cipher-suites=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384,TLS_CHACHA20_POLY1305_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
+				"--tls-min-version=VersionTLS12",
+			},
+		},
+		{
+			name:               "Empty APIServerConfig args to override",
+			apiServerConfig:    &APIServerConfig{},
+			initialArgs:        []string{"--foo=bar", "--tls-cipher-suites=toOverride", "--tls-min-version=toOverride"},
+			tlsCipherSuitesArg: "--tls-cipher-suites=",
+			tlsMinVersionArg:   "--tls-min-version=",
+			finalArgs: []string{
+				"--foo=bar",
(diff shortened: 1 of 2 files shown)


pkg/manifests/manifests_test.go
```diff
@@ -4953,3 +4953,82 @@ xx_omitted_before_deploy__test_file_name:foo.yaml`,
...
 	}
 }
+
+func TestSetTLSSecurityConfiguration(t *testing.T) {
+	tests := []struct {
+		name               string
+		apiServerConfig    *APIServerConfig
+		initialArgs        []string
+		tlsCipherSuitesArg string
+		tlsMinVersionArg   string
```
rexagod, line 4963: (nit) Since these fields remain unchanged throughout the cases here, can we hardcode them in the `require` instead? Or we could change them between `--web.tls-cipher-suites=` (dot-based flags) and `--tls-cipher-suites=` to test both cases that `setTLSSecurityConfiguration` handles.

test/e2e/tls_security_profile_test.go
```diff
@@ -37,97 +37,39 @@ func atLeastVersionTLS12(v string) string {
...
-		},
-		{
-			name: "intermediate profile",
-			profile: &configv1.TLSSecurityProfile{
-				Type:         configv1.TLSProfileIntermediateType,
-				Intermediate: &configv1.IntermediateTLSProfile{},
-			},
-			expectedCipherSuite:   configv1.TLSProfiles[configv1.TLSProfileIntermediateType].Ciphers,
-			expectedMinTLSVersion: "VersionTLS12",
-		},
```
rexagod, line 71: Not sure if upstream has tests for these two profiles, in which case it'd be good to configure the APIServer profile for them and keep these cases.

rexagod:
We could additionally test MCO's cordoning behavior in origin, but that's out of the scope of this patch, and IMHO something the MCO team should drive (if not already done)?

pkg/manifests/manifests_test.go
```diff
@@ -4953,3 +4953,82 @@ xx_omitted_before_deploy__test_file_name:foo.yaml`,
...
+				Spec: configv1.APIServerSpec{
+					TLSSecurityProfile: &configv1.TLSSecurityProfile{
+						Type: configv1.TLSProfileCustomType,
+						Custom: &configv1.CustomTLSProfile{
+							TLSProfileSpec: configv1.TLSProfileSpec{
+								Ciphers: []string{
+									"ECDHE-RSA-AES128-GCM-SHA256",
+									"ECDHE-ECDSA-AES256-GCM-SHA384",
+								},
+								MinTLSVersion: "VersionTLS10",
```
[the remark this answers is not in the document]
machine424: I think other tests already cover that, e.g. `TestGetTLSCiphers`, wdyt?

pkg/manifests/manifests_test.go
machine424: no need I think, `setArg` is already tested in `TestSetArg` e.g.

machine424: I can hard code them of course.

test/e2e/tls_security_profile_test.go
machine424: the intermediate profile is still tested as the default one in `TestDefaultTLSSecurityProfileConfiguration`

machine424:
> We could additionally test MCO's disruptive behavior in origin, to see if the updated cipher suites are observed post-cordoning.

Yes, but we'll need to make it serial and it'll add slow down many CIs that run serial. we could also consider having it on QE side, I can see similar ones here https://github.com/search?q=repo%3Aopenshift%2Fopenshift-tests-private%20TLSSecurityProfile&type=code
(when you have some time @juzhao, I'll create a ticket for this)

pkg/manifests/manifests_test.go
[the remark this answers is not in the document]
rexagod: I don't see a specific case in `TestGetTLSCiphers` that covers the post-min "ranges" as proposed above TBH.

rexagod:
I'm not sure if QE's suites are subject to pre-release payload testing cadence same as origin is, but, and correct me if I'm wrong, isn't this what disruptive tests are for? Would we contribute all disruptive tests to QE's suite going forward, and what about the ones already in origin?

test/e2e/tls_security_profile_test.go
rexagod: ACK, I saw they both use the same underlying API.

pkg/manifests/manifests_test.go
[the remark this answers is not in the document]
machine424: If it's not sth that was tested in the old e2e `TestTLSSecurityProfileConfiguration`, I think it's out of the scope of this PR and should be done in another PR.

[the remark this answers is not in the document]
rexagod: Agreed, this is more suited for `TestGetTLSCiphers` than here.

machine424:
> I'm not sure if QE's suites are subject to pre-release payload testing cadence same as origin is, but, and correct me if I'm wrong, isn't this what disruptive tests are for? Would we contribute all disruptive tests to QE's suite going forward, and what about the ones already in origin?

As we discussed (for collection profiles), having it as a disruptive test will require setting up a new job in our CI as disruptive tests aren't run by default.
Let's reconsider this once we start working on the collection profiles one.

machine424:
Note that MCO have their test there https://github.com/openshift/openshift-tests-private/pull/18525 e.g.

PR #2547, opened 2024-12-26
OCPBUGS-31356: chore(client): DeleteXXX short-circuit if the resource doesn't exist as a GET is less expensive than a no-op DELETE

machine424:


machine424:
/retitle OCPBUGS-31356: chore(client): DeleteXXX short-circuit if the resource doesn't exist as a GET is less expensive than a no-op DELETE

marioferh (commented):
/lgtm

Thanks!!

PR #2548, opened 2024-12-30
adding accelerator dashboard deployment

yevgeny-shnaidman:
The commit adds deployment of the accelerators dashboard. This dashboard will can be viewed in the OCP console and will present the vendor_model:node_accelerator_cards:sum metric.
Since we don't want the customer to see an empty dashboard, the CMO will deploy the dashboard configmap , but will enable it (label it with console.openshift.io/dashboard=true) only in case there is at least one node_accelerator_card_info metric present in Prometheus

1) adding dashboard configmap to jsonnnet
2) adding prometheus client initialization to operator 3) adding new task that deployes the dashboard configmap,checks the
   presence of the node_accelerator_card_info metrics and labels the
   configmap accordingly


pkg/tasks/nodeexporter_dashboard.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func NewNodeExporterDashboardTask(client *client.Client, factory *manifests.Factory) *NodeExporterDashboardTask {
+	return &NodeExporterDashboardTask{
+		client:  client,
+		factory: factory,
+	}
+}
+
+func (t *NodeExporterDashboardTask) Run(ctx context.Context) error {
```
machine424, line 37: I think this could be part of the existing node exporter task.

machine424 (commented):
Thanks Yevgeny for working on this,
Even for metrics exposed by the monitoring stack itself, we try to encourage stakeholders to deploy and maintain their dashboards on their own. Did you consider this? Would it be possible for you? (Do you have an operator or something that can handle that?)

I took a look at the PR at the "deploy the dashboard based on the existence of a metric" stage, and although I agree with the purpose (to avoid an empty dashboard for the majority of users, I assume), the approach may indeed be error-prone. Maybe we can consider other approaches.

pkg/tasks/nodeexporter_dashboard.go
machine424: or even have it deployed by CVO as for the others https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_90_cluster-monitoring-operator_02-dashboards.yaml if we agree on having it on all clusters.

yevgeny-shnaidman: @machine424 , i put it in a different task as a preparation for adding metrics existence check, although i am not sure if it will work: if the certificate used by  thanos-querier is stored in the secret, and that secret is deployed by the CVO in a dedicated task, then it kind of "chicken-egg" issue.

jan--f:
I agree with Ayoub here, CMO is not a good place for this artifact. Is there a Jira ticket or planing document where we can give some input?

yevgeny-shnaidman:
> I agree with Ayoub here, CMO is not a good place for this artifact. Is there a Jira ticket or planing document where we can give some input?

@jan--f i have added the jira to the title.  The option to deploy it via CMO was discussed with @simonpasquier

jan--f:
Got it...found the discussion around this.
This currently gets deployed on all clusters, i.e. every user will see this dashboard, sometimes it'll be empty. Is that the chosen option for now or are we still looking for the right way to improve the user experience?

PR #2549, opened 2025-01-06
MON-4200: disable --auto-gomemlimit for Prometheus on SNO until we can ensure it won't result in excessive CPU usage

machine424:
…il we can ensure it won't result in excessive CPU usage

requires https://github.com/openshift/prometheus/pull/227

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1456,7 +1456,7 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret) 
 	}
 
 	setupAlerting(p, platformAlertmanagerService, f.namespace)
-	f.setupGoGC(p)
+	f.adjustGoGCRelatedConfig(p)
 
 	for i, container := range p.Spec.Containers {
 		switch container.Name {
@@ -1495,7 +1495,7 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret) 
 	return p, nil
 }
 
-func (f *Factory) setupGoGC(p *monv1.Prometheus) {
+func (f *Factory) adjustGoGCRelatedConfig(p *monv1.Prometheus) {
 	if f.infrastructure.HighlyAvailableInfrastructure() {
 		return
 	}
@@ -1513,6 +1513,11 @@ func (f *Factory) setupGoGC(p *monv1.Prometheus) {
 		// be savvy on CPU hence set GOGC=100 (Go runtime default) in this
 		// case.
 		p.Spec.Containers[i].Env = append(p.Spec.Containers[i].Env, v1.EnvVar{Name: "GOGC", Value: "100"})
+
+		// Until we're certain setting GOMEMLIMIT to 0.9 (default ratio) of detected maximum
+		// container or system memory won't result in excessive CPU usage, we're disabling the
+		// auto setting for SNO.
+		p.Spec.AdditionalArgs = append(p.Spec.AdditionalArgs, monv1.Argument{Name: "no-auto-gomemlimit"})
 	}
 }
 
@@ -1805,7 +1810,7 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret) (*monv1.Prometheus,
 		}
 	}
 
-	f.setupGoGC(p)
+	f.adjustGoGCRelatedConfig(p)
 
 	if f.config.UserWorkloadConfiguration.Alertmanager.Enabled {
 		setupAlerting(p, userWorkloadAlertmanagerService, f.namespaceUserWorkload)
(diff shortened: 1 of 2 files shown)


pkg/manifests/manifests.go
```diff
@@ -1491,7 +1491,7 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret)
 	return p, nil
 }
 
-func (f *Factory) setupGoGC(p *monv1.Prometheus) {
+func (f *Factory) adjustGoGCConfig(p *monv1.Prometheus) {
```
rexagod, line 1494: Maybe something like:

```suggestion
func (f *Factory) adjustGoSettings(p *monv1.Prometheus) {
```
Since this affects the GOMEMLIMIT too now.

pkg/manifests/manifests_test.go
```diff
@@ -4256,14 +4259,16 @@ func TestPrometheusGoGC(t *testing.T) {
...
-			if tc.exp == "" {
+			if tc.expectedGOGC == "" {
 				for _, env := range c.Env {
 					require.NotEqual(t, env.Name, "GOGC")
 				}
 				return
 			}
 
-			require.Contains(t, c.Env, v1.EnvVar{Name: "GOGC", Value: tc.exp})
+			require.Contains(t, c.Env, v1.EnvVar{Name: "GOGC", Value: tc.expectedGOGC})
```
rexagod, line 4269: +100!

pkg/manifests/manifests_test.go
```diff
@@ -4256,14 +4259,16 @@ func TestPrometheusGoGC(t *testing.T) {
...
+
+			require.Equal(t, tc.autoGOMEMLIMITDisabled, argumentPresent(*c, "--no-auto-gomemlimit"))
```
rexagod, line 4271: We could drop the `tc.autoGOMEMLIMITDisabled` field as this could be safely derived from `ir.HighlyAvailableInfrastructure()`, as that's the only case where this is disabled for now (else enabled)?
```suggestion
			require.Equal(t, tc.ir.HighlyAvailableInfrastructure(), argumentPresent(*c, "--no-auto-gomemlimit"))
```

rexagod (commented):
Since there is no ticket linked to this, I was wondering if we saw any instances of 10% memory reduction bottleneck-ing the CPU on SNO?

rexagod:
I'm asking https://github.com/openshift/cluster-monitoring-operator/pull/2549#pullrequestreview-2533827847 as any observed insight should help me set a more meaningful buffer threshold in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/1010#discussion_r1903810069.

machine424:
Thanks for the review, this is still WIP actually, requires https://github.com/openshift/prometheus/pull/227. I've marked it as such. I'll get back to you later.

pkg/manifests/manifests.go
machine424: renaming as `adjustGoGCRelatedConfig` because `GOMEMLIMIT` is also related to Garbage collection.

pkg/manifests/manifests_test.go
machine424: yep, but I'm trying to avoid the same logic in tests as well in case sth is broken/wrong somewhere.
Also having it explicit makes reading the test cases easier.

rexagod:
Needs a `make versions`?

machine424:
> Needs a `make versions`?

I'd make the bot take care of that to not pollute the PR.

pkg/manifests/manifests.go
```diff
@@ -1513,6 +1513,11 @@ func (f *Factory) setupGoGC(p *monv1.Prometheus) {
 		// be savvy on CPU hence set GOGC=100 (Go runtime default) in this
 		// case.
 		p.Spec.Containers[i].Env = append(p.Spec.Containers[i].Env, v1.EnvVar{Name: "GOGC", Value: "100"})
+
+		// Until we're certain setting GOMEMLIMIT to 0.9 (default ratio) of detected maximum
+		// container or system memory won't result in excessive CPU usage, we're disabling the
+		// auto setting for SNO.
+		p.Spec.Containers[i].Args = append(p.Spec.Containers[i].Args, "--no-auto-gomemlimit")
```
simonpasquier, line 1520: should rather use `p.Spec.AdditionalArgs`?

machine424: Good catch, even though we're overriding all the args, no test seems to be failing...
e.g. https://github.com/openshift/cluster-monitoring-operator/blob/5ee3bd44ad0dd03b3e55a8693d9b0918aa613953/test/e2e/config_test.go#L231
I'll take a look
/hold

machine424: I'll check with the SNO team, we'd probably pick the same upgrade one than payload...

machine424:
/retitle OCPBUGS-55736: disable --auto-gomemlimit for Prometheus on SNO until we can ensure it won't result in excessive CPU usage

PR #2551, opened 2025-01-07
OCPBUGS-48074: Fixing telemetry rule for accelerators

yevgeny-shnaidman:
The list of the supported accelerators now includes ~50 devices, which makes it difficult to provide in the rule. this PR change the selection for vendors only, and adds the vendors currently supported by the node-exporter.
The list of the reported models is specified in the:
models are explicitly listed in https://github.com/openshift/node_exporter/blob/b7efb8c0f7d45f4ccb9120d4bcfb60ebe721ed69/collector/accelerators.go#L62


yevgeny-shnaidman:
/retitle OCPBUGS-48074: Fixing telemetry rule for accelerators

rexagod:
I can see this allows all models for an increased number of vendors. Is this intentional?

jan--f:
@yevgeny-shnaidman can you please give us some indication how many more time series we can expect from this?
Do we have an idea if some users will have multiple models or different vendors in a give cluster?
I guess this doesn't dramatically change the amount of telemetry metrics per cluster?

This also needs a `make generate` + commit.

yevgeny-shnaidman:
> I can see this allows all models for an increased number of vendors. Is this intentional?

@rexagod the rule is based on the metric created by node-exporter. This metric is created only for a specific devices of the specific vendors. So, this rule will only be activated only for vnedors and devices (models) specified in the node-exporter

yevgeny-shnaidman:
> @yevgeny-shnaidman can you please give us some indication how many more time series we can expect from this? Do we have an idea if some users will have multiple models or different vendors in a give cluster? I guess this doesn't dramatically change the amount of telemetry metrics per cluster?
> 
> This also needs a `make generate` + commit.

this rules just sums up the metrics created by node-exporter. Node-exporter creates the metrics based on the specific list of models and device. Basically we will send one telemetry metric per all the vendor_model metrics in the cluster ( if we have 3 NVIDIA 40T devices on one node, and 2 NVIDIA 40T devices on another , the telemetry should receive one NVIDIA-40T=5). I want to remove validation for list of models, since it is already checked in the node-exporter. 
We don't know if the user can have more then one type of GPUs in the cluster,  but if they do , we need to report it

PR #2552, opened 2025-01-08
OCPBUGS-48127:Fixing telemetry rule for accelerators

yevgeny-shnaidman:
The list of the supported accelerators now includes ~50 devices, which makes it difficult to provide in the rule. this PR change the selection for vendors only, and adds the vendors currently supported by the node-exporter.
The list of the reported models is specified in the: models are explicitly listed in https://github.com/openshift/node_exporter/blob/b7efb8c0f7d45f4ccb9120d4bcfb60ebe721ed69/collector/accelerators.go#L62


yevgeny-shnaidman:
/retitle OCPBUGS-48127:Fixing telemetry rule for accelerators

PR #2553, opened 2025-01-08
MON-4115: expose label metrics for jobs and cronjobs

rexagod:
Adds jobs and cronjobs to the exposed set of label metrics:
* `--metric-labels-allowlist=pods=[*],nodes=[*],namespaces=[*],persistentvolumes=[*],persistentvolumeclaims=[*],poddisruptionbudgets=[*],jobs=[*],cronjobs=[*]`

***

The linked issue is a feature request by a customer that requested these metrics to build their dashboards. I'm not sure what the qualifying factors were for the inclusion of the existing set of exposed label metrics `--metric-labels-allowlist=pods=[*],nodes=[*],namespaces=[*],persistentvolumes=[*],persistentvolumeclaims=[*],poddisruptionbudgets=[*]`, but I'm assuming we do so when there's an explicit and reasonable request to expose them.

Nonetheless, I've opened this PR to set a ground for discussion if we want to incorporate this, or not.


rexagod:
I'm having some issues running `make generate` on `macOS`, it's a gift that keeps giving.

machine424:
left a comment on the ticket.

rexagod:
@machine424 Can we merge this then (since configuring this is a broader scope that needs to be put up for discussion)? In the meantime this will not have any unintended effects in itself.

rexagod:
Support extending the labels-allowlist based on this [discussion](https://redhat-internal.slack.com/archives/C026QJY8WJJ/p1744291621985059?thread_ts=1740473272.376709&cid=C026QJY8WJJ).

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/openshift/cluster-monitoring-operator
 
-go 1.23.0
+go 1.24.0
 
-toolchain go1.24.0
```
rexagod: Interesting, `go mod tidy` seems to bump the `go` directive automatically.

rexagod: `GOTOOLCHAIN=local go mod tidy` gives the same result. Strange, maybe something's up with my env itself.

rexagod: So my local was 1.24 which is why this had no effect. Using `gvm` to match this.

rexagod: Ah so, toolchain version is preferred over the least supported one (`go` directive).

```
[cluster-monitoring-operator] gvm use go1.23.0                                                                                        MON-4115  ✱
Now using version go1.23.0
[cluster-monitoring-operator] go version                                                                                              MON-4115  ✱
go: downloading go1.24.4 (darwin/arm64)
go version go1.24.4 darwin/arm64
```

pkg/manifests/manifests.go
```diff
@@ -767,6 +768,19 @@ func (f *Factory) KubeStateMetricsDeployment(enableCRSMetrics bool) (*appsv1.Dep
 			if enableCRSMetrics {
 				d.Spec.Template.Spec.Containers[i].Args = append(container.Args, flagCRSConfigFile)
 			}
+			additionalAllowList := f.config.ClusterMonitoringConfiguration.KubeStateMetricsConfig.AdditionalLabelsAllowList
+			if additionalAllowList != nil && *additionalAllowList != "" {
+				allowListValidator := ksmpkg.LabelsAllowList{}
+				err = allowListValidator.Set(*additionalAllowList)
+				if err != nil {
+					return nil, fmt.Errorf("error parsing allowlist: %v", err)
+				}
```
rexagod: I see, the bump happens for this PR specifically, and not the `main` since, we pull KSM in which specifies a higher version. This is blocked until we move to `1.24.x`.

```
[cluster-monitoring-operator] GOTOOLCHAIN=go1.23.0 go mod tidy                                                                       2e7fc5398  ✭
go: k8s.io/kube-state-metrics/v2@v2.16.0 requires go >= 1.24.0 (running go 1.23.0; GOTOOLCHAIN=go1.23.0)
```

rexagod: Post go1.24 update, bumping dependencies reveals a missing `k8s.io/api/resource/v1beta2` package for the commit pinned in the `replace` directive. Upon updating it, it seems to clash with certain vendored dependencies, at which point, I believe it's simpler to copy the validation function from KSM here.

rexagod:
Ping @machine424 as this builds on his idea in [this thread](https://redhat-internal.slack.com/archives/C026QJY8WJJ/p1744291621985059?thread_ts=1740473272.376709&cid=C026QJY8WJJ).

rexagod:
(bump)

pkg/manifests/manifests.go
```diff
@@ -3599,3 +3643,65 @@ func hashStringMap(m map[string]string) string {
 	}
 	return hashByteMap(byteMap)
 }
+
+func validateLabelsAllowListFormat(value string) error {
```
rexagod: Taken from: https://github.com/kubernetes/kube-state-metrics/blob/5b7a11eed724511405a0148026ee96ac426f3787/pkg/options/types.go#L237, to ensure the additional list can be validated.

pkg/manifests/types.go
```diff
@@ -175,6 +175,11 @@ type KubeStateMetricsConfig struct {
 	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
 	// Defines a pod's topology spread constraints.
 	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+	// Defines label-metrics' allow list for resources in addition to the default one.
+	// Currently, this is only supported for `jobs` and `cronjobs`, due to cardinality concerns.
+	// This follows the format: resource1=[label1,label2,labelN...],...,resourceN=[...],
+	// which is then validated and appended to the default labels' allow list.
+	AdditionalLabelsAllowList *string `json:"additionalLabelsAllowList,omitempty"`
```
rexagod: @marioferh Following on @machine424's [comment](https://redhat-internal.slack.com/archives/C026QJY8WJJ/p1762261552991869?thread_ts=1762159131.948859&cid=C026QJY8WJJ), could you affirm if this would be okay to port to the CRD as is, or if you believe there are any changes to be made here? Thanks!

danielmellado: That's still a WIP with the api folks, so far I'd prefer to wait until we get an ack from them in Prometheus and PrometheusK8s resources before adding any additional fields. While it seems pretty much straightforward I'd hold this for now.

rexagod: IIUC I think we can revisit this now?

danielmellado: Yes, all Prometheus related CRDs are already merged, so feel free to take a look and adapt! ;)

pkg/manifests/manifests.go
```diff
@@ -762,6 +761,34 @@ func (f *Factory) KubeStateMetricsDeployment() (*appsv1.Deployment, error) {
 			if f.config.ClusterMonitoringConfiguration.KubeStateMetricsConfig.Resources != nil {
 				d.Spec.Template.Spec.Containers[i].Resources = *f.config.ClusterMonitoringConfiguration.KubeStateMetricsConfig.Resources
 			}
+			additionalAllowList := f.config.ClusterMonitoringConfiguration.KubeStateMetricsConfig.AdditionalLabelsAllowList
+			if additionalAllowList != nil && *additionalAllowList != "" {
+				err = validateLabelsAllowListFormat(*additionalAllowList)
```
simonpasquier: IMHO the validation should be part of `pkg/manifests/config.go` so the admission webhook can reject bad configurations.

pkg/manifests/types.go
```diff
@@ -175,6 +175,11 @@ type KubeStateMetricsConfig struct {
 	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
 	// Defines a pod's topology spread constraints.
 	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+	// Defines label-metrics' allow list for resources in addition to the default one.
+	// Currently, this is only supported for `jobs` and `cronjobs`, due to cardinality concerns.
+	// This follows the format: resource1=[label1,label2,labelN...],...,resourceN=[...],
```
simonpasquier: I would prefer to not map directly to the KSM CLI argument and adopt a more structured declaration which would translate easily to the config CRD.

````
additionalResourceLabels:
# resource values should be unique
# validation should ensure that only supported resources are accepted.
- resource: jobs
  labels:
  - foo
  - bar
 - resource: cronjobs
   labels:
   - *
````

File: pkg/manifests/types.go [hunk 7]
```diff
@@ -193,6 +203,11 @@ type KubeStateMetricsConfig struct {
 	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
 	// Defines a pod's topology spread constraints.
 	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+	// Defines additional resource labels to expose as metrics in addition to
+	// the default labels. Currently, only `jobs` and `cronjobs` resources are
+	// supported due to cardinality concerns. Each entry specifies a resource
+	// name and a list of label names (use `*` to expose all labels).
+	AdditionalResourceLabels []ResourceLabels `json:"additionalResourceLabels,omitempty"`
```
rexagod, line 210: Please note, this was renamed from `AdditionalLabelsAllowList` to this.

rexagod:
Updated commit and PR descriptions to reflect the newer field name.

pkg/manifests/manifests_test.go
```diff
@@ -3834,6 +3834,91 @@ func TestKubeStateMetrics(t *testing.T) {
...
+			c.SetImages(map[string]string{
+				"kube-state-metrics": "docker.io/openshift/origin-kube-state-metrics:latest",
+				"kube-rbac-proxy":    "docker.io/openshift/origin-kube-rbac-proxy:latest",
+			})
+
+			f := NewFactory("openshift-monitoring", "openshift-user-workload-monitoring", c, defaultInfrastructureReader(), &fakeProxyReader{}, NewAssets(assetsPath), &APIServerConfig{}, &configv1.Console{})
+
+			d, err := f.KubeStateMetricsDeployment()
+			if err != nil {
+				t.Fatal(err)
```
simonpasquier: (nit) can we use `require.NoError(t, err)` which is less verbose? Same for the other calls to t.Fatal(f).

pkg/manifests/config.go
```diff
@@ -363,9 +363,40 @@ func NewConfigFromString(content string) (*Config, error) {
...
+			return fmt.Errorf("%w: additionalResourceLabels: resource name must not be empty", ErrConfigValidation)
+		}
+		if !slices.Contains(supportedResourceLabelsResources, rl.Resource) {
+			return fmt.Errorf("%w: additionalResourceLabels: unsupported resource %q, supported resources are: %v", ErrConfigValidation, rl.Resource, supportedResourceLabelsResources)
+		}
+		if seen[rl.Resource] {
+			return fmt.Errorf("%w: additionalResourceLabels: duplicate resource %q", ErrConfigValidation, rl.Resource)
+		}
+		seen[rl.Resource] = true
+		if len(rl.Labels) == 0 {
```
simonpasquier, line 393: should we also check for non-empty values?

rexagod:
TODO: Introduce this field in https://github.com/openshift/api/pull/2778 once this is in.

pkg/manifests/config.go
rexagod: Added checks for empty and duplicate non-empty labels.

rexagod:
/unhold
/test e2e-hypershift-conformance

Failures do not seem to stem from this PR.

simonpasquier:
_comment deleted because wrong PR_

PR #2560, opened 2025-01-15
OCPBUGS-48381: jsonnet: switch telemeter to main

jan--f:


machine424:
/lgtm
thanks

PR #2569, opened 2025-02-11
OCPBUGS-50703: Add incidents metric to telemetry

falox:


Documentation/data-collection.md
```diff
@@ -1231,6 +1231,13 @@ data:
     #
     # openshift:openshift_network_operator_ipsec_state:info shows the cluster ipsec status (Disabled, External, Full) and whether the legacy or new API was used to set the status
     - '{__name__="openshift:openshift_network_operator_ipsec_state:info"}'
+    #
+    # owners: (https://github.com/openshift/cluster-health-analyzer)
+    #
+    # cluster:health:group_severity:count shows the total number of firing incidents by severity
+    # Expected labels:
+    # - severity: "critical", "warning", "info" or "none".
+    - '{__name__="cluster:health:group_severity:count"}'
```
jan--f, line 1240: If we're only expecting the 4 label values, let's match those for:
```suggestion
    # - severity: "critical", "warning", "info" or "none".
    - '{__name__="cluster:health:group_severity:count", severity=~"critical|warning|info|none"}'
```

jan--f (changes requested):
I have one small addition request. Also can you please link me to the recording rule for this? I couldn't find anything in https://github.com/openshift/cluster-health-analyzer

falox:
> I have one small addition request. Also can you please link me to the recording rule for this? I couldn't find anything in https://github.com/openshift/cluster-health-analyzer

@jan--f There is no recording rule, the metric is aggregated and served directly [here](https://github.com/openshift/cluster-health-analyzer/blob/main/pkg/processor/processor.go#L181). You can see the isolated code in its [PR](https://github.com/openshift/cluster-health-analyzer/pull/32/files). Let me know if it's enough

Documentation/data-collection.md
falox: Yep, updated

jan--f:
:+1: One last thing: since the intention is to backport this, I'd recommend to create the OCPBUGS for 4.19 right now and link the PR. We can then use the backport automations.

jan--f:
> > I have one small addition request. Also can you please link me to the recording rule for this? I couldn't find anything in https://github.com/openshift/cluster-health-analyzer
> 
> @jan--f There is no recording rule, the metric is aggregated and served directly [here](https://github.com/openshift/cluster-health-analyzer/blob/main/pkg/processor/processor.go#L181). You can see the isolated code in its [PR](https://github.com/openshift/cluster-health-analyzer/pull/32/files). Let me know if it's enough

I see, yes that suffices here. I would however encourage you to consider changing the metric names to something like `cluster_health_group_severity_count`. Using `:` is certainly legal but within OpenShift there is a convention that only recording rule names contain `:`. Hence my question. 
Functionally its not a problem, but you might end up with repeated questions about this "recording rule".

jan--f (approved):

falox:
/retitle OCPBUGS-50703: Add incidents metric to telemetry

jan--f:

A telemetry addition is very low risk.

/cherrypick release-4.18

PR #2571, opened 2025-02-14
WIP: MON-4150: Set fallbackScrapeProtocol for Prometheus v3

slashpai:
For Platform Prometheus using scrapeClass
For UWM Prometheus setting globally

Keeping it in draft and needs Prometheus v3 as well to work

### DIFF
--- a/assets/prometheus-k8s/prometheus.yaml
+++ b/assets/prometheus-k8s/prometheus.yaml
@@ -197,7 +197,8 @@ spec:
       openshift.io/cluster-monitoring: "true"
   ruleSelector: {}
   scrapeClasses:
-  - name: tls-client-certificate-auth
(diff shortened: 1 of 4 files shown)


assets/prometheus-k8s/prometheus.yaml
```diff
@@ -197,7 +197,8 @@ spec:
       openshift.io/cluster-monitoring: "true"
   ruleSelector: {}
   scrapeClasses:
-  - name: tls-client-certificate-auth
+  - fallbackScrapeProtocol: PrometheusText0.0.4
```
machine424, line 200: Let's just have it for UWM as we don't have control over its targets.
targets scraped by Platform Prom should be fixed instead.

PR #2574, opened 2025-03-03
NO-JIRA: chore: versions: kubeRbacProxy's primary branch isn't main yet

machine424:


jan--f:
oh right...krp is not a "Monitoring repo" anymore.
/lgtm

PR #2575, opened 2025-03-03
OU-515: remove old nginx configmap deletion

jgbernalp:
This PR removes the delete task of the ConfigMap that was used to hold the nginx configuration for the old version of the monitoring plugin. In 4.19 this is not necessary anymore.


machine424:
Thanks
/lgtm

PR #2580, opened 2025-03-13
ON-4043: Configuring external Alertmangers with proxy_url

marioferh:


pkg/manifests/types.go
```diff
@@ -780,6 +780,10 @@ type AdditionalAlertmanagerConfig struct {
 	Timeout *string `json:"timeout,omitempty"`
 	// Defines the TLS settings to use for Alertmanager connections.
 	TLSConfig TLSConfig `json:"tlsConfig,omitempty"`
+	// Defines proxy_url settings for Alertmanager
+	ProxyURL string `json:"proxyUrl,omitempty"`
+	// Defines no_proxy settings for Alertmanager
+	NoProxy string `json:"noProxy,omitempty"`
```
simonpasquier, line 786: IMHO we don't need to expose settings to the end-users and only rely on the cluster-wide proxy config.

marioferh: done

pkg/manifests/manifests.go
```diff
@@ -1161,6 +1161,17 @@ func (f *Factory) ThanosRulerAlertmanagerConfigSecret() (*v1.Secret, error) {
 		return nil, err
 	}
 
+	switch alertingConfiguration.Alertmanagers[0].Scheme {
```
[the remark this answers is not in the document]
marioferh: I don't understand your comment:
`alertingConfiguration` is: `var alertingConfiguration thanosAlertingConfiguration`

pkg/manifests/manifests.go
```diff
@@ -1460,6 +1471,8 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret)
 
 	for i, container := range p.Spec.Containers {
 		switch container.Name {
+		case "prometheus":
+			f.injectProxyVariables(&p.Spec.Containers[i])
```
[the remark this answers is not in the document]
marioferh: done

pkg/manifests/manifests_test.go
```diff
@@ -1988,12 +1988,6 @@ func TestPrometheusK8sAdditionalAlertManagerConfigsSecret(t *testing.T) {
 		expected       string
 		mountedSecrets []string
 	}{
-		{
-			name:           "empty config",
-			config:         "",
-			expected:       "[]\n",
-			mountedSecrets: []string{},
-		},
```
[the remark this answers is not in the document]
marioferh: Unintentional, done.

pkg/manifests/manifests.go
```diff
@@ -1795,6 +1809,7 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret) (*monv1.Prometheus,
 				FailureThreshold: 240,
 			}
 
+			f.injectProxyVariables(&p.Spec.Containers[i])
```
simonpasquier, line 1812: can we have a similar comment here?

pkg/manifests/amcfg.go
```diff
@@ -180,6 +183,16 @@ func ConvertToThanosAlertmanagerConfiguration(ta []AdditionalAlertmanagerConfig)
 
 		cfg.StaticConfigs = a.StaticConfigs
 
+		switch cfg.Scheme {
```
simonpasquier, line 186: ThanosRuler only supports the proxy URL config but CMO can take into account the NoProxy string to decide what is the correct value (HTTP Proxy, HTTPS Proxy or empty string). 
I'd use https://pkg.go.dev/golang.org/x/net/http/httpproxy#Config.ProxyFunc for this.

marioferh: done

pkg/manifests/manifests.go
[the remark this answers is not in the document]
marioferh: done

pkg/manifests/amcfg.go
```diff
@@ -180,6 +185,30 @@ func ConvertToThanosAlertmanagerConfiguration(ta []AdditionalAlertmanagerConfig)
...
 
+		httpConfig := httpproxy.Config{
+			HTTPProxy:  f.proxy.HTTPProxy(),
+			HTTPSProxy: f.proxy.HTTPSProxy(),
+			NoProxy:    f.proxy.NoProxy(),
+		}
+
+		proxyFunc := httpConfig.ProxyFunc()
+
+		if len(cfg.StaticConfigs) > 0 && cfg.StaticConfigs[0] != "" {
```
simonpasquier, line 196: It would be good to have a comment here explaining that we assume all hosts follow the same proxy policy (e.g. use a proxy or none). As well as a note in the configuration documentation to avoid user misunderstanding.

simonpasquier: (nit) we could also iterate over all hosts and stop at the first one returning a non-nil proxy URL. I feel that it's more predictable and simpler to document, WDYT?

marioferh: I did it like that because proxyFunc only return nil if no proxy is defined, then the scenario of a host returning nil is not possible. 

From api:
`// A nil URL and nil error are returned if no proxy is defined in the
// environment, or a proxy should not be used for the given request, as
// defined by NO_PROXY.
//
// As a special case, if req.URL.Host is "localhost" or a loopback address
// (with or without a port number), then a nil URL and nil error will be returned.

But your way it is better for the documentation

marioferh: done

simonpasquier (commented):
/lgtm

the documentation update on how CMO picks up the proxy URL in the case of ThanosRuler can go in another PR.

PR #2573, opened 2025-02-28
feat(auto_docs_examples): adjust and add more examples

machine424:
* [ ] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #2585, opened 2025-04-02
OCPBUGS-54211: return rules from tenancy endpoint

PeterYurkovich:
While investigating [OCPBUGS-54211](https://issues.redhat.com/browse/OCPBUGS-54211) I found two issues.

1. A frontend bug which didn't use silences from the selected namespace. Addressed in openshift/monitoring-plugin#374
2. No alert rules were returned from the tenancy endpoint for any namespace other than `openshift-monitoring`, even when the user had appropriate permissions.

When comparing the thanos-querier deployment in release-4.17 to the working one in release-4.18, I found that the `prom-label-proxy` image didn't use the `--rules-with-active-alerts` flag in release-4.17. Once the flag was enabled the thanos-querier deployment returns the appropriate alert rules when the tenancy endpoint is queried with appropriate permissions. 

---

### DIFF
--- a/assets/thanos-querier/deployment.yaml
+++ b/assets/thanos-querier/deployment.yaml
@@ -168,6 +168,7 @@ spec:
         - --label=namespace
         - --enable-label-apis
         - --error-on-replace
+        - --rules-with-active-alerts
         image: quay.io/prometheuscommunity/prom-label-proxy:v0.11.0
         name: prom-label-proxy
         resources:
--- a/jsonnet/components/thanos-querier.libsonnet
+++ b/jsonnet/components/thanos-querier.libsonnet
@@ -514,6 +514,7 @@ function(params)
                   '--label=namespace',
                   '--enable-label-apis',
                   '--error-on-replace',
+                  '--rules-with-active-alerts',
                 ],
                 resources: {
                   requests: {


machine424:
I wasn't able to have a test from https://github.com/openshift/cluster-monitoring-operator/pull/2573 fail on 4.17
The reproducer in https://github.com/openshift/cluster-monitoring-operator/pull/2573 doesn't seem to complain about alerts/rules, but maybe I'm missing sth.

PeterYurkovich:
I had performed a manual regression check on 4.17.15, and had found that following the reproduction in the story, the user was unable retrieve any rules from the tenancy path. The network request was simply returning an empty array rather than the expected alerts. I manually ran curl requests and used the UI to confirm and that adding and removing the argument was able to fix the issue.

I was also unable to figure out why the behavior happened, but if the reproducer didn't find the same issues on nightly then I will look to test there. Perhaps there was some unrelated regression in 4.17.15 which had caused some very odd behavior. I will look to investigate some more tomorrow.

machine424:
Thanks.
To be more precise, I only checked that the user wouldn't get a 403 on alerts/rules.
I didn't check whether the rules list is empty as I expect the behavior before the introduction of `--rules-with-active-alerts` in `4.18` to be consistent...
the `TestUserWorkloadMonitoringAlerting.assertTenancyForRulesAndAlerts` should be enforcing that behavior, if you can reproduce the problem there, it'd be great.

PR #2588, opened 2025-04-04
MON-4161: add topology telemetry signals

jan--f:


jan--f:
Accompanied by https://github.com/openshift/cluster-kube-apiserver-operator/pull/1829

machine424:
/lgtm
thanks
(nit: I thought the metrics were called `xxx:sum` instead of `xxx:info` as stated in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1829 title, I find `xxx:sum` clearer or even use `max()` and have them as `xxx:max`, if the sum isn't needed and we want them to be tolerant to duplicates...)

jan--f:
> /lgtm thanks (nit: I thought the metrics were called `xxx:sum` instead of `xxx:info` as stated in [openshift/cluster-kube-apiserver-operator#1829](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1829) title, I find `xxx:sum` clearer or even use `max()` and have them as `xxx:max`, if the sum isn't needed and we want them to be tolerant to duplicates...)

Good catch, I changed the name last minute prompted by a review and now fixed the PR title. We have prior art for `:info` metrics (e.g. https://github.com/openshift/cluster-monitoring-operator/blob/29b581be202cb1837960bc682d74f50a2e592979/manifests/0000_50_cluster-monitoring-operator_04-config.yaml#L1225) and since this reflects an `openshift/api` field, the sum is really just serving to aggregate away some labels. So I think the review comment was warranted and I changed it accordingly. We could use a different aggregation but the result won't change.

machine424:
Thanks for clarifying that.
(`openshift:openshift_network_operator_ipsec_state:info` uses a [group by()](https://github.com/openshift/cluster-network-operator/blob/2ca824bfbd32ad196d1ac8e4167bfd369f464f66/bindata/cluster-network-operator/prometheus.yaml#L15-L18) but yes let's keep it like that and change later if needed)

jan--f:
@machine424 mind re-lgtming this?

PR #2590, opened 2025-04-14
MON-4126: set fallbackScrapeProtocol: 'PrometheusText1.0.0' as default for all UWM Prometheus targets for backward compatibility with Prometheus v2 until a better migration process is available

machine424:

### DIFF
--- a/assets/prometheus-user-workload/prometheus.yaml
+++ b/assets/prometheus-user-workload/prometheus.yaml
@@ -260,6 +260,10 @@ spec:
       operator: In
       values:
       - leaf-prometheus
+  scrapeClasses:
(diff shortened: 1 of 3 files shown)


rexagod:
/title MON-4126: set fallbackScrapeProtocol: 'PrometheusText1.0.0' as default for all UWM Prometheus targets for backward compatibility with Prometheus v2 until a better migration process is available
/lgtm

rexagod:
/retitle MON-4126: set fallbackScrapeProtocol: 'PrometheusText1.0.0' as default for all UWM Prometheus targets for backward compatibility with Prometheus v2 until a better migration process is available

PR #2591, opened 2025-04-15
[release-4.18] OCPBUGS-56251: set upgradable false when alertmanager v1 is specified

slashpai:
Prometheus v3 no longer supports alertmanager v1

OCP 4.19 will fail early (InvalidConfigXXX) in case alertmanager apiVersion: v1 is still specified in one of the two configmaps.

To avoid that InvalidConfigXXX from triggering just after the 4.19 upgrade, we can make users bump their alertmanagers and only use apiVersion: v2 in the configmaps.

We can do that by making CMO set upgradable=false when it sees apiVersion: v1 in one of the configmaps and make sure clusters go through that 4.18.x before upgrading to 4.19.

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -52,7 +52,12 @@ const (
 	automaticBodySizeLimit = "automatic"
 
 	configKey = "config.yaml"
+
+	clusterMonitorConfigMapName      = "openshift-monitoring/cluster-monitoring-config"
+	userWorkloadMonitorConfigMapName = "openshift-user-workload-monitoring/user-workload-monitoring-config"
 )
+
+var errAlertmanagerV1NotSupported = errors.New("alertmanager's apiVersion=v1 is no longer supported, v2 has been available since Alertmanager 0.16.0")
 
 type InvalidConfigWarning struct {
 	ConfigMap string
@@ -209,6 +214,31 @@ func (u *UserWorkloadConfiguration) checkThanosRulerEvaluationInterval() error {
 	return nil
 }
 
+func (u *UserWorkloadConfiguration) checkAlertmanagerVersion() error {
+	if u.Prometheus != nil {
+		for _, amConfig := range u.Prometheus.AlertmanagerConfigs {
+			if alertmanagerV1(amConfig.APIVersion) {
+				return fmt.Errorf("%w: found in prometheus.additionalAlertmanagerConfigs", errAlertmanagerV1NotSupported)
+			}
+		}
+	}
+	if u.ThanosRuler != nil {
+		for _, amConfig := range u.ThanosRuler.AlertmanagersConfigs {
+			if alertmanagerV1(amConfig.APIVersion) {
+				return fmt.Errorf("%w: found in thanosRuler.additionalAlertmanagerConfigs", errAlertmanagerV1NotSupported)
+			}
+		}
+	}
+
+	return nil
+}
+
+func alertmanagerV1(version string) bool {
+	// Only meant to guide users by failing early in case v1 Alertmanager is still referenced,
+	// this is not meant to validate the apiVersion field.
+	return version == "v1"
+}
+
 func (u *UserWorkloadConfiguration) check() error {
 	if u == nil {
 		return nil
@@ -340,7 +370,7 @@ func newConfig(content []byte, collectionProfilesFeatureGateEnabled bool) (*Conf
 	wCmc := defaultClusterMonitoringConfiguration()
 	wErr := UnmarshalStrict(content, &wCmc)
 	if wErr != nil {
-		warning = &InvalidConfigWarning{ConfigMap: "openshift-monitoring/cluster-monitoring-config", Err: wErr}
+		warning = &InvalidConfigWarning{ConfigMap: clusterMonitorConfigMapName, Err: wErr}
 	}
 
 	cmc := defaultClusterMonitoringConfiguration()
@@ -351,6 +381,18 @@ func newConfig(content []byte, collectionProfilesFeatureGateEnabled bool) (*Conf
 
 	c.ClusterMonitoringConfiguration = &cmc
 	c.applyDefaults()
+
+	// Only to assist with the migration to Prometheus 3; fail early if Alertmanager v1 is still in use.
+	if wErr := c.checkAlertmanagerVersion(); wErr != nil {
+		if warning != nil {
+			wErr = errors.Join(warning.Err, wErr)
+		}
+		warning = &InvalidConfigWarning{
+			ConfigMap: clusterMonitorConfigMapName,
+			Err:       wErr,
+		}
+	}
+
 	c.UserWorkloadConfiguration = NewDefaultUserWorkloadMonitoringConfig()
 
 	return &c, warning, nil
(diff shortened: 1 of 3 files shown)


pkg/operator/operator.go
```diff
@@ -907,6 +907,16 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 		klog.Errorf("error occurred while setting status to done: %v", err)
 	}
 
+	// Set Upgradable=False if alertmanager apiVersion: v1 is still specified in cluster-monitoring-config configmap
+	if config.CheckAlertmanagerVersion() != nil {
+		err = o.client.StatusReporter().SetUpgradeable(ctx, configv1.ConditionFalse, "", manifests.ErrAlertmanagerV1NotSupported.Error())
```
juzhao, line 912: from golangci-lint job: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2591/pull-ci-openshift-cluster-monitoring-operator-main-golangci-lint/1912329457343926272
```
pkg/operator/operator.go:912:3: ineffectual assignment to err (ineffassign)
		err = o.client.StatusReporter().SetUpgradeable(ctx, configv1.ConditionFalse, "", manifests.ErrAlertmanagerV1NotSupported.Error())
		^
pkg/operator/operator.go:917:3: ineffectual assignment to err (ineffassign)
		err = o.client.StatusReporter().SetUpgradeable(ctx, configv1.ConditionFalse, "", manifests.ErrAlertmanagerV1NotSupported.Error())
		^
```

slashpai: addressed

pkg/operator/operator.go
```diff
@@ -907,6 +907,22 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 		klog.Errorf("error occurred while setting status to done: %v", err)
 	}
 
+	// Set Upgradable=False if alertmanager apiVersion: v1 is still specified in cluster-monitoring-config configmap
+	if config.CheckAlertmanagerVersion() != nil {
```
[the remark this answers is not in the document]
slashpai: I don't see https://github.com/openshift/cluster-monitoring-operator/blob/ff3a4f0218f4985874f2ab872f6f6f440a38e17d/pkg/operator/operator.go#L748 anylonger in main branch, am I missing something?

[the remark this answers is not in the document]
machine424: it was only done for `4.17`.
As the change in this PR should be done against `4.18`

slashpai:
working on changing the branch to 4.18 instead of main

pkg/manifests/config.go
```diff
@@ -709,6 +761,14 @@ func NewUserConfigFromString(content string) (*UserWorkloadConfiguration, *Inval
...
 	}
 
+	// Only to assist with the migration to Prometheus 3; fail early if Alertmanager v1 is still in use.
+	if err := u.checkAlertmanagerVersion(); err != nil {
+		warning = &InvalidConfigWarning{
+			ConfigMap: userWorkloadMonitorConfigMapName,
+			Err:       err,
+		}
+	}
+
```
machine424: To avoid overriding the existing warning

```suggestion
	if wErr := u.checkAlertmanagerVersion(); wErr != nil {
	       if warning != nil {
	          wErr = errors.Join(warning.Err, wErr)
	       }
		warning = &InvalidConfigWarning{
			ConfigMap: userWorkloadMonitorConfigMapName,
			Err:       wErr,
		}
	}

```

(we'll work on this later; introduce a permanent, easy to use warning signal)

pkg/manifests/config.go
```diff
@@ -351,6 +381,15 @@ func newConfig(content []byte, collectionProfilesFeatureGateEnabled bool) (*Conf
 
 	c.ClusterMonitoringConfiguration = &cmc
 	c.applyDefaults()
+
+	// Only to assist with the migration to Prometheus 3; fail early if Alertmanager v1 is still in use.
+	if err := c.checkAlertmanagerVersion(); err != nil {
```
machine424: see below to avoid overriding existing warning.

pkg/manifests/config_test.go
```diff
@@ -784,3 +784,47 @@ func TestDeprecatedConfig(t *testing.T) {
 		})
 	}
 }
+
+// TestUnsupportedAlertmanagerVersion
+// Verify it doesn't fail if Alertmanager v1 is still in use in 4.18.
+func TestUnsupportedAlertmanagerVersion(t *testing.T) {
```
machine424, line 791: to also cover UWM, we can loop over the two parsers as in https://github.com/openshift/cluster-monitoring-operator/blob/34568a9c7f7a0eeb1e8948d03cfc4049adf72362/pkg/manifests/manifests_test.go#L5214-L5226

pkg/manifests/config_test.go
```diff
@@ -784,3 +784,47 @@ func TestDeprecatedConfig(t *testing.T) {
...
+			config: `prometheusK8s:
+  additionalAlertmanagerConfigs:
+    - Scheme: foo
+  `,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			c, _, err := NewConfigFromString(tc.config, true)
+			require.NoError(t, err)
+			err = c.Precheck()
```
machine424: isn't using the warning from `NewConfigFromString` is enough?
I think `c.Precheck()` doesn't run the v1 checks and the test always go through `require.NoError(t, err)`

pkg/manifests/config_test.go
```diff
@@ -784,3 +784,47 @@ func TestDeprecatedConfig(t *testing.T) {
...
 	}
 }
+
+// TestUnsupportedAlertmanagerVersion
+// Verify it doesn't fail if Alertmanager v1 is still in use in 4.18.
+func TestUnsupportedAlertmanagerVersion(t *testing.T) {
+	for _, tc := range []struct {
+		name       string
+		config     string
+		shouldFail bool
```
machine424:
```suggestion
		shouldWarn bool
```

pkg/manifests/config_test.go
```diff
@@ -784,3 +784,47 @@ func TestDeprecatedConfig(t *testing.T) {
...
+	}{
+		{
+			name: "using unsupported Alertmanager v1 API",
+			config: `prometheusK8s:
+  additionalAlertmanagerConfigs:
+    - apiVersion: v1
+  `,
+			shouldFail: false,
```
machine424:
```suggestion
			shouldFail: true,
```

pkg/manifests/manifests_test.go
```diff
@@ -2182,6 +2182,10 @@ func TestPrometheusK8sAdditionalAlertManagerConfigsSecret(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
+			if c.checkAlertmanagerVersion() != nil {
```
machine424: I think it's better to add a `tt.shouldWarn` and just use the already existing warning from `NewConfigFromString`. And don't return early as even with a warning the following should work

pkg/manifests/manifests_test.go
```diff
@@ -2508,6 +2512,11 @@ func TestThanosRulerAdditionalAlertManagerConfigsSecret(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
+
+			if uwc.checkAlertmanagerVersion() != nil {
```
machine424: see above.

machine424 (commented):
thanks for this, some suggestions.

juzhao:
/retitle [release-4.18] OCPBUGS-56251: set upgradable false when alertmanager v1 is specified

juzhao:
LGTM, waiting for others to review

jan--f:
Thanks!
/lgtm
/label backport-risk-assessed

PR #2592, opened 2025-04-15
OCPBUGS-55005: chore: introduce context-rich anb case sensitive unmarshaller for the monitoring and  set CMO to Upgradeable=false if errors are found

machine424:
The strict unmarshaller is currently in advisory mode. Setting CMO to Upgradeable=false ensures that configurations meet the new unmarshaller checks in 4.18 before it is set to be the default in 4.19.

This is similar to what was done in https://github.com/openshift/cluster-monitoring-operator/pull/2494 when the strict unmarshaller was introduced.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -41,6 +41,7 @@ require (
 	k8s.io/metrics v0.31.1
 	k8s.io/utils v0.0.0-20240921022957-49e7df575cb6
 	sigs.k8s.io/controller-runtime v0.19.0
+	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd
 	sigs.k8s.io/yaml v1.4.0
 )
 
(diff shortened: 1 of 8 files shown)


machine424:
the e2e failures are legit as expected, I'll take a look.

juzhao:
LimitExceeded error for prow jobs, example
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2592/pull-ci-openshift-cluster-monitoring-operator-release-4.18-e2e-aws-ovn-techpreview/1914240921386356736
```
level=error msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed during pre-provisioning: failed to create IAM roles: failed to create IAM master role: failed to create master role: LimitExceeded: Cannot exceed quota for RolesPerAccount: 1000
```

PR #2593, opened 2025-04-17
MON-4207: Bump prometheus-operator to v0.81.0

slashpai:


slashpai:
@machine424 can I get review again?
rebased against main

PR #2595, opened 2025-04-24
MON-4282: Multi-tenant support for KSM's CRS

danwinship:
#2078 added support for VPA monitoring as a CRD type, and included changes to check for the existence of the VPA CRD, and only try to generate VPA metrics if the CRD was actually installed, because [if "the VPA CRD is not present in the cluster" then "enabling the custom-resource-state-based metrics will cause kube-state-metrics to error (affecting `KubeStateMetricsListErrors`)"](https://github.com/openshift/cluster-monitoring-operator/pull/2078/commits/bc1174a6f7746950a11bc405bc825a132c4957a7#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR205-R207).

But as far as I can tell, this comment isn't true; ksm's CRD support already does this internally, and doesn't try to monitor resources for CRD types when the corresponding CRD is not installed... (the absence of errors should be visible in the must-gather of the CI for this PR, since the VPA CRD is not present by default).

It seems like there were some bugs around CRD handling in older versions of ksm (eg, https://github.com/kubernetes/kube-state-metrics/issues/2142, https://github.com/kubernetes/kube-state-metrics/issues/2296), but they've been fixed now? So I think this code may have been needed originally, but is not needed any longer, and we can just make ksm do CRD handling always.

(In particular, this is needed if we want to use ksm to track any other CRD types, such as Gateway API.)

/assign @rexagod

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Note: This CHANGELOG is only for the monitoring team to track all monitoring related changes. Please see OpenShift release notes for official changes.
 
+## 4.20
+
+- [#2595](https://github.com/openshift/cluster-monitoring-operator/pull/2595) Multi-tenant support for KSM's CRS feature-set downstream.
+
 ## 4.18
 
 - [#2503](https://github.com/openshift/cluster-monitoring-operator/issues/2503) Expose `scrapeInterval` setting for UWM Prometheus.
--- a/assets/kube-state-metrics/cluster-role.yaml
+++ b/assets/kube-state-metrics/cluster-role.yaml
@@ -129,17 +129,17 @@ rules:
   - list
   - watch
 - apiGroups:
-  - autoscaling.k8s.io
-  resources:
-  - verticalpodautoscalers
-  verbs:
-  - list
-  - watch
-- apiGroups:
   - apiextensions.k8s.io
   resources:
   - customresourcedefinitions
   verbs:
   - get
+  - list
+  - watch
+- apiGroups:
+  - autoscaling.k8s.io
+  resources:
+  - verticalpodautoscalers
+  verbs:
   - list
   - watch
--- a/assets/kube-state-metrics/deployment.yaml
+++ b/assets/kube-state-metrics/deployment.yaml
@@ -36,6 +36,7 @@ spec:
         - --port=8081
         - --telemetry-host=127.0.0.1
         - --telemetry-port=8082
+        - --custom-resource-state-config-file=/etc/kube-state-metrics/custom-resource-state-configmap.yaml
         - |
           --metric-denylist=
           ^kube_secret_labels$,
--- a/jsonnet/components/kube-state-metrics.libsonnet
+++ b/jsonnet/components/kube-state-metrics.libsonnet
@@ -42,17 +42,17 @@ function(params)
 
     clusterRole+: {
       rules+: [
-        {
-          apiGroups: ['autoscaling.k8s.io'],
-          resources: ['verticalpodautoscalers'],
-          verbs: ['list', 'watch'],
-        },
         // CRD read permissions are required for kube-state-metrics to support the CRS feature-set.
         // Refer: https://github.com/kubernetes/kube-state-metrics/pull/1851/files#diff-916e6863e1245c673b4e5965c98dc27bafbd72650fdb38ce65ea73ee6304e027R45-R47
         {
           apiGroups: ['apiextensions.k8s.io'],
           resources: ['customresourcedefinitions'],
           verbs: ['get', 'list', 'watch'],
+        },
+        {
+          apiGroups: ['autoscaling.k8s.io'],
+          resources: ['verticalpodautoscalers'],
+          verbs: ['list', 'watch'],
         },
       ],
     },
--- a/jsonnet/components/kube-state-metrics.libsonnet
+++ b/jsonnet/components/kube-state-metrics.libsonnet
@@ -173,8 +173,8 @@ function(params)
     kubeRbacProxySecret: generateSecret.staticAuthSecret(cfg.namespace, cfg.commonLabels, 'kube-state-metrics-kube-rbac-proxy-config'),
 
     // This removes the upstream addon-resizer and all resource requests and
-    // limits. Additionally configures the kube-rbac-proxies to use the serving
-    // cert configured on the `Service` above.
+    // limits. Additionally, it configures the kube-rbac-proxies to use the serving
+    // cert configured on the `Service` above, and enables CRD metrics.
     //
     // The upstream kube-state-metrics Dockerfile defines a `VOLUME` directive
     // in `/tmp`. Although this is unused it will take some time for it to get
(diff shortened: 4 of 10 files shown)


rexagod:
The safe-guards in CMO's logic for handling KSM's CRS feature-set work around an issue in client-go. As it is right now:

- client-go [does not](https://github.com/kubernetes/kubernetes/issues/127329) support handling reflector errors,
- which was a [concern](https://github.com/openshift/cluster-monitoring-operator/pull/2078#issuecomment-1843256771) in the original patch as it not only unnecessarily alarmed users, but also,
- ended up incrementing [`kube_state_metrics_list_total{result="error"}`](https://github.com/kubernetes/kube-state-metrics/blob/9ea1495953853751b7affc0360e3f493d4318b55/pkg/watch/watch.go#L48),

owing to which, this was necessary.

machine424:
Thanks for the context Pranshu!
The way I understand it is that https://github.com/openshift/cluster-monitoring-operator/pull/2078#issuecomment-1843256771 only mentions cases where the CRs themselves are not complete; KSM will not appreciate some of the fields being missing.
But that doesn't cover the case where the CRD itself is missing, in that case I expect KSM to not even try to fetch the CRs (s they'll not be even present) and not complain about it. Isn't it the case?

rexagod:
So here's what I found.

There were some changes upstream between 4.18.13 and 4.19's latest nightly, due to which, errors such as the following, ceased to exist when the CRS config had definitions for a CRD that was not present **at startup**:
```
E1124 13:02:31.660297 1 reflector.go:148] k8s.io/kube-state-metrics/internal/store/builder.go:661: Failed to watch autoscaling.k8s.io/v1, Kind=VerticalPodAutoscaler: failed to list autoscaling.k8s.io/v1, Kind=VerticalPodAutoscaler: the server could not find the requested resource
```
By "upstream" above I mean `client-go` here, and not KSM itself, as I verified there were no such changes made in that duration (but we did bump all `k8s.io/...` deps from `v0.32.2` to `v0.32.3`). However, this "change" upstream only addresses the issue at startup, as we don't see any error logs or `{list,watch}_total{result="error"}` metric increments.

But this isn't the case when the CRD is deleted, in which case, both are exhibited:
```
E0513 03:10:15.260913   61388 reflector.go:166] "Unhandled Error" err="pkg/mod/k8s.io/client-go@v0.32.3/tools/cache/reflector.go:251: Failed to watch contoso.com/v1alpha1, Kind=MyPlatform: the server could not find the requested resource" logger="UnhandledError"
```
```
# HELP kube_state_metrics_watch_total Number of total resource watches in kube-state-metrics
# TYPE kube_state_metrics_watch_total counter
kube_state_metrics_watch_total{resource="*unstructured.Unstructured",result="error"} 1
kube_state_metrics_watch_total{resource="*unstructured.Unstructured",result="success"} 1
```

danwinship:
So what does that mean? You need to keep this code? If so, that means that ksm can only monitor _one_ kind of CRD, right? (At least until there are further client-go fixes.)

I can do the Gateway API metrics somewhere else if need be; it just seemed like it made the most sense to let ksm do them...

machine424:
Thanks for the details @rexagod.

I might have done something wrong during my tests then, I didn’t look too deeply into the code, but I expect https://github.com/kubernetes/kube-state-metrics/blob/8ec1fb3c15a902e17fef27af18d7079b1b593dca/internal/discovery/discovery.go#L194 to handle the updates (deletions in this case), don't you think this "watching the CRD" logic should/can live in KSM itself?
Maybe we get errors only until the next tick? apparently the updates only happen every 3s?

@danwinship, if we couldn't make changes upstream, is the Gateway CRD a resource that could be deleted once deployed/created?

rexagod:
I'm fairly positive https://github.com/kubernetes/kube-state-metrics/pull/2672 should enable us to merge this PR, and multi-resource support in CMO's CRS.

danwinship:
> is the Gateway CRD a resource that could be deleted once deployed/created?

Not in OCP; cluster-ingress-operator will [ensure it is always present](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/gateway-api-crd-life-cycle-management.md)

rexagod:
Hello, just wanted to update here that KSM v2.16 is GA upstream and we have a mirror PR downstream here: https://github.com/openshift/kube-state-metrics/pull/122. Once this is in, we can proceed with this PR.

rexagod:
https://github.com/openshift/kube-state-metrics/pull/122 is in, will review this once more.

This will also need a JIRA ticket, I'll open one on the monitoring end since this targets CMO.

rexagod:
/lgtm
/retitle [MON-4282](https://issues.redhat.com//browse/MON-4282): Multi-tenant support for KSM's CRS

Thank you, @danwinship! KSM's CRS in CMO can now safely support multiple CRs.

rexagod:
/jira refresh
/hold

Needs user-facing entry.

rexagod:
@danwinship This needs a CHANGELOG entry, like the one in the [VPA pull request](https://github.com/openshift/cluster-monitoring-operator/pull/2078/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR7).

If you're not around, I can create a new PR pulling in the commits from this one, or push to your fork (and this PR) directly if you can give me access to it, whatever works. 🙂

danwinship:
done

rexagod:
/test versions

Versions might have been updated in master since this failed.

danwinship:
@rexagod the changelog entry is OK? (And I got the release right? This is for 4.20 and there were apparently no new 4.19 features?)

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Note: This CHANGELOG is only for the monitoring team to track all monitoring related changes. Please see OpenShift release notes for official changes.
 
+## 4.20
+
+- [#2595](https://github.com/openshift/cluster-monitoring-operator/pull/2595) Allow generating additional CRD metrics beyond the VPA ones.
```
rexagod:
```suggestion
- [#2595](https://github.com/openshift/cluster-monitoring-operator/pull/2595) Multi-tenant support for KSM's CRS feature-set downstream.
```

(nit)

danwinship:
> I'm not quite sure what you meant by "no new 4.19 features" above

The last CHANGELOG entry before my new 4.20 entry is for 4.18, implying that there were no CHANGELOG-worthy changes in 4.19... or that I screwed up with branches somehow, which is why I was asking.

machine424:
/lgtm
We no longer rely on CHANGELOG.md.
For end users, we track the release notes entry on the tickets (which is not needed for this "internal" change).
thanks again @danwinship and @rexagod.

rexagod:
@machine424 I see, so the CHANGELOG is purely cosmetic and for book-keeping purposes now? I remember the doc-texts being used for release notes, but I thought we still maintain the CHANGELOG for the same reason. 🤔 

Also, thank you for the LGTM here!

PR #2579, opened 2025-03-13
WIP: OCPBUGS-18282: Reject reserved labels used as external labels

slashpai:
API documentation for `externalLabels` is like this

```
// The labels to add to any time series or alerts when communicating with
// external systems (federation, remote storage, Alertmanager).
// Labels defined by `spec.replicaExternalLabelName` and
// `spec.prometheusExternalLabelName` take precedence over this list.
```

We also [advise not to use reserved labels](https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/monitoring/configuring-core-platform-monitoring#attaching-additional-labels-to-your-time-series-and-alerts_configuring-alerts-and-notifications
) `prometheus` or `prometheus_replica` as externalLabels, because they are reserved and will be overwritten. It was not overwritten due to bug in PO which is fixed in [upstream](https://github.com/prometheus-operator/prometheus-operator/pull/5888/files ).

This change is for CMO to reject it beforehand and report to user if they use reserved labels as externalLabels

* [ ] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

PR #2604, opened 2025-05-21
OCPBUGS-18282: prevent use of reserved labels keys in Prometheus externalLabels

machine424:
implements https://github.com/openshift/cluster-monitoring-operator/pull/2579#issuecomment-2773042283

supersedes https://github.com/openshift/cluster-monitoring-operator/pull/2579

As this is done at the parsing level, the monitoringconfigmaps.openshift.io webhook would catch this early.


pkg/manifests/types.go
```diff
@@ -21,6 +21,7 @@ import (
 
 type CollectionProfile string
 type CollectionProfiles []CollectionProfile
+type ExternalLabels map[string]string
```
machine424, line 24: capitalized mainly to distinguish it from the filed name in the docs.

machine424: needs a doc entry like https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/monitoring/config-map-reference-for-the-cluster-monitoring-operator#tlsconfig

slashpai (commented):
/lgtm

PR #2602, opened 2025-05-15
NO-JIRA: metrics-client-ca: set owner annotation

vrutkovs:
* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

PR #2612, opened 2025-07-02
OCPBUGS-58427: pkg/manifests: don't reset annotations for metrics client CA

vrutkovs:
Previously this configmap didn't have annotations set, so these were being reset to an empty map. Now it has ownership annotation, so reset to an empty map is erasing them

Follow-up for https://github.com/openshift/cluster-monitoring-operator/pull/2602


vrutkovs:
/retitle OCPBUGS-58427: pkg/manifests: don't reset annotations for metrics client CA

PR #2618, opened 2025-07-11
OCPBUGS-58475: Enforce secure TLS settings in CMO

danielmellado:
This commit enforces secure TLS settings in order to eliminate insecure
cipher warnings in cluster-monitoring-operator logs. It changes
Alertmanager configurations used by Prometheus and Thanos.

- Set `min_version: TLS1.2` and explicitly define secure `cipher_suites`
- Applied to all additional Alertmanager configs
- Updated tests accordingly


marioferh (commented):
/lgtm

PR #2619, opened 2025-07-16
MON-4255: Bump prometheus-operator dependencies to v0.84.0

slashpai:


slashpai:
syncbot already updated po to v0.84.0 https://github.com/openshift/prometheus-operator/pull/333

PR #2621, opened 2025-07-21
MON-4296: Bump prometheus-operator libs to v0.84.0

slashpai:
Had to bump library-go and prometheus libs too fix errors


test/e2e/prometheus_test.go
```diff
@@ -20,7 +20,6 @@ import (
 	"testing"
 	"time"
 
-	_ "github.com/prometheus/prometheus/discovery/kubernetes" // required for promConfig.Load to parse kubernetes_sd_configs
```
machine424: could you tell us more?

pkg/manifests/manifests.go
```diff
@@ -2558,7 +2558,9 @@ func (f *Factory) NewService(manifest []byte) (*v1.Service, error) {
 	return &s, nil
 }
 
+//nolint:staticcheck // MON-4216 to track migration from Endpoints to EndpointSlice
```
machine424: out of curiosity, what was staticcheck complaining about? the method not being used?

slashpai: Depreciation warning for endpoints

machine424: I see, nit: maybe we can use `//nolint:staticcheck // SA1019` with a comment.

Or we can just remove that, I don't think we're using it anywhere, we can add it later if needed.

test/e2e/prometheus_test.go
slashpai: fixed, that should not have removed

PR #2623, opened 2025-07-22
MON-4318: Add debug image manifest to CMO

danielmellado:
This commit adds a new optional image to CMO manifest for debugging purposes,
it's based on ubi9, adding a workaround to enable smtp in libcurl, which is
needed by support KCS'


danielmellado:
/retitle [MON-4318](https://issues.redhat.com//browse/MON-4318): Add debug image manifest to CMO

PR #2624, opened 2025-07-22
OCPBUGS-57215: operator: increase wait time till degraded to max 4 times 5m

jan--f:
Fixes: https://issues.redhat.com/browse/OCPBUGS-57215


machine424:
thanks, given that it also applies to the `available` changes, let's explicitly state that in the title.
/lgmt

PR #1961, opened 2023-05-10
OCPBUGS-1626: update jsonnet dependencies

simonpasquier:
* [ ] I added CHANGELOG entry for this change.
* [X] No user facing changes, so no entry in CHANGELOG was needed.

This PR builds on top of #1915 which is the pull request generated by `openshift-monitoring-bot`. In addition to the normal jsonnet bumps, I made changes to avoid shipping the ScrapeConfig and PrometheusAgent artefacts since they are not supported within CMO.

PR #2626, opened 2025-07-28
MON-4299): chore(thanos): replace CLI args that will disappear in future versions

machine424:
--endpoint itself is deprecated now https://github.com/thanos-io/thanos/pull/7890/files but we can live with it for a moment

needed for https://github.com/openshift/thanos/pull/163


danielmellado:
Can we tag this along with a deprecation message/release note?

machine424:
> Can we tag this along with a deprecation message/release note?

users don't have to worry about that/access to that.

danielmellado:
> users don't have to worry about that/access to that.

fair point, but I'd still make a release note just in case and for the future us

machine424:
> > users don't have to worry about that/access to that.
> 
> fair point, but I'd still make a release note just in case and for the future us

https://github.com/openshift/cluster-monitoring-operator/blob/main/CHANGELOG.md is not longer used, it seems.
It's also a source of conflicts during backports.
For such a simple change, I'd just "git blame" to be honest.
The commit and the PR description give enough details.

danielmellado:
> https://github.com/openshift/cluster-monitoring-operator/blob/main/CHANGELOG.md is not longer used, it seems.

was about to suggest that but also noticed it being outdated, sigh. Let's get this merged and we'll discuss the changelog being updated further. Thanks Ayoub!

machine424:
For the thanos-querier, --target and --store are leftovers from https://github.com/openshift/cluster-monitoring-operator/pull/1961/files

one --endpoint is enough.

machine424:
@danielmellado if I can get another lgtm, thanks.

simonpasquier (commented):
/lgtm

PR #2628, opened 2025-07-29
OCPBUGS-58475: Enforce secure TLS settings in CMO server

danielmellado:
This PR completes the fix for insecure cipher warnings in
cluster-monitoring-operator logs by configuring the CMO's own
HTTPS server to use secure TLS settings.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -767,7 +767,7 @@ func (o *Operator) sync(ctx context.Context, key string) error {
 	var proxyConfig = getProxyReader(ctx, config, o.loadProxyConfig)
 
 	var apiServerConfig *manifests.APIServerConfig
-	apiServerConfig, err = o.loadApiServerConfig(ctx)
+	apiServerConfig, err = o.LoadApiServerConfig(ctx)
 
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/operator/operator.go
```diff
@@ -941,6 +941,12 @@ func (o *Operator) loadApiServerConfig(ctx context.Context) (*manifests.APIServe
 	return o.lastKnownApiServerConfig, nil
 }
 
+// GetAPIServerConfig returns the last known APIServerConfig.
+// This method should only be called after the operator has been initialized and run at least once.
+func (o *Operator) GetAPIServerConfig() *manifests.APIServerConfig {
+	return o.lastKnownApiServerConfig
```
jan--f: Should this not simply call `loadApiServerConfig` (or we call this directly from the `main` package)? ~From where this is called once on startup I think there isn't an APIServerConfig loaded yet. So does this actually set anything but the defaults?~
There is a chance the cache isn't populated before this is run.

danielmellado: Thanks for the catch, I was making this a bit overly complex, as I'm testing some assistants within this PR xD

PR #2632, opened 2025-07-30
Depppp

machine424:


machine424:
Prometheus Kube SD is still not compatible with k8s.io/client-go 1.33 
```
 # github.com/prometheus/prometheus/discovery/kubernetes
vendor/github.com/prometheus/prometheus/discovery/kubernetes/kubernetes.go:862:36: not enough arguments in call to cache.DefaultWatchErrorHandler
	have (*"k8s.io/client-go/tools/cache".Reflector, error)
	want (context.Context, *"k8s.io/client-go/tools/cache".Reflector, error) 
```
Will need to fix that upstream

juzhao:
see from https://github.com/openshift/cluster-monitoring-operator/pull/2656#discussion_r2358135425
for the ApplyNetworkPolicy func is not in 4.20 CMO library-go, will follow your advice
> If the openshift/library-go update (better done in a separate PR) will involve a lot of work and will take time, especially if it requires to bump the kube deps (I know that the prom deps would break on 1.33 e.g. [#2632 (comment)](https://github.com/openshift/cluster-monitoring-operator/pull/2632#issuecomment-3149691979)), you can just put the content of https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/networking.go#L19 in a file under https://github.com/openshift/cluster-monitoring-operator/tree/main/pkg/tasks and use it in the tasks.

machine424:
/close
superseded by other PRs

PR #2640, opened 2025-08-07
NO-JIRA: Use label matchers for Rules API in prom-label-proxy

simonpasquier:
There's no user visible change but it should bring performance and resource usage improvements in setups with large number of rules since the data is filtered at the source (Prometheus and Thanos) rather than by the proxy.

### DIFF
--- a/jsonnet/components/thanos-querier.libsonnet
+++ b/jsonnet/components/thanos-querier.libsonnet
@@ -513,6 +513,7 @@ function(params)
                   '--enable-label-apis',
                   '--error-on-replace',
                   '--rules-with-active-alerts',
+                  '--enable-label-matchers-for-rules-api',
                 ],
                 resources: {
                   requests: {
--- a/assets/thanos-querier/deployment.yaml
+++ b/assets/thanos-querier/deployment.yaml
@@ -167,6 +167,7 @@ spec:
         - --enable-label-apis
         - --error-on-replace
         - --rules-with-active-alerts
+        - --enable-label-matchers-for-rules-api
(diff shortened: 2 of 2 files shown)


jsonnet/components/thanos-querier.libsonnet
```diff
@@ -513,6 +513,7 @@ function(params)
                   '--enable-label-apis',
                   '--error-on-replace',
                   '--rules-with-active-alerts',
+                  '--enable-label-matchers-for-rules-api',
```
machine424, line 516: prometheus does expose the version in through /api/v1/status/buildinfo
thanos probably has a similar one.
Maybe we can make the proxy aware of that instead of new flags, to ensure compatibility.

machine424:
/lgtm
let's give it a try!
for other reviewers, Simon made the change here https://github.com/prometheus-community/prom-label-proxy/pull/295/files

jsonnet/components/thanos-querier.libsonnet
simonpasquier: I prefer an explicit opt-in: checking the reported version assumes that you use the upstream binaries or rely on the same build workflow.
In case the rules endpoint doesn't support label matchers, the outcome is quite bad since you get access to all the rules...

simonpasquier: I agree that the proxy discovering build information, logging it and providing useful advices (e.g. "Prometheus endpoint seems to support label matchers, maybe you want to try this flag?") would help troubleshooting and user experience.

PR #2641, opened 2025-08-08
OCPBUGS-60221: remove `managed_cluster` from reserved external labels

slashpai:


machine424:
/lgtm
thanks

PR #2643, opened 2025-08-13
[wip] MON-4321: Send telemetry via remote-write

jan--f:
We have a few things still to discuss. I have outlined this in https://issues.redhat.com/browse/MON-3875. To quote:

1. Will we reuse the [existing config fields](https://github.com/openshift/cluster-monitoring-operator/blob/2758b32dbbc49b73304046c80f19b501b24a5d51/pkg/manifests/types.go#L284) or introduce new ones and deprecate the old fields.
2. Will we keep generating the recording rule and remote write config from the current match list or use another mechanism of defining the cluster side telemetry allow list? We could for example offload the allow listing fully to the server side. Then we can add new telemetry metrics cluster side by simply adding the telemetry: prefix to a recording rule name.
3. Should we keep allow-listing telemetry metrics cluster side? The telemeter-client implementation explicitly matches the configured telemetry metrics. Cluster admin can't accidentally add telemetry metrics (though they still can intentionally, we have server side allowlisting to catch this). The cluster side allow listing makes for a burdensome telemetry addition process.


pkg/tasks/prometheus.go
```diff
@@ -346,16 +346,11 @@ func (t *PrometheusTask) create(ctx context.Context) error {
 		return fmt.Errorf("initializing Prometheus telemetry secret failed: %w", err)
 	}
 
-	if t.config.ClusterMonitoringConfiguration.TelemeterClientConfig.IsEnabled() && t.config.RemoteWrite {
+	if t.config.ClusterMonitoringConfiguration.TelemeterClientConfig.IsEnabled() {
```
juzhao: I see TelemeterClient deployment is removed, how to check TelemeterClientConfig is enabled

pkg/tasks/telemeter.go
simonpasquier, line 1: Don't we need to keep the telemeter task for at least the next minor release to remove telemeter resources?

pkg/manifests/manifests.go
```diff
@@ -1415,7 +1398,33 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret)
...
+				{
+					SourceLabels: []monv1.LabelName{"__name__"},
+					Regex:        "telemetry:.*",
+					Action:       "keep",
+				},
+				// To support a regex matcher we track the
+				// original metric name in the recording rule.
+				// Here we reinstate the original name and drop
+				// the temp name.
+				// See also jsonnet/components/telemetry-recording-rules.libsonnet
```
simonpasquier, line 1425: I was a bit confused by the `name_label` label at first. After reviewing more deeply, I understand why it's needed.

I have an alternative implementation in mind (though I haven't explored it fully): what if the telemetry metrics were all recorded with the same metric name (e.g. `telemetry:metric`) with the original name persisted in another label (e.g. `__telemetry_name__`)? It would avoid the special case with `label_name` and not be different in terms of cardinality.

simonpasquier:
Side comment: we could also "optimize" some of the legacy metrics which are transmitted from the raw source without any recording rule to aggregate away the "noisy" labels like pod or instance (e.g. `cluster_operator_conditions`).

jan--f:
> Side comment: we could also "optimize" some of the legacy metrics which are transmitted from the raw source without any recording rule to aggregate away the "noisy" labels like pod or instance (e.g. `cluster_operator_conditions`).

How about adding aggregating away those labels regardless?

pkg/manifests/manifests.go
jan--f: I opted for using the `original_metric_name` labe for now. While I know we use labels with the `__` prefix in other places  (`__id`), the [prometheus docs](https://prometheus.io/docs/concepts/data_model/) say
> Label names beginning with __ (two underscores) MUST be reserved for internal Prometheus use.

So not sure we should extend that practice. But overall no strong feeling about the label name.

jan--f:
I opted to introduce a new config field called `TelemetryConfig` and deprecate `TelemeterClientConfig`. I believe this is clearer then reusing some parts of `TelemeterClientConfig` and deprecating others.

jan--f:
> > Side comment: we could also "optimize" some of the legacy metrics which are transmitted from the raw source without any recording rule to aggregate away the "noisy" labels like pod or instance (e.g. `cluster_operator_conditions`).
> 
> How about adding aggregating away those labels regardless?

Perhaps just `sum without(pod, container, ...) ()`?

pkg/manifests/manifests.go
simonpasquier: Personally I like that the __ prefix indicates a label which doesn't come from the target but rather from relabeling.

simonpasquier:
> > > Side comment: we could also "optimize" some of the legacy metrics which are transmitted from the raw source without any recording rule to aggregate away the "noisy" labels like pod or instance (e.g. `cluster_operator_conditions`).
> > 
> > 
> > How about adding aggregating away those labels regardless?
> 
> Perhaps just `sum without(pod, container, ...) ()`?

I assume that in most cases, we'd want `max without(pod, container) (...)`. But I'd be worried to do it unilaterally as it may break some hidden contract.

simonpasquier:
@jan--f for awareness, Service Delivery uses this config for telemeter client:
https://github.com/openshift/managed-cluster-config/blob/3d39d600714aeae1743ba831f0ff3fe5a30e519b/resources/cluster-monitoring-config/config.yaml#L60-L67

jan--f:
> I assume that in most cases, we'd want `max without(pod, container) (...)`. But I'd be worried to do it unilaterally as it may break some hidden contract.

Fair point. Perhaps we better leave this for a follow up PR.

pkg/manifests/manifests.go
jan--f: added the __ prefix.

PR #2645, opened 2025-08-15
[wip] MON-4297 add networkpolicy for CMO/KSM/metrics-server/monitoring-plugin/OSM/prometheus-operator/telemeter-client

juzhao:
task: https://issues.redhat.com/browse/MON-4297
this PR added
1. default-deny-networkpolicy for all traffic, file: default-deny-networkpolicy.yaml
2. cluster-monitoring-operator networkpolicy, file: cluster-monitoring-operator-networkpolicy.yaml
3.  KSM networkpolicy, file: kube-state-metrics-networkpolicy.yaml
4.  metrics-server networkpolicy, file: metrics-server-networkpolicy.yaml
5. monitoring-plugin networkpolicy, file: monitoring-plugin-networkpolicy.yaml
6. OSM networkpolicy, file: openshift-state-metrics-networkpolicy.yaml
7. prometheus-operator networkpolicy, file: prometheus-operator-networkpolicy.yaml
8. prometheus-operator-admission-webhook networkpolicy, file: prometheus-operator-admission-webhook-networkpolicy.yaml
9. telemeter-client networkpolicy, file: telemeter-client-networkpolicy.yaml


juzhao:
/retitle [wip] [MON-4297](https://issues.redhat.com//browse/MON-4297) add networkpolicy for CMO/KSM/metrics-server/monitoring-plugin/OSM/prometheus-operator/telemeter-client

assets/telemeter-client/telemeter-client-networkpolicy.yaml
```diff
@@ -0,0 +1,24 @@
...
+  policyTypes:
+  - Ingress
+  - Egress
+  ingress:
+  # Allow access to ports 8443
+  - ports:
+    - port: 8443
+      protocol: TCP
+  egress:
+  - {}
```
jan--f, line 24: Telemeter client needs egress as well to send telemetry.

juzhao:
```
  egress:
  - {}
```
no ports/to settings, means no limit for outgoing traffic
```
$ oc explain netpol.spec.egress.
FIELDS:
  ports	<[]NetworkPolicyPort>
    ports is a list of destination ports for outgoing traffic. Each item in this
    list is combined using a logical OR. If this field is empty or missing, this
    rule matches all ports (traffic not restricted by port). If this field is
    present and contains at least one item, then this rule allows traffic only
    if the traffic matches at least one port in the list.

  to	<[]NetworkPolicyPeer>
    to is a list of destinations for outgoing traffic of pods selected for this
    rule. Items in this list are combined using a logical OR operation. If this
    field is empty or missing, this rule matches all destinations (traffic not
    restricted by destination). If this field is present and contains at least
    one item, this rule allows traffic only if the traffic matches at least one
    item in the to list.
```
if set
```
egress: []
```
this will block all outgoing traffic

jan--f:
We need a policy for the different Prometheus' too don't we? What if a user configures remote write?

juzhao:
> We need a policy for the different Prometheus' too don't we? What if a user configures remote write?

this PR does not include networkpolicy for Prometheus(will do in another PR), and I think
```
  egress:
  - {}
```
is enough for prometheus egress, since the remote write address is not a static ip, and above setting allows any ip/port and will use less cpu/memory

jan--f:
> > We need a policy for the different Prometheus' too don't we? What if a user configures remote write?
> 
> this PR does not include networkpolicy for Prometheus(will do in another PR), and I think
> 
> ```
>   egress:
>   - {}
> ```
> 
> is enough for prometheus egress, since the remote write address is not a static ip, and above setting allows any ip/port and will use less cpu/memory

Gotcha, thanks. We'll still need to generate the policies to get the `generate` target to pass.

simonpasquier:
Late to the party but we could reuse some bits from https://github.com/prometheus-operator/kube-prometheus which ships default network policies.
Right now our jsonnet removes all of them https://github.com/openshift/cluster-monitoring-operator/blob/3ebd14abcc91148afb6919d53ed223774ff93169/jsonnet/main.jsonnet#L516

juzhao:
/close
close it in favor of https://github.com/openshift/cluster-monitoring-operator/pull/2656 which include networkpolicy settings for alertmanager/prometheus/thanos-querier

PR #2647, opened 2025-08-18
OCPBUGS-58475: Enforce secure TLS settings in CMO server

danielmellado:
Set ocp intermediate tls settings to cmo server. This avoids throwing
unnecesary warning messages about TLS ciphers.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -24,6 +24,7 @@ import (
 	"github.com/openshift/library-go/pkg/authorization/hardcodedauthorizer"
 	"github.com/openshift/library-go/pkg/config/configdefaults"
 	"github.com/openshift/library-go/pkg/config/serving"
+	"github.com/openshift/library-go/pkg/crypto"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	"k8s.io/apiserver/pkg/authorization/union"
 	genericapiserver "k8s.io/apiserver/pkg/server"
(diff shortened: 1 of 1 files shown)


danielmellado:
The HyperShift test failure is an AWS infrastructure issue, not related to the TLS cipher changes.

02:38:16Z: AWS infrastructure (VPC, subnets, IAM roles) created successfully
02:39:20Z: AWS security group creation failed with "Waiting for AWS default security group to be created for hosted cluster"
02:40:08Z: Without security groups, no EC2 instances could be launched, leaving zero worker nodes
02:40:08Z - 03:23:46Z: CMO deployment failed 24 times with "the cluster does not have any nodes" - pods can't be scheduled without nodes

pkg/server/server.go
```diff
@@ -73,6 +74,10 @@ func (s *Server) Run(ctx context.Context, collectionProfilesEnabled bool) error
 	// Don't set a CA file for client certificates because the CA is read from
 	// the kube-system/extension-apiserver-authentication ConfigMap.
 	servingInfo.ServingInfo.ClientCA = ""
+	// Use intermediate TLS profile cipher suites to avoid insecure cipher warnings
+	// Convert OpenSSL cipher names to IANA names for Kubernetes validation
+	openSSLCiphers := configv1.TLSProfiles[configv1.TLSProfileIntermediateType].Ciphers
+	servingInfo.ServingInfo.CipherSuites = crypto.OpenSSLToIANACipherSuites(openSSLCiphers)
```
jan--f: Looking at https://github.com/openshift/library-go/blob/378de074fe7b73a4bc593ff807693c466889d3d5/pkg/config/configdefaults/config_default.go#L45-L56

I wonder if we should also use `CipherSuitesToNamesOrDie`? Or rather what is the difference to `OpenSSLToIANACipherSuites`.

Should we also set `MinTLSVersion`? We have some code in `pkg/manifests.manifests.go` already.

danielmellado: Here, as this is implemented, we'd need OpenSSLToIANACipherSuites because configv1.TLSProfiles contains OpenSSL format names, not Go constants, even though it's similar. We could convert that first, but it'd be adding more code IMHO.

I'll also add a MinTLSVersion, thanks!

PR #2648, opened 2025-08-18
MON-4343: Cleanup after 4.19

jan--f:

### DIFF
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -915,12 +915,14 @@ k8sPrometheusAdapter:
     profile: Request`
 	f.MustCreateOrUpdateConfigMap(t, f.BuildCMOConfigMap(t, data))
 	checkMetricValue(1)
+	f.AssertOperatorCondition(configv1.OperatorUpgradeable, configv1.ConditionFalse)(t)
 
 	// The metric should be reset to 0.
 	data = `
 k8sPrometheusAdapter:`
 	f.MustCreateOrUpdateConfigMap(t, f.BuildCMOConfigMap(t, data))
 	checkMetricValue(0)
+	f.AssertOperatorCondition(configv1.OperatorUpgradeable, configv1.ConditionTrue)(t)
 }
 
 // checks that the toleration is present


pkg/manifests/types.go
```diff
@@ -44,8 +44,6 @@ type ClusterMonitoringConfiguration struct {
 	UserWorkload *UserWorkloadConfig `json:"userWorkload,omitempty"`
 	// OmitFromDoc
 	HTTPConfig *HTTPConfig `json:"http,omitempty"`
-	// OmitFromDoc: `K8sPrometheusAdapter` defines settings for the Prometheus Adapter component.
```
machine424: I'm afraid that if we remove this, existing configs that still set it will start failing (especially that we have strict validation now), I think we should have a 4.20.x (or even a 4.19.x) that makes the operator upgreadable=false if prom-adapter's config is still set, to make sure the configs are removed, make sure all clusters go through that version before the upgrade, then we can remove its definition in the next version.

jan--f: isn't that (not upgradable) already the case? The `cluster_monitoring_operator_deprecated_config_in_use` triggers the `ClusterMonitoringOperatorDeprecatedConfig` alert. I thought clusters with firing alerts aren't upgradable...?

jan--f: ahh hmm maybe not for a `severety: info` alert :thinking:

machine424: yes, we usually rely on the operator's status.upgradeable for that
(we used that recently here https://github.com/openshift/cluster-monitoring-operator/pull/2591/files)

pkg/configvalidate/configvalidate.go
```diff
@@ -36,14 +36,14 @@ type parseConfig func(c *corev1.ConfigMap) error
 
 func configParser(collectionProfilesEnabled bool) parseConfig {
 	return func(c *corev1.ConfigMap) error {
-		_, err := manifests.NewConfigFromConfigMap(c, collectionProfilesEnabled)
+		_, _, err := manifests.NewConfigFromConfigMap(c, collectionProfilesEnabled)
```
machine424, line 39: in a follow up we can forward warnings as part of the api response via https://github.com/kubernetes-sigs/controller-runtime/blob/0f4e99e54a718b4b17452ef2925eb606fffa073b/pkg/webhook/admission/response.go#L121 so users are also informed/warned when editing the configmaps.

pkg/configvalidate/configvalidate_test.go
```diff
@@ -88,7 +88,7 @@ func TestHandle(t *testing.T) {
 				},
 			},
 			allowed: false,
-			message: "failed to parse data at key \"config.yaml\": error unmarshaling: unknown field \"prometheus_operator\"",
+			message: "failed to parse data at key \"config.yaml\": error unmarshaling JSON: while decoding JSON: json: unknown field \"prometheus_operator\"",
```
machine424, line 91: this looks like a regression, we already got rid of the repetitive `JSON: while decoding JSON: json`

pkg/manifests/config.go
```diff
@@ -358,19 +338,27 @@ func UnmarshalStrict(data []byte, v interface{}) error {
...
+	var warning *InvalidConfigWarning
+	wCmc := defaultClusterMonitoringConfiguration()
+	wErr := UnmarshalStrict(content, &wCmc)
+	if wErr != nil {
+		warning = &InvalidConfigWarning{ConfigMap: "openshift-monitoring/cluster-monitoring-config", Err: wErr}
+	}
+
 	cmc := defaultClusterMonitoringConfiguration()
-	err := UnmarshalStrict(content, &cmc)
+	err := k8syaml.UnmarshalStrict(content, &cmc)
```
machine424, line 352: I see now.
I don't think we want to reintroduce `k8syaml.UnmarshalStrict`.
Let's just use the existing `err := UnmarshalStrict(content, &cmc)`, given that the prom-adapter config is still defined, the unmarshalling should pass.
After that, in a `func (c *Config) Precheck()` or somewhere else after unmarshalling, we can emit a warning if the field is set (in a follow-up PR as you're planning to do).

same for `NewUserConfigFromString`.

test/e2e/config_test.go
```diff
@@ -71,16 +71,33 @@ func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
 	t.Log("asserting that CMO goes degraded after an invalid configuration is pushed")
 	f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionTrue)(t)
 	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionFalse)(t)
+	// Even when an invalid configuration is caught by both unmarshallers, the operator is still set
```
machine424, line 74: I think all the new checks below aren't needed.
let's add some in the PR that will apply the warning logic to prom-adapter.

pkg/manifests/config.go
```diff
@@ -202,31 +213,6 @@ func (u *UserWorkloadConfiguration) checkThanosRulerEvaluationInterval() error {
 	return nil
 }
 
-func (u *UserWorkloadConfiguration) checkAlertmanagerVersion() error {
```
machine424, line 205: if we could have "warning propagation" and "4.19 alertmanager code cleaning" in two separate it'd be great.
I can help with babysitting.

PR #2649, opened 2025-08-25
NO-JIRA: chore: bump jsonnet dependencies

simonpasquier:
This change updates jsonnet dependencies to their latest versions. In particular, it freezes the node_exporter and Prometheus dashboards to the latest versions supporting Grafana 11. From now on, these dashboards will no longer receive updates from the upstream mixins (until we migrate to the upstream Perses dashboards).

_The PR can wait after the 4.20 branch cut_


assets/node-exporter/daemonset.yaml
```diff
@@ -32,9 +32,10 @@ spec:
       automountServiceAccountToken: true
       containers:
       - args:
-        - --web.listen-address=127.0.0.1:9100
```
machine424, line 35: I agree, using the same port on different interfaces can confusing, plus a pod has plenty of them

assets/node-exporter/prometheus-rule.yaml
```diff
@@ -278,6 +278,15 @@ spec:
       for: 15m
       labels:
         severity: warning
+    - alert: NodeSystemdServiceCrashlooping
```
machine424: we'd need a ticket to understand and add a runbook for this one.

simonpasquier:
/hold cancel

removing the hold since we're in ` acknowledge-critical-fixes-only ` phase anyway.

assets/node-exporter/prometheus-rule.yaml
simonpasquier: I've opted for removing the alerting rule: the systemd collector isn't enabled by default + when it is, users can define their own AlertingRule custom resource (they probably do already).

simonpasquier:
/assign @juzhao 

Can you please verify that there's no regression with the console dashboards? The main change introduced by the PR is that we hardcode all dashboard definitions because the upstream dashboards are only compatible with Grafana 11+ format which isn't supported by the OCP console.

juzhao:
tested with PR, no regression bug, found one existing bug for 4.19+ https://issues.redhat.com/browse/OCPBUGS-62063, the bug is not caused by this PR
/verified by @juzhao

PR #2651, opened 2025-08-26
MON-4343: Cleanup deprecate pa config

jan--f:

### DIFF
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -64,6 +64,8 @@ func (e *InvalidConfigWarning) Warning() string {
 	return fmt.Sprintf("configuration in the %q ConfigMap is invalid and should be fixed: %s", e.ConfigMap, e.Err)
 }
 
+var errPrometheusAdapterDeprecated = errors.New("k8sPrometheusAdapter is deprecated and usage should be removed, use metricsServer instead")
+
 type Config struct {
 	Images                               *Images `json:"-"`
 	RemoteWrite                          bool    `json:"-"`
@@ -584,9 +586,9 @@ func (c *Config) LoadEnforcedBodySizeLimit(pcr PodCapacityReader, ctx context.Co
 	return nil
 }
 
-func (c *Config) Precheck() error {
+func (c *Config) Precheck() (*InvalidConfigWarning, error) {
 	if c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile != FullCollectionProfile && !c.CollectionProfilesFeatureGateEnabled {
-		return fmt.Errorf("%w: collectionProfiles is currently a TechPreview feature behind the \"MetricsCollectionProfiles\" feature-gate, to be able to use a profile different from the default (\"full\") please enable it first", ErrConfigValidation)
+		return nil, fmt.Errorf("%w: collectionProfiles is currently a TechPreview feature behind the \"MetricsCollectionProfiles\" feature-gate, to be able to use a profile different from the default (\"full\") please enable it first", ErrConfigValidation)
 	}
 
 	// Validate the configured collection profile iff tech preview is enabled, even if the default profile is set.
@@ -599,20 +601,22 @@ func (c *Config) Precheck() error {
 			metrics.CollectionProfile.WithLabelValues(string(profile)).Set(v)
 		}
 		if !slices.Contains(SupportedCollectionProfiles, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile) {
-			return fmt.Errorf(`%q is not supported, supported collection profiles are: %q: %w`, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile, SupportedCollectionProfiles.String(), ErrConfigValidation)
+			return nil, fmt.Errorf(`%q is not supported, supported collection profiles are: %q: %w`, c.ClusterMonitoringConfiguration.PrometheusK8sConfig.CollectionProfile, SupportedCollectionProfiles.String(), ErrConfigValidation)
 		}
 	}
 
 	// Highlight deprecated config fields.
 	var d float64
+	var warn *InvalidConfigWarning
 	if c.ClusterMonitoringConfiguration.K8sPrometheusAdapter != nil {
 		klog.Infof("k8sPrometheusAdapter is a deprecated config use metricsServer instead")
 		d = 1
+		warn = &InvalidConfigWarning{ConfigMap: "openshift-monitoring/cluster-monitoring-config", Err: errPrometheusAdapterDeprecated}
 	}
 	// Prometheus-Adapter is replaced with Metrics Server by default from 4.16
 	metrics.DeprecatedConfig.WithLabelValues("openshift-monitoring/cluster-monitoring-config", "k8sPrometheusAdapter", "4.16").Set(d)
 
-	return nil
+	return warn, nil
 }
 
 func calculateBodySizeLimit(podCapacity int) string {
--- a/pkg/manifests/config.go
+++ b/pkg/manifests/config.go
@@ -55,6 +55,15 @@ const (
 
 var reservedPrometheusExternalLabels = []string{"prometheus", "prometheus_replica", "cluster"}
 
+type InvalidConfigWarning struct {
+	ConfigMap string
+	Err       error
+}
+
+func (e *InvalidConfigWarning) Warning() string {
+	return fmt.Sprintf("configuration in the %q ConfigMap is invalid and should be fixed: %s", e.ConfigMap, e.Err)
+}
+
 type Config struct {
 	Images                               *Images `json:"-"`
 	RemoteWrite                          bool    `json:"-"`
@@ -328,6 +337,7 @@ func UnmarshalStrict(data []byte, v interface{}) error {
 
 func newConfig(content []byte, collectionProfilesFeatureGateEnabled bool) (*Config, error) {
 	c := Config{CollectionProfilesFeatureGateEnabled: collectionProfilesFeatureGateEnabled}
+
 	cmc := defaultClusterMonitoringConfiguration()
 	err := UnmarshalStrict(content, &cmc)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


jan--f:
The final bit of #2648

pkg/manifests/config.go
```diff
@@ -358,19 +369,27 @@ func UnmarshalStrict(data []byte, v interface{}) error {
 	}
 }
 
-func newConfig(content []byte, collectionProfilesFeatureGateEnabled bool) (*Config, error) {
+func newConfig(content []byte, collectionProfilesFeatureGateEnabled bool) (*Config, *InvalidConfigWarning, error) {
 	c := Config{CollectionProfilesFeatureGateEnabled: collectionProfilesFeatureGateEnabled}
+
+	var warning *InvalidConfigWarning
+	wCmc := defaultClusterMonitoringConfiguration()
+	wErr := UnmarshalStrict(content, &wCmc)
```
machine424: no need to propagate warnings for this.
we already capture the error below in `err := UnmarshalStrict(content, &cmc)`

pkg/manifests/config.go
```diff
@@ -720,38 +741,49 @@ func (u *UserWorkloadConfiguration) applyDefaults() {
...
-func NewUserConfigFromString(content string) (*UserWorkloadConfiguration, error) {
+func NewUserConfigFromString(content string) (*UserWorkloadConfiguration, *InvalidConfigWarning, error) {
 	if content == "" {
-		return NewDefaultUserWorkloadMonitoringConfig(), nil
+		return NewDefaultUserWorkloadMonitoringConfig(), nil, nil
 	}
+
+	var warning *InvalidConfigWarning
+	wU := &UserWorkloadConfiguration{}
+	wErr := UnmarshalStrict([]byte(content), &wU)
```
machine424: same here, having `err := UnmarshalStrict([]byte(content), &u)` below is enough, no need to have it as a warning as well.

pkg/manifests/config_test.go
```diff
@@ -265,7 +296,12 @@ thanosRuler:
 
 	for _, tc := range tcs {
 		t.Run(tc.name, func(t *testing.T) {
-			c, err := NewUserConfigFromString(tc.configString())
+			c, warning, err := NewUserConfigFromString(tc.configString())
```
machine424: I think that warnings in all tests that don't have a promadapter field set should nil

test/e2e/config_test.go
```diff
@@ -71,16 +71,33 @@ func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
...
 	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionFalse)(t)
+	// Even when an invalid configuration is caught by both unmarshallers, the operator is still set
+	// to Upgradeable=false, this keeps the logic simple and encourage users to fix their configs.
+	f.AssertOperatorCondition(configv1.OperatorUpgradeable, configv1.ConditionFalse)(t)
 	f.AssertOperatorConditionReason(configv1.OperatorDegraded, "InvalidConfiguration")
 	f.AssertOperatorConditionReason(configv1.OperatorAvailable, "InvalidConfiguration")
 	// Check that the previous setup hasn't been reverted
 	f.AssertStatefulsetExists("prometheus-user-workload", f.UserWorkloadMonitoringNs)(t)
 
+	t.Log("invalid configuration with field of the wrong case")
```
machine424: no need have those as we don't want to propagate the same unmarshalling errors as warnings, see above.

test/e2e/config_test.go
```diff
@@ -71,16 +71,33 @@ func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
...
+	f.AssertOperatorCondition(configv1.OperatorUpgradeable, configv1.ConditionFalse)(t)
+	f.AssertOperatorConditionReason(configv1.OperatorUpgradeable, "InvalidConfiguration")(t)
+	f.AssertOperatorConditionMessageContains(configv1.OperatorUpgradeable, `configuration in the "openshift-monitoring/cluster-monitoring-config" ConfigMap is invalid and should be fixed`)(t)
+
 	// Restore the first configuration.
 	f.MustCreateOrUpdateConfigMap(t, getUserWorkloadEnabledConfigMap(t, f))
 	t.Log("asserting that CMO goes back healthy after the configuration is fixed")
 	f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionFalse)(t)
 	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionTrue)(t)
+	// Once the config is adjusted, the operator becomes Upgradeable.
```
machine424: we could keep this and replace it with a config with promAdapter set; check it goes upgreadable=false when set and true when removed.

machine424: I see you added those checks in TestClusterMonitoringDeprecatedConfig

test/e2e/user_workload_monitoring_test.go
```diff
@@ -76,10 +75,62 @@ func TestUserWorkloadMonitoringInvalidConfig(t *testing.T) {
 	f.MustCreateOrUpdateConfigMap(t, cm)
 	defer f.MustDeleteConfigMap(t, cm)
 
+	t.Log("invalid UWM configuration with malformed YAML/JSON")
 	f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionTrue)(t)
 	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionFalse)(t)
 	f.AssertOperatorConditionReason(configv1.OperatorDegraded, "UserWorkloadInvalidConfiguration")
 	f.AssertOperatorConditionReason(configv1.OperatorAvailable, "UserWorkloadInvalidConfiguration")
+	// Even when an invalid configuration is caught by both unmarshallers, the operator is still set
```
machine424: no longer needed, we just have one unmarshaller now and it only yields errors, no need to have warnings for it. see above.

machine424 (commented):
thanks, some suggestions

pkg/manifests/config.go
jan--f: hmm I see, I suppose I didn't understand https://github.com/openshift/cluster-monitoring-operator/pull/2592 fully.

machine424: in https://github.com/openshift/cluster-monitoring-operator/pull/2592, we had 2 different unmarshallers, we propagated the failures of the newer/stricter one as warnings only to block upgrades so the configs are fixed.

machine424: now the newer/stricter one is the default/only one and we can (should) propagate its failures as errors.

pkg/manifests/config_test.go
```diff
@@ -791,7 +798,7 @@ func TestCollectionProfilePreCheck(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			c, err := NewConfigFromString(tc.config, true)
 			require.NoError(t, err)
-			err = c.Precheck()
+			_, err = c.Precheck()
```
machine424, line 722: nit: we can set the warning and ensure it's nil, an easy extra check is always welcome.

test/e2e/config_test.go
```diff
@@ -71,6 +71,9 @@ func TestClusterMonitoringOperatorConfiguration(t *testing.T) {
 	t.Log("asserting that CMO goes degraded after an invalid configuration is pushed")
 	f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionTrue)(t)
 	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionFalse)(t)
+	// Even when an invalid configuration is caught by both unmarshallers, the operator is still set
```
machine424: leftover? also I don't think we do/need to make Upgradeable=false, no for such case.

test/e2e/user_workload_monitoring_test.go
```diff
@@ -76,10 +75,22 @@ func TestUserWorkloadMonitoringInvalidConfig(t *testing.T) {
...
+	t.Log("invalid UWM configuration with malformed YAML/JSON")
 	f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionTrue)(t)
 	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionFalse)(t)
 	f.AssertOperatorConditionReason(configv1.OperatorDegraded, "UserWorkloadInvalidConfiguration")
 	f.AssertOperatorConditionReason(configv1.OperatorAvailable, "UserWorkloadInvalidConfiguration")
+
+	t.Log("restoring the initial configurations")
+	uwmCM.Data["config.yaml"] = ``
+	f.MustCreateOrUpdateConfigMap(t, uwmCM)
+	f.MustCreateOrUpdateConfigMap(t, getUserWorkloadEnabledConfigMap(t, f))
```
machine424: nit: do we need this `f.MustCreateOrUpdateConfigMap(t, getUserWorkloadEnabledConfigMap(t, f))` line?

test/e2e/user_workload_monitoring_test.go
```diff
@@ -76,10 +75,22 @@ func TestUserWorkloadMonitoringInvalidConfig(t *testing.T) {
...
+	f.AssertOperatorCondition(configv1.OperatorDegraded, configv1.ConditionFalse)(t)
+	f.AssertOperatorCondition(configv1.OperatorAvailable, configv1.ConditionTrue)(t)
+	// Once the config is adjusted, the operator becomes Upgradeable.
```
machine424: nit: I think the operator should always stay upgradeable for `cannot be deserialized`

test/e2e/config_test.go
jan--f: Yeah I see what you mean...thanks for your patience :sweat:

machine424:
/lgtm
thanks!

jan--f:
rebased. @machine424 requires your lgtm again.

jan--f:
/retitle [MON-4343](https://issues.redhat.com//browse/MON-4343): Cleanup deprecate pa config

dgoodwin:
Apologies but this took out payloads due to ROSA use of the deprecated config: https://issues.redhat.com/browse/OCPBUGS-61135

Per org policy I need to pursue a revert.

PR #2653, opened 2025-08-29
OCPBUGS-60948: Add selinux_warning_controller_selinux_volume_conflict to telemetry

jsafrane:
The metric shows number of pods that may fail to start when SELinuxMount feature gate is enabled.

While the other SELinux metrics sent through telemetry shows real conflicts that happen on a single node, the new metric shows _potential_ conflicts - the Pods run on different nodes, but if they land on the same one, they may not start when SELinuxMount is enabled.

Approval of the new telemetry metric: https://issues.redhat.com/browse/MON-4345


simonpasquier:
(suggestion) the metric name could be renamed to better convey the meaning. For instance, `cluster:pods_not_compatible_with_selinux_mount:count`.

jsafrane:
Renamed from `...:total` to `...:count`, as the value is not a counter.

jsafrane:
https://github.com/openshift/cluster-storage-operator/pull/609 has merged, PTAL at this PR.

jsafrane:
/retest-required
VolumeAttributesClass flake

PR #2656, opened 2025-09-01
OCPBUGS-61088: create networkpolicy settings for in-cluster monitoring

juzhao:
see https://issues.redhat.com/browse/OCPBUGS-61088, use this bug to set networkpolicy for in-cluster monitoring
this PR replaced https://github.com/openshift/cluster-monitoring-operator/pull/2645, 2645 does not include networkpolicy settings for alertmanager/prometheus/thanos-querier, since there is default deny network settings for all monitoring pods, if we don't include them, CI jobs would be failed for them

### DIFF
--- a/assets/admission-webhook/network-policy-downstream.yaml
+++ b/assets/admission-webhook/network-policy-downstream.yaml
@@ -4,22 +4,18 @@ metadata:
   labels:
     app.kubernetes.io/managed-by: cluster-monitoring-operator
     app.kubernetes.io/part-of: openshift-monitoring
-  name: alertmanager
+  name: prometheus-operator-admission-webhook
   namespace: openshift-monitoring
 spec:
   egress:
   - {}
   ingress:
   - ports:
-    - port: tenancy
-      protocol: TCP
-    - port: web
-      protocol: TCP
-    - port: metrics
+    - port: https
       protocol: TCP
   podSelector:
     matchLabels:
-      app.kubernetes.io/name: alertmanager
+      app.kubernetes.io/name: prometheus-operator-admission-webhook
   policyTypes:
   - Ingress
   - Egress
--- a/jsonnet/components/admission-webhook.libsonnet
+++ b/jsonnet/components/admission-webhook.libsonnet
@@ -168,4 +168,38 @@ function(params)
         },
       ],
     },
+    networkPolicyDownstream: {
+      apiVersion: 'networking.k8s.io/v1',
+      kind: 'NetworkPolicy',
+      metadata: {
+        name: 'prometheus-operator-admission-webhook',
+        namespace: 'openshift-monitoring',
+      },
+      spec: {
+        podSelector: {
+          matchLabels: {
+            'app.kubernetes.io/name': 'prometheus-operator-admission-webhook',
+          },
+        },
+        policyTypes: [
+          'Ingress',
+          'Egress',
+        ],
+        ingress: [
+          {
+            ports: [
+              {
+                // allow apiserver reach to prometheus-operator-admission-webhook
+                // 8443(port name: https) port to validate customresourcedefinitions
+                port: 'https',
+                protocol: 'TCP',
+              },
+            ],
+          },
+        ],
+        egress: [
+          {},
+        ],
+      },
+    },
   }
--- a/jsonnet/components/alertmanager.libsonnet
+++ b/jsonnet/components/alertmanager.libsonnet
@@ -440,8 +440,6 @@ function(params)
         ],
       },
     },
-    // Allow access to alertmanager 9092(port name: tenancy)/9095(port name: web)/9097(port name: metrics)
-    // and 9094(port name: udp-mesh for UDP, port name: tcp-mesh for TCP) ports
     networkPolicyDownstream: {
       apiVersion: 'networking.k8s.io/v1',
       kind: 'NetworkPolicy',
@@ -462,6 +460,12 @@ function(params)
         ingress: [
           {
             ports: [
+              {
+                // allow access to the Alertmanager endpoints restricted to a given project,
+                // port number 9092(port name: tenancy)
+                port: 'tenancy',
+                protocol: 'TCP',
+              },
               {
                 // allow prometheus to sent alerts to alertmanager, port number 9095(port name: web)
                 port: 'web',
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -1008,6 +1008,7 @@ func assertInClusterNetworkPolicyExists(t *testing.T) {
 		"monitoring-plugin",
 		"openshift-state-metrics",
 		"prometheus-operator",
+		"prometheus-operator-admission-webhook",
 		"telemeter-client",
 		"thanos-querier",
 	}
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -213,6 +213,7 @@ var (
 	AdmissionWebhookPodDisruptionBudget                 = "admission-webhook/pod-disruption-budget.yaml"
 	AdmissionWebhookService                             = "admission-webhook/service.yaml"
 	AdmissionWebhookServiceAccount                      = "admission-webhook/service-account.yaml"
+	AdmissionWebhookNetworkPolicy                       = "admission-webhook/network-policy-downstream.yaml"
 
 	PrometheusOperatorClusterRoleBinding  = "prometheus-operator/cluster-role-binding.yaml"
 	PrometheusOperatorClusterRole         = "prometheus-operator/cluster-role.yaml"
@@ -2169,6 +2170,10 @@ func (f *Factory) PrometheusOperatorNetworkPolicy() (*networkingv1.NetworkPolicy
 
 func (f *Factory) PrometheusOperatorAdmissionWebhookServiceAccount() (*v1.ServiceAccount, error) {
 	return f.NewServiceAccount(f.assets.MustNewAssetSlice(AdmissionWebhookServiceAccount))
+}
+
+func (f *Factory) AdmissionWebhookNetworkPolicy() (*networkingv1.NetworkPolicy, error) {
+	return f.NewNetworkPolicy(f.assets.MustNewAssetSlice(AdmissionWebhookNetworkPolicy))
 }
 
 func (f *Factory) PrometheusOperatorAdmissionWebhookService() (*v1.Service, error) {
--- a/pkg/tasks/prometheusoperator.go
+++ b/pkg/tasks/prometheusoperator.go
@@ -138,6 +138,16 @@ func (t *PrometheusOperatorTask) Run(ctx context.Context) error {
 }
 
 func (t *PrometheusOperatorTask) runAdmissionWebhook(ctx context.Context) error {
+	netpol, err := t.factory.AdmissionWebhookNetworkPolicy()
+	if err != nil {
+		return fmt.Errorf("initializing Prometheus Operator Admission Webhook NetworkPolicy failed: %w", err)
+	}
+
+	err = t.client.CreateOrUpdateNetworkPolicy(ctx, netpol)
+	if err != nil {
+		return fmt.Errorf("reconciling Prometheus Operator Admission Webhook NetworkPolicy failed: %w", err)
+	}
+
 	// Deploy manifests for the admission webhook service.
 	sa, err := t.factory.PrometheusOperatorAdmissionWebhookServiceAccount()
 	if err != nil {


juzhao:
/retitle [WIP] OCPBUGS-61088: create networkpolicy settings for in-cluster monitoring

jsonnet/components/admission-webhook.libsonnet
```diff
@@ -168,4 +168,42 @@ function(params)
...
+          'include.release.openshift.io/self-managed-high-availability': 'true',
+          'include.release.openshift.io/single-node-developer': 'true',
+        },
+        name: 'prometheus-operator-access',
+        namespace: 'openshift-monitoring',
+      },
+      spec: {
+        podSelector: {
+          matchLabels: {
+            'app.kubernetes.io/name': 'prometheus-operator',
```
simonpasquier: `prometheus-operator-admission-webhook` rather?

jsonnet/components/admission-webhook.libsonnet
```diff
@@ -168,4 +168,42 @@ function(params)
...
+          {
+            ports: [
+              {
+                port: '8443',
+                protocol: 'TCP',
+              },
+            ],
+          },
+        ],
+        egress: [
```
simonpasquier: I'm not sure that thee admission webhook needs any egress.

jsonnet/components/kube-state-metrics.libsonnet
```diff
@@ -307,4 +307,48 @@ function(params)
...
+                protocol: 'TCP',
+              },
+              {
+                port: '9443',
+                protocol: 'TCP',
+              },
+            ],
+          },
+        ],
+        egress: [
```
simonpasquier, line 343: Is there any recommendation to allow egress traffic only to the API service? This is the only destination which ksm should try to connect but I guess that it's not trivial to accommodate classic OCP (self-managed) and hosted clusters (HyperShift).

jsonnet/components/prometheus-operator.libsonnet
```diff
@@ -177,4 +177,42 @@ function(params)
...
+          'include.release.openshift.io/self-managed-high-availability': 'true',
+          'include.release.openshift.io/single-node-developer': 'true',
+        },
+        name: 'prometheus-operator-admission-webhook-access',
+        namespace: cfg.namespace,
+      },
+      spec: {
+        podSelector: {
+          matchLabels: {
+            'app.kubernetes.io/name': 'prometheus-operator-admission-webhook',
```
simonpasquier: typo?

juzhao: thanks, fixed

jsonnet/components/admission-webhook.libsonnet
juzhao: typo, fixed, thanks

jsonnet/components/admission-webhook.libsonnet
juzhao: this allows curl admission webhook enpoint from any pod of any space

jsonnet/components/kube-state-metrics.libsonnet
juzhao: networkpolicy applied to pod only, seems we can not define for API service, will check again.

jsonnet/components/admission-webhook.libsonnet
```diff
@@ -168,4 +169,45 @@ function(params)
...
+              ],
+            },
+          ],
+          egress: [
+            {},
+          ],
+        },
+      },
+    },
+    networkPolicyDownstream: renameNetworkPolicy.renameKey(netpol, 'networkPolicy', 'networkPolicyDownstream'),
```
juzhao: [ci/prow/generate](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2656/pull-ci-openshift-cluster-monitoring-operator-main-generate/1965972237311807488) error is
```
RUNTIME ERROR: Field does not exist: kind
	jsonnet/utils/configure-authentication-for-monitors.libsonnet:4:11-17	function <anonymous>
	jsonnet/utils/configure-authentication-for-monitors.libsonnet:43:10-39	object <anonymous>
	jsonnet/utils/remove-network-policy.libsonnet:3:10-14	object <anonymous>
	jsonnet/utils/remove-limits.libsonnet:19:22-26	thunk from <object <anonymous>>
	jsonnet/utils/remove-limits.libsonnet:3:28-29	function <anonymous>
	jsonnet/utils/remove-limits.libsonnet:19:10-27	object <anonymous>
	jsonnet/utils/sanitize-rules.libsonnet:648:73-77	thunk from <thunk from <thunk from <thunk from <object <anonymous>>>>>
	jsonnet/utils/sanitize-rules.libsonnet:590:19-20	function <anonymous>
	jsonnet/utils/sanitize-rules.libsonnet:648:59-78	thunk from <thunk from <thunk from <object <anonymous>>>>
	jsonnet/utils/sanitize-rules.libsonnet:605:17-18	function <anonymous>
	...
	jsonnet/utils/add-annotations.libsonnet:26:10-29	object <anonymous>
	jsonnet/utils/add-annotations.libsonnet:49:24-28	thunk from <object <anonymous>>
	jsonnet/utils/add-annotations.libsonnet:34:30-31	function <anonymous>
	jsonnet/utils/add-annotations.libsonnet:49:10-29	object <anonymous>
	jsonnet/utils/add-labels.libsonnet:12:10-14	object <anonymous>
	jsonnet/utils/set-terminationMessagePolicy.libsonnet:37:38-42	thunk from <object <anonymous>>
	jsonnet/utils/set-terminationMessagePolicy.libsonnet:3:44-45	function <anonymous>
	jsonnet/utils/set-terminationMessagePolicy.libsonnet:37:10-43	object <anonymous>
	Field "admission-webhook/networkPolicyDownstream"	
	During manifestation	
```
from jsonnet/components/admission-webhook.libsonnet code
`networkPolicyDownstream: renameNetworkPolicy.renameKey(netpol, 'networkPolicy', 'networkPolicyDownstream'),`
, the output is like below, which the top level networkPolicyDownstream is needless
```
{
  "networkPolicyDownstream": {
    "networkPolicyDownstream": {
      "apiVersion": "networking.k8s.io/v1",
      "kind": "NetworkPolicy",
      "metadata": {
        "name": "**",
        "namespace": "**"
      },
      "spec": {}
    }
  }
}
```
just want to get like
```
{
   "networkPolicyDownstream": {
      "apiVersion": "networking.k8s.io/v1",
      "kind": "NetworkPolicy",
      "metadata": {
         "name": "***",
         "namespace": "***"
      },
      "spec": { }
   }
}
```

change
`networkPolicyDownstream: renameNetworkPolicy.renameKey(netpol, 'networkPolicy', 'networkPolicyDownstream'),` to
`renameNetworkPolicy.renameKey(netpol, 'networkPolicy', 'networkPolicyDownstream'),`
get error
```
$ jsonnetfmt -i admission-webhook.libsonnet
STATIC ERROR: admission-webhook.libsonnet:212:24: expected token OPERATOR but got "."
```
don't know how to fix it

jsonnet/components/admission-webhook.libsonnet
```diff
@@ -168,4 +169,45 @@ function(params)
         },
       ],
     },
+    local netpol = {
+      networkPolicy: {
```
jan--f:
```suggestion
      networkPolicyDownstream: {
```
No need for creating a local and renaming it. Just create the networkPolicyDownstream field here.

Locals are not available outside of its definitio context.

jsonnet/components/admission-webhook.libsonnet
danielmellado: Besides Jan's comment, I think we may need to check whether o.kind exists or not and the manifest generation breaks. If we try to access `o.kind` one something that doesn't have it, jsonnet throws that error. 

For example, changing calls like `[if o.kind == 'ServiceMonitor' then 'spec']+: {` to `[if std.objectHas(o, 'kind') && o.kind == 'ServiceMonitor' then 'spec']+: {`

jsonnet/components/cluster-monitoring-operator.libsonnet
```diff
@@ -566,4 +566,44 @@ function(params) {
...
+              protocol: 'TCP',
+            },
+          ],
+        },
+      ],
+      egress: [
+        {},
+      ],
+    },
+  },
```
danielmellado, line 625: I'm getting a bit confused with all the changes, but shouldn't there be a defaultDenyNetworkPolicyDownstream defined here? `make verify` is failing as it expects that file to be in the assets.

juzhao: yes, forgot to add it, added it in next PR

jsonnet/components/admission-webhook.libsonnet
jan--f: No let's not add this check. That is what an LLM would suggest as well but this only masks an underlying error and we don't get the assets we expect.

juzhao:
rules [job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2656/pull-ci-openshift-cluster-monitoring-operator-main-rules/1966376820211191808) failed at
```
 + jsonnet -J jsonnet/vendor -J tmp/json-manifests jsonnet/main.jsonnet
RUNTIME ERROR: Unexpected type string, expected number
	jsonnet/utils/configure-authentication-for-monitors.libsonnet:4:11-17	function <anonymous>
	jsonnet/utils/configure-authentication-for-monitors.libsonnet:43:10-39	object <anonymous>
	jsonnet/utils/remove-network-policy.libsonnet:3:10-14	object <anonymous>
	jsonnet/utils/remove-limits.libsonnet:19:22-26	thunk from <object <anonymous>>
	jsonnet/utils/remove-limits.libsonnet:3:28-29	function <anonymous>
	jsonnet/utils/remove-limits.libsonnet:19:10-27	object <anonymous>
	jsonnet/utils/sanitize-rules.libsonnet:648:73-77	thunk from <thunk from <thunk from <thunk from <object <anonymous>>>>>
	jsonnet/utils/sanitize-rules.libsonnet:590:19-20	function <anonymous>
	jsonnet/utils/sanitize-rules.libsonnet:648:59-78	thunk from <thunk from <thunk from <object <anonymous>>>>
	jsonnet/utils/sanitize-rules.libsonnet:605:17-18	function <anonymous>
	...
	jsonnet/utils/add-annotations.libsonnet:26:10-29	object <anonymous>
	jsonnet/utils/add-annotations.libsonnet:49:24-28	thunk from <object <anonymous>>
	jsonnet/utils/add-annotations.libsonnet:34:30-31	function <anonymous>
	jsonnet/utils/add-annotations.libsonnet:49:10-29	object <anonymous>
	jsonnet/utils/add-labels.libsonnet:12:10-14	object <anonymous>
	jsonnet/utils/set-terminationMessagePolicy.libsonnet:37:38-42	thunk from <object <anonymous>>
	jsonnet/utils/set-terminationMessagePolicy.libsonnet:3:44-45	function <anonymous>
	jsonnet/utils/set-terminationMessagePolicy.libsonnet:37:10-43	object <anonymous>
	Field "cluster-monitoring-operator/networkPolicyDownstream"	
	During manifestation	
```
file jsonnet/components/cluster-monitoring-operator.libsonnet has 2 networkpolicies, one is for default deny all pods traffic, the other is allow access to CMO port 8443 (networkPolicyDownstream is array)
not sure how the [set-terminationMessagePolicy.libsonnet](https://github.com/openshift/cluster-monitoring-operator/blob/main/jsonnet/utils/set-terminationMessagePolicy.libsonnet) would throw errors for cluster-monitoring-operator networkpolicies.

maybe I need to update set-terminationMessagePolicy.libsonnet to consider the array, not sure if it will bring side effect. another solution is split the networkpolicy array, example
networkPolicyDownstream1 for default-deny, networkPolicyDownstream2 for CMO

jan--f:
> maybe I need to update set-terminationMessagePolicy.libsonnet to consider the array, not sure if it will bring side effect. another solution is split the networkpolicy array, example
> networkPolicyDownstream1 for default-deny, networkPolicyDownstream2 for CMO

We implicitly assume that the whole structure we manifest is a large map. Adding a list into the mix might also break in other places.

I think the default deny policy is sufficiently special to warrant its own key. Maybe something more expressive though? 
networkPolicyDownstream and networkPolicyDefaultDeny or so perhaps?

juzhao:
/retitle OCPBUGS-61088: create networkpolicy settings for in-cluster monitoring

assets/alertmanager/network-policy-downstream.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  annotations:
+    include.release.openshift.io/hypershift: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+  labels:
+    app.kubernetes.io/managed-by: cluster-monitoring-operator
+    app.kubernetes.io/part-of: openshift-monitoring
+  name: alertmanager-access
```
machine424: do we need to add the `-access` because of a convention or sth?
I'd use the same name as the component CR name/Deployment name etc. So the mapping is easier for users.
If we want to add another one for that convention, we can think about naming

machine424:
/hold
once the PR is ready, we need to ensure blocking (and maybe also informing) ci and payload jobs pass.

assets/alertmanager/network-policy-downstream.yaml
juzhao: thanks, just a name, not convention, maybe we can use component name for short

machine424: yes we can try the `app.kubernetes.io/name` of each.

assets/admission-webhook/network-policy-downstream.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  annotations:
+    include.release.openshift.io/hypershift: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
```
simonpasquier: (non-blocking) I believe that `include.release.openshift.io/*` annotations are only relevant for manifests deployed by CVO. Do we need them here?

simonpasquier (commented):
/lgtm

assets/admission-webhook/network-policy-downstream.yaml
juzhao: I saw the annnotations from the existing networkpolicy, example
```
$ oc -n openshift-cluster-storage-operator get netpol allow-all-egress -oyaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  annotations:
    capability.openshift.io/name: Storage
    include.release.openshift.io/hypershift: "true"
    include.release.openshift.io/ibm-cloud-managed: "true"
    include.release.openshift.io/self-managed-high-availability: "true"
    include.release.openshift.io/single-node-developer: "true"
```
so added to these networkpolicies

pkg/client/client.go
```diff
@@ -1585,6 +1600,11 @@ func (c *Client) CreateOrUpdatePodDisruptionBudget(ctx context.Context, pdb *pol
 	return err
 }
 
+func (c *Client) CreateOrUpdateNetworkPolicy(ctx context.Context, netpol *networkingv1.NetworkPolicy) error {
+	_, _, err := resourceapply.ApplyNetworkPolicy(ctx, c.kclient.NetworkingV1().NetworkPolicies(netpol.Namespace), c.eventRecorder, netpol)
```
juzhao: golangci-lint [job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2656/pull-ci-openshift-cluster-monitoring-operator-main-golangci-lint/1968583817073528832) failed for
```
pkg/client/client.go:1604:29: undefined: resourceapply.ApplyNetworkPolicy) (typecheck)
```

checked from https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/networking.go#L19
ApplyNetworkPolicy func is added in openshift/library-go, but not in CMO vendor openshift/library-go, see from: https://github.com/openshift/cluster-monitoring-operator/tree/main/vendor/github.com/openshift/library-go/pkg/operator/resource/resourceapply

we need to update the CMO vendor openshift/library-go to use the ApplyNetworkPolicy func in openshift/library-go

machine424: If the openshift/library-go update (better done in a separate PR) will involve a lot of work and will take time, especially if it requires to bump the kube deps (I know that the prom deps would break on 1.33 e.g. https://github.com/openshift/cluster-monitoring-operator/pull/2632#issuecomment-3149691979), you can just put the content of https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/networking.go#L19 in a file under https://github.com/openshift/cluster-monitoring-operator/tree/main/pkg/tasks and use it in the tasks.

In that case, please leave me a comment in https://github.com/openshift/cluster-monitoring-operator/pull/2632 so I adjust the wiring once the kube deps and library-go are updated.

I don't think that updating library-go in 4.20 with all the deps that should follow is a good idea, so yes, let's bring to CMO what you need and clean stuff up on main later.

juzhao: @machine424 thanks, main branch(4.21) will bump the CMO vendor openshift/library-go, see PR:
https://github.com/openshift/cluster-monitoring-operator/pull/2674

> In that case, please leave me a comment in https://github.com/openshift/cluster-monitoring-operator/pull/2632 
> so I adjust the wiring once the kube deps and library-go are updated.

answered in PR 2632

juzhao:
e2e-aws-ovn [job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2656/pull-ci-openshift-cluster-monitoring-operator-main-e2e-aws-ovn/1969400381104459776) is failed, checked must-gather
other components' networkpolicy are created, but only metrics-server networkpolicy is not created, that caused metrics-server pod CrashLoopBackOff(due to the default-deny all pods' traffic and the metrics-server 10250 port is not allowed access ), do you know why the metrics-server networkpolicy file is not created, I did not find why
```
$ omc -n openshift-monitoring get po | grep metrics-server
metrics-server-fd7cf46d8-5p8fl                           0/1     CrashLoopBackOff   17         44m
metrics-server-fd7cf46d8-s2rhd                           0/1     CrashLoopBackOff   17         44m

$ omc -n openshift-monitoring get networkpolicy
NAME                                    POD-SELECTOR                                                   AGE
alertmanager                            app.kubernetes.io/name=alertmanager                            44m
cluster-monitoring-operator             app.kubernetes.io/name=cluster-monitoring-operator             44m
default-deny                            <none>                                                         44m
kube-state-metrics                      app.kubernetes.io/name=kube-state-metrics                      44m
monitoring-plugin                       app.kubernetes.io/name=monitoring-plugin                       43m
node-exporter                           app.kubernetes.io/name=node-exporter                           44m
openshift-state-metrics                 app.kubernetes.io/name=openshift-state-metrics                 44m
prometheus                              app.kubernetes.io/name=prometheus                              44m
prometheus-operator                     app.kubernetes.io/name=prometheus-operator                     44m
prometheus-operator-admission-webhook   app.kubernetes.io/name=prometheus-operator-admission-webhook   44m
telemeter-client                        app.kubernetes.io/name=telemeter-client                        38m
thanos-querier                          app.kubernetes.io/name=thanos-query                            43m
```
metrics-server networkpolicy name is metrics-server, file: assets/metrics-server/network-policy-downstream.yaml
libsonnet file: jsonnet/components/metrics-server.libsonnet
@slashpai @machine424 can you help to check? thanks!

machine424:
> e2e-aws-ovn [job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2656/pull-ci-openshift-cluster-monitoring-operator-main-e2e-aws-ovn/1969400381104459776) is failed, checked must-gather other components' networkpolicy are created, but only metrics-server networkpolicy is not created, that caused metrics-server pod CrashLoopBackOff(due to the default-deny all pods' traffic and the metrics-server 10250 port is not allowed access ), do you know why the metrics-server networkpolicy file is not created, I did not find why
> 
> ```
> $ omc -n openshift-monitoring get po | grep metrics-server
> metrics-server-fd7cf46d8-5p8fl                           0/1     CrashLoopBackOff   17         44m
> metrics-server-fd7cf46d8-s2rhd                           0/1     CrashLoopBackOff   17         44m
> 
> $ omc -n openshift-monitoring get networkpolicy
> NAME                                    POD-SELECTOR                                                   AGE
> alertmanager                            app.kubernetes.io/name=alertmanager                            44m
> cluster-monitoring-operator             app.kubernetes.io/name=cluster-monitoring-operator             44m
> default-deny                            <none>                                                         44m
> kube-state-metrics                      app.kubernetes.io/name=kube-state-metrics                      44m
> monitoring-plugin                       app.kubernetes.io/name=monitoring-plugin                       43m
> node-exporter                           app.kubernetes.io/name=node-exporter                           44m
> openshift-state-metrics                 app.kubernetes.io/name=openshift-state-metrics                 44m
> prometheus                              app.kubernetes.io/name=prometheus                              44m
> prometheus-operator                     app.kubernetes.io/name=prometheus-operator                     44m
> prometheus-operator-admission-webhook   app.kubernetes.io/name=prometheus-operator-admission-webhook   44m
> telemeter-client                        app.kubernetes.io/name=telemeter-client                        38m
> thanos-querier                          app.kubernetes.io/name=thanos-query                            43m
> ```
> 
> metrics-server networkpolicy name is metrics-server, file: assets/metrics-server/network-policy-downstream.yaml libsonnet file: jsonnet/components/metrics-server.libsonnet @slashpai @machine424 can you help to check? thanks!

I can see that even the metrics-server deployment isn't "deployed": `reconciling MetricsServer Deployment failed: updating Deployment object failed: waiting for DeploymentRollout of openshift-monitoring/metrics-server`

Because the networkpolicy is deployed after that, that explains why you cannot see it (the task isn't able to reach that step).

Regarding the Deployment not being able to rollout, probably the metrics-server Pods try to connect to/ or to be reached by some other Pods, probably APIserver ones.

Maybe enabling debug log level on metrics-server can tell us more about that, you can also compare the debug logs with the ones from a startup without NP...

danielmellado:
I think that ths issue isn't actually about missing create permissions - it's a task execution order problem.

What's happening is: 
1) ClusterMonitoringOperatorTask creates the default-deny NetworkPolicy (blocking all traffic),
2) MetricsServerTask tries to deploy metrics-server pods concurrently, but the pods fail their health checks on port 10250 because it's blocked by the default-deny policy
3) the deployment rollout times out waiting for readiness, and 
4) the NetworkPolicy creation code in MetricsServerTask never gets reached because the task fails at step 2. The CI logs show exactly this - readiness probe failures with "context deadline exceeded" and the deployment stuck with 3 pods (old + new) instead of the expected 2.

One option for example could be to move the component network policy creation to ClusterMonitoringOperatorTask?

danielmellado:
**ClusterMonitoringOperatorTask starts:**
```go
// Creates default-deny NetworkPolicy (blocks ALL pod traffic)
denyNetpol, err := t.factory.ClusterMonitoringDenyAllTraffic()
err = t.client.CreateOrUpdateNetworkPolicy(ctx, denyNetpol)
```
- `default-deny` NetworkPolicy created, blocking all traffic in openshift-monitoring namespace

**MetricsServerTask:**
```go
// Line 194-202: Creates deployment FIRST
dep, err := t.factory.MetricsServerDeployment(...)
err = t.client.CreateOrUpdateDeployment(ctx, dep)
```
- metrics-server deployment created with `replicas: 2`
- Pods start but can't pass health checks on port 10250 (blocked by default-deny)

**Deployment rollout waits for readiness:**
- **Expected**: 2 pods ready
- **Actual**: 0 pods ready (health checks timeout due to network blocking)
- **Timeout**: Deployment rollout fails after waiting

**NetworkPolicy creation never reached:**
```go
// Line 229-240: This code NEVER executes because earlier step fails
netpol, err := t.factory.MetricsServerNetworkPolicy()
err = t.client.CreateOrUpdateNetworkPolicy(ctx, netpol)
```
- No allow rule for metrics-server port 10250

**Rolling update**
```
Old ReplicaSet (metrics-server-699b4fd74d): 2 pods → scaled down to 1 pod
New ReplicaSet (metrics-server-7bfbdf65cb): 0 pods → scaled up to 2 pods
Total: 3 pods (1 old + 2 new)
```

What I proposed earlier should fix this as this would change the execution flow to something along the lines of

```
1. ClusterMonitoringOperatorTask:
   ├── Create default-deny (blocks all)
   ├── Create metrics-server allow rule (allows port 10250)
   └── Create all other component allow rules

2. MetricsServerTask:
   ├── Create deployment
   ├── Pods start and pass health checks  (port 10250 allowed)
   ├── Deployment rollout succeeds 
   └── Continue with other resources
```

machine424:
We can just keep the NP creation in each component's task and just move the code that deploy them to the top; you're right that NP should be deployed first.
I doubt that would be enough for metrics-server, I think its NP needs to allow more traffic, but let's re-order first and see.

danielmellado:
go-proxy CI issues are unrelated to the patch

juzhao:
e2e CI error is
```
: Run multi-stage test e2e-agnostic-operator expand_less 	14s
{  failed to acquire lease for "aws-2-quota-slice": status 503 Service Unavailable, status code 503}
```
seems prow issue

juzhao:
1. [e2e-aws-ovn-techpreview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2656/pull-ci-openshift-cluster-monitoring-operator-main-e2e-aws-ovn-techpreview/1971031731083415552) failed
```
: [sig-instrumentation][Late] Alerts shouldn't exceed the series limit of total series sent via telemetry from each cluster [Suite:openshift/conformance/parallel]
    [
        <*errors.errorString | 0xc0074b8030>{
            s: "promQL query returned unexpected results:\navg_over_time(cluster:telemetry_selected_series:count[1h37m56s]) >= 780\n[\n  {\n    \"metric\": {\n      \"prometheus\": \"openshift-monitoring/k8s\"\n    },\n    \"value\": [\n      1758776352.201,\n      \"788.4744897959183\"\n    ]\n  }\n]",
        },
    ]
```
same issue as https://github.com/openshift/cluster-monitoring-operator/pull/2681#issuecomment-3332362127

2. [e2e-agnostic-operator](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2656/pull-ci-openshift-cluster-monitoring-operator-main-e2e-agnostic-operator/1971031730949197824)
failed at image_registry_test.go, PR https://github.com/openshift/cluster-monitoring-operator/pull/2681 will remove the tests
```
    image_registry_test.go:52: 
        	Error Trace:	/go/src/github.com/openshift/cluster-monitoring-operator/test/e2e/image_registry_test.go:52
        	            				/go/src/github.com/openshift/cluster-monitoring-operator/test/e2e/image_registry_test.go:18
        	Error:      	Not equal: 
        	            	expected: "registry.build05.ci.openshift.org"
        	            	actual  : "quay-proxy.ci.openshift.org"
```

juzhao:
tested with PR, no regression issues.
/verified by @juzhao

assets/admission-webhook/network-policy-downstream.yaml
machine424: probably those NPs are deployed by CVO.
If NPs are not required to have those annotations, I agree with Simon, let's remove them

pkg/tasks/alertmanager.go
```diff
@@ -53,6 +53,18 @@ func (t *AlertmanagerTask) Run(ctx context.Context) error {
 }
 
 func (t *AlertmanagerTask) create(ctx context.Context) error {
+	netpol, err := t.factory.AlertmanagerNetworkPolicy()
+	if err != nil {
+		return fmt.Errorf("initializing Alertmanager NetworkPolicy failed: %w", err)
+	}
+
+	if netpol != nil {
```
machine424: no need for those checks as they may hide broken factories, if a factory returns a nil NP with no err, we want the code to break so we fix the factory

pkg/tasks/alertmanager.go
```diff
@@ -329,6 +341,18 @@ func (t *AlertmanagerTask) destroy(ctx context.Context) error {
 		}
 	}
 
+	netpol, err := t.factory.AlertmanagerNetworkPolicy()
+	if err != nil {
+		return fmt.Errorf("initializing Alertmanager NetworkPolicy object failed: %w", err)
+	}
+
+	if netpol != nil {
+		err = t.client.DeleteNetworkPolicy(ctx, netpol)
```
machine424: we can add a check that the NP disappears here https://github.com/openshift/cluster-monitoring-operator/blob/709ab51a727c8d2f8d2755d4e3c263021eaf5079/test/e2e/alertmanager_test.go#L679

pkg/tasks/alertmanager.go
juzhao: thanks, will remove `if netpol != nil` check for all components

jsonnet/components/alertmanager.libsonnet
```diff
@@ -440,4 +440,58 @@ function(params)
...
+        },
+        policyTypes: [
+          'Ingress',
+          'Egress',
+        ],
+        ingress: [
+          {
+            ports: [
+              {
+                port: 9092,
```
machine424: I'll not suggest referencing the port numbers from existing values instead of hard-coding them, I know you had enough jsonnet ;)
But can you add a comment per port explaining why we should open it to improve readability/reviewing?

(If you could create a ticket to try/think about referencing to make sure the NPs stay up to date, that'd be great)

danielmellado: Totally agree with this, as of now it may seems totally straightforward but the future us will deff appreciate ;)

jsonnet/components/admission-webhook.libsonnet
machine424: > this allows curl admission webhook enpoint from any pod of any space

IIUC, the egress is with regard to the admission-webhook Pod, the igress will already allows curling it from outside.
the egress is to stop admission-webhook initialize stuff (connect to others etc), and I agree with Simon, the webhook doesn't need to initialize anything so we can close its egress

jsonnet/components/kube-state-metrics.libsonnet
machine424: we can use pod/namespace selectors in rgress rules https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors
but yes, it could be tricky to have it right for all the flavors.

maybe we can create a ticket for this, and try to narrow down egress for ksm and others in a follow up PR, with some manual/payload testing we should be able to get it right. (maybe in 4.21)

jsonnet/components/alertmanager.libsonnet
juzhao: thanks, will update

assets/admission-webhook/network-policy-downstream.yaml
juzhao: removed

jsonnet/components/alertmanager.libsonnet
juzhao: added e2e cases and referenced the port numbers from existing values instead of hard-coding them

pkg/tasks/alertmanager.go
juzhao: added, thanks

jsonnet/components/admission-webhook.libsonnet
juzhao: removed

jsonnet/components/admission-webhook.libsonnet
```diff
@@ -168,4 +168,33 @@ function(params)
         },
       ],
     },
+    // Allow access to prometheus-operator-admission-webhook 8443(port name is https) port
```
machine424: I think it’d be better to put those comments next to the ports they refer to and to also share why we should expose it.

jsonnet/components/alertmanager.libsonnet
```diff
@@ -440,4 +440,54 @@ function(params)
...
+        ],
+        ingress: [
+          {
+            ports: [
+              {
+                port: 'tenancy',
+                protocol: 'TCP',
+              },
+              {
+                port: 'tcp-mesh',
```
machine424: I don't think the mesh ports need to be exposed, AFAIK they're only used for the alertmanager clustering (only alertmanagers need to reach each other there)

jsonnet/components/alertmanager.libsonnet
```diff
@@ -440,4 +440,54 @@ function(params)
...
+              {
+                port: 'udp-mesh',
+                protocol: 'UDP',
+              },
+              {
+                port: 'web',
+                protocol: 'TCP',
+              },
+              {
+                port: 'metrics',
```
machine424, line 476: we probably don't need to expose those internal metrics ports as only prometheus from the same namespace should be able to use them.
But maybe some users/layers need to do so from outside the namespace

test/e2e/config_test.go
```diff
@@ -995,6 +995,35 @@ func expectVolumeMountsInContainer(containerName, mountName string) framework.Po
...
 }
 
+// assertNetworkPolicyExists ensures that the NetworkPolicies
+// are deployed under openshift-monitoring namespace
+func assertNetworkPolicyExists(t *testing.T) {
+	assertions := []struct {
+		name      string
+		assertion framework.AssertionFunc
+	}{
+		{name: "assert default deny networkpolicy exists", assertion: f.AssertNetworkPolicyExists("default-deny", f.Ns)},
```
machine424: nit: we can just loop over a slice of names and call the `AssertNetworkPolicyExists` once in the test

jsonnet/components/admission-webhook.libsonnet
```diff
@@ -168,4 +168,33 @@ function(params)
...
+          },
+        },
+        policyTypes: [
+          'Ingress',
+        ],
+        ingress: [
+          {
+            ports: [
+              {
+                port: 'https',
```
machine424, line 194:
```suggestion
               // expose the webhook endpoints to the apiserver pods for example.
                port: 'https',
```

jsonnet/components/admission-webhook.libsonnet
machine424: see example below, it'd be great to have the why for all ports, if we can't think of any external client, maybe the port shouldn't be exposed...

jsonnet/components/alertmanager.libsonnet
juzhao: the prometheus metrics port need not to be exposed, but the alertmanager metrics port should be exposed, otherwise the alertmanager target is down, see from picture
Image

machine424: okk, I think we can refine that as only prometheus pods should be able to reach that port, but let's do that later.

jsonnet/components/alertmanager.libsonnet
juzhao: yes, confirmed, no need to open, removed

jsonnet/components/admission-webhook.libsonnet
juzhao: updated

juzhao:
reason for remove node-exporter networkpolicy to open 9100 port: even not open 9100 port for node-exporter, the node-expoert targets are UP and can scrape node-exporter metrics from prometheus, seems the port is open in node, but did not find where it's opened

juzhao:
@machine424 can `lgtm` added to this PR?

test/e2e/config_test.go
```diff
@@ -995,6 +995,33 @@ func expectVolumeMountsInContainer(containerName, mountName string) framework.Po
...
+		"prometheus-operator",
+		"prometheus-operator-admission-webhook",
+		"telemeter-client",
+		"thanos-querier",
+	}
+
+	t.Run("check in-cluster monitoring NetworkPolicies", func(t *testing.T) {
+		for _, name := range networkPolicyNames {
+			t.Run(fmt.Sprintf("assert %s networkpolicy exists", name), func(t *testing.T) {
+				f.AssertNetworkPolicyExists(name, f.Ns)
```
machine424, line 1019: in a follow up maybe:  let's also check that the count of deployed NPs is equal to `len(networkPolicyNames)`

machine424:
I looked at the failing payloads and I can see that some of the aggregates are passing though, so we cannot conclude that the NPs are breaking some workflows in the tests. 

Let's wait until we stabilize this on main before deciding on backporting to 4.20

We'll also try to to propose fixes instead of reverting the whole change in case it breaks some workflow that isn't tested. I'll try to keep an eye on that.

Thanks again for working on this!

/lgtm

pkg/client/client.go
```diff
@@ -1585,6 +1600,11 @@ func (c *Client) CreateOrUpdatePodDisruptionBudget(ctx context.Context, pdb *pol
 	return err
 }
 
+func (c *Client) CreateOrUpdateNetworkPolicy(ctx context.Context, netpol *networkingv1.NetworkPolicy) error {
+	_, _, err := resourceapply.ApplyNetworkPolicy(ctx, c.kclient.NetworkingV1(), c.eventRecorder, netpol)
```
lunarwhite, line 1604: (non-blocking, could be addressed as a follow-up)

FYI there is a recent hotfix landed in `ApplyNetworkPolicy` utils function to address few reconciliation flaws: https://github.com/openshift/library-go/pull/2042. This patch has been cherrypicked onto `release-4.20` branch as well. You might want to adapt this fix by bumping deps with latest `o/library-go`.

juzhao: thanks, will do

PR #2657, opened 2025-09-02
OCPBUGS-61113: add flag `--watch-referenced-objects-in-all-namespaces` to prometheus-operator

slashpai:


simonpasquier:
/lgtm
/hold

we need to assess the impact on the memory usage before/after switching the flag.

slashpai:
@simonpasquier are we good to unhold?

PR #2658, opened 2025-09-02
OCPBUGS-61135: Revert "MON-4343: Cleanup deprecate pa config"

dgoodwin:
Reverts openshift/cluster-monitoring-operator#2651

See https://issues.redhat.com/browse/OCPBUGS-61135


dgoodwin:
/verified

Revert to known good state.

/cherry-pick release-4.20

dgoodwin:
/verified by revert to known good state

PR #2664, opened 2025-09-10
OCPBUGS-60751: Fix Prometheus UI redirect for port-forward access

danielmellado:
When accessing Prometheus via port-forward, the UI incorrectly redirects
to /monitoring/query instead of /query, causing a 404 error.


juzhao:
tested with PR, still see the same issue, details see from bug commnent

juzhao:
tested again with the new change, still same issue

PR #2665, opened 2025-09-10
OCPBUGS-15430: Remove KubeAPIDown alert

ingvagabund:
KubeAPIDown alert fires only when either all kube-apiserver instances are down or connection to Prometheus is down. From the use perspective accessing the Console is a common scenario. Thus, when API is down the console/Prometheus can not be accessed. Thus, this alert is not observed unless an external observer is installed. There are also other alerts like KubeAPIErrorBudgetBurn that are expected to trigger before the API is down.

### DIFF
--- a/assets/control-plane/prometheus-rule.yaml
+++ b/assets/control-plane/prometheus-rule.yaml
@@ -352,16 +352,6 @@ spec:
       for: 15m
       labels:
         severity: warning
-    - alert: KubeAPIDown
-      annotations:
-        description: KubeAPI has disappeared from Prometheus target discovery.
-        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-monitoring-operator/KubeAPIDown.md
-        summary: Target disappeared from Prometheus target discovery.
-      expr: |
-        absent(up{job="apiserver"} == 1)
-      for: 15m
-      labels:
-        severity: critical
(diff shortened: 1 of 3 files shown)


assets/control-plane/prometheus-rule.yaml
```diff
@@ -352,16 +352,6 @@ spec:
       for: 15m
       labels:
         severity: warning
-    - alert: KubeAPIDown
```
ingvagabund, line 355: Alternatively, `KubeAPIDown` could be renamed to `KubeAPIAbsentInPrometheus` or similar.

simonpasquier:
I think that it deserves a bigger discussion with the API team. Ideally they should be the one owning all API-related rules (CMO ships the KubeAPIDown rule only because it comes from https://github.com/kubernetes-monitoring/kubernetes-mixin/ which is the upstream source of CMO).
Depending on their answer I see 2 options:
* The alerting rule should be kept (or renamed) => we need to organize a migration of the rule to github.com/openshift/cluster-kube-apiserver-operator/
* The alerting rule can go away => let's remove it from CMO.

As an aside, I see other alerting rules that would probably need to move to cluster-kube-apiserver-operator
* KubeClientErrors
* KubeAggregatedAPIErrors
* KubeAggregatedAPIDown
* KubeAPITerminatedRequests

cc @dgrisonnet @jan--f

PR #2675, opened 2025-09-18
MON-4361,MON-4380: Optional Monitoring Capability

rexagod:
Metric rules and metrics exporters have not been opted-in to keep the telemetry rules functioning.


simonpasquier (commented):
It might be good to wait for https://github.com/openshift/cluster-monitoring-operator/pull/2649 since it's migrating all the dashboards to static assets.

On the jsonnet implementation side, I wonder if it wouldn't easier to read/maintain if we inject the annotation into each component that needs it.
E.g. here for components for which all resources are OptionalMonitoring

https://github.com/openshift/cluster-monitoring-operator/blob/ea9a533f05318f51114d2e98e5172f1dcd769eb4/jsonnet/main.jsonnet#L520-L539

Or at the level of the jsonnet component file in case it's per resource.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
   - `KubePdbNotEnoughHealthyPods`
   - `KubeNodePressure`
   - `KubeNodeEviction`
+- []() Allow cluster-admins to opt-into optional monitoring using the `OptionalMonitoring` capability.
```
simonpasquier: I realize that adding the annotation to the manifests under the `assets/` directory will have no direct effect since there's no logic in CMO to deploy these resources conditionally, right?

rexagod: I've raised a PR for that: https://github.com/openshift/cluster-monitoring-operator/pull/2688/files.

rexagod:
Reverted the `capability.openshift.io/name: Console` to respect [being able to support dashboards in optional monitoring](https://redhat-internal.slack.com/archives/C6A3NV5J9/p1753979917745059?thread_ts=1753965750.365569&cid=C6A3NV5J9) since we'll still be scraping all targets anyway (to not break any telemetry rules).

CHANGELOG.md
simonpasquier: Thanks!

assets/admission-webhook/alertmanager-config-validating-webhook.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   annotations:
+    capability.openshift.io/name: OptionalMonitoring
```
simonpasquier, line 5: If the service is optional (*), shouldn't we apply the annotation to all admission-webhook resources? 

(*) there could be an argument that we still want the admission webhook for PrometheusRule resources because of telemetry?

assets/monitoring-plugin/deployment.yaml
simonpasquier, line 1: Not directly related to this change but if the console is disabled, wouldn't it be logical to avoid deploying the monitoring plugin resources?

manifests/0000_50_cluster-monitoring-operator_00_0alertingrules-custom-resource-definition.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   annotations:
     api-approved.openshift.io: https://github.com/openshift/api/pull/1406
     api.openshift.io/merged-by-featuregates: "true"
+    capability.openshift.io/name: OptionalMonitoring
```
simonpasquier: not sure that CMO will start if the CRDs aren't present.

manifests/0000_50_cluster-monitoring-operator_00_0alertmanager-config-custom-resource-definition.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    capability.openshift.io/name: OptionalMonitoring
```
simonpasquier, line 5: same here: IIRC the Prometheus operator will (at the minimum) log errors if the CRDs aren't installed.

rexagod:
_x-posting from #2688:_

> Speaking out loud but wouldn't it be possible for the operator to look at the `capability.openshift.io/name` annotation on each managed resource and skip the action if it's marked as OptionalMonitoring and the capability is actually disabled? This way we wouldn't have to change the tasks orchestration.

I believe that would require us to bake that behavior into several `t.client` methods, but as of now, most of the exclusions are per-component, with per-object exclusions only being limited to CMO and PO tasks, for which I've appended the capability check into their surrounding conditionals.

I also think a per-component exclusion at component initialization lets us dodge the need for a cache-based `hasOptionalMonitoringCapability` check so it's not querying the API for all object requests under a component.

EDIT: I've refactored the patch to not modify the tasks' orchestration, but the tasks themselves as I realised that we don't actually need to think about cleaning these up, quoting the third commit:

> Enabling the `OptionalMonitoring` capability translates to enabling all
optional monitoring components under CMO. Note that since capabilities
cannot be disabled once enabled, so cleanup for optional monitoring
resources is not necessary. To clarify further, there are two possible
paths at install time:
> * capability is disabled -> enabled (no need to cleanup)
> * capability is enabled -/> (cannot be disabled) (no need to cleanup)

assets/monitoring-plugin/deployment.yaml
rexagod: True, I've moved the plugin to be independent of the `OptionalMonitoring` capability and instead made it dependent on the `Console` one, which is in-line with its [task's behavior](https://github.com/openshift/cluster-monitoring-operator/blob/main/pkg/tasks/monitoring_plugin.go#L42-L56), PTAL at commit: https://github.com/openshift/cluster-monitoring-operator/pull/2675/commits/55d6da0e17313626b46823ebe1da6592a0d59432.

assets/admission-webhook/alertmanager-config-validating-webhook.yaml
rexagod: Yes that was my understanding as well, so I limited this to AM strictly.

manifests/0000_50_cluster-monitoring-operator_00_0alertingrules-custom-resource-definition.yaml
rexagod: Excluded.

manifests/0000_50_cluster-monitoring-operator_00_0alertmanager-config-custom-resource-definition.yaml
rexagod: Excluded.

pkg/client/client.go
```diff
@@ -1748,6 +1748,10 @@ func (c *Client) HasConsoleCapability(ctx context.Context) (bool, error) {
 	return c.HasClusterCapability(ctx, configv1.ClusterVersionCapabilityConsole)
 }
 
+func (c *Client) HasOptionalMonitoringCapability(ctx context.Context) (bool, error) {
+	return c.HasClusterCapability(ctx, "")
```
rexagod: /hold

Needs https://github.com/openshift/api/pull/2468

assets/admission-webhook/alertmanager-config-validating-webhook.yaml
simonpasquier: Ah I didn't realize that this webhook configuration was specifically for AlertmanagerConfig resources. I would still recommend that we document the rationale for not having all admission-webhook resources marked as optional (not sure where it should happen though).

jsonnet/utils/opt-into-capability.libsonnet
```diff
@@ -0,0 +1,31 @@
+{
+  local addAnnotationToChild(parent, annotationKeyCapability, annotationValueOptionalMonitoringCapability) =
+    parent {
+      metadata+: {
+        annotations+: {
+          [annotationKeyCapability]: annotationValueOptionalMonitoringCapability,
+        },
+      },
+    },
+  local addAnnotationToChildren(parent, annotationKeyCapability, annotationValueOptionalMonitoringCapability) =
```
simonpasquier:
```suggestion
  local addAnnotationToChildren(parent, key, value) =
```

jsonnet/utils/opt-into-capability.libsonnet
```diff
@@ -0,0 +1,31 @@
+{
+  local addAnnotationToChild(parent, annotationKeyCapability, annotationValueOptionalMonitoringCapability) =
```
simonpasquier: (nit) `o` (for object) instead of parent?
```suggestion
  local addAnnotationToChild(o, key, value) =
```

jsonnet/utils/opt-into-capability.libsonnet
```diff
@@ -0,0 +1,31 @@
...
+          }
+        else
+          addAnnotationToChild(parent[k], annotationKeyCapability, annotationValueOptionalMonitoringCapability)
+      for k in std.objectFields(parent)
+    },
+
+  local annotationKeyCapability = 'capability.openshift.io/name',
+  local annotationValueConsoleCapability = 'Console',
+  local annotationValueOptionalMonitoringCapability = 'OptionalMonitoring',
+  consoleForObject(o): addAnnotationToChild(o, annotationKeyCapability, annotationValueConsoleCapability),
```
simonpasquier, line 29: (suggestion) can we add a small comment before each function to describe what it does?

pkg/tasks/alertmanager.go
```diff
@@ -44,7 +44,11 @@ func NewAlertmanagerTask(
 }
 
 func (t *AlertmanagerTask) Run(ctx context.Context) error {
-	if t.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.IsEnabled() {
+	optionalMonitoringEnabled, err := t.client.HasOptionalMonitoringCapability(ctx)
```
simonpasquier: Can we have the operator retrieve the information about monitoring capability once before reconciling and provide the information to each task runner?
In the same vein, we create a small helper which would wrap a "Task" and run it only if the monitoring capability is enabled (e.g. useful for "Task"s for which all managed resources are dependent on the capability).

jsonnet/versions.yaml
```diff
@@ -21,6 +21,6 @@ versions:
     monitoringPlugin: 1.0.0
     nodeExporter: 1.9.1
     promLabelProxy: 0.12.1
-    prometheus: 3.5.0
+    prometheus: 3.6.0
```
simonpasquier: (nit) don't bother about updating the versions.yaml file in this PR.

manifests/0000_50_cluster-monitoring-operator_00_0alertmanager-custom-resource-definition.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    capability.openshift.io/name: OptionalMonitoring
```
wking, line 5: Nit: `OptionalMonitoring` seems like a fairly redundant name for an optional capability?  Is there no more specific/descriptive name we can use for the portion of the monitoring component that's covered by this capability name?  If the naming has already been hashed out in an enhancement or something, just point me at that discussion.  I'm not trying to re-open something that's already settled.

rexagod: `OptionalMonitoring` indicates a capability that targets the optional components of the monitoring stack. Internally, we've used the same terminology for all components that are not required for telemetry. In retrospect, this seemed better than, say, `NonTelemetryComponents`, but open to ideas.

rexagod: (I followed [these steps](https://github.com/openshift/enhancements/blob/master/enhancements/installer/component-selection.md#how-to-implement-a-new-capability) to go about implementing this capability, so I didn't create an EP, but PLMK if that's necessary)

assets/admission-webhook/alertmanager-config-validating-webhook.yaml
rexagod: Added a description annotation on the object.

rexagod:
Trying to spin up a cluster for the better part of today, but no luck, so I still need to test out the latest commit(s) just to be sure.

rexagod:
> Can we have the operator retrieve the information about monitoring capability once before reconciling and provide the information to each task runner?

Missed this part, on it.

rexagod:
/remove-lifecycle stale

Will revisit this once the related telemetry-profile effort is complete.

PR #2680, opened 2025-09-22
WIP: fix(prometheus): use legacy validation and escaping schemes until fully utf8 support

machine424:
This doesn't guarantee utf8 is disabled everywhere, promtool e.g. will continue using utf8.

This only concerns scraping. 

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -25,8 +25,11 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/stretchr/testify/require"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/clientcmd"
 )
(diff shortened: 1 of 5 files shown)


test/e2e/main_test.go
```diff
@@ -188,3 +191,70 @@ func TestMemoryUsageRecordingRule(t *testing.T) {
...
+				{
+					Name: "example",
+					Rules: []monitoringv1.Rule{
+						{
+							Record: "foo.bar:count",
+							Expr:   intstr.FromString(`count(foo.bar)`),
+						},
+						{
+							Alert: "Alert",
+							Expr:  intstr.FromString("foo.bar == 1"),
```
simonpasquier, line 251: `{"foo.bar"}` rather?

test/e2e/main_test.go
```diff
@@ -188,3 +191,70 @@ func TestMemoryUsageRecordingRule(t *testing.T) {
...
+							For:   func(d monitoringv1.Duration) *monitoringv1.Duration { return &d }("1s"),
```
simonpasquier, line 252: (nit)
```suggestion
```

test/e2e/main_test.go
```diff
@@ -188,3 +191,70 @@ func TestMemoryUsageRecordingRule(t *testing.T) {
 		"count(namespace:container_memory_usage_bytes:sum)",
 	)
 }
+
+func TestPrometheusUTF8Support(t *testing.T) {
```
simonpasquier, line 195: Ideally we should deploy an application exposing UTF-8 label names and verify that the metric/label names are escaped properly.

machine424: Yes, I thought about that, but deploying a custom target in CMO's e2e will require some work, for now we should just ensure those utf8 targets aren't scraped (aren't scraped using utf8), so it's not worth it, once we fully support utf8, that setup will make a good regression/2e2 test indeed.

machine424: Given that I'd need to bump some po deps (at least .85 is needed) and thus some kube deps in 4.20 to better control prometheus global config regarding escaping/validation schema. I think I will just disable utf8 on prometheus side for now.

machine424: Opened https://github.com/openshift/prometheus/pull/272

machine424:
/close
in favor of https://github.com/openshift/prometheus/pull/272

PR #2682, opened 2025-09-25
NO-ISSUE: chore: remove mentions of prometheus Adapter

simonpasquier:

### DIFF
--- a/assets/cluster-monitoring-operator/cluster-role-aggregated-metrics-reader.yaml
+++ b/assets/cluster-monitoring-operator/cluster-role-aggregated-metrics-reader.yaml
@@ -2,7 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   labels:
-    app.kubernetes.io/component: metrics-adapter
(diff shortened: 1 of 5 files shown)


assets/cluster-monitoring-operator/cluster-role-aggregated-metrics-reader.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   labels:
-    app.kubernetes.io/component: metrics-adapter
+    app.kubernetes.io/component: metrics-server
```
machine424, line 5: One could argue that sth like
```suggestion
    app.kubernetes.io/component: operator
```
is ok as well...

machine424:
thanks!
/lgtm

assets/cluster-monitoring-operator/cluster-role-aggregated-metrics-reader.yaml
simonpasquier: yeah, my rationale was that metrics-server replaces metrics-adapter hence the choice.

PR #2681, opened 2025-09-23
OCPBUGS-62109: test: remove image registry e2e tests

simonpasquier:
This commit removes end-to-end tests which verified that all container images originate from the same registry. With recent CI changes, it isn't true anymore.

For context, the e2e test was added due to https://github.com/openshift/cluster-monitoring-operator/pull/1392#pullrequestreview-776018095 and added in https://github.com/openshift/cluster-monitoring-operator/pull/1430. We can discuss later if we need to revive the test but for now all our PRs are blocked.

* [ ] I added CHANGELOG entry for this change.
* [x] No user facing changes, so no entry in CHANGELOG was needed.

PR #2683, opened 2025-09-25
chore: remove image locations in assets

simonpasquier:
The image locations should be injected by CMO from the information provided by CVO. By removing the default values in the static manifests, end-to-end tests should catch bugs in CMO forgetting to pass the expected image locations.

Follow-up of #2681 


machine424:
/lgtm
thanks!

PR #2684, opened 2025-09-26
WIP: OCPBUGS-62160: Remove AlertManager endpoints when disabled

danielmellado:
This commit adds a specific validation to remove AlertManager endpoints
if it's disabled

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -1448,7 +1448,11 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret) 
 		p.Spec.Thanos.Image = &f.config.Images.Thanos
 	}
 
-	setupAlerting(p, platformAlertmanagerService, f.namespace)
+	if f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.IsEnabled() {
+		setupAlerting(p, platformAlertmanagerService, f.namespace)
+	} else {
+		p.Spec.Alerting.Alertmanagers = []monv1.AlertmanagerEndpoints{}
+	}
 	f.adjustGoGCRelatedConfig(p)
 
 	for i, container := range p.Spec.Containers {
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -1954,6 +1954,71 @@ func TestPrometheusRetentionConfigs(t *testing.T) {
 	}
 }
 
+func TestPrometheusK8sAlertmanagerDisabled(t *testing.T) {
+	c, err := NewConfigFromString(`alertmanagerMain:
+  enabled: false`, false)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+	c.SetImages(map[string]string{
+		"prometheus":       "docker.io/openshift/origin-prometheus:latest",
+		"kube-rbac-proxy":  "docker.io/openshift/origin-kube-rbac-proxy:latest",
+		"prom-label-proxy": "docker.io/openshift/origin-prom-label-proxy:latest",
+	})
+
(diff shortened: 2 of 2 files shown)


pkg/manifests/manifests.go
```diff
@@ -1448,7 +1448,11 @@ func (f *Factory) PrometheusK8s(grpcTLS *v1.Secret, telemetrySecret *v1.Secret)
 		p.Spec.Thanos.Image = &f.config.Images.Thanos
 	}
 
-	setupAlerting(p, platformAlertmanagerService, f.namespace)
+	if f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.IsEnabled() {
+		setupAlerting(p, platformAlertmanagerService, f.namespace)
```
simonpasquier: should we extend `setupAlerting()` to take an `enabled` bool argument?

simonpasquier:
Can we extend `TestAlertmanagerDisabling()` which already covers disabling the platform Alertmanager?

pkg/manifests/manifests.go
danielmellado: yes, I think it'd end cleaner that way, let me update this.

danielmellado:
> Can we extend `TestAlertmanagerDisabling()` which already covers disabling the platform Alertmanager?

added a new assertion and for now removed the tests I originally add, but was wondering whether they'd still be useful as unit tests or not, wdyt?

simonpasquier:
> > Can we extend `TestAlertmanagerDisabling()` which already covers disabling the platform Alertmanager?
> 
> added a new assertion and for now removed the tests I originally add, but was wondering whether they'd still be useful as unit tests or not, wdyt?

I find that for configuration testing, end-to-end is usually the more robust option because it will ensure that all components are properly deployed and interact together as expected.

pkg/manifests/manifests.go
```diff
@@ -1805,9 +1810,9 @@ func (f *Factory) PrometheusUserWorkload(grpcTLS *v1.Secret) (*monv1.Prometheus,
 	f.adjustGoGCRelatedConfig(p)
 
 	if f.config.UserWorkloadConfiguration.Alertmanager.Enabled {
-		setupAlerting(p, userWorkloadAlertmanagerService, f.namespaceUserWorkload)
+		setupAlerting(p, userWorkloadAlertmanagerService, f.namespaceUserWorkload, true)
 	} else {
-		setupAlerting(p, platformAlertmanagerService, f.namespace)
+		setupAlerting(p, platformAlertmanagerService, f.namespace, true)
```
simonpasquier: shouldn't we pass `f.config.ClusterMonitoringConfiguration.AlertmanagerMainConfig.IsEnabled()`?

test/e2e/framework/assertions.go
```diff
@@ -311,6 +311,26 @@ func (f *Framework) AssertPrometheusExists(name, namespace string) func(t *testi
...
 
+func (f *Framework) AssertPrometheusAlertmanagerEndpointsEmpty(name, namespace string) func(t *testing.T) {
+	return func(t *testing.T) {
+		err := Poll(time.Second, 5*time.Minute, func() error {
+			prom, err := f.MonitoringClient.Prometheuses(namespace).Get(ctx, name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+
+			if len(prom.Spec.Alerting.Alertmanagers) != 0 {
```
simonpasquier: Can we check the number of discovered Alertmanager endpoints from the Prometheus config?

danielmellado: Done!

pkg/manifests/manifests.go
danielmellado: Done ;) I also think it's more resilient that way

PR #2688, opened 2025-09-29
[WIP] MON-4380: Add optional monitoring logic

rexagod:
Optional components include:
* Alertmanager
* AlertmanagerUWM
* ClusterMonitoringOperatorDeps (partially, for AM)
* MonitoringPlugin
* PromtheusOperator (partially, for AM)
* PromtheusOperatorUWM
* ThanosRuler

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1748,6 +1748,11 @@ func (c *Client) HasConsoleCapability(ctx context.Context) (bool, error) {
 	return c.HasClusterCapability(ctx, configv1.ClusterVersionCapabilityConsole)
 }
 
+func (c *Client) HasOptionalMonitoringCapability(_ context.Context) (bool, error) {
+	//return c.HasClusterCapability(ctx, configv1.ClusterVersionCapabilityOptionalMonitoring)
+	return false, nil
+}
+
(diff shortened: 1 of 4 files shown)


pkg/client/client.go
```diff
@@ -1748,6 +1748,11 @@ func (c *Client) HasConsoleCapability(ctx context.Context) (bool, error) {
 	return c.HasClusterCapability(ctx, configv1.ClusterVersionCapabilityConsole)
 }
 
+func (c *Client) HasOptionalMonitoringCapability(_ context.Context) (bool, error) {
+	//return c.HasClusterCapability(ctx, configv1.ClusterVersionCapabilityOptionalMonitoring)
```
rexagod, line 1752: Waiting on this.

rexagod:
Also, this needs logic to drop existing optional workloads from the cluster, when the capability is enabled.

simonpasquier (commented):
Speaking out loud but wouldn't it be possible for the operator to look at the `capability.openshift.io/name` annotation on each managed resource and skip the action if it's marked as `OptionalMonitoring` and the capability is actually disabled? This way we wouldn't have to change the tasks orchestration.

rexagod:
/close

Moving all raised comments and patches to #2675

PR #2691, opened 2025-10-02
NO-JIRA: clarify ExternalLabels doc

machine424:
cc @eromanova97 
backporting this to release-4.20 in this repo will take days, I don't think that's worth the effort/time.
But you can have this applied to the 4.20>= docs


pkg/manifests/types.go
```diff
@@ -193,7 +193,7 @@ type PrometheusK8sConfig struct {
 	// automatically calculated based on cluster capacity.
 	// The default value is empty, which indicates no limit.
 	EnforcedBodySizeLimit string `json:"enforcedBodySizeLimit,omitempty"`
-	// Defines labels to be added to any time series or alerts when
+	// A map[string]string that defines labels to be added to any time series or alerts when
```
danielmellado: nit: no strong opinion but I was thinking maybe about something like this to somehow keep track of the original short formatting. 

// Defines labels to be added to any time series or alerts when communicating with external systems.
// A map[string]string; by default, no labels are added.

wdyt?

machine424: we agreed on this with @eromanova97 
but I have no preference.
I'd keep the `systems such as federation, remote storage, and Alertmanager` part though

eromanova97: No strong opinions either 👍 It can stay as is, but if we want to keep that consistency, then I would apply the second suggestion

Documentation/api.md
```diff
@@ -424,7 +424,7 @@ The `PrometheusK8sConfig` resource defines settings for the Prometheus component
...
 | enforcedBodySizeLimit | string | Enforces a body size limit for Prometheus scraped metrics. If a scraped target's body response is larger than the limit, the scrape will fail. The following values are valid: an empty value to specify no limit, a numeric value in Prometheus size format (such as `64MB`), or the string `automatic`, which indicates that the limit will be automatically calculated based on cluster capacity. The default value is empty, which indicates no limit. |
-| externalLabels | ExternalLabels | Defines labels to be added to any time series or alerts when communicating with external systems such as federation, remote storage, and Alertmanager. By default, no labels are added. |
+| externalLabels | ExternalLabels | Defines labels to be added to any time series or alerts when communicating with external systems such as federation, remote storage, and Alertmanager. A map[string]string. By default, no labels are added. |
```
juzhao:
```suggestion
| externalLabels | ExternalLabels | Defines labels to be added to any time series or alerts when communicating with external systems such as federation, remote storage, and Alertmanager. The type is map[string]string. By default, no labels are added. |
```

pkg/manifests/types.go
```diff
@@ -659,7 +660,7 @@ type PrometheusRestrictedConfig struct {
 	// Defines labels to be added to any time series or alerts when
 	// communicating with external systems such as federation, remote storage,
 	// and Alertmanager.
-	// By default, no labels are added.
+	// A map[string]string. By default, no labels are added.
```
rexagod:
```suggestion
	// A map[string]string; by default, no labels are added.
```

Documentation/api.md
machine424: cc @danielmellado @eromanova97 wdty?
A Pandora box was opened (context https://github.com/openshift/cluster-monitoring-operator/pull/2691#discussion_r2398491814).

pkg/manifests/types.go
machine424: cc @danielmellado @eromanova97 @juzhao wdty?
A Pandora box was opened (context https://github.com/openshift/cluster-monitoring-operator/pull/2691#discussion_r2398491814).

machine424:
I'll probably just close this PR, and let you @eromanova97 decide and open another one :)

machine424:
or we can agree on the final wording so I don't have to change it three times.

juzhao:
> or we can agree on the final wording so I don't have to change it three times.

agree, I think we could let @eromanova97 decide on the format

eromanova97:
This is the wording that was peer reviewed from the docs side and used in the docs:
`The type is map[string]string. By default, no labels are added.`

machine424:
/override ci/prow/okd-scos-e2e-aws-ovn
only doc changes
/skip

machine424:
I meant
/override ci/prow/e2e-aws-ovn-techpreview

PR #2693, opened 2025-10-05
NO-ISSUE: set openshift-state-metrics as default container

lance5890:

As we use `kubectl logs -n openshift-monitoring openshift-state-metrics-xxx`,we see the default container logs is `kube-rbac-proxy-main`, like :
```
Defaulted container "kube-rbac-proxy-main" out of: kube-rbac-proxy-main, kube-rbac-proxy-self, openshift-state-metrics
I0824 20:06:46.135475       1 kube-rbac-proxy.go:530] Reading config file: /etc/kube-rbac-policy/config.yaml
I0824 20:06:46.137801       1 kube-rbac-proxy.go:233] Valid token audiences: 
I0824 20:06:46.138411       1 kube-rbac-proxy.go:347] Reading certificate files
I0824 20:06:46.138657       1 dynamic_cafile_content.go:157] "Starting controller" name="client-ca::/etc/tls/client/client-ca.crt"
I0824 20:06:46.139062       1 kube-rbac-proxy.go:395] Starting TCP socket on :8443
I0824 20:06:46.139651       1 kube-rbac-proxy.go:402] Listening securely on :8443
```

But most of the time we just care about the container `openshift-state-metrics` logs , not the`kube-rbac-proxy-main`

### DIFF
--- a/assets/openshift-state-metrics/deployment.yaml
+++ b/assets/openshift-state-metrics/deployment.yaml
@@ -17,6 +17,7 @@ spec:
   template:
     metadata:
       annotations:
+        kubectl.kubernetes.io/default-container: openshift-state-metrics
         openshift.io/required-scc: restricted-v2
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
--- a/jsonnet/components/openshift-state-metrics.libsonnet
+++ b/jsonnet/components/openshift-state-metrics.libsonnet
@@ -34,6 +34,7 @@ function(params) {
           } + cfg.commonLabels,
           annotations+: {
             'openshift.io/required-scc': 'restricted-v2',
+            'kubectl.kubernetes.io/default-container': 'openshift-state-metrics',
           },
         },
         spec+: {


simonpasquier:
thanks for your contribution! The YAML manifest is generated from jsonnet so you'd need to inject the annotation here and regenerate the code:

https://github.com/openshift/cluster-monitoring-operator/blob/fb823bae1e43811b0c8920b640dee2609550dde2/jsonnet/components/openshift-state-metrics.libsonnet#L24-L28

If it's too complicated for you, we can handle the change.

assets/openshift-state-metrics/deployment.yaml
```diff
@@ -17,6 +17,7 @@ spec:
   template:
     metadata:
       annotations:
+        kubectl.kubernetes.io/default-container: openshift-state-metrics
```
lance5890, line 20: regenerated @simonpasquier

simonpasquier:
/retitle NO-ISSUE: set openshift-state-metrics as default container

PR #2694, opened 2025-10-05
MON-4386,MON-4387: `telemetry` Collection Profile

rexagod:

***

#### Both of these changes need to be merged together. Additionally, [MON-4788](https://issues.redhat.com/browse/MON-4388) also needs to be merged with these (once it's pushed).

***

Marked as draft for now as this needs all non-in-cluster-monitoring-stack recording rules ported to monitors as well.

Not 100% sure right now without investigating more on this, but I believe we can keep the existing machinery, i.e., the whitelisted machinery in CMO and everything from that point on as is, as that would serve as (a) a safeguard to make sure we pipe out the same rules and (b) a check to make sure the metrics (rules) being collected match the rules being sent out. But this will be worked on later once we start adding tests for this in origin (after the aforementioned porting effort is done).


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Note: This CHANGELOG is only for the monitoring team to track all monitoring related changes. Please see OpenShift release notes for official changes.
 
+## 4.XX
```
rexagod: ~Since [adding external support](https://issues.redhat.com/browse/MON-4388) may take a while.~

rexagod:
I wanted to talk a bit regarding the `telemetry` collection profile's expected nature.

The `minimal` collection profile started off with introducing corresponding monitors for our components, which can be extended by external component owners to cast a wider net, and cut down on metric ingestion.

The `telemetry` collection profile veers off from this convention as it demands a complete set of telemetry monitors (to fulfill the exact set of metrics required by the telemetry rules) from the get-go. This has already been done for internal components in this draft, but doing so for external components is complicated, even when we have the [complete list of metrics](https://github.com/rexagod/cluster-monitoring-operator/blob/930d757e1b3243b3c4a2995ca42a1331d6501703/manifests/telemetry-metrics.md) now.

If CMO runs with remote-write capability (no telemeter-client), targets may be injected with the same relabellings that are done for the telemetry one. This may have performance implications since every target would have it's metrics filtered through the same relabelling rules' regexes, but it'd be more dynamic than keeping a record of monitor keys for the ones that are needed for telemetry, since that wouldn't be immune to breaking changes (renaming or migration). However, as it is right now we'd need an approach that not only covers this but also the legacy (federated) telemeter-client-based approach, since users should also see the cut-down in metric volumes in not just their LTS storage systems, but also the PVs hooked up for data retention, which brings us back to constraining at the monitor level.

A more manageable and fitting way would be to adhere to the same pattern as the `minimal` collection profile, and start constraining metrics within the monitoring stack, i.e., with the monitors we control. Teams would be encouraged (in the handbook documentation, and maybe an AI-based check on the PR) to curate monitors on their end with a specific annotation so eventually the telemetry metric volume comes closer to what the telemetry rules really need. This wouldn't be subject to breaking changes, and happen at its own pace in the wider OpenShift ecosystem.

@simonpasquier Curious to hear your thoughts on this. I believe the last passage makes the most sense going forward for this collection profile, but PLMK if I missed something.

rexagod:
⬆️ This was resolved over Slack. We'll move forward with the same approach as we did for the `minimal` profile.

rexagod:
Need to test a couple of things on 4.22 build that has https://github.com/openshift/cluster-monitoring-operator/pull/2769.

rexagod:
Ah, this doesn't seem to have a final commit from my local (actually deploying telemetry assets, etc.). I'll move this to draft in the meantime, test it on the aforementioned build, and push the changes.

jsonnet/utils/generate-service-monitors.libsonnet
```diff
@@ -2,13 +2,16 @@
...
+  // 1. Add the profile prefix to the ServiceMonitor name
+  // 2. Add the profile label "monitoring.openshift.io/collection-profile: <profile>"
   // 3. Add a metricRelabelings with action keep and regex equal to metrics
-  local minimal(sm, metrics) = sm {
+  local run(sm, metrics, label) = sm {
     metadata+: {
-      name+: '-minimal',
-      labels+: minimalLabel,
+      name+: '-' + label['monitoring.openshift.io/collection-profile'],
+      labels+: label,
```
simonpasquier: (nit)
```suggestion
  local run(sm, metrics, profile) = sm {
metadata+: {
      name+: '-' + profile,
      labels+: {
          'monitoring.openshift.io/collection-profile': profile,
       },
```

pkg/tasks/telemeter.go
```diff
@@ -187,14 +187,16 @@ func (t *TelemeterClientTask) create(ctx context.Context) error {
...
 		return fmt.Errorf("initializing Telemeter client ServiceMonitor failed: %w", err)
 	}
 
-	err = t.client.CreateOrUpdateServiceMonitor(ctx, sm)
-	if err != nil {
-		return fmt.Errorf("reconciling Telemeter client ServiceMonitor failed: %w", err)
+	for _, sm := range sms {
+		err = t.client.CreateOrUpdateServiceMonitor(ctx, sm)
+		if err != nil {
+			return fmt.Errorf("reconciling Telemeter client ServiceMonitor failed: %w", err)
```
simonpasquier: (nit) should we print the name of the resource?

pkg/manifests/manifests.go
```diff
@@ -430,10 +438,22 @@ func (f *Factory) AlertmanagerUserWorkloadClusterRole() (*rbacv1.ClusterRole, er
 	return f.NewClusterRole(f.assets.MustNewAssetSlice(AlertmanagerUserWorkloadClusterRole))
 }
 
+func (f *Factory) AlertmanagerServiceMonitors() ([]*monv1.ServiceMonitor, error) {
+	return serviceMonitors(
+		f.AlertmanagerServiceMonitor,
+		nil,
```
simonpasquier, line 450: is it expected that we have a nil servicemonitor for the minimal profile?

jsonnet/components/cluster-monitoring-operator.libsonnet
```diff
@@ -159,6 +160,12 @@ function(params) {
     },
   },
 
+  telemetryServiceMonitor: generateServiceMonitor.telemetry(
+    self.serviceMonitor, std.join('|', [
+      'cluster_monitoring_operator_collection_profile',
```
simonpasquier: how do we ensure that the we keep the list up-to-date?

simonpasquier (commented):
Good work! I'm still worried that the metrics list get out of date compared to what telemetry expects. What's the plan to avoid the drift?

jsonnet/components/cluster-monitoring-operator.libsonnet
rexagod: The following rules from the whitelist are currently reliant on us.

```
  '{__name__="cluster:cpu_usage_cores:sum"}'
  '{__name__="ALERTS",alertstate="firing",severity=~"critical|warning|info|none"}'
  '{__name__="cluster:alertmanager_integrations:max"}'
  '{__name__="cluster:capacity_cpu_cores:sum"}'
  '{__name__="cluster:capacity_memory_bytes:sum"}'
  '{__name__="cluster:memory_usage_bytes:sum"}'
  '{__name__="cluster:node_instance_type_count:sum"}'
  '{__name__="cluster:telemetry_selected_series:count"}'
  '{__name__="cluster:virt_platform_nodes:sum"}'
  '{__name__="count:up0"}'
  '{__name__="count:up1"}'
  '{__name__="monitoring:container_memory_working_set_bytes:sum"}'
  '{__name__="namespace_job:scrape_samples_post_metric_relabeling:topk3"}'
  '{__name__="namespace_job:scrape_series_added:topk3_sum1h"}'
  '{__name__="node_role_os_version_machine:cpu_capacity_cores:sum"}'
  '{__name__="node_role_os_version_machine:cpu_capacity_sockets:sum"}'
  '{__name__="openshift:cpu_usage_cores:sum"}'
  '{__name__="openshift:memory_usage_bytes:sum"}'
  '{__name__="openshift:prometheus_tsdb_head_samples_appended_total:sum"}'
  '{__name__="openshift:prometheus_tsdb_head_series:sum"}'
  '{__name__="profile:cluster_monitoring_operator_collection_profile:max"}'
  '{__name__="vendor_model:node_accelerator_cards:sum",vendor=~"NVIDIA|AMD|GAUDI|INTEL|QUALCOMM|Marvell|Mellanox"}'
  '{__name__="workload:cpu_usage_cores:sum"}'
  '{__name__="workload:memory_usage_bytes:sum"}'
  '{__name__=~"cluster:usage:.*"}'
```

I'm thinking of marking these rules in the telemetry ConfigMap (that we use as the source of truth for the test introduced in https://github.com/openshift/origin/pull/30643). We'll extract and query these to ensure they are working. If not, it would likely mean that the SMs are outdated and not updated with relevant metrics.

We'll document this in the handbook so teams can mark rules they introduce when any of them rely on one of our components, as well as update the `jsonnet` code. However, this does make reviewing a bit more involved for telemetry rule addition PRs.

rexagod: The variance-driven test that I have in the aforementioned PR is more of a good measure. I believe the approach above will be more robust.

rexagod: IMHO the test we discussed above makes more sense in CMO, since we need to run this on every PR that touches the involved assets (whitelist, SMs, etc.), and IIUC there's no way to configure this in Origin directly. I could house this under extended tests so we can track this in Sippy, and link this to release payloads.

EDIT: The variance and other telemetry tests seem fine for origin. I've added the "check for metrics used in our telemetry rules" test in CMO.

jan--f: Maybe its worth thinking about a central authoritative list of telemetry metrics that includes metadata, e.g the owning team and service monitor. We could then generate the various artifacts from this single source (our own allowlists and SM configs, Telemeter server allow list). That could also become useful as an index for when data is accessed (as data do we have from which source).

rexagod: That makes a lot of sense, I'll whip something up.

rexagod:
Please note that the fixup commit (https://github.com/openshift/cluster-monitoring-operator/pull/2694/commits/c78bc38c8b5f723d8ab3848750dab34eeeed288e) includes all generated data, the one before that has the actual changes. PLMK if there's anything else I can do here to make reviewing easier.

pkg/manifests/manifests.go
rexagod: I'd say so. Currently AM has no metrics that are applied under `minimal` circumstances. However, certain telemetry rules do rely on some of it's metrics so we have one for that profile.

rexagod:
- [x] Add test that queries ALL telemetry rules when the profile is applied, to make sure the e2e behavior is preserved.

EDIT: While writing the telemetry test on the origin side, some rules weren't available, likely due to certain operators not being deployed. So instead of testing rules, I'm thinking of testing the metrics that they depend on instead.

test/e2e/telemeter_test.go
```diff
@@ -18,13 +18,34 @@ import (
...
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	k8syaml "sigs.k8s.io/yaml"
 )
 
+// Some metrics may only be conditionally present depending on the environment, or the configuration.
+// For such metics, we skip the presence check.
+var ignoredMetrics = map[string]struct{}{
+	"kube_pod_restart_policy":                                       {}, // Disabled through --metric-denylist.
```
rexagod, line 39: `cluster:usage:pods:terminal:workload:sum` uses this metric, and currently is specified in the [whitelist](https://github.com/rexagod/cluster-monitoring-operator/blob/be9fa0a82919ec12ac985fcebc323a191d854cb0/manifests/0000_50_cluster-monitoring-operator_04-config.yaml#L340-L353).

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -1,1281 +1,2777 @@
...
-    #
-    # cluster:mtv_migrations_status_total:sum is the total number of VM migrations running on the cluster,
-    # labeled with {status}, {provider}, {mode}, and {target}.
-    # Expected labels:
-    # - provider: "ova", "vsphere", "openstack", "openshift", "ovirt" or "awsec2"
-    # - target: "Local" or "Remote"
-    # - mode: "Cold", "Warm" or "RCM"
-    # - status: "Suceeded", "Failed" or "Canceled"
-    - '{__name__="cluster:mtv_migrations_status_total:sum", provider=~"ova|vsphere|openstack|openshift|ovirt|awsec2", target=~"Local|Remote", mode=~"Cold|Warm|RCM", status=~"Succeeded|Failed|Canceled"}'
+        #
```
rexagod: extraneous tabs, will drop

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
rexagod, line 1: Comparing the rules here with the initial commit in this PR (which had unchanged whitelist from main at that time), there is no difference between the rules exposed (only their metadata).

rexagod:
Rebase errors are in the generated docs, and shouldn't block reviews. I'll add the newer telemetry metrics in the jsonnet file and generate once this is good to go.

pkg/manifests/manifests.go
simonpasquier: Hmm we do have Alertmanager's rules which would require metrics when the minimal profile is enabled? IIUC we haven't implemented the minimal profile for Alertmanager but we should before proceeding here, no?

simonpasquier: Also if we have a telemetry profile, there's no way that we don't have a minimal profile since minimal is a superset of telemetry.

pkg/manifests/manifests.go
```diff
@@ -752,7 +772,11 @@ func (f *Factory) KubeStateMetricsClusterRole() (*rbacv1.ClusterRole, error) {
 }
 
 func (f *Factory) KubeStateMetricsServiceMonitors() ([]*monv1.ServiceMonitor, error) {
-	return serviceMonitors(f.config.CollectionProfilesFeatureGateEnabled, f.KubeStateMetricsServiceMonitor, f.KubeStateMetricsMinimalServiceMonitor)
```
simonpasquier, line 765: IIUC this is due to the fact that `CollectionProfilesFeatureGateEnabled` is always enabled in 4.22? Can we have this change in a separate PR for better isolation?

jsonnet/utils/telemetry-allowlist-and-monitors-entries.libsonnet
simonpasquier, line 1: IMHO the switch from a flat YAML file for the telemetry allow-list to jsonnet should be its own PR so we can discuss the implications (developer documentation/education, import on backports, etc.) separately.

simonpasquier (commented):
I commented already on the need of implementing the Alertmanager's minimal profile before but it applies to all components that send metrics to telemetry.

test/e2e/telemeter_test.go
```diff
@@ -18,13 +18,34 @@ import (
...
+	"github.com/openshift/cluster-monitoring-operator/pkg/manifests"
+	"github.com/openshift/cluster-monitoring-operator/test/e2e/framework"
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	k8syaml "sigs.k8s.io/yaml"
 )
 
+// Some metrics may only be conditionally present depending on the environment, or the configuration.
+// For such metics, we skip the presence check.
```
juzhao:
```suggestion
// For such metrics, we skip the presence check.
```

jsonnet/utils/telemetry-allowlist-and-monitors.libsonnet
```diff
@@ -0,0 +1,205 @@
...
+// associated with in-cluster components.
+//
+// Monitors for components external to what's housed under this repository are
+// not generated, but that information is still maintained.
+local o = import './telemetry-allowlist-and-monitors-entries.libsonnet';
+local entries = o.entries;
+local cmoMonitors = o.cmoMonitors;
+
+// Each entry in the whitelist will continue to adhere to the existing pattern,
+// with he exception of marker comments, i.e.,
```
juzhao:
```suggestion
// with the exception of marker comments, i.e.,
```

jsonnet/utils/telemetry-allowlist-and-monitors.libsonnet
```diff
@@ -0,0 +1,205 @@
+// This file acts as the single source of truth for the telemetry whitelist as
+// well as the associated monitors for the telemetry collection profile. Before
+// adding a telemery entry, please read the following notes:
+// * Teams are advised to add entries for telemetry (as they did ealier in
+// manifests/0000_50_cluster-monitoring-operator_04-config.yaml), here instead.
+// * Each entry must have the following fields, the absence of which will cause
+// generation to fail:
+//   * metadata: The metadata associated with the rule. This consists of:
+//     * owners: The entities that owns this rule.
```
juzhao:
```suggestion
//     * owners: The entities that own this rule.
```

jsonnet/utils/telemetry-allowlist-and-monitors.libsonnet
```diff
@@ -0,0 +1,205 @@
+// This file acts as the single source of truth for the telemetry whitelist as
+// well as the associated monitors for the telemetry collection profile. Before
+// adding a telemery entry, please read the following notes:
+// * Teams are advised to add entries for telemetry (as they did ealier in
```
juzhao:
```suggestion
// * Teams are advised to add entries for telemetry (as they did earlier in
```

jsonnet/utils/telemetry-allowlist-and-monitors.libsonnet
```diff
@@ -0,0 +1,205 @@
+// This file acts as the single source of truth for the telemetry whitelist as
+// well as the associated monitors for the telemetry collection profile. Before
+// adding a telemery entry, please read the following notes:
```
juzhao:
```suggestion
// adding a telemetry entry, please read the following notes:
```

pkg/manifests/manifests.go
rexagod: Right. It seems we've identified a gap in the `minimal` implementation.

The `minimal` implementation had and has no way to update itself based on the telemetry metrics being added over time. I'll make some jsonnet changes here to merge all `telemetry` profile metrics into `minimal`, so the former is truly a subset of the latter, and we do not miss any metrics as they get added.

pkg/manifests/manifests.go
rexagod: Sure thing, will do.

jsonnet/utils/telemetry-allowlist-and-monitors-entries.libsonnet
rexagod: I'll create a seperate PR.

rexagod:
Thank you for the typo check, @juzhao! I've recently got decent at blind typing but sometimes I'd misstype and Vim would end up editing stuff it wasn't supposed to 😓

rexagod:
I've added the pending TODOs in the PR description.

rexagod:
@simonpasquier Regarding https://github.com/openshift/cluster-monitoring-operator/pull/2694#discussion_r2773330802, IIUC besides the feature-gate removal, and [one additional Jsonnet fix](https://github.com/openshift/cluster-monitoring-operator/pull/2694/changes/e87b3f41f412cf92e6fd0a96912e74cbcae66873), this PR ends up only reflecting changes coming from the static YAML to Jsonnet based whitelist generation, so a separate PR may not be needed? PLMK if I'm missing something and I'll go ahead and create one anyway.

I've raised a PR to isolate the feature-gate removal, and the change mentioned in your review should go away once that's merged, see https://github.com/openshift/cluster-monitoring-operator/pull/2812/changes#diff-1f9d2d18d4cabd8262e585b932330183f7cb9b544ac9f27c8bcb3ef9c8d14376L897. I've also rebased that PR under this one (the first commit here coming from that PR).

If this makes sense the only thing left for the effort after both of the PRs are in would be to ensure that the minimal profile is a true superset of all metrics in its corresponding telemetry profile (merge latter and former during their Jsonnet build). I'm planning to do that in a separate PR following this one, as that is not strictly related to the YAML-Jsonnet whitelist generation being done here, for better isolation of concerns.

test/e2e/telemeter_test.go
```diff
@@ -18,13 +18,34 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	k8syaml "sigs.k8s.io/yaml"
 )
 
+// Some metrics may only be conditionally present depending on the environment, or the configuration.
+// For such metrics, we skip the presence check.
+var ignoredMetrics = map[string]struct{}{
+	"kube_pod_restart_policy":                                       {}, // Disabled through --metric-denylist.
+	"kubelet_volume_stats_used_bytes":                               {},
```
juzhao, line 40: `kubelet_volume_stats_used_bytes` is used in recording rule `cluster:kubelet_volume_stats_used_bytes:provisioner:sum`, 
```
      - expr: |
          sum  by (provisioner) (
            topk by (namespace, persistentvolumeclaim) (
              1, kubelet_volume_stats_used_bytes
            ) * on (namespace,persistentvolumeclaim) group_right()
            topk by (namespace, persistentvolumeclaim) (
              1, kube_persistentvolumeclaim_info * on(storageclass) group_left(provisioner) topk by(storageclass) (1, max by(storageclass, provisioner) (kube_storageclass_info))
            )
          )
```
the recording rule is in the whitelist

rexagod: Every metric ignored here is used in the whitelist, however, they are not exposed OOTB in a fresh cluster without additional configuration.

Please let me know if you find otherwise.

pkg/manifests/manifests.go
simonpasquier: For a component that needs to support the `telemetry` profile, it should also support the `minimal` profile which would be a copy of the `full` profile if we don't have time to carve the correct `minimal` ServiceMonitor.

simonpasquier: For clarity it might be better to add first the `minimal` ServiceMonitors for all components for which we know a `telemetry` profile will be needed. WDYT?

jan--f:
> @simonpasquier Regarding [#2694 (comment)](https://github.com/openshift/cluster-monitoring-operator/pull/2694#discussion_r2773330802), IIUC besides the feature-gate removal, and [one additional Jsonnet fix](https://github.com/openshift/cluster-monitoring-operator/pull/2694/changes/e87b3f41f412cf92e6fd0a96912e74cbcae66873), this PR ends up only reflecting changes coming from the static YAML to Jsonnet based whitelist generation, so a separate PR may not be needed? PLMK if I'm missing something and I'll go ahead and create one anyway.

I think the idea would be to have only the static to jsonnet changes in its own PR to make review a bit easier (and reflect this purpose of the PR accordingly). I agree its worth a discussion and that shouldn't block the progress of the telemetry profile.

pkg/manifests/manifests.go
rexagod: Right. The following components need a `telemetry` profile (based on the rules in the whitelist), which was added here:

* Alertmanager
* CMO
* Control Plane
* KSM
* Node Exporter
* OSM
* Prometheus
* Telemeter Client

I believe all of these have a revised `minimal` profile either added or updated in https://github.com/openshift/cluster-monitoring-operator/pull/2814. PLMK if I'm missing anything here.

simonpasquier:
I believe that this PR can be closed?

rexagod:
/close

I'll open a new PR to salvage the whitelist generation from here.

PR #2734, opened 2025-11-03
OCPBUGS-85115, MON-4420: Gateway API telemetry

rikatz:

This PR adds the following changes required for Gateway API telemetry collection:
* Enables the collection of Gateway API resources (Gateway Class and Gateway) using kube-state-metrics from `openshift-monitoring` namespace
* Adds a PrometheusRule on the same namespace (as the collection comes from ksm) to aggregate the metrics and reduce labels.

### DIFF
--- a/assets/kube-state-metrics/custom-resource-state-configmap.yaml
+++ b/assets/kube-state-metrics/custom-resource-state-configmap.yaml
@@ -507,14 +507,6 @@ data:
                 "gateway_class":
                 - "metadata"
                 - "name"
-                "namespace":
-                - "metadata"
-                - "namespace"
-                "reason":
-                - "status"
-                - "conditions"
-                - "[type=Accepted]"
-                - "reason"
             "type": "Info"
           "help": "Information about GatewayClasses"
           "name": "gateway_class"
@@ -540,11 +532,6 @@ data:
                 - "conditions"
                 - "[type=Programmed]"
                 - "status"
-                "reason":
-                - "status"
-                - "conditions"
-                - "[type=Programmed]"
-                - "reason"
             "type": "Info"
           "help": "Information about Gateways"
           "name": "gateway"
--- a/assets/kube-state-metrics/custom-resource-state-configmap.yaml
+++ b/assets/kube-state-metrics/custom-resource-state-configmap.yaml
@@ -501,7 +501,7 @@ data:
                 - "conditions"
                 - "[type=Accepted]"
                 - "status"
-                "controller_name":
+                "controller":
                 - "spec"
                 - "controllerName"
                 "gateway_class":
@@ -509,7 +509,7 @@ data:
                 - "name"
             "type": "Info"
           "help": "Information about GatewayClasses"
-          "name": "gateway_class"
+          "name": "gateway_class_info"
       - "groupVersionKind":
           "group": "gateway.networking.k8s.io"
           "kind": "Gateway"
@@ -521,7 +521,7 @@ data:
                 "gateway":
                 - "metadata"
                 - "name"
-                "gateway_class_name":
+                "gateway_class":
                 - "spec"
                 - "gatewayClassName"
                 "namespace":
@@ -534,7 +534,7 @@ data:
                 - "status"
             "type": "Info"
           "help": "Information about Gateways"
-          "name": "gateway"
+          "name": "gateway_info"
 kind: ConfigMap
 metadata:
   labels:
--- a/assets/kube-state-metrics/custom-resource-state-configmap.yaml
+++ b/assets/kube-state-metrics/custom-resource-state-configmap.yaml
@@ -488,6 +488,66 @@ data:
             - "metadata"
             - "name"
           "name": "verticalpodautoscaler_spec_resourcepolicy_container_policies_maxallowed_memory"
+      - "groupVersionKind":
+          "group": "gateway.networking.k8s.io"
+          "kind": "GatewayClass"
+          "version": "v1"
+        "metrics":
+        - "each":
+            "info":
+              "labelsFromPath":
+                "accepted":
+                - "status"
+                - "conditions"
+                - "[type=Accepted]"
+                - "status"
+                "controller_name":
+                - "spec"
+                - "controllerName"
+                "gateway_class":
+                - "metadata"
+                - "name"
+                "namespace":
+                - "metadata"
+                - "namespace"
+                "reason":
+                - "status"
+                - "conditions"
+                - "[type=Accepted]"
+                - "reason"
+            "type": "Info"
+          "help": "Information about GatewayClasses"
+          "name": "gateway_class"
+      - "groupVersionKind":
+          "group": "gateway.networking.k8s.io"
+          "kind": "Gateway"
+          "version": "v1"
+        "metrics":
+        - "each":
+            "info":
+              "labelsFromPath":
+                "gateway":
+                - "metadata"
+                - "name"
+                "gateway_class_name":
+                - "spec"
+                - "gatewayClassName"
+                "namespace":
+                - "metadata"
+                - "namespace"
+                "programmed":
+                - "status"
+                - "conditions"
+                - "[type=Programmed]"
+                - "status"
+                "reason":
+                - "status"
+                - "conditions"
+                - "[type=Programmed]"
+                - "reason"
+            "type": "Info"
+          "help": "Information about Gateways"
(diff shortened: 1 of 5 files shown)


juzhao:
jsonnet-fmt job failed, you can use jsonnetfmt or other jsonnet CLI to formate libsonnet files

assets/cluster-monitoring-operator/prometheus-rule.yaml
```diff
@@ -373,6 +373,33 @@ spec:
...
+            label_replace(
+              (
+                kube_customresource_gateway{programmed="True"}
+                and on(gateway_class_name) label_replace(kube_customresource_gateway_class{accepted="True", controller_name!="openshift.io/gateway-controller/v1"}, "gateway_class_name", "$1", "gateway_class", "(.+)")
+              ),
+              "gateway_class_type", "not-openshift", "", ""
+            )
+          )
+        )
+      record: openshift:gateway_api_usage:count
```
juzhao: I think it's better to  add the rule under your namespace

simonpasquier: 👍

simonpasquier:
Can we have a Jira ticket for the telemetry addition?
https://rhobs-handbook.netlify.app/products/openshiftmonitoring/telemetry.md/#request-approval

simonpasquier:
/retitle [MON-4420](https://issues.redhat.com//browse/MON-4420): Gateway API telemetry

rikatz:
@simonpasquier just confirmong, [MON-4420](https://issues.redhat.com//browse/MON-4420) is enough as an approval request?

rikatz:
I will also separate the PRs here, adding just @danwinship commit (+ a change I did to remove the namespace from Gateway Class) and move the Prometheus rule to our repo.

assets/cluster-monitoring-operator/prometheus-rule.yaml
rikatz: removed, will do on openshift-ingress-operator namespace

simonpasquier:
/cc @rexagod 

for expert :eyes: on the kube-state-metrics config

rikatz:
Executed a test with clusterbot here:
- Created a cluster using this PR
- Applied the PrometheusRule of https://github.com/openshift/cluster-ingress-operator/pull/1300
- Created some GatewayClass and Gateway
after some time I can see the `openshift:gateway_api_usage:count` metric being created properly

simonpasquier (commented):
Looking at the Prometheus data captured from https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2734/pull-ci-openshift-cluster-monitoring-operator-main-e2e-aws-ovn/1985715323818479616, I'm not sure why the `kube_customresource_gateway_class` series have a smaller lifetime than the `kube_customresource_gateway` series?

Image

Image

rikatz:
@simonpasquier I may be wrong, but they seem two different metrics (at least 2 different cardinalities, as one has accepted=Unknown and the other accepted=True

Would it be some e2e test of Gateway API running at this same time, and maybe erasing the class after the test?

assets/kube-state-metrics/custom-resource-state-configmap.yaml
```diff
@@ -488,6 +488,53 @@ data:
...
+                - "status"
+                "controller_name":
+                - "spec"
+                - "controllerName"
+                "gateway_class":
+                - "metadata"
+                - "name"
+            "type": "Info"
+          "help": "Information about GatewayClasses"
+          "name": "gateway_class"
```
rexagod:
```suggestion
          "name": "gateway_class_info"
```

Needs to be [suffixed](https://github.com/kubernetes/kube-state-metrics/blob/807f8bd184ed4325c1a04214039bb0bd51e1878e/pkg/customresourcestate/registry_factory.go#L76) with an [`_info`](https://prometheus.io/docs/practices/naming/#:~:text=foobar_build_info%20(for%20a%20pseudo%2Dmetric%20that%20provides%20metadata%20about%20the%20running%20binary)).

assets/kube-state-metrics/custom-resource-state-configmap.yaml
```diff
@@ -488,6 +488,53 @@ data:
...
+                - "metadata"
+                - "namespace"
+                "programmed":
+                - "status"
+                - "conditions"
+                - "[type=Programmed]"
+                - "status"
+            "type": "Info"
+          "help": "Information about Gateways"
+          "name": "gateway"
```
rexagod: Same here.

rexagod (commented):
I'm seeing [errors](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2734/pull-ci-openshift-cluster-monitoring-operator-main-e2e-aws-ovn/1985715323818479616/artifacts/e2e-aws-ovn/gather-extra/artifacts/pods/openshift-ingress_gateway-jklf6-openshift-default-7cd9b4c668-sxf4z_istio-proxy.log) in one of the associated pods, which is `killed` around the same time the `gateway_class` metric goes down:

```
17:07:30	openshift-ingress	kubelet	gateway-xv69v-openshift-default-588b5ff6fb-qbpdc	
Killing

Stopping container istio-proxy
```

assets/kube-state-metrics/custom-resource-state-configmap.yaml
```diff
@@ -488,6 +488,53 @@ data:
...
+        "metrics":
+        - "each":
+            "info":
+              "labelsFromPath":
+                "gateway":
+                - "metadata"
+                - "name"
+                "gateway_class_name":
+                - "spec"
+                - "gatewayClassName"
```
rexagod: (nit) I get that this has an additional `_name` suffix in contrast to the `gateway_class` label name above, due to it stemming from `gatewayClassName`, but since label renaming is always a breaking change, I wanted to raise the idea if it'd be better to keep both label names consistent.

rexagod: Going by KSM convention, we usually assign `_name` to non-workloads (like service, [port](https://github.com/kubernetes/kube-state-metrics/blob/807f8bd184ed4325c1a04214039bb0bd51e1878e/internal/store/endpoint.go#L134), etc.) whereas native workloads are kept [without the suffix](https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/persistentvolumeclaim.go#L40) in the labelset.

assets/kube-state-metrics/custom-resource-state-configmap.yaml
rikatz: doing. I assume the same applies for gateway_info?

assets/kube-state-metrics/custom-resource-state-configmap.yaml
rikatz: answered here :D

assets/kube-state-metrics/custom-resource-state-configmap.yaml
rikatz: hum makes sense. 

So I am removing the _name from both this label and https://github.com/openshift/cluster-monitoring-operator/pull/2734/files#diff-c818ed42e1a55cb4f22e6878f5c7fea95af6fb2237db4d4eefdbb612e795b96bR504

rikatz:
Failing because of https://github.com/openshift/cluster-monitoring-operator/pull/2736

rexagod:
/approved
/test ginkgo-tests
/test versions

The [anomaly](https://github.com/openshift/cluster-monitoring-operator/pull/2734#pullrequestreview-3432480273) above is most likely an Ingress issue, so we'd be okay moving forward here. However, please note that the same issue could interfere with the overall `gateway` metrics' exposition being introduced here and hinder observability for its consumers, so it'd make sense (for the Ingress team) to triage the issue down the line.

rikatz:
Created OCPBUGS-65775 for further verification of the proxy being killed during tests

rikatz:
/retitle OCPBUGS-85115,[MON-4420](https://redhat.atlassian.net/browse/MON-4420): Gateway API telemetry

rikatz:
/retitle OCPBUGS-85115, [MON-4420](https://redhat.atlassian.net/browse/MON-4420): Gateway API telemetry

PR #2814, opened 2026-02-10
MON-4517: Minimal and telemetry CP monitors

rexagod:
Bases minimal monitors on the same specs as the "full" ones.
***
This is rebased over #2821

## Summary by CodeRabbit

* **New Features**
  * Added a new "telemetry" collection profile and telemetry-specific ServiceMonitors across core monitoring components (Alertmanager, Prometheus, kube-state-metrics, node-exporter, telemeter-client, control-plane, openshift-state-metrics).
  * Introduced "minimal" collection-profile ServiceMonitors for many components to reduce metric volume and provide TLS-authenticated scrape endpoints.

* **Tests**
  * Updated collection-profile tests to include the new telemetry profile in selection logic.


simonpasquier:
> Contractually, [minimal profile is supposed to support telemetry operations as well](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/metrics-collection-profiles.md?plain=1#L191-L192). This makes sure that minimal monitors always
> have the expected set of telemetry metrics as well.

IMHO it should be the other way around: unless defined more precisely, minimal should be the same as full (because of dashboards and alerting rules).

jsonnet/utils/generate-service-monitors.libsonnet
```diff
@@ -6,7 +6,8 @@
   // 2. Add the profile prefix to the ServiceMonitor name
   // 3. Add the profile label "monitoring.openshift.io/collection-profile: <profile>"
   // 4. Add a metricRelabelings with action keep and regex equal to metrics
-  local run(sm, metrics, profile) =
+  local run(smWithDrop, metrics, profile, shouldRemoveDrop) =
+    local sm = if shouldRemoveDrop then removeDrop(smWithDrop) else smWithDrop;
```
rexagod:
```
~/bases/work/cluster-monitoring-operator/assets MON-4517 ≡
❯ for i in alertmanager cluster-monitoring-operator openshift-state-metrics prometheus-k8s telemeter-client; do diff $i/minimal-service-monitor.yaml $i/service-monitor.yaml; done
11,12c11
<     monitoring.openshift.io/collection-profile: minimal
<   name: alertmanager-main-minimal
---
>   name: alertmanager-main
8,9c8
<     monitoring.openshift.io/collection-profile: minimal
<   name: cluster-monitoring-operator-minimal
---
>   name: cluster-monitoring-operator
8,9c8
<     monitoring.openshift.io/collection-profile: minimal
<   name: openshift-state-metrics-minimal
---
>   name: openshift-state-metrics
11,12c11
<     monitoring.openshift.io/collection-profile: minimal
<   name: prometheus-k8s-minimal
---
>   name: prometheus-k8s
8,9c8
<     monitoring.openshift.io/collection-profile: minimal
<   name: telemeter-client-minimal
---
>   name: telemeter-client
```

jsonnet/utils/generate-service-monitors.libsonnet
```diff
@@ -6,7 +6,8 @@
   // 2. Add the profile prefix to the ServiceMonitor name
   // 3. Add the profile label "monitoring.openshift.io/collection-profile: <profile>"
   // 4. Add a metricRelabelings with action keep and regex equal to metrics
```
danielmellado: I'd replace this with something such as "// 4. If metrics is non-null, add a metricRelabelings with action keep and regex equal to metrics" as the logic somehow changed

jsonnet/utils/generate-service-monitors.libsonnet
danielmellado: was this a diff of an output for the libsonnet generation? xD

rexagod: Ah sorry. I pasted this here to show that nothing was changed in the newly generated minimal monitors from their full counterparts except names and the CP label.

jsonnet/utils/generate-service-monitors.libsonnet
```diff
@@ -39,5 +61,6 @@
     },
   },
 
-  minimal(sm, metrics): minimal(removeDrop(sm), metrics),
+  minimal(sm, metrics, removeDrop=true): run(sm, metrics, profiles[0], removeDrop),
```
simonpasquier: (nit) I feel that adding more arguments makes it harder to understand what happens at the caller site. Can we have a more explicit API?

````
// Returns a copy of the input service monitor with the correct annotation.
serviceMonitorForMinimalProfile(sm): ...
serviceMonitorForTelemetryProfile(sm): ...


// Returns a copy of the input service monitor with the list of metrics.
keepOnlyMetrics(sm, metrics): ... 
````

so we would call it like:

````
utils.serviceMonitorForTelemetryProfile(
    utils.keepOnlyMetrics(sm, ["foo", "bar"]),
)
````

or

````
// just use the full service monitor (no customization for minimal).
utils.serviceMonitorForMinimalProfile(sm)
````

File: pkg/tasks/alertmanager.go [hunk 4]
```diff
@@ -223,14 +223,16 @@ func (t *AlertmanagerTask) create(ctx context.Context) error {
...
+	smams, err := t.factory.AlertmanagerServiceMonitors()
 	if err != nil {
 		return fmt.Errorf("initializing Alertmanager ServiceMonitor failed: %w", err)
 	}
 
-	err = t.client.CreateOrUpdateServiceMonitor(ctx, smam)
-	if err != nil {
-		return fmt.Errorf("reconciling Alertmanager ServiceMonitor failed: %w", err)
+	for _, smam := range smams {
+		err = t.client.CreateOrUpdateServiceMonitor(ctx, smam)
```
simonpasquier: should we have a CreateOrUpdateServiceMonitors() function to reduce duplication?

assets/prometheus-k8s/telemetry-service-monitor.yaml
```diff
@@ -0,0 +1,35 @@
...
+    monitoring.openshift.io/collection-profile: telemetry
+  name: prometheus-k8s-telemetry
+  namespace: openshift-monitoring
+spec:
+  endpoints:
+  - bearerTokenFile: ""
+    interval: 30s
+    metricRelabelings:
+    - action: keep
+      regex: (ALERTS|prometheus_tsdb_head_samples_appended_total|prometheus_tsdb_head_series|scrape_samples_post_metric_relabeling|scrape_series_added|up)
```
simonpasquier: (nit) up and scrape_* metrics aren't coming from the scraped target.

assets/control-plane/telemetry-service-monitor-kubelet.yaml
```diff
@@ -0,0 +1,119 @@
...
+      sourceLabels:
+      - endpoint
+      targetLabel: endpoint
+    - action: replace
+      replacement: crio
+      targetLabel: job
+    scheme: https
+    tlsConfig:
+      caFile: /etc/prometheus/configmaps/kubelet-serving-ca-bundle/ca-bundle.crt
+      insecureSkipVerify: false
```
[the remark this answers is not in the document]
rexagod: I think this should be addressed in a separate PR.

assets/cluster-monitoring-operator/telemetry-service-monitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - action: keep
+      regex: (cluster_monitoring_operator_collection_profile)
+      sourceLabels:
+      - __name__
+    port: https
+    scheme: https
+    tlsConfig:
+      insecureSkipVerify: false
+      serverName: cluster-monitoring-operator.openshift-monitoring.svc
+  scrapeClass: tls-client-certificate-auth
```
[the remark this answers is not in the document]
rexagod: It'd make sense doing this in a separate PR, as this is pulled from CMO's default SM, which is not affected by this patch.

jsonnet/utils/generate-service-monitors.libsonnet
```diff
@@ -39,5 +15,56 @@
...
+        name: baseName + '-' + profile,
+        labels+: {
+          'monitoring.openshift.io/collection-profile': profile,
+        },
+      },
+    },
+
+  // Returns a copy of the input service monitor with the provided list of metrics.
+  // The metrics parameter is an array of metric names (e.g., ["metric1", "metric2", "metric3"]).
+  // This function removes existing "drop" metricRelabelings before adding the keep filter.
```
simonpasquier: (nit) can we explicit the motivation for removing the drop action(s)? I believe that it's for optimization purposes (no need to drop if we know that we have a keep-only strategy after).

pkg/manifests/manifests.go
```diff
@@ -436,10 +449,26 @@ func (f *Factory) AlertmanagerUserWorkloadClusterRole() (*rbacv1.ClusterRole, er
 	return f.NewClusterRole(f.assets.MustNewAssetSlice(AlertmanagerUserWorkloadClusterRole))
 }
 
+func (f *Factory) AlertmanagerServiceMonitors() ([]*monv1.ServiceMonitor, error) {
+	return serviceMonitors(
+		f.AlertmanagerServiceMonitor,
```
simonpasquier: we should be able to get rid of the `f.Alertmanager*ServiceMonitor()` methods and pass file paths (e.g. AlertmanagerServiceMonitor, ...) directly to serviceMonitors().

pkg/manifests/manifests.go
```diff
@@ -3494,16 +3611,34 @@ func makeConsoleURL(c *configv1.Console, path string) (string, error) {
...
 }
 
-func serviceMonitors(fullServiceMonitor, minimalServiceMonitor func() (*monv1.ServiceMonitor, error)) ([]*monv1.ServiceMonitor, error) {
-	sMonitor, err := fullServiceMonitor()
-	if err != nil {
-		return nil, err
+func serviceMonitors(fullServiceMonitor, minimalServiceMonitor, telemetryServiceMonitor func() (*monv1.ServiceMonitor, error)) ([]*monv1.ServiceMonitor, error) {
+	var sms []*monv1.ServiceMonitor
+
+	if fullServiceMonitor != nil {
```
simonpasquier: i wouldn't check for nil: our contract is that when supporting collection profiles, a component must implement 1 service monitor for each.

assets/cluster-monitoring-operator/telemetry-service-monitor.yaml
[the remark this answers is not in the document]
simonpasquier: there's no concrete reason to define a scrape interval here if it is equal to the default value.

assets/control-plane/telemetry-service-monitor-kubelet.yaml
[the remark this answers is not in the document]
simonpasquier: one could argue that it's still valuable to exercise these additional metrics endpoints even if we collect no metric from them.

pkg/manifests/manifests.go
```diff
@@ -3494,16 +3514,16 @@ func makeConsoleURL(c *configv1.Console, path string) (string, error) {
...
 
-func serviceMonitors(fullServiceMonitor, minimalServiceMonitor func() (*monv1.ServiceMonitor, error)) ([]*monv1.ServiceMonitor, error) {
-	sMonitor, err := fullServiceMonitor()
-	if err != nil {
-		return nil, err
-	}
-	sMonitorMinimal, err := minimalServiceMonitor()
-	if err != nil {
-		return nil, err
+func (f *Factory) serviceMonitors(assetPaths ...string) ([]*monv1.ServiceMonitor, error) {
```
simonpasquier, line 3553: (nit) should we assert that we always 3 service monitors corresponding to our 3 supported collection profiles? should we also ensure that the collection profile's label value maps to a supported value?

rexagod: Right, we can enforce that expectation now. Patched.

pkg/manifests/manifests.go
```diff
@@ -3494,16 +3514,33 @@ func makeConsoleURL(c *configv1.Console, path string) (string, error) {
...
-		return nil, err
+
+	sms := make([]*monv1.ServiceMonitor, 0, len(assetPaths))
+	for _, path := range assetPaths {
+		sm, err := f.NewServiceMonitor(f.assets.MustNewAssetSlice(path))
+		if err != nil {
+			return nil, err
+		}
+
+		// Absence of the collection profile label implies the "full" profile.
```
simonpasquier: (nit) is this something that we should allow? I'd prefer to be explicit and require all service monitor assets to carry a collection profile label.

pkg/manifests/manifests.go
```diff
@@ -3494,16 +3514,33 @@ func makeConsoleURL(c *configv1.Console, path string) (string, error) {
...
+		profileValue := sm.Labels[collectionProfileLabel]
+		if profileValue == "" {
+			profileValue = FullCollectionProfile
+		}
+		if !slices.Contains(SupportedCollectionProfiles, CollectionProfile(profileValue)) {
```
simonpasquier: (nit) should we also validate that we have all profiles covered (e.g. 2 service monitors could have the same profile label by mistake)?

simonpasquier (commented):
/lgtm
/hold

I'm fine merging it as-is and address my 2 minor comments in a follow-up.

simonpasquier:
@rexagod this needs a rebase.

rexagod:
Squashed all except the latest commit for the last review, and rebased.

rexagod:
Renamed `fixup!` commits to be more specific, and sourced telemetry metrics for each minimal and telemetry SMs from a single hoisted telemetry list of metrics for all components that do not have their minimal profiles same as the full ones.

test/e2e/telemeter_test.go
simonpasquier: this doesn't seem to be relevant?

simonpasquier, line 148: My recommendation would be to keep the test as high-level as possible. IMHO trying to reverse-engineer the operator's logic is going to be brittle.
I'd rather turn on the telemetry collection profile and verify that Prometheus scrapes "some" metrics from the different jobs in the openshift-monitoring namespace (e.g. `sum by(job) (scrape_samples_post_metric_relabeling{namespace="openshift-monitoring"})`).

rexagod: ACK, I've dropped most of the test and we only check the aforementioned query post-application now.

simonpasquier:
@rexagod the generate job fails

simonpasquier (commented):
/lgtm

PR #2838, opened 2026-03-04
NO-JIRA: add metrics for the validating webhook

simonpasquier:


simonpasquier:
/hold

Unfortunately it doesn't work out of the box because the k8s metrics framework doesn't play nicely with controller-runtime :-/
They use different Prometheus registries by default. I'll see if we can do something about this.

simonpasquier:
/hold cancel

I had no choice other than re-implementing the controller-runtime metrics in CMO.

jan--f:
Nice, thank you!
Two questions/comments:
Is it worth leaving a comment about the choice of bucket boundaries for future readers?
Any reason to not set a `NativeHistogramBucketFactor`? I'm afraid I'm not too knowledgeable about what it would take for use to use native histograms, but the client we use should support it..?

simonpasquier:
@jan--f addressed your 2 (valid) points in 8ac3965514a751369c1e59b02476a438abad1f06

simonpasquier:
> Any reason to not set a `NativeHistogramBucketFactor`? I'm afraid I'm not too knowledgeable about what it would take for use to use native histograms, but the client we use should support it..?

I realize that it isn't yet supported by the `k8s.io/component-base/metrics` wrapper which we use instead of `github.com/prometheus/client_golang/prometheus`. We'll have to wait for https://github.com/kubernetes/kubernetes/issues/128842

simonpasquier:
/retest-required
/verified by @simonpasquier 

image

PR #2843, opened 2026-03-09
Add dashboard-kernel-stack-af-packet-standalone.yaml 

midu16:
Standalone ConfigMap for Node Exporter kernel stack / AF_PACKET correlation dashboard . Apply to openshift-config-managed for console discovery.


karampok:
AFAICT The dashboard covers the overall softirq and kernel CPU cost per node, but it cannot answer a per-isolated-CPU question like "which kernel function is consuming cycles on CPUs dedicated to DPDK workloads.

The insight I am looking is
```
For CPUs that are isolated and pinned to a workload (e.g. DPDK), I want to measure 
what percentage of their time is spent on kernel work (e.g. kernel networking), and 
break that down by  function type (e.g. NET_RX) to identify which kernel function is
 consuming cycles on CPUs that should be dedicated to userspace packet processing.
```

manifests/dashboard-kernel-stack-af-packet-standalone.yaml
```diff
@@ -0,0 +1,1092 @@
...
+                "legend": {
+                    "show": true,
+                    "alignAsTable": true,
+                    "current": true,
+                    "values": true
+                },
+                "lines": true,
+                "targets": [
+                    {
+                        "expr": "sum(rate(node_zoneinfo_numa_miss_total{}[1m])) by (instance, node) / (sum(rate(node_zoneinfo_numa_hit_total{}[1m])) by (instance, node) + sum(rate(node_zoneinfo_numa_miss_total{}[1m])) by (instance, node))",
```
juzhao, line 1050:
```suggestion
                        "expr": "sum(rate(node_zoneinfo_numa_miss_total[1m])) by (instance, node) / (sum(rate(node_zoneinfo_numa_hit_total[1m])) by (instance, node) + sum(rate(node_zoneinfo_numa_miss_total[1m])) by (instance, node))",
```
no need to add `{}` if there is not label specified

simonpasquier:
discussed offline but there's little appetite to include the dashboard in the current form because some of the metrics used come from node_exporter's collectors which can't be enabled in the CMO configuration.

PR #2876, opened 2026-04-02
manifests/0000_50_cluster-monitoring-operator_04-config: Replace cluster_version_payload with cluster_version_risk_conditions

wking:
I occasionally use `cluster_version_payload` to look at individual clusters, but `cluster_version_risk_conditions` would allow easier fleet-level aggregation, to detect issues like buggy [PromQL in conditional update risks][1].  For single-cluster debugging, Insights archives will still have the information I used to get from `cluster_version_payload` (and more, like the failure message about why the CVO is having trouble applying a resource).  Expected cardinality for `cluster_version_risk_conditions` is one condition per risk per cluster-evaluating-updates-where-that-risk-matters.  Which is likely going to come in higher than `cluster_version_payload`'s cardinality of around-two-per-cluster. But hopefully still manageable.

[1]: https://github.com/openshift/cincinnati-graph-data/blob/22e24128e69eb0e8649b28190c2a923463a441d8/README.md#block-edges


Documentation/data-collection.md
```diff
@@ -99,12 +99,13 @@ data:
...
-    # version operator has progressed and can be used to identify whether
-    # a particular payload entry is causing failures during upgrade.
+    # cluster_version_risk_conditions captures update risk evaluation
+    # from the cluster version operator, and can be used to identify
+    # whether a particular update risk condition is prevalent, and how
+    # successfully that risk is being evaluated.
     #
     # consumers: (@openshift/openshift-team-cluster-manager)
-    - '{__name__="cluster_version_payload"}'
+    - '{__name__="cluster_version_risk_conditions"}'
```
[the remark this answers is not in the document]
wking: Is there a recommneded way to update `Documentation/sample-metrics.md`?  Do I just grab a few of these from an example tech-preview cluster and stuff them in by hand?  Do I update the whole file with the results from that tech-preview cluster?  Other?

wking:
Rebased onto the current `main` tip with 9045e66f8 -> 2e410f50f

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
```diff
@@ -91,12 +91,13 @@ data:
...
-    # cluster_version_payload captures how far through a payload the cluster
-    # version operator has progressed and can be used to identify whether
-    # a particular payload entry is causing failures during upgrade.
+    # cluster_version_risk_conditions captures update risk evaluation
+    # from the cluster version operator, and can be used to identify
+    # whether a particular update risk condition is prevalent, and how
+    # successfully that risk is being evaluated.
     #
     # consumers: (@openshift/openshift-team-cluster-manager)
-    - '{__name__="cluster_version_payload"}'
```
simonpasquier, line 99: why remove `cluster_version_payload` now? If merged, telemeter server would stop accepting metrics from old clusters that don't have `cluster_version_risk_conditions`.

simonpasquier (commented):
We'd need a recording rule to aggregate away labels like pod, instance, service, ... that aren't interesting and would create unnecessary churn.
IIUC the only [expected](https://github.com/openshift/cluster-version-operator/blob/dd0a8410467a09cd181216e191825d4088ea2115/pkg/cvo/metrics.go#L527-L529) value for the condition label is `Applies` so it would be good (as a safety measure) to filter on this value in the allow-list.
Also for future ourselves, how many combinations of risk x reason could there be in total? While there should be only a limited number of risks per cluster, I'd like us to estimate the overall cardinality.

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
wking: > why remove `cluster_version_payload` now? 

I tried to cover this in [my commit/pull message][1], but

* I don't use `cluster_version_payload` often.
* I'm not aware of anyone else using it, although obviously that doesn't mean that it's impossible for anyone else to be using it.
* I'm trying to find ways to save cardinality costs to buy room for the new and more valuable `cluster_version_risk_conditions`.

> If merged, telemeter server would stop accepting metrics from old clusters that don't have `cluster_version_risk_conditions`.

I get so little value out of `cluster_version_payload` that I'm ok with that, and it saves us the trouble of remembering to come back around and remove it at a later date.

[1]: https://github.com/openshift/cluster-monitoring-operator/pull/2876#issue-4194521332

wking:
> We'd need a recording rule to aggregate away labels like pod, instance, service...

Where do you want this recording rule?  E.g. the current `cluster_operator_conditions` metric is flowing into Telemetry with labels like:

```console
cluster_operator_conditions{_id="...", condition="Available", endpoint="metrics", instance="10.0....:9099", job="cluster-version-operator", name="authentication", namespace="openshift-cluster-version", pod="cluster-version-operator-5f5cb985c7-n9mqr", prometheus="openshift-monitoring/k8s", reason="AsExpected", receive="true", service="cluster-version-operator", tenant_id="..."}
```

> IIUC the only [expected](https://github.com/openshift/cluster-version-operator/blob/dd0a8410467a09cd181216e191825d4088ea2115/pkg/cvo/metrics.go#L527-L529) value for the condition label is `Applies` so it would be good (as a safety measure) to filter on this value in the allow-list.

Fine with me.  2e410f50f -> 0593a6d57 rebases on the current tip and adds `condition="Applies"` filtering.

> Also for future ourselves, how many combinations of risk x reason could there be in total? While there should be only a limited number of risks per cluster, I'd like us to estimate the overall cardinality.

The ClusterVersion API sets a maximum [at 500 risks per cluster][1].  But so far in practice, checking https://github.com/openshift/cincinnati-graph-data risk declarations:

```console
$ git log --oneline --date=short --format='%ad %h %s' -1
2026-07-02 505afb32 Merge pull request #9864 from openshift-ota-bot/promote-4.18.47-to-candidate-4.20
$ grep '^name: ' blocked-edges/* | sed 's|[^/]*/\([0-9]*[.][0-9]*\)[.].*name: \(.*\)|\1 \2|' | sort | uniq | cut -d' ' -f1 | sort -V | uniq -c
      3 4.7
      3 4.10
     15 4.11
     15 4.12
     18 4.13
     21 4.14
     18 4.15
     20 4.16
     15 4.17
     19 4.18
     24 4.19
      7 4.20
      5 4.21
      1 4.22
```

which puts the upper bound of risks visible to any current cluster at 43 for a theoretical 4.18 cluster in a channel where all 19 current 4.18 risks and all 24 current 4.19 risks are visible as next-hop updates.

[1]: https://github.com/openshift/api/blob/ef71f942ef6cca617c5ec7344ec6813949232597/config/v1/types_cluster_version.go#L214

simonpasquier:
> Where do you want this recording rule?

answered elsewhere but the rule should be deployed by CVO: https://rhobs-handbook.netlify.app/products/openshiftmonitoring/telemetry.md/#configure-recording-rules

Thanks for the additional details on the cardinality. I don't see an issue for the Telemeter server but heads-up to @moadz in any case.

manifests/0000_50_cluster-monitoring-operator_04-config.yaml
simonpasquier: I'm fine with that. The process isn't well defined on our side (we tend to always append to the allow-list) but for consistency we probably want to remove the metric from all supported z-stream releases.
cc @jan--f for additional input :)

jan--f: Agree completely. We tend to shy away from removing telemetry metrics cause we don't know who uses them for what.
In this case I think we have the main user proposing this so no concerns.
+1 for backports.

PR #2145, opened 2023-11-03
OCPBUGS-21621: Set the new --disable-http2 flag for prometheus-adapter to disable HTTP2

machine424:
* [ ] I added CHANGELOG entry for this change.
* [ ] No user facing changes, so no entry in CHANGELOG was needed.

/cc @simonpasquier

PR #2882, opened 2026-04-10
CNTRLPLANE-3217: Enable configurable PKI for GRPC TLS certificates

sanchezl:
## Summary

- Bump `library-go`, `api`, and `client-go` to bring in configurable PKI support (library-go PRs #2127, #2145)
- When the `ConfigurablePKI` feature gate is enabled, resolve key algorithms from the cluster PKI profile for the self-signed GRPC TLS certificates used for Thanos GRPC communication
- When the feature gate is disabled (default), existing RSA-2048 behavior is preserved

## Details

Certificate names registered with the PKI profile:
- `monitoring.grpc-tls-signer` — self-signed CA
- `monitoring.grpc-tls-client` — Thanos querier client cert
- `monitoring.grpc-tls-serving` — Prometheus GRPC server cert

The implementation switches from the legacy `MakeSelfSignedCAConfig`/`MakeServerCert`/`MakeClientCertificateForDuration` APIs to the new parameterized `NewSigningCertificate`/`NewServerCertificate`/`NewClientCertificate` APIs which properly handle SubjectKeyID computation and KeyUsage flags for all key types.

CSR-based certificates (metrics-client-certs, federate-client-certs, metrics-server-client-certs) are not affected — their keys are already ECDSA and signing is handled by the kube-apiserver.

## Test plan

- [ ] Unit tests pass (`make test-unit`)
- [ ] CI passes
- [ ] With `ConfigurablePKI` disabled (default): GRPC TLS certs continue to use RSA-2048
- [ ] With `ConfigurablePKI` enabled and PKI resource configured: GRPC TLS certs use the configured algorithm


simonpasquier (commented):
:wave: thanks for doing this! 

LGTM but doesn't the operator need to react and trigger a reconciliation when the pki configuration changes?

sanchezl:
> LGTM but doesn't the operator need to react and trigger a reconciliation when the pki configuration changes?

No. The new config just gets picked up at the next rotation cycle. This is intentional.

pkg/manifests/tls_test.go
simonpasquier, line 1: Can we add some tests with a non-nil PKI provider?

simonpasquier (commented):
> > LGTM but doesn't the operator need to react and trigger a reconciliation when the pki configuration changes?
> 
> No. The new config just gets picked up at the next rotation cycle. This is intentional.

Do we even need to an informer then? At the very least I'd expect a comment in operator.go explaining that it doesn't need to reconcile the TLS materials ASAP.

pkg/manifests/tls_test.go
sanchezl: Done.

sanchezl:
> Do we even need to an informer then? At the very least I'd expect a comment in operator.go explaining that it doesn't need to reconcile the TLS materials ASAP.

Yes, we need the informer so that the PKIProfileProvider can read the PKI resource from the lister cache instead of hitting the API server on every reconciliation. I've added a comment explaining why no event handler is registered: PKI profile changes are picked up at the next certificate rotation cycle, not via immediate reconciliation. 
I also switched from starting the informer manually to adding it to o.informers so its cache is synced before the first reconciliation runs.

simonpasquier:
/approve

The tech preview job fails because the CMO service accounat has no permissions to list/watch the PKI configuration.
RBAC should be added to `jsonnet/components/cluster-monitoring-operator.libsonnet` then run `make generate` and commit the changes.

https://github.com/openshift/cluster-monitoring-operator/blob/88f9787dc4151c93d324829f645a1269b16d4c8c/jsonnet/components/cluster-monitoring-operator.libsonnet#L176-L283

simonpasquier (commented):
/lgtm
/approve

simonpasquier:
/verified by e2e tests

Origin e2e tests run with the feature gates enabled (`e2e-aws-ovn-techpreview`), proving that the change doesn't break when the new code path is activated.

PR #2886, opened 2026-04-16
MON-4558: enable zoneinfo node-exporter collector via config

midu16:
Add optional NodeExporter zoneinfo collector toggle (disabled by default).

Wire --collector.zoneinfo/--no-collector.zoneinfo in updateNodeExporterArgs, extend unit and e2e tests, and update API and AsciiDoc documentation.

PR #2887, opened 2026-04-16
MON-4560: enable softirqs node-exporter collector via config

midu16:
Add optional NodeExporter softirqs collector toggle (disabled by default).

Wire --collector.softirqs/--no-collector.softirqs in updateNodeExporterArgs, extend unit and e2e tests, and update API and AsciiDoc documentation.


simonpasquier (changes requested):
same comment as https://github.com/openshift/cluster-monitoring-operator/pull/2886#issuecomment-4260069130

simonpasquier (approved):
/lgtm

PR #2888, opened 2026-04-16
MON-4559: enable interrupts node-exporter collector via config

midu16:
Add optional NodeExporter interrupts collector toggle (disabled by default).

Wire --collector.interrupts/--no-collector.interrupts in updateNodeExporterArgs, extend unit and e2e tests, and update API and AsciiDoc documentation.


pkg/manifests/types.go
```diff
@@ -562,6 +565,14 @@ type NodeExporterCollectorSystemdConfig struct {
 	Units []string `json:"units,omitempty"`
 }
 
+// The `NodeExporterCollectorInterruptsConfig` resource works as an on/off switch for
+// the `interrupts` collector of the `node-exporter` agent.
+// By default, the `interrupts` collector is disabled.
+type NodeExporterCollectorInterruptsConfig struct {
+	// A Boolean flag that enables or disables the `interrupts` collector.
+	Enabled bool `json:"enabled,omitempty"`
```
simonpasquier, line 573: Given the cardinality of the metrics, I'd expect that we offer a way for end-users to select/discard interrupts (similar to what we have for systemd units in NodeExporterCollectorSystemdConfig).
In general I'm more inclined towards an allow-list approach which limits the risk of the collector generating tons of metrics and  overloading Prometheus.

pkg/manifests/types.go
```diff
@@ -577,6 +580,26 @@ type NodeExporterCollectorSystemdConfig struct {
 	Units []string `json:"units,omitempty"`
 }
 
+// The `NodeExporterCollectorInterruptsConfig` resource works as an on/off switch for
+// the `interrupts` collector of the `node-exporter` agent.
+// By default, the `interrupts` collector is disabled.
+// If you enable this collector, use `include` to allow-list which interrupt lines are
```
simonpasquier: given that the collector generates no metrics if you only enable it without setting an include list, I'd be inclined to remove the `Enabled` field and rely on the size of the `Include`array to enable/disable the collector.

@danielmellado @jan--f for inputs.

danielmellado: +1 on dropping Enabled. Having enabled: true with an empty include is a confusing no-op. Users would expect metrics and get nothing.

I think it'd be simpler to just use include as the implicit toggle: non-empty list enables the collector, empty/absent disables it, in the same way Simon's suggests.

simonpasquier: it will nicely with the config CRD because we can enforce that when configuring the interrupts collector, the list needs to be > 0.

pkg/manifests/types.go
```diff
@@ -590,6 +593,23 @@ type NodeExporterCollectorSystemdConfig struct {
...
 }
 
+// The `NodeExporterCollectorInterruptsConfig` resource configures the `interrupts`
+// collector of the `node-exporter` agent.
+// By default, the collector is disabled.
+// A non-empty `include` list enables the collector and allow-lists which interrupt
+// lines are exposed (each entry is a regular expression; they are combined with OR).
+// This limits metric cardinality compared to exposing every line from `/proc/interrupts`.
+// If `include` is empty, the collector is disabled.
+// To match all interrupt lines (high cardinality), set `include` to one broad regexp
```
simonpasquier: I wouldn't promote this usage. A lazy user would configure this pattern and complain that they see too many metrics.

pkg/manifests/types.go
```diff
@@ -590,6 +593,23 @@ type NodeExporterCollectorSystemdConfig struct {
 	Units []string `json:"units,omitempty"`
 }
 
+// The `NodeExporterCollectorInterruptsConfig` resource configures the `interrupts`
+// collector of the `node-exporter` agent.
+// By default, the collector is disabled.
+// A non-empty `include` list enables the collector and allow-lists which interrupt
+// lines are exposed (each entry is a regular expression; they are combined with OR).
+// This limits metric cardinality compared to exposing every line from `/proc/interrupts`.
+// If `include` is empty, the collector is disabled.
```
simonpasquier: already stated above
```suggestion
```

pkg/manifests/types.go
```diff
@@ -590,6 +593,23 @@ type NodeExporterCollectorSystemdConfig struct {
 	Units []string `json:"units,omitempty"`
 }
 
+// The `NodeExporterCollectorInterruptsConfig` resource configures the `interrupts`
+// collector of the `node-exporter` agent.
+// By default, the collector is disabled.
+// A non-empty `include` list enables the collector and allow-lists which interrupt
+// lines are exposed (each entry is a regular expression; they are combined with OR).
+// This limits metric cardinality compared to exposing every line from `/proc/interrupts`.
```
simonpasquier: the details about the regex matching should be explained with the `include` field.
```suggestion
// A non-empty `include` list enables the collector and defines which interrupt metrics should be collected.
```

pkg/manifests/types.go
```diff
@@ -590,6 +593,23 @@ type NodeExporterCollectorSystemdConfig struct {
...
+// If `include` is empty, the collector is disabled.
+// To match all interrupt lines (high cardinality), set `include` to one broad regexp
+// such as `.*`.
+type NodeExporterCollectorInterruptsConfig struct {
+	// A list of regular expression patterns. Each line in `/proc/interrupts` is matched
+	// against the same string node-exporter uses: the IRQ name, info, and devices fields
+	// joined with `;`, for example `LOC;77;IO-APIC …`. Patterns are combined with OR.
```
simonpasquier: Can we provide some example patterns and tell what they will match? E.g.

```
FOO;.* matches all interrupts named FOO
.*;some_dev matches all interrupts on some_dev device
```

pkg/manifests/types.go
```diff
@@ -590,6 +593,23 @@ type NodeExporterCollectorSystemdConfig struct {
...
+// collector of the `node-exporter` agent.
+// By default, the collector is disabled.
+// A non-empty `include` list enables the collector and allow-lists which interrupt
+// lines are exposed (each entry is a regular expression; they are combined with OR).
+// This limits metric cardinality compared to exposing every line from `/proc/interrupts`.
+// If `include` is empty, the collector is disabled.
+// To match all interrupt lines (high cardinality), set `include` to one broad regexp
+// such as `.*`.
+type NodeExporterCollectorInterruptsConfig struct {
+	// A list of regular expression patterns. Each line in `/proc/interrupts` is matched
```
simonpasquier, line 602: it should be noted that patterns are anchored on both ends.

pkg/manifests/types.go
midu16: @simonpasquier can we have a call to braimstorm this together ?

pkg/manifests/types.go
jan--f: This doesn't seem to be addressed?

midu16: @jan--f i have update this as part of the last commit.

pkg/manifests/types.go
```diff
@@ -580,6 +583,23 @@ type NodeExporterCollectorSystemdConfig struct {
...
+// By default, the collector is disabled.
+// A non-empty `include` list enables the collector and defines which interrupt metrics should be collected.
+type NodeExporterCollectorInterruptsConfig struct {
+	// A list of regular expression patterns. Each line in `/proc/interrupts` is matched
+	// against the same string node-exporter uses: the IRQ name, info, and devices fields
+	// joined with `;`, for example `LOC;77;IO-APIC 2-edge …`. Patterns are combined with OR
+	// into a single expression anchored on both ends, so each pattern must match the entire
+	// string (use `.*` where needed). An empty list disables the collector.
+	// Examples: `FOO;.*` matches all interrupts named FOO; `.*;some_dev` matches all interrupts
+	// on device `some_dev`; `.;.;(eth|eno|ens|em[0-9]|bond|team|mlx|.TxRx.).*` matches network
```
danielmellado: The single . in the first two positions only matches a single character, but node-exporter's /proc/interrupts format joins fields that can be multi-character (e.g. LOC;77;IO-APIC...). Should this be someting along te lines of .*;.*;(eth|eno|ens|em[0-9]|bond|team|mlx|.*TxRx.*).* instead?

If the single-char match is intentional for some reason, it would help to call that out explicitly in the comment.

pkg/manifests/manifests_test.go
```diff
@@ -3479,6 +3481,41 @@ nodeExporter:
 			argsPresent: []string{"--collector.ethtool", "--collector.ethtool.device-exclude=^(br-int|lo)$"},
 			argsAbsent:  []string{"--no-collector.ethtool"},
 		},
+		{
+			name: "enable interrupts collector with include pattern",
+			config: `
+nodeExporter:
+  collectors:
+    interrupts:
+      include: ["^LOC;.*"]
```
danielmellado: regexListToArg already wraps the combined pattern with ^(...)$, so the ^ in ^LOC;.* is somehoe redundant IMHO. It produces ^(^LOC;.*)$.

The doc examples in types.go correctly use patterns without the leading anchor (e.g. FOO;.*). Aligning the tests with the documented guidance would be less confusing for future readers.

Same applies to lines 3502 (^LOC;.*, ^NMI;.*)

pkg/manifests/manifests_test.go
```diff
@@ -3479,6 +3481,41 @@ nodeExporter:
...
+			name: "disable interrupts collector when include is empty",
+			config: `
+nodeExporter:
+  collectors:
+    interrupts:
+      include: []
+`,
+			argsPresent: []string{"--no-collector.interrupts"},
+			argsAbsent:  []string{"--collector.interrupts"},
+		},
```
danielmellado, line 3568: nit: There's no test case validating that an invalid regex in include returns an error. The systemd collector has this coverage via TestNodeExporterSystemdUnits and can be used as an example.

danielmellado (changes requested):
Besides some nits it overall LGTM, I'll wait for Simon to take a second run as well. Thanks!

Note that now you'd need to add the corresponding interrupts field to the ClusterMonitoring CRD in [openshift/api](https://github.com/openshift/api) (types_cluster_monitoring.go) to keep the CR path in sync with the ConfigMap.

It should follow the systemd union pattern (collectionPolicy + collect.include) since you have sub-configuration, plus the corresponding merge logic in pkg/manifests/config_merge.go and tests in config_merge_test.go. Happy to help with that in a follow-up PR once this one lands.

simonpasquier:
> Note that now you'd need to add the corresponding interrupts field to the ClusterMonitoring CRD in [openshift/api](https://github.com/openshift/api) (types_cluster_monitoring.go) to keep the CR path in sync with the ConfigMap.

I created https://redhat.atlassian.net/browse/MON-4608 as a sub-task of [MON-4559](https://redhat.atlassian.net/browse/MON-4559)

pkg/manifests/manifests.go
```diff
@@ -970,6 +970,17 @@ func (f *Factory) updateNodeExporterArgs(args []string) ([]string, error) {
 		args = setArg(args, "--no-collector.ethtool", "")
 	}
 
+	if len(f.config.ClusterMonitoringConfiguration.NodeExporterConfig.Collectors.Interrupts.Include) > 0 {
+		args = setArg(args, "--collector.interrupts", "")
+		pattern, err := regexListToArg(f.config.ClusterMonitoringConfiguration.NodeExporterConfig.Collectors.Interrupts.Include)
```
simonpasquier, line 975: it'd be better to catch the issue when loading the configuration as it would be notified to users when they update the configmap (thanks to the admission webhook servicce) rather than late in the reconciliation cycle.

I've submitted https://github.com/openshift/cluster-monitoring-operator/pull/3048 to adopt this approach for the existing systemd collector.

pkg/manifests/types.go
```diff
@@ -591,6 +594,23 @@ type NodeExporterCollectorSystemdConfig struct {
...
```
simonpasquier: @danielmellado had a comment about the single dots in the regexp but I see no reply, not sure if the comment being resolved is on purpose or not.

pkg/manifests/types.go
midu16: Good catch, fixed. Updated the examples to use .*;.*;(eth|eno|ens|em[0-9]|bond|team|mlx|.*TxRx.*).* — the single dots were a mistake, the name and info fields can be multi-character.

cc: @simonpasquier

pkg/manifests/config.go
```diff
@@ -387,6 +388,22 @@ func (c *Config) validate() error {
...
+	return nil
+}
+
+func validateNodeExporterConfig(config *NodeExporterConfig) error {
+	if config == nil {
+		return nil
+	}
+	for _, pattern := range config.Collectors.Interrupts.Include {
+		if _, err := regexp.Compile(pattern); err != nil {
+			return fmt.Errorf("%w: interrupts: invalid regexp pattern: %s", ErrConfigValidation, pattern)
```
simonpasquier: the top-level func already wraps the returned error with `ErrConfigValidation`

simonpasquier (commented):
let's wait for https://github.com/openshift/cluster-monitoring-operator/pull/3048 to merge and rebase.

Also pinging @shreyasiddhartha to review the documentation part.

shreyasiddhartha:
> let's wait for #3048 to merge and rebase.
> 
> Also pinging @shreyasiddhartha to review the documentation part.

@simonpasquier Looks good from docs perspective. Thanks!

pkg/manifests/manifests.go
```diff
@@ -970,6 +970,13 @@ func (f *Factory) updateNodeExporterArgs(args []string) ([]string, error) {
 		args = setArg(args, "--no-collector.ethtool", "")
 	}
 
+	if include := f.config.ClusterMonitoringConfiguration.NodeExporterConfig.Collectors.Interrupts.Include; len(include) > 0 {
+		args = setArg(args, "--collector.interrupts", "")
+		args = setArg(args, "--collector.interrupts.name-include=", "^("+strings.Join(include, "|")+")$")
```
simonpasquier: we still need to call regexListToArg(), just in case the validation missed something.

simonpasquier (commented):
/lgtm

PR #2889, opened 2026-04-17
MON-4492: ClusterMonitoring alertmanagerConfig logic

marioferh:
Apply the same Phase 1 rules as metricsServer: when cluster-monitoring-config does not define alertmanagerMain, merge spec.alertmanagerConfig from the ClusterMonitoring CR (deploymentMode Disabled/DefaultConfig/CustomConfig and customConfig fields). ConfigMap values still win when alertmanagerMain is present.

Leave AlertmanagerMainConfig unset in applyDefaults until merge, default it in EnsureSafeDefaults, and initialize a non-nil struct in NewFactory for code paths that build manifests without MergeClusterMonitoringCRD.

Share CRD []ContainerResource → corev1.ResourceRequirements conversion via containerResourcesFromCRD for metrics server and alertmanager.

Add unit tests (config_merge_test) and e2e tests for CR-driven settings and ConfigMap vs CR precedence.

Made-with: Cursor

### DIFF
--- a/pkg/manifests/config_merge.go
+++ b/pkg/manifests/config_merge.go
@@ -163,9 +163,7 @@ func (c *Config) mergeMetricsServerConfiguration(msc configv1alpha1.MetricsServe
 	}
 	cfg.NodeSelector = msc.NodeSelector
 	cfg.Tolerations = msc.Tolerations
-	if res := containerResourcesFromCRD(msc.Resources); res != nil {
-		cfg.Resources = res
-	}
+	cfg.Resources = containerResourcesFromCRD(msc.Resources)
 	if msc.Audit.Profile != "" {
 		cfg.Audit = &Audit{
 			Profile: auditv1.Level(string(msc.Audit.Profile)),
@@ -191,9 +189,7 @@ func (c *Config) mergePrometheusOperatorConfiguration(poc configv1alpha1.Prometh
 	}
 	cfg.NodeSelector = poc.NodeSelector
 	cfg.Tolerations = poc.Tolerations
-	if res := containerResourcesFromCRD(poc.Resources); res != nil {
-		cfg.Resources = res
-	}
+	cfg.Resources = containerResourcesFromCRD(poc.Resources)
 	cfg.TopologySpreadConstraints = poc.TopologySpreadConstraints
 
 	c.ClusterMonitoringConfiguration.PrometheusOperatorConfig = cfg
@@ -228,9 +224,7 @@ func mergeAlertmanagerCustomConfigFromCRD(dst *AlertmanagerMainConfig, cc config
 	if len(cc.NodeSelector) > 0 {
 		dst.NodeSelector = cc.NodeSelector
 	}
-	if res := containerResourcesFromCRD(cc.Resources); res != nil {
-		dst.Resources = res
-	}
+	dst.Resources = containerResourcesFromCRD(cc.Resources)
 	if len(cc.Secrets) > 0 {
 		for _, s := range cc.Secrets {
 			dst.Secrets = append(dst.Secrets, string(s))
@@ -259,6 +253,7 @@ func persistentVolumeClaimToEmbedded(pvc *v1.PersistentVolumeClaim) *monv1.Embed
 	}
 	em.EmbeddedObjectMetadata.Name = pvc.Name
 	em.EmbeddedObjectMetadata.Labels = pvc.Labels
+	em.EmbeddedObjectMetadata.Annotations = pvc.Annotations
 	em.Spec = pvc.Spec
 	return em
 }
--- a/pkg/manifests/config_merge.go
+++ b/pkg/manifests/config_merge.go
@@ -16,7 +16,9 @@ package manifests
 
 import (
 	configv1alpha1 "github.com/openshift/api/config/v1alpha1"
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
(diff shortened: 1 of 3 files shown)


pkg/manifests/config_merge.go
```diff
@@ -97,33 +161,104 @@ func (c *Config) mergeMetricsServerConfiguration(msc configv1alpha1.MetricsServe
...
-				resources.Requests[v1.ResourceName(res.Name)] = res.Request
-			}
-			if !res.Limit.IsZero() {
-				resources.Limits[v1.ResourceName(res.Name)] = res.Limit
-			}
-		}
-		cfg.Resources = resources
+	if res := containerResourcesFromCRD(msc.Resources); res != nil {
+		cfg.Resources = res
 	}
```
simonpasquier: (nit) we don't even need the test, right?
```suggestion
	cfg.Resources = containerResourcesFromCRD(msc.Resources)
```

pkg/manifests/config_merge.go
```diff
@@ -97,33 +161,104 @@ func (c *Config) mergeMetricsServerConfiguration(msc configv1alpha1.MetricsServe
...
+	}
+
+	cfg := &PrometheusOperatorConfig{}
+
+	if poc.LogLevel != "" {
+		cfg.LogLevel = logLevelCRDToManifest(poc.LogLevel)
+	}
+	cfg.NodeSelector = poc.NodeSelector
+	cfg.Tolerations = poc.Tolerations
+	if res := containerResourcesFromCRD(poc.Resources); res != nil {
```
simonpasquier: same here

pkg/manifests/config_merge.go
```diff
@@ -97,33 +161,104 @@ func (c *Config) mergeMetricsServerConfiguration(msc configv1alpha1.MetricsServe
...
+	if clusterMonitoringAlertmanagerSpecEmpty(ac) {
+		return
+	}
+
+	cfg := &AlertmanagerMainConfig{}
+	switch ac.DeploymentMode {
+	case configv1alpha1.AlertManagerDeployModeDisabled:
+		cfg.Enabled = ptr.To(false)
+	case configv1alpha1.AlertManagerDeployModeDefaultConfig:
+		// Deploy with platform defaults (Enabled is nil → true via IsEnabled).
```
simonpasquier, line 211: (nit) should we set explicitly to true?

pkg/manifests/config_merge.go
```diff
@@ -97,33 +161,104 @@ func (c *Config) mergeMetricsServerConfiguration(msc configv1alpha1.MetricsServe
...
+}
+
+func mergeAlertmanagerCustomConfigFromCRD(dst *AlertmanagerMainConfig, cc configv1alpha1.AlertmanagerCustomConfig) {
+	if ll := logLevelCRDToManifest(cc.LogLevel); ll != "" {
+		dst.LogLevel = ll
+	}
+	if len(cc.NodeSelector) > 0 {
+		dst.NodeSelector = cc.NodeSelector
+	}
+	if res := containerResourcesFromCRD(cc.Resources); res != nil {
```
simonpasquier: same here

pkg/manifests/config_merge.go
```diff
@@ -97,33 +155,104 @@ func (c *Config) mergeMetricsServerConfiguration(msc configv1alpha1.MetricsServe
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: v1.SchemeGroupVersion.String(),
+			Kind:       "PersistentVolumeClaim",
+		},
+	}
+	em.EmbeddedObjectMetadata.Name = pvc.Name
+	em.EmbeddedObjectMetadata.Labels = pvc.Labels
+	em.Spec = pvc.Spec
+	return em
+}
```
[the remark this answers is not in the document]
simonpasquier: It looks like a valid concern. If not, we should at least document why.

simonpasquier (commented):
The API lacks this field from the ConfigMap:

https://github.com/openshift/cluster-monitoring-operator/blob/890e5fb783f7db0b89ffebd8120168ceb9ee7d7f/pkg/manifests/types.go#L110-L115

Is it on purpose?

marioferh:
> The API lacks this field from the ConfigMap:
> 
> https://github.com/openshift/cluster-monitoring-operator/blob/890e5fb783f7db0b89ffebd8120168ceb9ee7d7f/pkg/manifests/types.go#L110-L115
> 
> Is it on purpose?

it is, it is explained somewhere here, I don't remember where is the exact comment: https://github.com/openshift/api/pull/2148

pkg/manifests/config_merge.go
marioferh: done

pkg/manifests/config_merge.go
marioferh: done

pkg/manifests/config_merge.go
marioferh: done

pkg/manifests/config_merge.go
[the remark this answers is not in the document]
marioferh: fixed

pkg/manifests/config_merge.go
```diff
@@ -78,6 +108,40 @@ func verbosityLevelToNumeric(level configv1alpha1.VerbosityLevel) uint8 {
...
+	for _, res := range resources {
+		if !res.Request.IsZero() {
+			out.Requests[v1.ResourceName(res.Name)] = res.Request
+		}
+		if !res.Limit.IsZero() {
+			out.Limits[v1.ResourceName(res.Name)] = res.Limit
+		}
+	}
+	return out
+}
```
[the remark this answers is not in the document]
marioferh: not needed

pkg/manifests/config_merge.go
simonpasquier: it can be a follow-up.

simonpasquier (commented):
/lgtm

PR #2891, opened 2026-04-23
OCPBUGS-84215: Fix alertmanager SCC race condition during deployment

ropatil010:
HI Team, 

PTAL on this PR. 

Add WaitForSCCAccess to prevent race condition where alertmanager-main StatefulSet creates pods before RBAC propagates to kube-apiserver.

Problem:
The cluster-monitoring-operator creates ClusterRole, ClusterRoleBinding, and ServiceAccount for alertmanager, then immediately creates the Alertmanager StatefulSet. The StatefulSet attempts to create pods before the kube-apiserver's RBAC authorizer cache has propagated the SCC access, resulting in pod creation failures with "unable to validate against any security context constraint: provider 'nonroot': Forbidden".

This causes the test "[sig-auth][Feature:SCC][Early] should not have pod creation failures during install" to fail when the SCC denials exceed the 5-second threshold (typically seeing 11 occurrences).

Solution:
- Add WaitForSCCAccess() method to pkg/client/client.go that uses SubjectAccessReview to verify the serviceaccount can "use" the SCC
- Call WaitForSCCAccess immediately after ServiceAccount creation and before creating any other resources
- Poll every 1 second with 2-minute timeout for RBAC to propagate
- Fail fast if RBAC doesn't propagate, avoiding creation of other resources

Changes:
- pkg/client/client.go: Add WaitForSCCAccess method with SubjectAccessReview
- pkg/tasks/alertmanager.go: Wait for alertmanager-main SA to get nonroot SCC access

This follows the same pattern used to fix OCPBUGS-77899 for the operator-controller-controller-manager component.
Failure logs: 
```
https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-origin-31002-nightly-4.22-e2e-metal-ovn-two-node-fencing-serial-3of3/2047152852652003328
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -44,6 +44,7 @@ import (
 	monitoring "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
 	admissionv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
+	authorizationv1 "k8s.io/api/authorization/v1"
 	v1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	policyv1 "k8s.io/api/policy/v1"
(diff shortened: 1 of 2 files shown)


ropatil010:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-metal-ovn-two-node-fencing-serial-techpreview
/payload-job periodic-ci-openshift-release-main-nightly-4.23-e2e-metal-ovn-two-node-fencing-serial-techpreview
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ovn-two-node-fencing-serial-techpreview

simonpasquier (commented):
we have other operands (prometheus, thanos ruler, user-defined stack) which runs with non-restricted SCCs. I assume that those need to be "fixed" too.

Is there any upstream issue documenting the race condition due to incorrect cache?

ropatil010:
> we have other operands (prometheus, thanos ruler, user-defined stack) which runs with non-restricted SCCs. I assume that those need to be "fixed" too.
> 
> Is there any upstream issue documenting the race condition due to incorrect cache?

Hi Simon, 
I'm planning to close this PR actually. I tried to check certain CI profiles but did not hit this kind of issue. 
The logs which i shared in failure logs could be flake.

ropatil010:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-techpreview

ropatil010:
Closing this PR, unable to reproduce the issue. Earlier failure could be due to env issue/Flaky. 
/close

PR #2899, opened 2026-04-28
NO-ISSUE: Update AGENTS.md

simonpasquier:


hack/build-jsonnet.sh
```diff
@@ -42,8 +42,15 @@ done
 # wait for all jobs to finish
 wait
 
+add_generated_header() {
+	local tmp
+	tmp=$(mktemp)
+	printf '# Code generated by hack/build-jsonnet.sh. DO NOT EDIT.\n' | cat - "$1" > "$tmp"
```
danielmellado, line 48: nit: I would've put this on a different PR or at least mention this in the description, but given that it's on a split commit it's fine.

simonpasquier: yeah I needed the change to tell the agent `don't modify directly files which contain "DO NOT EDIT"`

PR #2900, opened 2026-04-28
MON-4520: follow up AlertManagerConfig and PrometheusOp PR changes

marioferh:
Changes:
- Set cfg.Enabled = ptr.To(true) explicitly for AlertManagerDeployModeDefaultConfig instead of relying on IsEnabled() treating nil as true.
- Add e2e tests for PrometheusOperator CRD configuration and Phase 1 ConfigMap-over-CRD precedence.

Made-with: Cursor

### DIFF
--- a/pkg/manifests/config_merge.go
+++ b/pkg/manifests/config_merge.go
@@ -208,8 +208,9 @@ func (c *Config) mergeAlertmanagerConfiguration(ac configv1alpha1.AlertmanagerCo
 	case configv1alpha1.AlertManagerDeployModeDisabled:
 		cfg.Enabled = ptr.To(false)
 	case configv1alpha1.AlertManagerDeployModeDefaultConfig:
-		// Deploy with platform defaults (Enabled is nil → true via IsEnabled).
+		cfg.Enabled = ptr.To(true)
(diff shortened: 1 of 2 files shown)


pkg/manifests/config_merge.go
```diff
@@ -208,7 +208,7 @@ func (c *Config) mergeAlertmanagerConfiguration(ac configv1alpha1.AlertmanagerCo
 	case configv1alpha1.AlertManagerDeployModeDisabled:
 		cfg.Enabled = ptr.To(false)
 	case configv1alpha1.AlertManagerDeployModeDefaultConfig:
-		// Deploy with platform defaults (Enabled is nil → true via IsEnabled).
+		cfg.Enabled = ptr.To(true)
 	case configv1alpha1.AlertManagerDeployModeCustomConfig:
 		mergeAlertmanagerCustomConfigFromCRD(cfg, ac.CustomConfig)
```
simonpasquier, line 214: mergeAlertmanagerCustomConfigFromCRD() should also set Enabled to true or we should set it there

marioferh: done

PR #2907, opened 2026-04-30
MON-4524: MetricsServerConfig resources merge 

marioferh:

PR #2919, opened 2026-05-06
MON-4527: ClusterMonitoring NodeExporterConfig logic 

marioferh:


marioferh:
/hold 
include this commit: https://github.com/openshift/cluster-monitoring-operator/pull/2907 wait for merge and rebase

pkg/manifests/types.go
```diff
@@ -360,6 +360,10 @@ type ThanosQuerierConfig struct {
 
 // The `NodeExporterConfig` resource defines settings for the `node-exporter` agent.
 type NodeExporterConfig struct {
+	// Defines the nodes on which the Pods are scheduled.
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+	// Defines tolerations for the pods.
+	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
```
simonpasquier: should we hide theses settings from the "official" doc?
I understand that we added these fields for consistency but I have a hard time seeing a use case for node selector and tolerations in the scope of a daemonset like node_exporter.

marioferh: yep, I was trying to clarify this with @danielmellado this morning.
Do you agree with Simon?

danielmellado: The api design added that for consistency among all the other components as requested by the api tram. That said, we might for now not directly use them in CMO and that wouldn't hurt.

simonpasquier: we should probably revisit if there's no concrete use case because while the API would be consistent, it would still be confusing for users. I agree that for now, we should just drop the fields and not port them to the existing config.

marioferh: agree

pkg/manifests/config_merge.go
```diff
@@ -15,8 +15,11 @@
 package manifests
 
 import (
+	"strings"
+
 	configv1alpha1 "github.com/openshift/api/config/v1alpha1"
 	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	yamlv3 "gopkg.in/yaml.v3"
```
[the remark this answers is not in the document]
danielmellado: That's ok for now, we'll handle in a follow-up

pkg/manifests/config_merge.go
```diff
@@ -40,6 +43,159 @@ func (c *Config) mergeClusterMonitoringCRD(clusterMonitoring *configv1alpha1.Clu
 	c.mergePrometheusOperatorConfiguration(clusterMonitoring.Spec.PrometheusOperatorConfig)
 	c.mergeAlertmanagerConfiguration(clusterMonitoring.Spec.AlertmanagerConfig)
 	c.mergeMonitoringPluginConfiguration(clusterMonitoring.Spec.MonitoringPluginConfig)
+	c.mergeNodeExporterConfiguration(clusterMonitoringConfigYAML, clusterMonitoring.Spec.NodeExporterConfig)
+}
+
+// configMapYAMLDeclaresNodeExporter reports whether the cluster-monitoring-config body includes a
+// top-level nodeExporter key (including explicit null). Phase 1: if set, the ConfigMap wins for
+// the whole nodeExporter stanza and the ClusterMonitoring CR's nodeExporterConfig is ignored.
+func configMapYAMLDeclaresNodeExporter(cmYAML string) bool {
```
simonpasquier: The merge logic shouldn't depend on the input YAML. Can we refactor the config creation logic to avoid this?

pkg/manifests/types.go
danielmellado: https://github.com/openshift/api/pull/2837

pkg/manifests/config.go
```diff
@@ -299,10 +299,26 @@ func NewConfigFromString(content string) (*Config, error) {
...
+	return ok
+}
+
 // NewConfigFromStringAndClusterMonitoringResource returns the Config
 // initialized from the provided string and merged with the ClusterMonitoring
 // resource.
 func NewConfigFromStringAndClusterMonitoringResource(content string, cmr *configv1alpha1.ClusterMonitoring) (*Config, error) {
+	nodeExporterDeclaredByConfigMap := configMapDeclaresNodeExporter(content)
+
 	cmc := ClusterMonitoringConfiguration{
```
simonpasquier: would it be possible to avoid initialization of the node_exporter fields and move this to applyDefaults()? it would remove the need for a different treatment when merging the node_exporter configuration.

marioferh: I think so. Done

simonpasquier (commented):
another option would be turn the bool fields controlling the enabled-by-default collector to pointers.

```
type NodeExporterCollectorNetDevConfig struct {
	// A Boolean flag that enables or disables the `netdev` collector.
	Enabled *bool `json:"enabled,omitempty"`
}
...
type NodeExporterCollectorNetClassConfig struct {
	// A Boolean flag that enables or disables the `netclass` collector.
	Enabled *bool `json:"enabled,omitempty"`
	// A Boolean flag that activates the `netlink` implementation of the `netclass` collector.
	// Its default value is `true`: activating the netlink mode.
	// This implementation improves the performance of the `netclass` collector.
	UseNetlink *bool `json:"useNetlink,omitempty"`
}
```

And inside applyDefaults(), set the fields to true if they are undefined/nil.

(the systemd collector doesn't need to be included because it's disabled by default)

pkg/manifests/config_merge.go
```diff
@@ -51,9 +51,148 @@ func (c *Config) mergeClusterMonitoringCRD(clusterMonitoring *configv1alpha1.Clu
 		return err
 	}
 
+	c.mergeNodeExporterConfiguration(clusterMonitoring.Spec.NodeExporterConfig)
 	return nil
 }
 
+func clusterMonitoringNodeExporterCollectorsEmpty(col configv1alpha1.NodeExporterCollectorConfig) bool {
+	if col.CpuFreq.CollectionPolicy != "" {
```
simonpasquier: (suggestion) maybe we can use a loop to reduce the code


```
for _, pol := range []configv1alpha1.NodeExporterCollectorCollectionPolicy{
    col.CpuFreq,
    col.TcpStat,
    ...} {
    if pol != "" {
        return false
    }
}
return true
```

marioferh: done

PR #2750, opened 2025-11-19
OCPBUGS-85115, MON-4420: Enable Gateway API collection by telemeter

rikatz:
This change enables the collection of Gateway API count metric by telemeter, providing insights of the usage of Gateway API on Openshift.

The generated metric contains only a count of the amount of created Gateway resources, aggregated by the Gateway Class. In case the Gateway Class is the one managed by Openshift, the metric will have a label 'gateway_class_type' with value 'openshift'.

In case the Gateway uses a Gateway Class that is not managed by Openshift Gateway API offer, the label 'gateway_class_type' will contain the value 'not-openshift'.

* [ ] I added CHANGELOG entry for this change.
* [X] No user facing changes, so no entry in CHANGELOG was needed.

PR #2920, opened 2026-05-06
[release-4.20] OCPBUGS-85130: Enable Gateway API telemetry metrics

rikatz:
Backport from release 4.21:

This PR backports following changes required for Gateway API telemetry collection:

From https://github.com/openshift/cluster-monitoring-operator/pull/2734

* Enables the collection of Gateway API resources (Gateway Class and Gateway) using kube-state-metrics from openshift-monitoring namespace
* Adds a PrometheusRule on the same namespace (as the collection comes from ksm) to aggregate the metrics and reduce labels.

From https://github.com/openshift/cluster-monitoring-operator/pull/2750

* Enables the collection of Gateway API count metric by telemeter, providing insights of the usage of Gateway API on Openshift. The generated metric contains only a count of the amount of created Gateway resources, aggregated by the Gateway Class. In case the Gateway Class is the one managed by Openshift, the metric will have a label 'gateway_class_type' with value 'openshift', otherwise it will have a label `not-openshift`


rexagod:
/approve
/backport-risk-assessed
/verified by @rexagod 
/lgtm

@rikatz In the future, a follow-up to https://github.com/openshift/cluster-monitoring-operator/blob/main/test/e2e/kube_state_metrics_test.go#L76 for the metrics introduced in the original PR (and this one) should help us detect any regressions. Thank you for this backport!

PR #2926, opened 2026-05-15
chore: restrict NetworkPolicy ingress rules for internal-only ports

machine424:


machine424:
/retitle [MON-4577](https://redhat.atlassian.net/browse/MON-4577): chore: restrict NetworkPolicy ingress rules for internal-only ports

test/e2e/main_test.go
```diff
@@ -346,6 +346,11 @@ func TestNetworkPolicy(t *testing.T) {
 
 				var allowed []int32
 				for _, rule := range np.Spec.Ingress {
+					if len(rule.From) > 0 {
```
danielmellado, line 349: can you elaborate this a little bit more? I understand the reason, but feels totally hacky in the code, thanks! besides this it lgtm

machine424: there is
						// Won't go this way; this test is a safeguard,
						// not a validation of NP enforcement semantics.
below
(we'll not be considering those NPs, thus continue)

machine424:
cc @danielmellado

test/e2e/main_test.go
danielmellado: Instead of silently skipping rules with from selectors, we could collect those restricted ports into a separate slice and add a require.Subset check against the pod's container ports, same as you already do for allowed at line 378.

You still wouldn't probe connectivity (the test pod won't match the from selector anyway), but you'd catch stale NP rules referencing ports the pod no longer exposes.

Besides this and a bit more longer description of the PR it LGTM, but I won't block this to merge, It can be done in a follow-up. Thanks Ayoub!

danielmellado:
Also note that I'm explictly not talking about IPv6 here, as I assume out of the scope for now xD

simonpasquier:
/hold

I've got no concern with the implementation but we have an RFE about allowing an external Thanos Querier to integrate with the in-cluster monitoring components (https://redhat.atlassian.net/browse/OBSDA-1412). I'd prefer to wait for the investigation to complete before we commit on locking down the gRPC ports (it doesn't say that we can't have proper network policies but it might require some adaptation).

machine424:
opened https://github.com/openshift/cluster-monitoring-operator/pull/3032 for alertmanager only.

machine424:
test
/jira backport release-5.0

machine424:
test
/jira refresh

PR #2928, opened 2026-05-18
MON-4525: manifests: merge TelemeterClientConfig from ClusterMonitoring CRD

danielmellado:
Apply Phase 1 CR merge when cluster-monitoring-config omits
telemeterClient. Map NodeSelector, Tolerations, Resources and
TopologySpreadConstraints from the CRD.

Signed-off-by: Daniel Mellado <dmellado@fedoraproject.org>


simonpasquier:
there's a merge conflict but otherwise LGTM.

One thing that we need to clarify: there's an (undocumented) way to configure HTTP proxy in the CMO configuration and it is especially used for allowing the telemeter client to reach the Telemeter server (we had a recent thread about it internally). We need to decide if we want to support this officially or not.

simonpasquier:
@danielmellado still a conflict to resolve.

danielmellado:
> @danielmellado still a conflict to resolve.

done! missed the notification xD thanks!

simonpasquier (commented):
/lgtm

PR #2932, opened 2026-05-25
OCPBUGS-86352: jsonnet: exclude ReplicationController from catch-all …

danielmellado:
When a `DeploymentConfig` is used, the pod ownership chain is `DeploymentConfig → ReplicationController → Pod`. Two recording rules both produce a `namespace_workload_pod:kube_pod_owner:relabel` series for the same `(namespace, pod)`:
1. Our custom DeploymentConfig rule walks from `ReplicationController` to `DeploymentConfig` via `kube_replicationcontroller_owner` and emits `workload_type=deploymentconfig`.
2. The upstream catch-all rule (from kubernetes-mixin) matches any `owner_kind` not explicitly excluded. It excludes `ReplicaSet`, `DaemonSet`, `StatefulSet`, `Job`, and `Node` — but not `ReplicationController`. So it also fires and emits `workload_type=ReplicationController`.
Dashboard queries that join on `(namespace, pod)` with `group_left` then fail with: `many-to-many matching not allowed: matching labels must be unique on one side`.
Reproduced on 4.21.0-0.nightly-2026-05-24-163444.

## Fix
Exclude `owner_kind=ReplicationController` from the upstream catch-all rule via the existing `patchedRules` mechanism in `sanitize-rules.libsonnet`, so that only our custom DeploymentConfig rule produces the recording rule result for these pods.

## Testing
- Added `test/rules/OCPBUGS-86352.yaml` — promtool unit test simulating a DeploymentConfig pod, asserting exactly one series per `(namespace, pod)` with `workload_type=deploymentconfig`.
- Full `make test-rules` passes with no regressions.
- Verified fix on a live 4.21 cluster: before the fix, 2 series per pod; after, only 1.

### DIFF
--- a/assets/control-plane/prometheus-rule.yaml
+++ b/assets/control-plane/prometheus-rule.yaml
@@ -781,7 +781,7 @@ spec:
             OR
             label_replace(
               group by (cluster, namespace, pod, owner_name, owner_kind) (
-                kube_pod_owner{job="kube-state-metrics", owner_kind!="ReplicaSet", owner_kind!="DaemonSet", owner_kind!="StatefulSet", owner_kind!="Job", owner_kind!="Node", owner_kind!=""}
+                kube_pod_owner{job="kube-state-metrics", owner_kind!~"ReplicaSet|ReplicationController|DaemonSet|StatefulSet|Job|Node|"}
               )
               , "workload", "$1", "owner_name", "(.+)"
             )
--- a/jsonnet/utils/sanitize-rules.libsonnet
+++ b/jsonnet/utils/sanitize-rules.libsonnet
@@ -482,6 +482,24 @@ local patchedRules = [
       },
     ],
   },
+  {
+    // OCPBUGS-86352: exclude ReplicationController from the catch-all pod_owner rule
+    // to prevent duplicate namespace_workload_pod:kube_pod_owner:relabel series when
+    // a DeploymentConfig (which creates a ReplicationController) is used. Our custom
+    // deploymentconfig rule already handles this owner_kind.
+    name: 'k8s.rules.pod_owner',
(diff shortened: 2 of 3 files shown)


simonpasquier (commented):
It may useful for upstream to allow the customization of non-standard types?

Instead of 

```
 kube_pod_owner{
job="kube-state-metrics", 
owner_kind!="ReplicaSet", owner_kind!="ReplicationController", owner_kind!="DaemonSet", owner_kind!="StatefulSet", owner_kind!="Job", owner_kind!="Node", owner_kind!=""}
```

we could have

```
 kube_pod_owner{
job="kube-state-metrics", 
owner_kind!~"ReplicaSet|ReplicationController|DaemonSet|StatefulSet|Job|Node|"}
```

danielmellado:
> It may useful for upstream to allow the customization of non-standard types?
> 
> Instead of
> 
> ```
>  kube_pod_owner{
> job="kube-state-metrics", 
> owner_kind!="ReplicaSet", owner_kind!="ReplicationController", owner_kind!="DaemonSet", owner_kind!="StatefulSet", owner_kind!="Job", owner_kind!="Node", owner_kind!=""}
> ```
> 
> we could have
> 
> ```
>  kube_pod_owner{
> job="kube-state-metrics", 
> owner_kind!~"ReplicaSet|ReplicationController|DaemonSet|StatefulSet|Job|Node|"}
> ```

Good point, thanks Simon. Agreed on both counts.

I'll update the PR and also open a PR upstream at kubernetes-monitoring/kubernetes-mixin to propose making the excluded owner kinds configurable, so distributions that add custom owner kind rules (like our DeploymentConfig one) can extend the exclusion list without patching.

PR #2938, opened 2026-05-29
OCPBUGS-86782: fix: wait for SCC use before creating Alertmanager operands

eggfoobar:
Alertmanager pods could be denied at install when ClusterRoleBinding SCC permissions had not yet propagated to admission, causing the openshift Early SCC test to fail on slow installs (e.g. two-node fencing metal jobs).

Add WaitForClusterRoleBindingSCCUse to poll SubjectAccessReview until each bound service account can use the SCCs granted by its ClusterRole. Call it from platform and user-workload Alertmanager tasks after RBAC is applied and before the Alertmanager CR is created.

Model: Composer

## Telemetry report

```telemetry_report
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1642,8 +1642,10 @@ func (c *Client) WaitForClusterRoleBindingSCCUse(ctx context.Context, cr *rbacv1
 		namespace string
 		name      string
 		sccName   string
+		ok        bool
 	}
 
+	// Generate bucket of SCC objects to check, skip all non ServiceAccountKind
 	var checks []sccUseCheck
 	for _, subject := range crb.Subjects {
 		if subject.Kind != rbacv1.ServiceAccountKind {
@@ -1662,64 +1664,62 @@ func (c *Client) WaitForClusterRoleBindingSCCUse(ctx context.Context, cr *rbacv1
 		return nil
 	}
 
+	// Helper function to execute check, currently only needed by WaitForClusterRoleBindingSCCUse
+	subjectAccessReviewSCCAllowed := func(ctx context.Context, user, sccName string) (bool, string, error) {
+		sar, err := c.kclient.AuthorizationV1().SubjectAccessReviews().Create(ctx, &authzv1.SubjectAccessReview{
+			Spec: authzv1.SubjectAccessReviewSpec{
+				User: user,
+				ResourceAttributes: &authzv1.ResourceAttributes{
+					Group:    secv1.GroupName,
+					Resource: "securitycontextconstraints",
+					Name:     sccName,
+					Verb:     "use",
+				},
+			},
+		}, metav1.CreateOptions{})
+		if err != nil {
+			return false, "", err
+		}
+
+		return sar.Status.Allowed, sar.Status.Reason, nil
+	}
+
 	pending := append([]sccUseCheck(nil), checks...)
 	var lastErr error
 	if err := Poll(ctx, func(ctx context.Context) (bool, error) {
-		// avoid creating access reviews on already successful runs
-		var stillPending []sccUseCheck
-		for _, check := range pending {
-			allowed, reason, err := c.subjectAccessReviewSCCAllowed(ctx, check.namespace, check.name, check.sccName)
-			if err != nil {
-				lastErr = err
-				klog.V(4).ErrorS(err, "WaitForClusterRoleBindingSCCUse: failed to create SubjectAccessReview")
-				stillPending = append(stillPending, check)
-				continue
-			}
-			if allowed {
+		successfullyCheckedAll := true
+		for i, check := range pending {
+			// skip successful runs
+			if pending[i].ok {
 				continue
 			}
 
 			user := fmt.Sprintf("system:serviceaccount:%s:%s", check.namespace, check.name)
-			if reason != "" {
-				lastErr = fmt.Errorf("%s cannot use SCC %q: %s", user, check.sccName, reason)
-			} else {
-				lastErr = fmt.Errorf("%s cannot use SCC %q", user, check.sccName)
+			allowed, reason, err := subjectAccessReviewSCCAllowed(ctx, user, check.sccName)
+			if err != nil {
+				lastErr = fmt.Errorf("failed to check access for service account %s/%s and SCC %s: %w", check.namespace, check.name, check.sccName, err)
+				klog.V(4).ErrorS(err, "WaitForClusterRoleBindingSCCUse: failed to create SubjectAccessReview")
+				successfullyCheckedAll = false
+				continue
 			}
-			stillPending = append(stillPending, check)
+			if allowed {
+				pending[i].ok = true
+				continue
+			}
+
+			successfullyCheckedAll = false
+			if reason == "" {
+				reason = "no reason"
+			}
+			lastErr = fmt.Errorf("%s cannot use SCC %q: %s", user, check.sccName, reason)
 		}
 
-		pending = stillPending
-		if len(pending) == 0 {
-			return true, nil
-		}
-
-		return false, nil
+		return successfullyCheckedAll, nil
 	}, WithPollInterval(time.Second), WithPollTimeout(30*time.Second), WithLastError(&lastErr)); err != nil {
 		return fmt.Errorf("waiting for ClusterRoleBinding %q SCC permissions: %w", crb.Name, err)
 	}
 
 	return nil
-}
-
-func (c *Client) subjectAccessReviewSCCAllowed(ctx context.Context, namespace, name, sccName string) (bool, string, error) {
-	user := fmt.Sprintf("system:serviceaccount:%s:%s", namespace, name)
-
-	sar, err := c.kclient.AuthorizationV1().SubjectAccessReviews().Create(ctx, &authzv1.SubjectAccessReview{
-		Spec: authzv1.SubjectAccessReviewSpec{
-			User: user,
-			ResourceAttributes: &authzv1.ResourceAttributes{
-				Group:    secv1.GroupName,
-				Resource: "securitycontextconstraints",
-				Name:     sccName,
-				Verb:     "use",
-			},
-		},
-	}, metav1.CreateOptions{})
-	if err != nil {
-		return false, "", err
-	}
-
-	return sar.Status.Allowed, sar.Status.Reason, nil
 }
 
 func (c *Client) CreateOrUpdateServiceAccount(ctx context.Context, sa *v1.ServiceAccount) error {
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1607,25 +1607,28 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
 	return err
 }
 
-func (c *Client) WaitForClusterRoleBindingSCCUse(ctx context.Context, crb *rbacv1.ClusterRoleBinding) error {
-	if crb.RoleRef.Kind != "ClusterRole" {
+func (c *Client) WaitForClusterRoleBindingSCCUse(ctx context.Context, cr *rbacv1.ClusterRole, crb *rbacv1.ClusterRoleBinding) error {
+	if cr == nil || crb == nil {
+		return nil
+	}
+	if crb.RoleRef.Kind != "ClusterRole" || crb.RoleRef.Name != cr.Name {
 		return nil
 	}
 
-	cr, err := c.kclient.RbacV1().ClusterRoles().Get(ctx, crb.RoleRef.Name, metav1.GetOptions{})
-	if err != nil {
-		return fmt.Errorf("getting ClusterRole %q for SCC propagation wait: %w", crb.RoleRef.Name, err)
+	// account for wildcard rules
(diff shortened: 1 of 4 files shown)


eggfoobar:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-metal-ovn-two-node-fencing-upgrade periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ovn-two-node-fencing-dualstack periodic-ci-openshift-release-main-nightly-4.23-e2e-metal-ovn-two-node-fencing-dualstack

eggfoobar:
/payload-aggregate periodic-ci-openshift-release-main-nightly-4.22-e2e-metal-ovn-two-node-fencing-upgrade 3

pkg/client/client.go
```diff
@@ -1606,6 +1607,86 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
 	return err
 }
 
+func (c *Client) WaitForClusterRoleBindingSCCUse(ctx context.Context, crb *rbacv1.ClusterRoleBinding) error {
+	if crb.RoleRef.Kind != "ClusterRole" {
+		return nil
+	}
+
+	cr, err := c.kclient.RbacV1().ClusterRoles().Get(ctx, crb.RoleRef.Name, metav1.GetOptions{})
```
simonpasquier: do we really need to get the cluster role? maybe the task can pass the cluster role definition?

pkg/client/client.go
```diff
@@ -1606,6 +1607,86 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+	if err != nil {
+		return fmt.Errorf("getting ClusterRole %q for SCC propagation wait: %w", crb.RoleRef.Name, err)
+	}
+
+	var sccNames []string
+	for _, rule := range cr.Rules {
+		if !slices.Contains(rule.Verbs, "use") {
+			continue
+		}
+		if !slices.Contains(rule.APIGroups, "security.openshift.io") {
```
simonpasquier: we can use `secv1.GroupName`

pkg/client/client.go
```diff
@@ -1606,6 +1607,86 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+	}
+
+	return nil
+}
+
+func (c *Client) waitForServiceAccountSCCUse(ctx context.Context, namespace, name, sccName string) error {
+	user := fmt.Sprintf("system:serviceaccount:%s:%s", namespace, name)
+	var lastReason string
+
+	err := Poll(ctx, func(ctx context.Context) (bool, error) {
```
simonpasquier: we'd need to use WithLastError() and return no error from the condition function directly (see WaitForThanosRuler() for an example).

pkg/client/client.go
```diff
@@ -1606,6 +1607,86 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
 	return err
 }
 
+func (c *Client) WaitForClusterRoleBindingSCCUse(ctx context.Context, crb *rbacv1.ClusterRoleBinding) error {
```
simonpasquier: I'd expect that Poll() to wrap the whole body of the function.

pkg/client/client.go
```diff
@@ -1606,6 +1607,86 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+	var sccNames []string
+	for _, rule := range cr.Rules {
+		if !slices.Contains(rule.Verbs, "use") {
+			continue
+		}
+		if !slices.Contains(rule.APIGroups, "security.openshift.io") {
+			continue
+		}
+		if !slices.Contains(rule.Resources, "securitycontextconstraints") {
+			continue
```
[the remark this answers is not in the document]
eggfoobar: Done

pkg/client/client.go
eggfoobar: Oh thats a great suggestion, updated, thanks!

pkg/client/client.go
eggfoobar: Updated, thanks!

pkg/client/client.go
eggfoobar: Oh interesting, sounds good, I went ahead and updated the polling function to match the existing patterns better, thanks!

pkg/client/client.go
eggfoobar: I went ahead and moved the polling into the main calling function, was that what you were thinking here?

pkg/client/client.go
```diff
@@ -1606,6 +1607,121 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+	for _, rule := range cr.Rules {
+		if !matches(rule.Verbs, "use") {
+			continue
+		}
+		if !matches(rule.Verbs, secv1.GroupName) {
+			continue
+		}
+		if !matches(rule.Verbs, "securitycontextconstraints") {
+			continue
+		}
```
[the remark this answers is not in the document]
eggfoobar: oops

pkg/client/client.go
```diff
@@ -1606,6 +1607,121 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+		sccNames = append(sccNames, rule.ResourceNames...)
+	}
+	if len(sccNames) == 0 {
+		return nil
+	}
+
+	type sccUseCheck struct {
+		namespace string
+		name      string
+		sccName   string
```
simonpasquier, line 1644: (suggestion) add a bool field to record when we know that the permission has been granted and we have to retry. It avoids maintaining 2 additional slices (pending & stillPending).
```suggestion
		sccName   string
		ok        bool
```

pkg/client/client.go
```diff
@@ -1606,6 +1607,121 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+
+	pending := append([]sccUseCheck(nil), checks...)
+	var lastErr error
+	if err := Poll(ctx, func(ctx context.Context) (bool, error) {
+		// avoid creating access reviews on already successful runs
+		var stillPending []sccUseCheck
+		for _, check := range pending {
+			allowed, reason, err := c.subjectAccessReviewSCCAllowed(ctx, check.namespace, check.name, check.sccName)
+			if err != nil {
+				lastErr = err
```
simonpasquier:
```suggestion
				lastErr = fmt.Errorf("failed to check access for service account %s/%s and SCC %s: %w", namespace, name, sccName, err)
```

pkg/client/client.go
```diff
@@ -1606,6 +1607,121 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+				klog.V(4).ErrorS(err, "WaitForClusterRoleBindingSCCUse: failed to create SubjectAccessReview")
+				stillPending = append(stillPending, check)
+				continue
+			}
+			if allowed {
+				continue
+			}
+
+			user := fmt.Sprintf("system:serviceaccount:%s:%s", check.namespace, check.name)
```
simonpasquier: (nit) subjectAccessReviewSCCAllowed() could take a username instead of namespace & name.

pkg/client/client.go
```diff
@@ -1606,6 +1607,121 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+			if reason != "" {
+				lastErr = fmt.Errorf("%s cannot use SCC %q: %s", user, check.sccName, reason)
+			} else {
+				lastErr = fmt.Errorf("%s cannot use SCC %q", user, check.sccName)
+			}
```
simonpasquier: (nit)
```suggestion
			if reason == "" {
				reason = "no reason"
			}
			lastErr = fmt.Errorf("%s cannot use SCC %q: %s", user, check.sccName, reason)
```

pkg/client/client.go
```diff
@@ -1606,6 +1607,121 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+
+		return false, nil
+	}, WithPollInterval(time.Second), WithPollTimeout(30*time.Second), WithLastError(&lastErr)); err != nil {
+		return fmt.Errorf("waiting for ClusterRoleBinding %q SCC permissions: %w", crb.Name, err)
+	}
+
+	return nil
+}
+
+func (c *Client) subjectAccessReviewSCCAllowed(ctx context.Context, namespace, name, sccName string) (bool, string, error) {
```
simonpasquier: the function becomes less useful if we pass it a username instead of namespace & name (e.g. I would be ok if it's inlined in WaitForClusterRoleBindingSCCUse).

pkg/client/client.go
eggfoobar: Ah great suggestion, updated

pkg/client/client.go
eggfoobar: Thats a good point, I went ahead and moved things here to be more inside of this function call since this is the only one that is currently using it and took your suggestion for formatting this user type.

pkg/client/client.go
eggfoobar: I like it, updated

pkg/client/client.go
eggfoobar: Thats a fair point, I moved this inside the function so it's a bit more self contained now

pkg/client/client.go
eggfoobar: Updated

eggfoobar:
/payload-aggregate periodic-ci-openshift-release-main-nightly-4.22-e2e-metal-ovn-two-node-fencing-upgrade 3

eggfoobar:
/verified by CI

Aggregated runs all passed

pkg/client/client.go
```diff
@@ -1606,6 +1607,121 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+			user := fmt.Sprintf("system:serviceaccount:%s:%s", check.namespace, check.name)
+			allowed, reason, err := subjectAccessReviewSCCAllowed(ctx, user, check.sccName)
+			if err != nil {
+				lastErr = fmt.Errorf("failed to check access for service account %s/%s and SCC %s: %w", check.namespace, check.name, check.sccName, err)
+				klog.V(4).ErrorS(err, "WaitForClusterRoleBindingSCCUse: failed to create SubjectAccessReview")
+				successfullyCheckedAll = false
+				continue
+			}
+			if allowed {
+				check.ok = true
```
simonpasquier: otherwise we modify the copy, not the original value.
```suggestion
				check[i].ok = true
```

pkg/client/client.go
```diff
@@ -1606,6 +1607,121 @@ func (c *Client) CreateOrUpdateClusterRoleBinding(ctx context.Context, crb *rbac
...
+				continue
+			}
+
+			successfullyCheckedAll = false
```
simonpasquier, line 1710: we could return on the first error to simplify the flow.

simonpasquier:
btw we have other components using SCCs other than the default `restricted-v2` (namely Prometheus, node_exporter and ThanosRuler). Shouldn't we also implement a similar check for them?

pkg/client/client.go
eggfoobar: Ah good point, fixed

pkg/client/client.go
eggfoobar: I think this might be fine as is, so we still loop to other items in the list and just flag that we're not done yet when we've looped through all, wdyt?

eggfoobar:
> btw we have other components using SCCs other than the default `restricted-v2` (namely Prometheus, node_exporter and ThanosRuler). Shouldn't we also implement a similar check for them?

That's a good point, I kept this focused on only this small part that was being triggered to avoid over complicating other aspects of the code unless needed, since we might want to backport this to 4.22. Ideally, unless we see more odd behavior in CI I would keep other flows as is to help identify things that might warrant attention, wdyt?

eggfoobar:
/payload-aggregate periodic-ci-openshift-release-main-nightly-4.22-e2e-metal-ovn-two-node-fencing-upgrade 3

simonpasquier:
> > btw we have other components using SCCs other than the default `restricted-v2` (namely Prometheus, node_exporter and ThanosRuler). Shouldn't we also implement a similar check for them?
> 
> That's a good point, I kept this focused on only this small part that was being triggered to avoid over complicating other aspects of the code unless needed, since we might want to backport this to 4.22. Ideally, unless we see more odd behavior in CI I would keep other flows as is to help identify things that might warrant attention, wdyt?

Agreed. Can we still capture the information in the Jira though?

pkg/client/client.go
simonpasquier: My personal preference would be to keep a simple algorithm (no need for the intermediate successfullyCheckedAll variable if you return on the first error) and also because a transient error shouldn't be a frequent event. But since the code is already working, no need to change it.

simonpasquier (commented):
/lgtm

PR #2904, opened 2026-04-28
MON-4581: Revert TRT-2644, drop `kube_running_pod_ready`, and introduce telemetry changes

rexagod:
#### **Revisit https://github.com/openshift/cluster-monitoring-operator/pull/2895 once this is in.**
***
#### Revisit this once https://github.com/openshift/cluster-monitoring-operator/pull/2950 is in. That's a subset of this that needs to go in first. Putting this in draft for now.
***
This reverts commit e33a447, and addresses the issue.

`kube_running_pod_ready` is not a metric owned by KSM, as the prefix in
the name would suggest, but a recording rule in CMO. This led to it
being included in the telemetry metrics list which KSM's telemetry, and,
by association, minimal service monitors use.

While this had no effect just because of that, the test in origin that
covers this case extracted these metrics from SMs, and performed a count
over them, which gave rise to the [additional 301 samples](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview-serial-1of3/2048838049839714305) being present,
see [1] for more info.

[1]: https://github.com/openshift/origin/blob/main/test/extended/prometheus/collection_profiles.go#L298

PR #2950, opened 2026-06-08
MON-4581: Update component minimal profiles

rexagod:
#### Once this is merged, https://github.com/openshift/cluster-monitoring-operator/pull/2904 will go in with the reduced set of changes (PTAL at the second point below).
***
Steps taken for this commit:
* Squash all commits in #2904
* Only keep changes related to minimal profiles for CMO components

The commit messages below are from the first step, and kept for the record.

chore: `make generate`

[TRT-2644](https://redhat.atlassian.net/browse/TRT-2644): Drop `kube_running_pod_ready` from KSM's minimal and telemetry SMs

`kube_running_pod_ready` is not a metric owned by KSM, as the prefix in the name would suggest, but a recording rule in CMO. This led to it being included in the telemetry metrics list which KSM's telemetry, and, by association, minimal service monitors use.

While this had no effect just because of that, the test in origin that covers this case extracted these metrics from SMs, and performed a count over them, which gave rise to the additional 301 samples being present, see [1] for more info.

[1]: https://github.com/openshift/origin/blob/main/test/extended/prometheus/collection_profiles.go#L298

fix: Comment on collection profiles definitions

chore: `make versions`

chore: `make generate`


pkg/manifests/types.go
```diff
@@ -30,7 +30,7 @@ const (
 	// FullCollectionProfile collects all metrics.
 	FullCollectionProfile = "full"
 
-	// MinimalCollectionProfile collects only metrics used by recording/alerting, dashboards and Telemetry.
+	// MinimalCollectionProfile collects metrics used by recording, alerting, dashboards and Telemetry.
```
simonpasquier:
```suggestion
	// MinimalCollectionProfile collects metrics used by recording and alerting rules, dashboards and Telemetry.
```

simonpasquier (commented):
/lgtm

rexagod:
Resolved conflicts and rebased.

simonpasquier (commented):
/lgtm
/verified by tests

PR #2953, opened 2026-06-08
MON-4523:  ClusterMonitoring prometheusConfig

marioferh:


pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,466 @@
...
+}
+
+func collectionProfileCRDToManifest(cp configv1alpha1.CollectionProfile) CollectionProfile {
+	switch cp {
+	case configv1alpha1.CollectionProfileFull:
+		return FullCollectionProfile
+	case configv1alpha1.CollectionProfileMinimal:
+		return MinimalCollectionProfile
+	default:
+		return ""
```
simonpasquier: should we return an error or default to full? Similar remarks for the other conversion functions.

pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,466 @@
...
+		Profile: sigv4.Profile,
+		RoleArn: sigv4.RoleArn,
+	}
+	out.AccessKey = secretKeySelectorFromCRD(sigv4.AccessKey)
+	out.SecretKey = secretKeySelectorFromCRD(sigv4.SecretKey)
+	return out
+}
+
+func relabelActionString(action configv1alpha1.RelabelAction) string {
+	return strings.ToLower(string(action))
```
simonpasquier: (nit) no need to lower I suspect

pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,466 @@
...
+	out := make([]monv1.RelabelConfig, 0, len(configs))
+	for _, rc := range configs {
+		out = append(out, relabelConfigFromCRD(rc))
+	}
+	return out
+}
+
+func applyRemoteWriteAuthorizationFromCRD(auth configv1alpha1.RemoteWriteAuthorization, dst *RemoteWriteSpec) {
+	switch auth.Type {
+	case configv1alpha1.RemoteWriteAuthorizationTypeBearerToken:
```
simonpasquier: in practice, I don't think that we need this option because it's functionally equivalent to Authorization

pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,466 @@
...
+			}
+		}
+	case configv1alpha1.RemoteWriteAuthorizationTypeOAuth2:
+		dst.OAuth2 = oauth2FromCRD(auth.OAuth2)
+	case configv1alpha1.RemoteWriteAuthorizationTypeSigV4:
+		dst.Sigv4 = sigv4FromCRD(auth.Sigv4)
+	case configv1alpha1.RemoteWriteAuthorizationTypeSafeAuthorization:
+		if auth.SafeAuthorization != nil {
+			dst.Authorization = &monv1.SafeAuthorization{
+				Type:        "Bearer",
```
simonpasquier: Type shouldn't be hardcoded

pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,466 @@
...
+				Username: *username,
+				Password: *password,
+			}
+		}
+	case configv1alpha1.RemoteWriteAuthorizationTypeOAuth2:
+		dst.OAuth2 = oauth2FromCRD(auth.OAuth2)
+	case configv1alpha1.RemoteWriteAuthorizationTypeSigV4:
+		dst.Sigv4 = sigv4FromCRD(auth.Sigv4)
+	case configv1alpha1.RemoteWriteAuthorizationTypeSafeAuthorization:
+		if auth.SafeAuthorization != nil {
```
simonpasquier: I would have expected that SafeAuthorization is never nil when the type is Authorization. Also the field doesn't need to include `Safe`.

pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,466 @@
...
+	cfg.Tolerations = pc.Tolerations
+	cfg.TopologySpreadConstraints = pc.TopologySpreadConstraints
+	cfg.Resources = containerResourcesFromCRD(pc.Resources)
+	cfg.QueryLogFile = pc.QueryLogFile
+	cfg.ExternalLabels = externalLabelsFromCRD(pc.ExternalLabels)
+	cfg.AlertmanagerConfigs = additionalAlertmanagerConfigsFromCRD(pc.AdditionalAlertmanagerConfigs)
+	cfg.RemoteWrite = remoteWriteSpecsFromCRD(pc.RemoteWrite)
+
+	if pc.EnforcedBodySizeLimitBytes > 0 {
+		cfg.EnforcedBodySizeLimit = strconv.FormatInt(pc.EnforcedBodySizeLimitBytes, 10)
```
simonpasquier: (nit)
```suggestion
		cfg.EnforcedBodySizeLimit = strconv.Itoa(pc.EnforcedBodySizeLimitBytes)
```

pkg/manifests/config_merge_prometheus.go
marioferh: unknown enum values now return an error during merge (consistent with ConfigMap validation). 

Empty fields are still left unset so applyDefaults() applies full.

agree?

pkg/manifests/config_merge_prometheus.go
marioferh: done

pkg/manifests/config_merge_prometheus.go
marioferh: I remember some discussion around this.

But maybe we can remove the BearerToken branch, and SafeAuthorization covers the same case. 

Users should use type: SafeAuthorization with the safeAuthorization secret reference instead.

pkg/manifests/config_merge_prometheus.go
marioferh: done

pkg/manifests/config_merge_prometheus.go
marioferh: CMO now returns an error when authorization credentials are missing instead of silently skipping. 

The safe removal needs an API change. I will follow up a change there.

marioferh:
https://github.com/openshift/api/pull/2901

pkg/manifests/config_merge_prometheus.go
simonpasquier: :+1:

pkg/manifests/config_merge_prometheus.go
simonpasquier: Yes I would avoid having 2 fields with one (Authorization) being a superset of the other (BearerToken). They both exist upstream because support for Authorization was added after and we can't remove a field from a stable API.

marioferh:
Waiting for API

pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,551 @@
...
+}
+
+func applyRemoteWriteAuthorizationFromCRD(auth configv1alpha1.RemoteWriteAuthorization, dst *RemoteWriteSpec) error {
+	switch auth.Type {
+	case "":
+		return nil
+	case configv1alpha1.RemoteWriteAuthorizationTypeBasicAuth:
+		username := secretKeySelectorFromCRD(auth.BasicAuth.Username)
+		password := secretKeySelectorFromCRD(auth.BasicAuth.Password)
+		if username != nil && password != nil {
```
danielmellado: If only one of username or password is provided (the other Secret name is empty), this silently does nothing.This should return an error instead:

pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,551 @@
...
+			dst.BasicAuth = &monv1.BasicAuth{
+				Username: *username,
+				Password: *password,
+			}
+		}
```
danielmellado, line 413: Are there any tests for this?

pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,551 @@
...
+	}
+	cfg.AlertmanagerConfigs = alertmanagerConfigs
+	remoteWrite, err := remoteWriteSpecsFromCRD(pc.RemoteWrite)
+	if err != nil {
+		return err
+	}
+	cfg.RemoteWrite = remoteWrite
+
+	if pc.EnforcedBodySizeLimitBytes > 0 {
+		cfg.EnforcedBodySizeLimit = strconv.Itoa(int(pc.EnforcedBodySizeLimitBytes))
```
danielmellado: (nit) pc.EnforcedBodySizeLimitBytes is int64 in the API but gets narrowed through int() here. The API caps it at 1GB so it's safe in practice, but `strconv.FormatInt(pc.EnforcedBodySizeLimitBytes, 10)` avoids the narrowing cast entirely and is more explicit.

pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,551 @@
...
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
+)
+
+const serviceAccountTokenPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+
+// clusterMonitoringPrometheusSpecEmpty reports whether the CR's prometheusConfig stanza
+// contains no user-set field.
+func clusterMonitoringPrometheusSpecEmpty(pc configv1alpha1.PrometheusConfig) bool {
```
danielmellado, line 33: (nit) As we're manually mirroring every field of PrometheusConfig. If new fields are added to the API type, this will silently drift. IMHO we can add a comment like // NOTE: must be updated when PrometheusConfig fields change in openshift/api.

pkg/manifests/config_merge_prometheus.go
```diff
@@ -0,0 +1,551 @@
...
+			Headers:  remoteWriteHeadersFromCRD(rw.Headers),
+		}
+		if rw.RemoteTimeoutSeconds > 0 {
+			cfg.RemoteTimeout = fmt.Sprintf("%ds", rw.RemoteTimeoutSeconds)
+		}
+		if rw.ExemplarsMode != "" {
+			switch rw.ExemplarsMode {
+			case configv1alpha1.ExemplarsModeSend:
+				cfg.SendExemplars = ptr.To(true)
+			case configv1alpha1.ExemplarsModeDoNotSend:
```
danielmellado, line 463: Worth a brief comment stating this is intentional and that nil/unset means "do not send" in the prometheus-operator API, so future readers don't think it was forgotten.

danielmellado (changes requested):
Overall LGTM, just a few nits, thanks!

pkg/manifests/config_merge_prometheus.go
marioferh: Agree, added comment

pkg/manifests/config_merge_prometheus.go
marioferh: Agree.
added a brief comment on the DoNotSend

pkg/manifests/config_merge_prometheus.go
marioferh: added

pkg/manifests/config_merge_prometheus.go
marioferh: Fixed — partial BasicAuth now returns an error

pkg/manifests/config_merge_prometheus.go
marioferh: Ok.
Using strconv.FormatInt directly.

PR #2957, opened 2026-06-11
OCPBUGS-87800: fix(TestDocExamples) flake: use internal registry for test pods

machine424:
The ose-cli image (~584MB!!) pulled from registry.redhat.io can take over
    60s, causing TestDocExamples to flake when the pod stays
    in Pending past the poll timeout. Switch to the cli image from the
    cluster's internal registry (openshift/cli) to avoid external pulls.

ALSO

    fix(TestNetworkPolicy): use internal registry for test pods
    
    Same as the TestDocExamples fix: switch runProbePod from
    registry.redhat.io/openshift4/ose-cli:latest to the internal registry
    path to avoid slow external image pulls in CI.

## Telemetry report

```telemetry_report
```

### DIFF
--- a/test/e2e/doc_examples_test.go
+++ b/test/e2e/doc_examples_test.go
@@ -109,7 +109,7 @@ func TestDocExamples(t *testing.T) {
 							Containers: []corev1.Container{
 								{
 									Name:            containerName,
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
/lgtm

PR #2958, opened 2026-06-11
OCPBUGS-87799: fix TestTelemetryReport flake by waiting for rate() on  raw metrics

machine424:
The test previously waited for recording rules that apply rate() on other recording rules (3s evaluation interval), which need far fewer samples than rate() on raw scraped metrics (30s scrape interval).

When Prometheus restarts and because emptyDir is used the TSDB, the old waitForQuery calls could return before a second raw scrape occurred, causing rate-based recording rules to report 0 timeseries and fail the golden file comparison.

In the linked job run, Prometheus restarted right before TestTelemetryReport. The fresh pod had an empty TSDB with only one scrape worth of data.

Wait for e2etelemetry:suffix_ok:rate instead, which applies rate() directly on a raw counter. This won't return data until at least two scrapes have landed, which also implicitly confirms that the PrometheusRule is loaded and targets are being scraped.

## Telemetry report

```telemetry_report
```

### DIFF
--- a/test/e2e/telemetry_report_test.go
+++ b/test/e2e/telemetry_report_test.go
@@ -55,11 +55,11 @@ func TestTelemetryReport(t *testing.T) {
 	t.Cleanup(cleanup)
 	promURL := fmt.Sprintf("http://%s", host)
 
-	// Wait for recording rules to produce data: a trivial one and a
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
/lgtm

PR #2962, opened 2026-06-16
OCPBUGS-90089: chore(test): clean up remote write config in TestPrometheusRemoteWrite

machine424:
TestPrometheusRemoteWrite was not cleaning up the cluster-monitoring-config ConfigMap after its subtests. The last subtest deleted the rwe2e Prometheus receiver but left the remoteWrite section in the ConfigMap, causing Prometheus to accumulate failed remote write samples in its queue_manager.

When a later test (e.g. TestTelemeterRemoteWrite) reset the ConfigMap, Prometheus Operator removed the TLS certs from its TLS assets secret simultaneously with the config update. Because Prometheus re-reads cert files on every send attempt (prometheus/common/config.tlsRoundTripper), the queue_manager drain failed with "no such file or directory" and retried until the flush-deadline expired (2x1m for samples + metadata = ~2m), blocking pod shutdown and causing downstream test timeouts.

Add a t.Cleanup that resets the ConfigMap, waits for the remote write config to be removed from the Prometheus config secret, then force-deletes the Prometheus pods to skip the stalled drain.

In practice, TestTelemeterRemoteWrite ran next and had to absorb this ~2m penalty. Increasing its timeout is not a robust fix: test ordering can change and a different test would inherit the same problem.

Failing run (if link still works):
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-monitoring-operator/2959/pull-ci-openshift-cluster-monitoring-operator-main-e2e-agnostic-operator/2065784368860762112

## Telemetry report

```telemetry_report
```

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -30,6 +30,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
 )
 
 func TestPrometheusMetrics(t *testing.T) {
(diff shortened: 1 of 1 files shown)


machine424:
/retitle OCPBUGS-90089: chore(test): clean up remote write config in TestPrometheusRemoteWrite

simonpasquier (commented):
/lgtm

PR #2972, opened 2026-06-25
OCPBUGS-92085: set Prometheus shards value explicitly

simonpasquier:
As explained in
https://github.com/prometheus-operator/prometheus-operator/issues/6291, the VPA expects the `.spec.shards` field to be set to an explicit value to scale the resource.

### DIFF
--- a/assets/prometheus-k8s/prometheus.yaml
+++ b/assets/prometheus-k8s/prometheus.yaml
@@ -222,6 +222,7 @@ spec:
     matchLabels:
       openshift.io/cluster-monitoring: "true"
   serviceMonitorSelector: {}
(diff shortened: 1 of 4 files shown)


machine424:
/lgtm
/hold
If you could add some context (maybe in https://github.com/prometheus-operator/prometheus-operator/issues/6291) on why making prom-operator set the default value as it does for `replicas` is not possible/that straightforward. Thanks.

simonpasquier:
/unhold

I'll check more closely if we can have a default value for shards in upstream but I'm always a bit nervous to change the API definition (yeah I know, it should be safe but that might still create issues in some edge situations).

PR #2976, opened 2026-06-29
OCPBUGS-93722:[release-4.22] Prometheus: Information disclosure of Azure OAuth client secret via config API 

Pranavgarud06:
While working on the CVE-2026-42151 I have observed (github.com/prometheus/prometheus) was using the vulnerable version. So, in order to resolve this issue I have bumped the prometheus package to the patched version (v0.312.0).


simonpasquier:
/hold 

we need the fix to land on the main branch first.

PR #2977, opened 2026-06-29
MON-4413: support MessageVersion v2.0 for remote-write

simonpasquier:


pkg/manifests/config_test.go
```diff
@@ -868,6 +868,69 @@ func TestAdditionalResourceLabelsValidation(t *testing.T) {
...
+		{
+			name: "cluster monitoring: omitted messageVersion is valid",
+			config: `prometheusK8s:
+  remoteWrite:
+  - url: http://example.com
+`,
+		},
+		{
+			name:   "cluster monitoring: V1.0 is valid",
+			config: "prometheusK8s:\n  remoteWrite:\n  - url: http://example.com\n    messageVersion: V1.0\n",
```
machine424: nit: could we format these as multiline strings instead? It would make them more readable for humans :)

pkg/manifests/manifests.go
```diff
@@ -3560,6 +3560,9 @@ func addRemoteWriteConfigs(clusterID string, rw []monv1.RemoteWriteSpec, rwTarge
 				SafeAuthorization: *target.Authorization,
 			}
 		}
+		if target.MessageVersion != "" {
+			rwConf.MessageVersion = new(monv1.RemoteWriteMessageVersion(target.MessageVersion))
```
danielmellado, line 3564: FYI for backports: `new(expr)` needs Go 1.26. On a 1.25 branch this would need `ptr.To(...)` instead,if at some point we'd like to backport this to pre go1.26, we'd need to change it.

test/e2e/prometheus_test.go
```diff
@@ -257,6 +257,24 @@ func TestPrometheusRemoteWrite(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "assert remote write with message version v2.0",
+			rwSpec: `
+  - url: https://%[1]s/api/v1/write
```
danielmellado, line 212: Wouldn't we need also `tlsConfig.ca`, which the other HTTPS remote-write cases in this table all set? IIUC Without that, we might end up testing TLS failure against the self-signed receiver rather than `messageVersion` itself.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/cluster-monitoring-operator
 
-go 1.25.0
+go 1.26.0
```
danielmellado, line 3: again, question on the backport here ;)

danielmellado (commented):
Left some comments, thanks!

pkg/manifests/manifests.go
simonpasquier: I don't think that we'll backport since it's a new feature and not a bug fix.

test/e2e/prometheus_test.go
simonpasquier: hmm you're right that it should fail the test but I suspect that we have a pre-existing flaw in the test. Checking right now.

simonpasquier: ah since the in-cluster Prometheus uses the config from the previous sub-test so as soon as recreated the Prometheus receiver will get the metrics...

simonpasquier: updated now

go.mod
simonpasquier: same answer :)

danielmellado:
Thanks Simon ;)

pkg/manifests/manifests.go
```diff
@@ -3560,6 +3560,9 @@ func addRemoteWriteConfigs(clusterID string, rw []monv1.RemoteWriteSpec, rwTarge
 				SafeAuthorization: *target.Authorization,
 			}
 		}
+		if target.MessageVersion != "" {
```
machine424, line 3563: yes, I think CMO should enforce the default version value in case prometheus-operator starts using v2 as the default one or if prometheus changes the default value to v2.

pkg/manifests/manifests_test.go
```diff
@@ -1452,6 +1452,50 @@ func TestRemoteWriteAuthorizationConfig(t *testing.T) {
...
+		expectedVersion *monv1.RemoteWriteMessageVersion
+	}{
+		{
+			name:           "defaults to V1.0 when not set",
+			messageVersion: "",
+		},
+		{
+			name:            "V1.0 is passed through",
+			messageVersion:  "V1.0",
+			expectedVersion: new(monv1.RemoteWriteMessageVersion1_0),
```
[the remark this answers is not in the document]
machine424: are you sure abut this?

test/e2e/framework/prometheus.go
```diff
@@ -49,7 +46,7 @@ func (f Framework) MakePrometheusWithWebTLSRemoteReceive(name, tlsSecretName str
 		Spec: monitoringv1.PrometheusSpec{
 			CommonPrometheusFields: monitoringv1.CommonPrometheusFields{
 				Image:                     image,
-				Replicas:                  &replicas,
+				Replicas:                  new(int32(1)),
```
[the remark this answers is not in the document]
machine424: are you really sure about this??

go.mod
```diff
@@ -1,185 +1,186 @@
...
```
machine424, line 3: I assume this was needed for some new syntax or sth :)

machine424 (commented):
thanks, just some propositions.

pkg/manifests/manifests.go
simonpasquier: Right now, the operator assumes no default and delegates to Prometheus. I assume that changing the default in Prometheus will only happen in a new major version at which time we would make the proper decision anyway.

go.mod
simonpasquier: [yep](https://go.dev/doc/go1.26#language) and coding assistants are confused by the new syntax as they lack 1.26 knowledge :)

pkg/manifests/manifests.go
simonpasquier: > the operator assumes no default

This isn't quite right: it needs to rely on the Prometheus default value to make some decisions anyway though it won't explicitly set the parameter in the generated configuration if unset in the resource.

machine424: I see, but I still think that being explicit on CMO side (like we already do for other defaults) is safer and should cost nothing.

test/e2e/framework/prometheus.go
```diff
@@ -33,10 +35,29 @@ import (
 	"k8s.io/utils/ptr"
 )
 
+// WaitForPrometheusUpdate reads the current generation from the given
+// Prometheus resource and polls the Kubernetes API until the resource's
+// generation is incremented and it matches with the status observedGeneration.
+func (f Framework) WaitForPrometheusUpdate(ctx context.Context, p *monitoringv1.Prometheus) error {
```
machine424, line 41: I assume this is only needed for a "clearer failure message" (even though the more explicit would be in the pods logs).
Otherwise the config propagation delay would be absorbed into the remoteWriteCheckMetrics 6 min wait window.

(I like the idea of WaitForPrometheusUpdate's generations comparisons, I may use it for a new test )

machine424 (commented):
/lgtm
/hold
we can always follow up.

test/e2e/framework/prometheus.go
simonpasquier: correct this is to distinguish between "the CMO config couldn't be applied" (for instance a secret ref is invalid) vs. "the config is applied but it didn't produce the expected result".

simonpasquier:
/test e2e-agnostic-operator

CMO tests were affected by a Telemeter outage.

simonpasquier:

the test is permafailing for reasons unrelated to the change.

PR #2986, opened 2026-07-14
MON-4558: enable zoneinfo node-exporter collector via config

midu16:
Add optional NodeExporter zoneinfo collector toggle (disabled by default).

Wire --collector.zoneinfo/--no-collector.zoneinfo in updateNodeExporterArgs, extend unit and e2e tests, and update API and AsciiDoc documentation.

Addressing the isssue raised on https://github.com/openshift/cluster-monitoring-operator/pull/2910


hack/openshift-origin-collection-profiles-https-main.patch
```diff
@@ -0,0 +1,36 @@
...
+
+scrape_samples_post_metric_relabeling is summed for endpoint https-main
+only; count() must use the same job/endpoint or https-self inflates got.
+
+The ~301 sample delta that motivated reverting PR #2886 was caused by
+kube_running_pod_ready in the KSM minimal allowlist: it is a CMO recording
+rule, not a kube-state-metrics metric. MON-4581 removed it from the KSM
+minimal and telemetry ServiceMonitors; zoneinfo itself is disabled by
+default and is dropped by the minimal keep regex when enabled.
+
```
danielmellado: Is there a corresponding PR open against openshift/origin for this fix? The hack/ patch unblocks CI here but shouldn't get orphaned. Please link it in the description if one exists.

pkg/manifests/types.go
```diff
@@ -588,6 +591,14 @@ type NodeExporterCollectorSoftirqsConfig struct {
 	Enabled bool `json:"enabled,omitempty"`
 }
 
+// The `NodeExporterCollectorZoneinfoConfig` resource works as an on/off switch for
+// the `zoneinfo` collector of the `node-exporter` agent.
+// By default, the `zoneinfo` collector is disabled.
```
danielmellado, line 598: nit: Consider being slightly more specific about what /proc/zoneinfo exposes (e.g. "per-zone page counts, watermarks, and protection thresholds") to help users decide whether to enable it.

danielmellado (changes requested):
Commits 1 and 2 (hack: add openshift/origin patch... + hack: refresh origin collection-profiles patch (drop stray trailer)) should be squashed before merge. Commit 2 just removes a stray trailer from commit 1. 

Same openshift/api note as #2888: you'll need to add zoneinfo to NodeExporterCollectorConfig in openshift/api types_cluster_monitoring.go (simple collectionPolicy-only struct like softirqs), plus the merge line in config_merge.go. Can be a follow-up.

Besides this, it LGTM. I'll also wait for Simon to take a look.

pkg/manifests/types.go
simonpasquier: +1

simonpasquier:
> Same openshift/api note as #2888: you'll need to add zoneinfo to NodeExporterCollectorConfig in openshift/api types_cluster_monitoring.go (simple collectionPolicy-only struct like softirqs), plus the merge line in config_merge.go. Can be a follow-up.

I created https://redhat.atlassian.net/browse/MON-4607 as a sub-task of [MON-4558](https://redhat.atlassian.net/browse/MON-4558)

jsonnet/versions.yaml
simonpasquier, line 1: Please don't bother about updating the versions.yaml file because it creates unnecessary changes to review. The versions job is optional and we have an automated task to keep it in the file in sync.

midu16: i updated it because ci/prow/generate failure. was failing because of the versions not matching.

simonpasquier: I think that these are 2 unrelated issues. Please revert the changes to versions.yaml and regenerate.

midu16: reverted and all local generations are passing.

simonpasquier (commented):
/lgtm

PR #2993, opened 2026-07-17
OCPBUGS-99019: jsonnet: exclude AlertmanagerClusterFailedPeers from shipped rules

danielmellado:
The upstream alertmanager mixin added a new AlertmanagerClusterFailedPeers
alert (prometheus/alertmanager@34fb323) which was pulled in via a jsonnet
dependency bump in PR #2953. The alert fires when
alertmanager_cluster_failed_peers > 0 for 15 minutes.

In CI (and any environment where Alertmanager pods are recreated), the
alert fires on transient peer failures during StatefulSet rollouts:
DNS resolution briefly fails for the old pod, or the stale peer IP
times out, causing the metric to stay above zero long enough to trigger
the alert. This is expected operational behavior, not a real HA problem.

Origin e2e tests reject any new alert without historical test data
("[sig-trt][invariant] No new alerts should be firing"), which caused
techpreview serial jobs to fail after the merge.

Exclude the alert from the shipped PrometheusRule, consistent with how
we already exclude AlertmanagerClusterCrashlooping. Sustained HA issues
remain covered by AlertmanagerMembersInconsistent and
AlertmanagerClusterDown.

Reference: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-main-ci-5.0-e2e-gcp-ovn-techpreview-serial-1of2/2077851826295672832

### DIFF
--- a/assets/alertmanager/prometheus-rule.yaml
+++ b/assets/alertmanager/prometheus-rule.yaml
@@ -101,14 +101,3 @@ spec:
       for: 5m
       labels:
         severity: warning
-    - alert: AlertmanagerClusterFailedPeers
-      annotations:
-        description: Alertmanager {{ $labels.namespace }}/{{ $labels.pod}} has {{ $value }} failed peers in the {{$labels.job}} cluster.
-        summary: An Alertmanager instance has failed peers in the cluster.
-      expr: |
-        # Without max_over_time, failed scrapes could create false negatives, see
-        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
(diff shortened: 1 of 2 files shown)


danielmellado:
cc @marioferh @simonpasquier

machine424:
I think something's not right here.
`alertmanager_cluster_failed_peers > 0` for more than 15 minutes makes me think the metric is broken or our clustering config is messed up.
Ok to remove it temp to fix the test, but I think we need to investigate this.

danielmellado:
> I think something's not right here. `alertmanager_cluster_failed_peers > 0` for more than 15 minutes makes me think the metric is broken or our clustering config is messed up. Ok to remove it temp to fix the test, but I think we need to investigate this.

that's the idea, I see you already commented and saw the thread ;)

PR #2998, opened 2026-07-22
 [release-4.22] Add e2e regression test for CronJob timezone panic

slashpai:
Adds TestCronJobWithTimezone to verify kube-state-metrics handles CronJobs with .spec.timeZone without crashing. 

Regression test for openshift/kube-state-metrics#147.

Expected test to fail before openshift/kube-state-metrics#147 is merged and expected to succeed after merged

Assissted-By Claude Opus 4.6

### DIFF
--- a/test/e2e/kube_state_metrics_test.go
+++ b/test/e2e/kube_state_metrics_test.go
@@ -17,11 +17,6 @@ package e2e
 import (
 	"errors"
 	"fmt"
-	v1 "k8s.io/api/autoscaling/v1"
-	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	vpav1 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
 	"path"
 	"regexp"
 	"testing"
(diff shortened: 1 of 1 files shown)


test/e2e/kube_state_metrics_test.go
```diff
@@ -168,3 +172,79 @@ func deleteVPACRD(t *testing.T, vpaCRD interface{}) {
...
+		t,
+		timeout,
+		fmt.Sprintf(`group(kube_cronjob_next_schedule_time{cronjob="tz-regression-test",namespace="%s"} > 0)`, f.Ns),
+	)
+
+	// Verify the schedule is not marked as invalid.
+	f.ThanosQuerierClient.WaitForQueryReturnOne(
+		t,
+		timeout,
+		fmt.Sprintf(`group(kube_cronjob_schedule_invalid{cronjob="tz-regression-test",namespace="%s"} == 0)`, f.Ns),
```
simonpasquier, line 241: kube_cronjob_schedule_invalid only exists if the timezone is invalid.

slashpai: As we saw in 5.0 test without the fix for tzdata it was setting this metric as 1 even though correct schedule.

simonpasquier: yes but the CI base image have the /usr/share/zoneinfo directory

simonpasquier:
/close

unfortunately the e2e test doesn't help surfacing the issue because of differences between CI and product images.

PR #2999, opened 2026-07-23
NO-JIRA: test: adjust e2e tests on retention settings

simonpasquier:
This commit updates the e2e tests exercising the retention settings to check the Prometheus custom resource's spec instead of the Prometheus container's arguments, the reason being that the settings now live in the configuration file (and checking the generated configuration duplicates testing already performed at the Prometheus operator level).

Required for https://github.com/openshift/prometheus-operator/pull/385

### DIFF
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -235,7 +235,7 @@ func TestClusterMonitorPrometheusOperatorConfig(t *testing.T) {
 	}
 }
 
-func TestClusterMonitorPrometheusK8Config(t *testing.T) {
+func TestPrometheusK8sConfig(t *testing.T) {
 	const (
 		pvcClaimName    = "prometheus-k8s-db-prometheus-k8s-0"
 		statefulsetName = "prometheus-k8s"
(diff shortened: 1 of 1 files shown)


simonpasquier:
/retitle NO-JIRA: test: adjust e2e tests on retention settings

test/e2e/config_test.go
```diff
@@ -235,7 +235,7 @@ func TestClusterMonitorPrometheusOperatorConfig(t *testing.T) {
 	}
 }
 
-func TestClusterMonitorPrometheusK8Config(t *testing.T) {
+func TestPrometheusK8sConfig(t *testing.T) {
```
danielmellado, line 238: I assume you change the name here so it relates further to the current config location?

slashpai (commented):
/lgtm

test/e2e/config_test.go
simonpasquier: correct

simonpasquier:

the job permafails but it's not related to the change.

PR #3000, opened 2026-07-24
Fix Prometheus rule expression for KubeJobNotCompleted

BacLuc:
kube_job_status_active is 0 when completed, but it exists.
So the metric is still included and the time() - max still grows leading to an alert.
AI discovered this.

## Telemetry report

```telemetry_report
```

### DIFF
--- a/assets/control-plane/prometheus-rule.yaml
+++ b/assets/control-plane/prometheus-rule.yaml
@@ -181,7 +181,7 @@ spec:
       expr: |
         time() - max by(namespace, job_name, cluster) (kube_job_status_start_time{namespace=~"(openshift-.*|kube-.*|default)",job="kube-state-metrics"}
           and
-        kube_job_status_active{namespace=~"(openshift-.*|kube-.*|default)",job="kube-state-metrics"} > 0) > 43200
+        ( kube_job_status_active{namespace=~"(openshift-.*|kube-.*|default)",job="kube-state-metrics"} > 0 ) > 0) > 43200
       labels:
         severity: warning
     - alert: KubeJobFailed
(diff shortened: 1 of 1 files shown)


assets/control-plane/prometheus-rule.yaml
```diff
@@ -181,7 +181,7 @@ spec:
       expr: |
         time() - max by(namespace, job_name, cluster) (kube_job_status_start_time{namespace=~"(openshift-.*|kube-.*|default)",job="kube-state-metrics"}
           and
-        kube_job_status_active{namespace=~"(openshift-.*|kube-.*|default)",job="kube-state-metrics"} > 0) > 43200
+        (kube_job_status_active{namespace=~"(openshift-.*|kube-.*|default)",job="kube-state-metrics"} > 0) > 0) > 43200
```
danielmellado, line 184: In PromQL, comparison operators (>) bind tighter than set operators (and), so the evaluation order is:

kube_job_status_active{...} > 0 is evaluated first — this is a filter that drops all series where the value is 0 (i.e., completed jobs are removed from the result set entirely).
and then performs a set intersection, keeping kube_job_status_start_time only for jobs that survived step 1.
So completed jobs (where kube_job_status_active == 0) are already correctly excluded. The fact that the metric exists in Prometheus doesn't matter — the > 0 filter eliminates it before and ever sees it.

Changing kube_job_status_active{...} > 0 to (kube_job_status_active{...} > 0) > 0 is redundant — the first > 0 already returns only values greater than zero, so applying > 0 again has no effect.

Besides this, this shouldn't be edited directly, but changing the jsonnet file.

danielmellado (changes requested):
Thanks for the contribution @BacLuc, but I don't think this PR is mergeable as-is. I explained it in the review, but it mostly looks like an AI hallucination.

assets/control-plane/prometheus-rule.yaml
BacLuc: ah ok. I added some filters and these filters might have ended in the wrong place.
Sorry for the trouble.

PR #3005, opened 2026-07-28
MON-4624: enable the nvmesubsystem collector by default

simonpasquier:
This commit enables the nvmesubsystem collector by default (disabled by default in upstream) and adds an entry to the CMO configmap to toggle it.


pkg/manifests/config.go
```diff
@@ -316,6 +316,9 @@ func defaultNodeExporterConfig() *NodeExporterConfig {
 			DmMultipath: NodeExporterCollectorDmMultipathConfig{
 				Enabled: ptr.To(true),
 			},
+			NvmeSubsystem: NodeExporterCollectorNvmeSubsystemConfig{
```
machine424, line 319: nit: not really related to the PR.
for simplicity, not sure if we should still do so is he consumer is already doing
```
if collectors.NvmeSubsystem.Enabled == nil {
		collectors.NvmeSubsystem.Enabled = ptr.To(true)
	}
```

machine424 (commented):
thanks
/lgtm

simonpasquier:

the job failed on a router test case, unrelated to the PR.

PR #3009, opened 2026-07-28
OCPBUGS-99988: fix: convert OpenShift TLS version to Prometheus format for additional alertmanager configs

jan--f:
The APIServerConfig.MinTLSVersion() returns the OpenShift constant format (e.g. "VersionTLS12") which Prometheus cannot parse, causing prometheus-k8s pods to crash with "unknown TLS version: VersionTLS12" when additionalAlertmanagerConfigs use scheme: https.

Use the existing convertTLSVersionToMonitoringV1() function to convert to the Prometheus-compatible format (e.g. "TLS12") in both the Prometheus and Thanos Ruler additional alertmanager config paths.

## Telemetry report

```telemetry_report
```

### DIFF
--- a/pkg/manifests/amcfg.go
+++ b/pkg/manifests/amcfg.go
@@ -43,9 +43,15 @@ func (a PrometheusAdditionalAlertmanagerConfigs) MarshalYAMLWithTLSConfig(cipher
 func (f *Factory) MarshalPrometheusAdditionalAlertmanagerConfigs(amConfigs []AdditionalAlertmanagerConfig) ([]byte, error) {
 	prometheusAmConfigs := PrometheusAdditionalAlertmanagerConfigs(amConfigs)
 	cipherSuites := f.APIServerConfig.TLSCiphers()
-	minTLSVersion := f.APIServerConfig.MinTLSVersion()
 
-	result, err := prometheusAmConfigs.MarshalYAMLWithTLSConfig(cipherSuites, minTLSVersion)
+	// Convert the OpenShift TLS version format (e.g. "VersionTLS12") to the
+	// Prometheus-compatible format (e.g. "TLS12").
+	promTLSVersion, err := convertTLSVersionToMonitoringV1(f.APIServerConfig.MinTLSVersion())
+	if err != nil {
+		return nil, fmt.Errorf("converting TLS version for additional alertmanager config: %w", err)
+	}
+
+	result, err := prometheusAmConfigs.MarshalYAMLWithTLSConfig(cipherSuites, string(*promTLSVersion))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/manifests/amcfg.go
```diff
@@ -213,7 +225,7 @@ func (f *Factory) ConvertToThanosAlertmanagerConfiguration(ta []AdditionalAlertm
 
 		// Apply TLS security settings only for HTTPS connections without explicit TLS config
 		if needsTLSSecuritySettings {
-			cfg.HTTPConfig.TLSConfig.MinVersion = minTLSVersion
+			cfg.HTTPConfig.TLSConfig.MinVersion = string(*promTLSVersion)
```
machine424, line 228: Thanks for taking care of this, I was about to assign the ticket to myself.

shouldn't we be applying `APIServerConfig` to servers only? (here prometheus/thanos-ruler are clients to that external alertmanager, and here we're/se can only touching client config).

Saying that because in many other places, we're not trying to enforce client TLS configs.

Thant being said, I don't quit get `needsTLSSecuritySettings`, if we're enforcing, we must enforce even if the user is setting some fields in `TLSConfig`...

jan--f: hmm @danielmellado can you chime in here? The issue was introduced via https://redhat.atlassian.net/browse/OCPBUGS-58475

I could see a case for enforcing TLS configs for connections that go off cluster but I agree that we should only do it on wider consensus.

Also is thanos even seeing the TLSConfig? https://thanos.io/tip/components/rule.md/#alertmanager has no MinVersion or Cipher fields. Prometheus however does have those fields, so perhaps its just a matter of reusing the same configs here.

danielmellado: Hey, thanks for pinging me.

So the background on OCPBUGS-58475: the goal was just to get rid of insecure cipher warnings that were showing up in CMO logs whenever Prometheus/Thanos connected to external alertmanagers over HTTPS. Setting a TLS floor on the client side doesn't hurt security, it just means we won't negotiate anything below TLS 1.2.

@machine424 has a fair point though, we don't enforce client TLS configs anywhere else, and the APIServerConfig profile is really meant for servers. Plus the needsTLSSecuritySettings guard makes it basically a no-op for anyone who sets a CA or cert (which is most real configs).

On the Thanos question: IIUC Thanos uses prometheus/common for its HTTP client config internally, so it should actually honor min_version and cipher_suites even though the docs don't explicitly list them. So reusing the same config for both paths makes sense.

I'd say let's get this merged as a crash fix since the format conversion is clearly the right thing to do. I'll open a follow-up to clean up the enforcement logic, either drop it entirely (BoringCrypto already handles the TLS 1.2 floor at the Go binary level) or at least make it consistent.

simonpasquier: https://github.com/openshift/enhancements/blob/master/enhancements/security/centralized-tls-config.md#non-goals says in the non-goals:

> Enforcing client TLS settings. This enhancement applies only to the server-side TLS configuration of managed components.

For external Alertmanager configuration, I don't think that CMO should enforce anything. For in-cluster Alertmanager, it should be fine because the server will enforce the TLS requirements anyway.

danielmellado: Thanks for that link @simonpasquier, I've put up https://github.com/openshift/cluster-monitoring-operator/pull/3012

PR #3011, opened 2026-07-29
NO-JIRA: increase golangci-lint timeout

simonpasquier:


machine424:
/lgtm
Ok for the band-aid, it's still one of the fastest tests. Not sure if we need to change any timeouts in release/ as well (we could probably increase the cpu requests as well...)
We can dig deeper if needed later.

PR #3016, opened 2026-07-30
NO-JIRA: e2e: defer subtest resource cleanup in TestPrometheusRemoteWrite

machine424:
Deleting TLS secrets between subtests while the Prometheus CR still references them causes Prometheus Operator to enter exponential backoff on the openshift-monitoring/k8s key. The backoff compounds across subtests and can block CMO's ValidatePrometheus past its 5-minute timeout, making the parent cleanup poll expire.

Move per-subtest resource deletions (secrets, services, routes, receivers) into the parent cleanup, after the ConfigMap is reset and the remote write config is confirmed removed, so the CR never references missing resources.

## Telemetry report

```telemetry_report
```

### DIFF
--- a/test/e2e/prometheus_test.go
+++ b/test/e2e/prometheus_test.go
@@ -146,6 +146,13 @@ func TestPrometheusRemoteWrite(t *testing.T) {
 	}
 	image := k8sProm.Spec.Image
 
+	// Subtests register resource deletions here instead of calling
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
/lgtm

good catch!

machine424:
/retitle NO-JIRA: e2e: defer subtest resource cleanup in TestPrometheusRemoteWrite
/verified by CI

PR #3025, opened 2026-08-03
NO-JIRA: tasks: pass context to MetricsServerTask methods instead of storing it

machine424:
MetricsServerTask stored context.Context as a struct field, captured at construction time. Currently both contexts are the same, but storing context in a struct is a Go anti-pattern, it decouples the work's lifetime from the caller's intent and breaks if the caller ever passes a different context (e.g. a per-sync cancellable context).

## Telemetry report

```telemetry_report
```


machine424:
/retitle NO-JIRA: tasks: pass context to MetricsServerTask methods instead of storing it
/verified by CI

simonpasquier (commented):
/lgtm

PR #3032, opened 2026-08-04
MON-4577: restrict alertmanager gossip mesh ports to same-instance pods

machine424:
first part of https://github.com/openshift/cluster-monitoring-operator/pull/2926

---

Split the alertmanager NetworkPolicy ingress rules so that ports 9094 TCP/UDP (used for the HA gossip mesh) are only accessible from alertmanager pods of the same instance, while the service ports (9092, 9095, 9097) remain open to all sources.

This applies to both the platform alertmanager (openshift-monitoring) and the user-workload alertmanager (openshift-user-workload-monitoring).

The e2e network policy connectivity test is updated to skip ingress rules that have a 'from' selector since the probe pod won't match the restricted source criteria.

## Telemetry report

```telemetry_report
```

### DIFF
--- a/assets/alertmanager-user-workload/network-policy-downstream.yaml
+++ b/assets/alertmanager-user-workload/network-policy-downstream.yaml
@@ -17,6 +17,13 @@ spec:
       protocol: TCP
     - port: 9097
       protocol: TCP
+  - from:
+    - podSelector:
+        matchLabels:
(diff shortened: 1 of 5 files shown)


simonpasquier (commented):
/lgtm

machine424:
MachineConfig rolling reboots during tests

/retest-required

PR #3033, opened 2026-08-04
NO-JIRA: Allow to pass custom arguments to e2e test command

simonpasquier:
Needed by https://github.com/openshift/monitoring-plugin which doesn't need to run all CMO tests but only the ones exercising the monitoring plugin.


jgbernalp (approved):
/lgtm

PR #3037, opened 2026-08-05
NO-JIRA: Add test descriptions

simonpasquier:
This commit adds descriptions for tests which the monitoring plugin CI depends on. In particular they advertise not changing the names to avoid side effects.


simonpasquier:
/override ci/prow/e2e-agnostic-operator 

the PR only updates code comments.

PR #3042, opened 2026-08-06
OCPBUGS-105310: [release-4.19] wrap library-go resourceCache with mutex for thread safety

machine424:
## Summary

The library-go `resourceCache` is not thread-safe and CMO runs tasks concurrently with the same client, which causes data races.

Replace the cache with a no-op implementation until we can use a library-go that ships a thread-safe version (see https://github.com/openshift/library-go/pull/2380).

Disabling the cache is the simplest fix to backport, especially since we were not heavily relying on it.

## Changes

- Replace `resourceapply.NewResourceCache()` with a `noOpResourceCache{}` that always returns `false` from `SafeToSkipApply`

/jira OCPBUGS-99769
/cherry-pick main

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -23,6 +23,7 @@ import (
 	"reflect"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/imdario/mergo"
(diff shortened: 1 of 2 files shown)


machine424:
/retitle OCPBUGS-105310: replace resourceCache with no-op to fix concurrent map access

machine424:
/retitle OCPBUGS-105310: [release-4.19] replace resourceCache with no-op to fix concurrent map access

PR #3039, opened 2026-08-05
OCPBUGS-105389: enable read-only rootfs for all containers

simonpasquier:

PR #3047, opened 2026-08-10
OCPBUGS-105469: enable read-only rootfs for all containers

simonpasquier:
Manual cherry-pick of #3039

### DIFF
--- a/assets/admission-webhook/deployment.yaml
+++ b/assets/admission-webhook/deployment.yaml
@@ -67,7 +67,8 @@ spec:
           requests:
             cpu: 5m
             memory: 30Mi
-        securityContext: {}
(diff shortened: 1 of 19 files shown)


simonpasquier:
/hold

let's wait for #3039 to soak in a bit.

machine424:
I see why this is important, but I'm hesitant to backport it.
We could break unknown use cases (especially around /tmp) in a z stream.
I think it makes more sense to target 5.0 only and mention the changes in release notes

But maybe I'm missing some context.

/lgtm
/hold

simonpasquier:
/hold cancel

Clarified out of band that the backport makes sense. We have good confidence that it's not going to introduce regression.

PR #3048, opened 2026-08-10
NO-JIRA: feat: validate systemd units

simonpasquier:
This commit moves the validation of the system units from the operator's reconciliation to the configuration loading. It improves the user experience because invalid values will now be flagged by the admission webhook service.


danielmellado:
Thanks Simon, much better than to have a reconciliation loop failing around!

PR #3051, opened 2026-08-11
ART-22049: remove references to configmap-reloader image

simonpasquier:
This commit removes the last references to the configmap-reloader image which was used for the telemeter-client deployment but has been replaced by the prometheus-config-reloader a long time ago.

https://github.com/openshift/cluster-monitoring-operator/pull/970


simonpasquier:
/hold

I'll investigate what it will take to remove configmap-reload from openshift/release (if possible).

PR #3054, opened 2026-08-13
NO-JIRA: test: produce JUnit XML from e2e runs in CI

machine424:
When ARTIFACT_DIR is set (Prow/ci-operator), use gotestsum to write JUnit XML alongside normal stdout. The existing CI pipeline ingests these files into BigQuery, giving us searchable failure messages, flake rates, and duration trends via Search.CI and Sippy.

gotestsum is added to `hack/tools/` (same as promtool, jb, etc.) and built via the existing `$(TOOLING)` target. `--format=standard-verbose` preserves the same per-test output as plain `go test -v`.

No change for local runs — `GOTEST` falls back to plain `go test`.


simonpasquier (commented):
/lgtm

jan--f:
Oh nice! I wonder why we missed this until now.
/lgtm
/verified by @jan--f

PR #3055, opened 2026-08-13
NO-JIRA: hack/tools: remove unused vendor directory

machine424:
## Summary

The `hack/tools/vendor/` directory was added inadvertently in fba67bcc9 (CVE-2026-33186 grpc bump, 2026-04-01) but the `$(TOOLING)` Makefile target has always used `-mod=mod`, meaning it downloads from the module proxy and never reads `hack/tools/vendor/`.

This dead directory adds ~7600 tracked files and 3.2M lines with no functional purpose.

Fixes https://github.com/openshift/cluster-monitoring-operator/pull/2872

## Changes

- Remove `hack/tools/vendor/` (7664 files, never used by the build)
- Add `hack/tools/vendor/` to `.gitignore` to prevent reintroduction

The tools sub-module still has `go.mod` + `go.sum` for version pinning; builds continue to work via `-mod=mod` (module cache / GOPROXY).


simonpasquier (commented):
/lgtm

PR #3059, opened 2026-08-19
NO-JIRA: Fix staticcheck warnings across multiple packages

ankimaha-sys:
## Summary

This PR fixes several staticcheck warnings across multiple packages:

**Source code fixes:**
- **ST1005** (`pkg/alert/relabel_controller.go`, `pkg/alert/rule_controller.go`, `pkg/client/client.go`): Lowercase error strings per Go conventions. Also removes trailing punctuation from one error string.
- **ST1019** (`pkg/manifests/manifests.go`): Consolidate duplicate import of `k8s.io/api/core/v1`.

**Test code fixes:**
- **S1000** (`pkg/client/client_test.go`): Simplify single-case `select` blocks to direct channel receives.
- **S1008** (`pkg/operator/operator_test.go`): Simplify redundant `if-return-false-return-true` to direct boolean return.
- **S1004** (`pkg/manifests/tls_test.go`): Use `bytes.Equal` instead of `bytes.Compare(...) == 0`.
- **S1039** (`test/e2e/alertmanager_test.go`): Remove unnecessary `fmt.Sprintf` for a string literal.
- **S1003** (`test/e2e/framework/assertions.go`): Use `strings.Contains` instead of `strings.Index(...) >= 0`.

All changes are mechanical. `go vet` and `go build` pass cleanly.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -28,8 +28,6 @@ linters:
         - "-QF*"     # quickfix suggestions (style preference)
         - "-ST1000"  # package comments (style preference)
         - "-ST1003"  # naming conventions like API vs Api (style preference)
-        - "-ST1005"  # error string capitalization (style preference)
-        - "-ST1019"  # duplicate imports (style preference)
         - "-ST1020"  # comment format on functions (style preference)
         - "-ST1021"  # comment format on types (style preference)
     depguard:
--- a/pkg/alert/relabel_controller.go
+++ b/pkg/alert/relabel_controller.go
@@ -182,7 +182,7 @@ func (c *RelabelConfigController) processNextWorkItem(ctx context.Context) bool 
 	defer c.queue.Done(key)
 
 	if err := c.sync(ctx, key); err != nil {
-		utilruntime.HandleError(fmt.Errorf("Error syncing AlertRelabelConfig (%s): %w", key, err))
+		utilruntime.HandleError(fmt.Errorf("error syncing AlertRelabelConfig (%s): %w", key, err))
 
 		// Re-queue failed sync.
 		c.queue.AddRateLimited(key)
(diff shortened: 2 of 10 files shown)


ankimaha-sys:
Hi team, this is a small staticcheck cleanup PR — 9 files, +21/-32 lines, all mechanical fixes. Could someone /ok-to-test when you get a chance? Thanks!

machine424:
Thanks for this.
staticcheck does run as part of golangci https://github.com/openshift/cluster-monitoring-operator/blob/ce6cc5c03fd43ea531eff82d9f416e624a819672/.golangci.yaml#L18-L34
But the checks you mentioned was disabled here https://github.com/openshift/cluster-monitoring-operator/commit/7db8393cc1c0f28c26f760de6f5f2fdfeb01d74d
I don;t see a reason for that, especially that most of the violations can be auto-fixed. 

Would you be open to re-enabling those checks?

ankimaha-sys:
Thanks for the context! Yeah, I noticed those checks were disabled in the golangci config so ran standalone staticcheck to find them. Happy to adjust if any of the changes don't look right.

machine424:
Actually, we'd like to avoid doing that manually every now and then.

you can try to re-enable checks from the list

```
       - "-QF*"     # quickfix suggestions (style preference) 
       - "-ST1000"  # package comments (style preference) 
       - "-ST1003"  # naming conventions like API vs Api (style preference) 
       - "-ST1005"  # error string capitalization (style preference) 
       - "-ST1019"  # duplicate imports (style preference) 
       - "-ST1020"  # comment format on functions (style preference) 
       - "-ST1021"  # comment format on types (style preference) 
```

by removing the corresponding lines.

ankimaha-sys:
Re-enabled ST1005 and ST1019 checks in `.golangci.yaml` so they'll be enforced going forward. Verified with `golangci-lint run` — no new violations from these checks.

machine424:
Thanks!

Feel free to re-enable more of them in follow-up PRs, just make sure they stay manageable and easy to review.

/lgtm
/verified by machine424 and CI

PR #3060, opened 2026-08-21
MON-4640: wire interrupts node-exporter collector from ClusterMonitor…

midu16:
[MON-4640](https://redhat.atlassian.net/browse/MON-4640): Wire the interrupts node-exporter collector from the ClusterMonitoring CR

  ## What / Why

  The `ClusterMonitoring` CR (`config.openshift.io/v1alpha1`, behind the
  `ClusterMonitoringConfig` feature gate) exposes `nodeExporterConfig.collectors`
  so node-exporter collectors can be toggled without the legacy
  `cluster-monitoring-config` ConfigMap.

  `mergeNodeExporterCollectorsFromCRD` in `pkg/manifests/config_merge.go`, however,
  only translated **softirqs** among the newer collectors. Setting
  `collectors.interrupts` on the CR was silently a no-op — the CRD schema accepted
  and stored it, but CMO never emitted the corresponding `--collector.interrupts`
  flags. This PR closes that gap so the interrupts collector configured via the CR
  actually takes effect.

  Parent: MON-4559. Depends on the API field added in MON-4608.

  ## Changes

  - **`pkg/manifests/config_merge.go`**
    - `mergeNodeExporterCollectorsFromCRD`: translate the CR's interrupts
      `CollectionPolicy` into the internal ConfigMap-derived include-list model —
      `Collect` copies the CR include patterns into `Interrupts.Include`;
      `DoNotCollect` clears the list (a non-empty list enables the collector, an
      empty list disables it, matching how `manifests.go` emits
      `--collector.interrupts` / `--collector.interrupts.name-include=`).
    - `clusterMonitoringNodeExporterCollectorsEmpty`: include
      `Interrupts.CollectionPolicy` so an interrupts-only CR is not treated as an
      empty spec and skipped.
  - **`pkg/manifests/config_merge_test.go`**: add coverage —
    - interrupts `Collect` maps to the include list,
    - interrupts `DoNotCollect` maps to an empty include list,
    - ConfigMap-declared `collectors.interrupts` takes precedence (CR ignored),
    - interrupts case added to the collectors-empty test.
  - **vendor/github.com/openshift/api/config/v1alpha1**: vendor the `Interrupts`
    field and its types (`NodeExporterCollectorInterruptsConfig`,
    `NodeExporterCollectorInterruptsCollectConfig`,
    `NodeExporterInterruptsIncludePattern`) plus generated deepcopy, mirroring the
    API change from MON-4608.

  ## Precedence note

  Node-exporter merge is all-or-nothing at the `nodeExporter` block boundary: if
  the `cluster-monitoring-config` ConfigMap declares **any** `nodeExporter` field,
  `mergeNodeExporterConfiguration` returns early and the CR's entire
  `nodeExporterConfig` is ignored. This PR does not change that behavior; the new
  "ConfigMap wins" test documents it for interrupts.

  ## Testing

  - `go build ./pkg/manifests/...` — OK
  - `go test ./pkg/manifests/` — PASS (incl. the 4 new interrupts assertions)
  - `go vet ./pkg/manifests/` — clean

  ## Caveat / merge order

  The vendored `openshift/api` change anticipates MON-4608. Until that API PR
  merges upstream and CMO bumps `go.mod` + re-vendors, `make verify-vendor` will
  flag the vendor tree. This PR should merge (or be rebased) **after** the API
  bump; the vendored diff here is only to make the merge logic compile and testable
  now.

  ## Follow-up

  `zoneinfo` remains unwired in the CR path (same class of gap). Out of scope for
  [MON-4640](https://redhat.atlassian.net/browse/MON-4640); tracked separately.


simonpasquier:
/close

duplicate of https://github.com/openshift/cluster-monitoring-operator/pull/3056

PR #3062, opened 2026-08-21
MON-4638: align ThanosQueryOverload description with for=1h

prithvipatil97:
## Summary
- OpenShift sets `for: 1h` for all `thanos-query` alerts (since #1087), but the upstream `ThanosQueryOverload` description still said the query had been overloaded for more than **15 minutes**.
- Add a named sanitize-rules patch for `ThanosQueryOverload` (before the wildcard `alert: ''` entry) so the description says **1 hour**, matching the OpenShift `for` duration.
- Update the generated `assets/thanos-querier/prometheus-rule.yaml` accordingly.
- Cosmetic / docs-in-alert change only: expression and firing behavior are unchanged.

## Test plan
- [x] Diff shows only the description duration wording (`15 minutes` → `1 hour`) plus the sanitize-rules override
- [ ] After merge / on a cluster with the new CMO build, confirm:
  ```bash
  oc -n openshift-monitoring get prometheusrule thanos-querier -o yaml | grep -A2 ThanosQueryOverload
  ```
  Description should say "more than 1 hour" while `for` remains `1h`.
- [ ] Optional: ask OWNERS whether Minor backports to release branches are desired (not assumed).

Fixes [MON-4638](https://redhat.atlassian.net/browse/MON-4638)

### DIFF
--- a/jsonnet/utils/sanitize-rules.libsonnet
+++ b/jsonnet/utils/sanitize-rules.libsonnet
@@ -479,6 +479,20 @@ local patchedRules = [
   {
     name: 'thanos-query',
     rules: [
+      // MON-4638: OpenShift sets for=1h for thanos-query alerts, but the upstream
+      // ThanosQueryOverload description still says "15 minutes". Override the
+      // description to match the OpenShift for duration. Named patch must come
+      // before the wildcard (alert: '') patch because matching stops at first hit.
+      {
+        alert: 'ThanosQueryOverload',
+        'for': '1h',
+        labels: {
+          severity: 'warning',
+        },
+        annotations: {
+          description: 'Thanos Query {{$labels.job}} in {{$labels.namespace}} has been overloaded for more than 1 hour. This may be a symptom of excessive simultaneous complex requests, low performance of the Prometheus API, or failures within these components. Assess the health of the Thanos query instances, the connected Prometheus instances, look for potential senders of these requests and then contact support.',
+        },
+      },
       {
         alert: '',
         'for': '1h',
--- a/assets/thanos-querier/prometheus-rule.yaml
+++ b/assets/thanos-querier/prometheus-rule.yaml
@@ -82,7 +82,7 @@ spec:
         severity: warning
     - alert: ThanosQueryOverload
       annotations:
-        description: Thanos Query {{$labels.job}} in {{$labels.namespace}} has been overloaded for more than 15 minutes. This may be a symptom of excessive simultaneous complex requests, low performance of the Prometheus API, or failures within these components. Assess the health of the Thanos query instances, the connected Prometheus instances, look for potential senders of these requests and then contact support.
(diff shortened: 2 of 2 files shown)


prithvipatil97:
Hi @machine424 @slashpai,

Could you please review/approve this change when you have a moment?

It only updates the ThanosQueryOverload alert description from “15 minutes” to “1 hour” so it matches OpenShift’s for: 1h (via a named sanitize-rules override). Expression and firing behavior are unchanged.

Kindly let me know if you want me to perform any changes.

Thanks in advance.

Regards,
Prithviraj Patil

jsonnet/utils/sanitize-rules.libsonnet
```diff
@@ -479,6 +479,20 @@ local patchedRules = [
   {
     name: 'thanos-query',
     rules: [
+      // MON-4638: OpenShift sets for=1h for thanos-query alerts, but the upstream
+      // ThanosQueryOverload description still says "15 minutes". Override the
+      // description to match the OpenShift for duration. Named patch must come
+      // before the wildcard (alert: '') patch because matching stops at first hit.
+      {
+        alert: 'ThanosQueryOverload',
+        'for': '1h',
```
machine424, line 488: do we need to set for and labels here?

how about moving this after the wildcard actually?

Also, would you be interested in opening a PR upstream to remove the 15 mins from the description https://github.com/thanos-io/thanos/blob/51afa7e58085c82d11ad4e6eafee0f12ef9aa96f/mixin/alerts/query.libsonnet#L148?; it's not handy for consumers that override `for`, no other alert in there hardcodes it in the description plus the description is long enough...

machine424:
Thanks Prithviraj for taking care of this.

jsonnet/utils/sanitize-rules.libsonnet
simonpasquier: 👍

prithvipatil97: Hello @machine424 ,
Thanks for your reveiew.

Yes, we need to set for and labels here.
As we know, with the current first-match patch logic we need them on the named ThanosQueryOverload entry. If we only override annotations.description, the wildcard (alert: '') never applies.
Moving the named patch after the wildcard wouldn’t work for the same reason, alert: '' matches first, so the description override would never run.

Yes, I will create an upstream PR in thanos-io/thanos to remove the hardcoded “15 minutes” from the ThanosQueryOverload description so users that override for aren’t stuck with mismatched text.

Please let me know if we are good with this approach.

Regards,
Prithviraj Patil

machine424:
/lgtm
/verified by @machine424

I see, thanks for the details!

prithvipatil97:
Hello @machine424 ,
Thanks for your review and providing the approval.

Now, the PR approved + lgtm + verified. But, the six red e2e jobs died assembling the OpenShift release payload because rhel-coreos / rhel-coreos-10 has a broken image label ("${DESCRIPTION}".

Could you please confirm if we have any option here to get this job succeeded, or do we need to wait until nightly can build a payload.

Regards,
Prithviraj Patil

simonpasquier:
> Could you please confirm if we have any option here to get this job succeeded, or do we need to wait until nightly can build a payload.

second option

PR #3081, opened 2026-09-08
RFE-8850: add app.kubernetes.io/name label to cluster-monitoring-view

dhananjay316:
## Summary
- Add `app.kubernetes.io/name: cluster-monitoring-view` to the CMO-owned `cluster-monitoring-view` ClusterRole so it can be selected for RBAC aggregation, matching `cluster-logging-*-view`.
- This covers the monitoring half of [RFE-8850](https://redhat.atlassian.net/browse/RFE-8850). `cluster-reader` remains owned by openshift-apiserver bootstrap policy.

## Test plan
- [ ] Confirm generated `assets/cluster-monitoring-operator/cluster-role-view.yaml` includes `app.kubernetes.io/name: cluster-monitoring-view`
- [ ] `go test ./pkg/manifests -run TestClusterMonitoringClusterRoleViewNameLabel`
- [ ] After install, `oc get clusterrole cluster-monitoring-view -o jsonpath='{.metadata.labels}'` shows the name label

### DIFF
--- a/assets/cluster-monitoring-operator/cluster-role-view.yaml
+++ b/assets/cluster-monitoring-operator/cluster-role-view.yaml
@@ -3,6 +3,7 @@ kind: ClusterRole
 metadata:
   labels:
     app.kubernetes.io/managed-by: cluster-monitoring-operator
+    app.kubernetes.io/name: cluster-monitoring-view
     app.kubernetes.io/part-of: openshift-monitoring
   name: cluster-monitoring-view
 rules:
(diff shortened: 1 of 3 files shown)


simonpasquier:
From https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-name

> Used on: All Objects **(typically used on workload resources)**.

While the change would satisfy the immediate customer's request, we want consistency across all resources. It also creates an expectation on us to maintain the label over time. Has it been suggested to the customer that they had their own label to the cluster-monitoring-view cluster role? CMO should preserve it (if not, it's a bug).

```
oc label clusterrole cluster-monitoring-view aggregate-to-monitoring=true
```

dhananjay316:
> While the change would satisfy the immediate customer's request, we want consistency across all resources. It also creates an expectation on us to maintain the label over time. Has it been suggested to the customer that they had their own label to the cluster-monitoring-view cluster role? CMO should preserve it

@simonpasquier  thanks, right. I'll see if the customer agrees to it.

PR #3080, opened 2026-09-08
MON-4685: add option to disable the dmmultipath collector

simonpasquier:
This commit adds an entry to the CMO configmap to disable the dmmultipath collector (enabled by default in upstream node_exporter).

PR #3082, opened 2026-09-09
MON-4679: Add option to disable the nvmesubsystem collector

simonpasquier:


simonpasquier:
/hold

needs #3080 first as well as https://github.com/openshift/node_exporter/pull/188

